PR #18, opened 2021-09-10
Ignore Pull Request. This change is for testing purposes only. Fix typo in Makefile

sreyasn:


janitha (approved):

Makefile
```diff
@@ -265,3 +283,70 @@ verify: prereqs ## Run static code analysis
 .PHONY: verify-codegen
 verify-codegen: ## Verify generated code
 	hack/verify-codegen.sh
+
+## --------------------------------------
+## Development - kind
+## --------------------------------------
+# Kind cluster name used in integration tests. Please note this name must
+# match the value in the Groovy CI script or else the CI process won't
```
akutz, line 291: Please update this comment to no longer reflect the note about Groovy. Because `KIND_CLUSTER_NAME` is already something we can set externally due to the `?=`, we can simply ensure we export `KIND_CLUSTER_NAME=kind-it` in the internal code that bootstraps this.

I would also change the name _here_ to something more unique so anyone deploying locally won't have a conflict with other Kind clusters.

akutz (changes requested):
Aside from a few small things, looks good. I voiced one concern to Vikram, but other than that please just address the two issues I raised here and then I will approve.

PR #29, opened 2022-12-05
Introduce VirtualMachineImage Firmware status field and utilize with deployment Config Spec

sreyasn:
The firmware status field for a virtualmachineimage gets populated when creating the vmi. The firmware config string if populated in the ovf image will be consumed when the vm class config spec does not specify firmware type.

This helps prevent issues with deploying images that don't support a specific type of firmware (ie) efi,bios and default to the image as the baseline unless otherwise specified by the vi admin via the vmclass.

Testing: 
- make test
- Manual testing with centos marketplace using OvfEnv transport

### DIFF
--- a/api/v1alpha1/virtualmachineimage_types.go
+++ b/api/v1alpha1/virtualmachineimage_types.go
@@ -136,7 +136,12 @@ type VirtualMachineImageStatus struct {
 	// ContentVersion describes the observed content version of this VirtualMachineImage that was last successfully
 	// synced with the vSphere content library item.
 	// +optional
-	ContentVersion string `json:"contentVersion"`
+	ContentVersion string `json:"contentVersion,omitempty"`
+
+	// Firmware describe the firmware type used by this VirtualMachineImage.
+	// eg: bios, efi.
+	// +optional
+	Firmware string `json:"firmware,omitempty"`
 }
 
 func (vmImage *VirtualMachineImage) GetConditions() Conditions {
--- a/pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go
+++ b/pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go
@@ -20,6 +20,7 @@ func CreateConfigSpec(
 	name string,
 	vmClassSpec *v1alpha1.VirtualMachineClassSpec,
 	minFreq uint64,
+	imageFirmware string,
 	vmClassConfigSpec *vimtypes.VirtualMachineConfigSpec) *vimtypes.VirtualMachineConfigSpec {
 
 	var configSpec *vimtypes.VirtualMachineConfigSpec
@@ -80,6 +81,11 @@ func CreateConfigSpec(
 		configSpec.MemoryAllocation.Limit = &lim
 	}
 
+	// Use firmware type from the image if config spec doesn't have it.
+	if configSpec.Firmware == "" && imageFirmware != "" {
+		configSpec.Firmware = imageFirmware
+	}
+
 	return configSpec
 }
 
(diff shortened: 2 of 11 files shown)


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -137,6 +137,10 @@ type VirtualMachineImageStatus struct {
 	// synced with the vSphere content library item.
 	// +optional
 	ContentVersion string `json:"contentVersion"`
+
+	// Firmware describe the firmware type used by this VirtualMachineImage. eg: bios, efi.
```
akutz, line 141: Please do not exceed 80chars for line length.

api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -137,6 +137,10 @@ type VirtualMachineImageStatus struct {
...
+	// +optional
+	Firmware string `json:"firmware"`
```
akutz, line 143: If a field is optional then you this should be `json:"firmware,omitempty"`. That's true of the `ContentVersion` field above as well.

akutz (changes requested):

pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -90,9 +96,10 @@ func CreateConfigSpecForPlacement(
 	vmClassSpec *v1alpha1.VirtualMachineClassSpec,
 	minFreq uint64,
 	storageClassesToIDs map[string]string,
+	imageFirmware string,
 	vmClassConfigSpec *vimtypes.VirtualMachineConfigSpec) *vimtypes.VirtualMachineConfigSpec {
 
-	configSpec := CreateConfigSpec(vmCtx.VM.Name, vmClassSpec, minFreq, vmClassConfigSpec)
+	configSpec := CreateConfigSpec(vmCtx.VM.Name, vmClassSpec, minFreq, imageFirmware, vmClassConfigSpec)
```
akutz, line 102: TODO: make the function take `CreateConfigSpecOpts` or something like that. Not right now, just to simplify it later.

pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -80,6 +81,11 @@ func CreateConfigSpec(
 		configSpec.MemoryAllocation.Limit = &lim
 	}
 
+	// Use firmware type from the image if config spec doesn't have it.
+	if configSpec.Firmware == "" && imageFirmware != "" {
```
akutz, line 85: I thought we said the image would override the ConfigSpec since that is what AMIs do?

akutz (changes requested):
PTAL at the note about order of precedence

pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go
sreyasn: I thought the order would be VMI status.Firmware > configSpec.Firmware > CL setting BIOS/EFI firmware ?. 

meaning we should still let the vi admin specify configSpec.Firmware and we'll honor it?

pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go
sreyasn: will mark a TODO here. Thanks!

pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go
akutz: Based on our conversation with the GUI team, they will work with us to ensure the firmware selection is optional in the VM Class Create/Edit workflow in vSphere. This means the above logic is fine, since an admin will be able to omit selecting a firmware.

akutz (approved):
lgtm, thanks!

PR #30, opened 2022-12-06
Add security compliance check for clustercontentlibraryitem and contentlibraryitem controllers

cyclexuxu:
This change includes:

-  Update `external-crds` yaml files.
-  Add `securityCompliance` check for `clustercontentlibraryitem` and `conternlibraryitem` controllers. 

   - When the `securityCompliance` field of a `ContentLibraryItem` or `ClusterContentLibraryItem` is not set or is set to be false, the controller should not create corresponding `VirtualMachineImage` or `ClusterVirtualMachineImage`. If already exists, should delete `VirtualMachineImage` or `ClusterVirtualMachineImage` since Content Library doesn't support deploying such VM images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,11 @@ KUBECTL            := $(TOOLS_BIN_DIR)/kubectl
 ETCD               := $(TOOLS_BIN_DIR)/etcd
 
 # Allow overriding manifest generation destination directory
-MANIFEST_ROOT ?= config
-CRD_ROOT      ?= $(MANIFEST_ROOT)/crd/bases
-WEBHOOK_ROOT  ?= $(MANIFEST_ROOT)/webhook
-RBAC_ROOT     ?= $(MANIFEST_ROOT)/rbac
+MANIFEST_ROOT     ?= config
+CRD_ROOT          ?= $(MANIFEST_ROOT)/crd/bases
+EXTERNAL_CRD_ROOT ?= $(MANIFEST_ROOT)/crd/external-crds
+WEBHOOK_ROOT      ?= $(MANIFEST_ROOT)/webhook
+RBAC_ROOT         ?= $(MANIFEST_ROOT)/rbac
 
 # Image URL to use all building/pushing image targets
 IMAGE ?= vmoperator-controller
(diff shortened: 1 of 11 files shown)


dilyar85 (approved):
LGTM!

akutz (changes requested):
Please see the comment I left on the change. Also, what is `SecurityCompliance`? The comment is unhelpful:

> SecurityCompliance shows the security compliance of the library item.

That's like saying `Foo is the foo of the item`. Unless I know what `foo` means, such a comment is useless. So what is `SecurityCompliance` and why should we care what it says? Why is the item even surfaced if it's not usable? Please work with the IaaS Image Registry folks and suggest they should not even surface library items that are not accessible in a given namespace. 

If a `ContentLibraryItem` or `ClusterContentLibraryItem` resource exists, it should be something we can leverage. This seems like a design flaw with the image registry APIs.

cyclexuxu:
> 



> Please see the comment I left on the change. Also, what is `SecurityCompliance`? The comment is unhelpful:
> 
> > SecurityCompliance shows the security compliance of the library item.
> 
> That's like saying `Foo is the foo of the item`. Unless I know what `foo` means, such a comment is useless. So what is `SecurityCompliance` and why should we care what it says? Why is the item even surfaced if it's not usable? Please work with the IaaS Image Registry folks and suggest they should not even surface library items that are not accessible in a given namespace.
> 
> If a `ContentLibraryItem` or `ClusterContentLibraryItem` resource exists, it should be something we can leverage. This seems like a design flaw with the image registry APIs.

Reached out to the Image Registry team on this topic:
 
What is `SecurityCompliance` and why should we care what it says? 

_The security compliance comes from security policy feature from CL, it shows if a library item is security compliant per the security policy configured on the library, currently only OVF policy is supported in CL._

Why is the item even surfaced if it's not usable? 

_The reason to show the content library item with security compliance false is to  provide some additional visibility on such OVF templates in the Supervisor cluster or namespace. Depending on the security compliance of the item or security policy enforcement on the library, the state may change to true or false. It's helpful to provide a way to expose such state info in the cluster so that devOps can have a way to tell  why the corresponding VM image either show up or disappear._

The security compliance state can be changed from `true` to `false`, and vice versa, either because the item is updated with valid/invalid cert, or the security policy is turned on or off.

akutz:
>  It's helpful to provide a way to expose such state info in the cluster so that devOps can have a way to tell why the corresponding VM image either show up or disappear.

But if the image disappears, the DevOps user won't know why because the resource is gone. I still maintain this has no value. cc @ericvmw

akutz (approved):

PR #33, opened 2022-12-06
Fix e-mail link in README

ihgann:
This change fixes a minor link typo in `README.md`. Previously, the link for Ellen Mei would point to https://github.com/vmware-tanzu/vm-operator/blob/main/meie@vmware.com, which does not exist.


akutz (approved):

akutz:
Thanks @ihgann !!

PR #34, opened 2022-12-07
Removed internalID from required field

ljoyceVMware:
This change removed internalID from required fields in VMI/CVMI Status since it's already deprecated.

### DIFF
--- a/api/v1alpha1/virtualmachineimage_types.go
+++ b/api/v1alpha1/virtualmachineimage_types.go
@@ -109,7 +109,7 @@ type VirtualMachineImageStatus struct {
 	Uuid string `json:"uuid,omitempty"` //nolint:revive,stylecheck
 
 	// Deprecated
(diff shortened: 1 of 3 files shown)


api/v1alpha1/virtualmachineimage_types.go
```diff
@@ -109,7 +109,7 @@ type VirtualMachineImageStatus struct {
 	Uuid string `json:"uuid,omitempty"` //nolint:revive,stylecheck
 
 	// Deprecated
-	InternalId string `json:"internalId"` //nolint:revive,stylecheck
+	InternalId string `json:"internalId,omitempty"` //nolint:revive,stylecheck
```
akutz, line 112: * Please mark this as deprecated in the correct manner -- https://book.kubebuilder.io/reference/markers/crd.html
* Should also have `// +optional`

akutz (changes requested):

PR #35, opened 2022-12-09
Propagate TKG labels to cvmi

zyiyi11:
After below discussion, image registry will use the label `servicetype.iaas.vmware.com/tkg` for TKG ClusterContentLibrary and ClusterContentLibraryItem in order to distinguish them from the other cluster scoped content library and content library item resources.

This change propagates the label to ClusterVirtualMachineImage resources corresponding to the TKG content library items as well.

Test done:
```
# k get cvmi --show-labels
NAME                    PROVIDER-NAME              CONTENT-LIBRARY-NAME   IMAGE-NAME                                                                  VERSION                           OS-TYPE               FORMAT   AGE   LABELS
vmi-0b5e4855a0f850ba8   clitem-0b5e4855a0f850ba8   cl-cbe1f4d63cf1d1877   ob-17861429-photon-3-k8s-v1.20.2---vmware.1-tkg.1.1d4f79a                                                                           Ovf      20h   type.services.vmware.com/tkg=
```

### DIFF
--- a/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
+++ b/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
@@ -202,6 +202,17 @@ func (r *Reconciler) setUpCVMIFromCCLItem(cvmi *vmopv1a1.ClusterVirtualMachineIm
 		return err
 	}
 
+	if cvmi.Labels == nil {
+		cvmi.Labels = make(map[string]string)
+	}
+
+	// Only watch for service type labels from ClusterContentLibraryItem
+	for label := range cclItem.Labels {
+		if strings.HasPrefix(label, "type.services.vmware.com/") {
+			cvmi.Labels[label] = ""
+		}
+	}
+
 	// Do not initialize the Spec or Status directly as it might overwrite the existing fields.
 	cvmi.Spec.Type = string(cclItem.Status.Type)
 	cvmi.Spec.ImageID = cclItem.Spec.UUID
--- a/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller_unit_test.go
+++ b/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller_unit_test.go
@@ -170,6 +170,7 @@ func unitTestsReconcile() {
 					utils.PopulateRuntimeFieldsTo(expectedCVMI, &createdCVMI)
 
 					Expect(createdCVMI.Name).To(Equal(expectedCVMI.Name))
+					Expect(createdCVMI.Labels).To(Equal(expectedCVMI.Labels))
 					Expect(createdCVMI.OwnerReferences).To(Equal(expectedCVMI.OwnerReferences))
 					Expect(createdCVMI.Spec).To(Equal(expectedCVMI.Spec))
 					Expect(createdCVMI.Status).To(Equal(expectedCVMI.Status))
@@ -203,6 +204,7 @@ func unitTestsReconcile() {
 					utils.PopulateRuntimeFieldsTo(expectedCVMI, &updatedCVMI)
 
 					Expect(updatedCVMI.Name).To(Equal(expectedCVMI.Name))
+					Expect(updatedCVMI.Labels).To(Equal(expectedCVMI.Labels))
 					Expect(updatedCVMI.OwnerReferences).To(Equal(expectedCVMI.OwnerReferences))
 					Expect(updatedCVMI.Spec).To(Equal(expectedCVMI.Spec))
 					Expect(updatedCVMI.Status).To(Equal(expectedCVMI.Status))
@@ -232,6 +234,7 @@ func unitTestsReconcile() {
 					utils.PopulateRuntimeFieldsTo(expectedCVMI, &currentCVMI)
 
 					Expect(currentCVMI.Name).To(Equal(expectedCVMI.Name))
+					Expect(currentCVMI.Labels).To(Equal(expectedCVMI.Labels))
 					Expect(currentCVMI.OwnerReferences).To(Equal(expectedCVMI.OwnerReferences))
 					Expect(currentCVMI.Spec.HardwareVersion).To(BeZero())
 					Expect(currentCVMI.Status.ImageSupported).To(BeNil())
--- a/controllers/contentlibrary/utils/test_utils.go
+++ b/controllers/contentlibrary/utils/test_utils.go
@@ -17,6 +17,9 @@ import (
 	imgregv1a1 "github.com/vmware-tanzu/vm-operator/external/image-registry/api/v1alpha1"
 )
 
+// ContentLibraryServiceTypeLabelKey is used to differentiate a TKG resource from a VM service resource.
+const ContentLibraryServiceTypeLabelKey = "type.services.vmware.com/tkg"
+
 func DummyClusterContentLibraryItem(name string) *imgregv1a1.ClusterContentLibraryItem {
 	cclItem := &imgregv1a1.ClusterContentLibraryItem{
 		TypeMeta: metav1.TypeMeta{
@@ -25,6 +28,10 @@ func DummyClusterContentLibraryItem(name string) *imgregv1a1.ClusterContentLibra
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
+			Labels: map[string]string{
+				ContentLibraryServiceTypeLabelKey: "",
+				"dummy-not-service-label":         "",
+			},
 		},
 		Spec: imgregv1a1.ClusterContentLibraryItemSpec{
 			UUID: "dummy-ccl-item-uuid",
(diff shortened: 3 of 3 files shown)


akutz:
Hey @ericvmw , the labels don't match what I thought we discussed. Why is the label not?

```yaml
metadata:
  labels:
    imageregistry.vmware.com/servicetype: tkg 
```

As it is:

* The `kind` should not be part of the label as other resources may wish to use the label as well
* If something can only have one service type then the type of service should be the label's value
* If there are multiple service types, then we should use either:
    * a comma-delimited value:

        ```yaml
        metadata:
          labels:
            servicetype.imageregistry.vmware.com/servicetype: tkg, service1, service2
        ```
    * or label existence:

        ```yaml
        metadata:
          labels:
            servicetype.imageregistry.vmware.com/tkg:      # can be empty, label presence is enough
            servicetype.imageregistry.vmware.com/service1: # can be empty, label presence is enough
            servicetype.imageregistry.vmware.com/service2: # can be empty, label presence is enough
        ```

akutz (changes requested):
Hey @ericvmw , the labels don't match what I thought we discussed. Why is the label not?

```yaml
metadata:
  labels:
    imageregistry.vmware.com/servicetype: tkg 
```

As it is:

* The `kind` should not be part of the label as other resources may wish to use the label as well
* If something can only have one service type then the type of service should be the label's value
* If there are multiple service types, then we should use either:
    * a comma-delimited value:

        ```yaml
        metadata:
          labels:
            servicetype.imageregistry.vmware.com/servicetype: tkg, service1, service2
        ```
    * or label existence:

        ```yaml
        metadata:
          labels:
            servicetype.imageregistry.vmware.com/tkg:      # can be empty, label presence is enough
            servicetype.imageregistry.vmware.com/service1: # can be empty, label presence is enough
            servicetype.imageregistry.vmware.com/service2: # can be empty, label presence is enough
        ```

pkg/vmprovider/providers/vsphere/constants/constants.go
```diff
@@ -114,4 +114,6 @@ const (
 	V1alpha1SubnetMask = "V1alpha1_SubnetMask"
 	// V1alpha1FormatNameservers is an alias for versioned templating function V1alpha1_FormatNameservers.
 	V1alpha1FormatNameservers = "V1alpha1_FormatNameservers"
+
+	ContentLibraryServiceTypeLabelKey = "contentlibrary.imageregistry.vmware.com/servicetype"
```
akutz, line 118: Please add comments to all exported symbols.

akutz:
> 👋 Hi @ericvmw / @acharyasreej,
> 
> I am confused. The labels don't match what I thought we discussed. For example, why not the following:
> 
> ```yaml
> metadata:
>   labels:
>     imageregistry.vmware.com/servicetype: tkg 
> ```
> 
> As it is:
> 
>     * The `kind` should not be part of the label as other resources may wish to use the label as well
> 
>     * If something can only have one service type then the type of service should be the label's value
> 
>     * If there are multiple service types, then we should use either:
>       
>       * a comma-delimited value:
>         ```yaml
>         metadata:
>           labels:
>             imageregistry.vmware.com/servicetype: tkg, service1, service2
>         ```
>         
>         
>             
>               
>             
>         
>               
>             
>         
>             
>           
>         or
>         ```yaml
>         metadata:
>           labels:
>             servicetype.imageregistry.vmware.com: tkg, service1, service2
>         ```
>         
>         
>             
>               
>             
>         
>               
>             
>         
>             
>           
>         The downside of the CSV approach is that label selectors do not allow for wildcard matches on the value. That is why I prefer the next option...
>       * label existence:
>         ```yaml
>         metadata:
>           labels:
>             servicetype.imageregistry.vmware.com/tkg:      # can be empty, label presence is enough
>             servicetype.imageregistry.vmware.com/service1: # can be empty, label presence is enough
>             servicetype.imageregistry.vmware.com/service2: # can be empty, label presence is enough
>         ```

Alternatively @zyiyi11, we could the label existence method in a more generic API group, i.e.  `servicetype.iaas.vmware.com/<service>:`. For example:

```yaml
metadata:
  labels:
    servicetype.iaas.vmware.com/tkg:      # can be empty, label presence is enough
    servicetype.iaas.com/service1: # can be empty, label presence is enough
    servicetype.iaas.com/service2: # can be empty, label presence is enough
``` 

This way the same label key pattern is used across resource types and operators. It makes **_no_** sense for `contentlibrary` to be in the label name, and when I think about it, `imageregistry` probably should not be either. I vote we go with the above approach in VM Operator, and our downstream consumers can rally to this pattern.

akutz:
Hi @zyiyi11,

Regarding https://github.com/vmware-tanzu/vm-operator/pull/35#pullrequestreview-1212220691...we could use the label existence method in a more generic API group, i.e.  `servicetype.iaas.vmware.com/<service>:`. For example:

```yaml
metadata:
  labels:
    servicetype.iaas.vmware.com/tkg:      # can be empty, label presence is enough
    servicetype.iaas.vmware.com/service1: # can be empty, label presence is enough
    servicetype.iaas.vmware.com/service2: # can be empty, label presence is enough
``` 

This way the same label key pattern is used across resource types and operators. It makes **_no_** sense for `contentlibrary` to be in the label name, and when I think about it, `imageregistry` probably should not be either. I vote we go with the above approach in VM Operator, and our downstream consumers can rally to this pattern.

acharyasreej:
> Hi @zyiyi11,
> 
> Regarding [#35 (review)](https://github.com/vmware-tanzu/vm-operator/pull/35#pullrequestreview-1212220691), we could use the label existence method in a more generic API group, i.e. `servicetype.iaas.vmware.com/<service>:`. For example:
> 
> ```yaml
> metadata:
>   labels:
>     servicetype.iaas.vmware.com/tkg:      # can be empty, label presence is enough
>     servicetype.iaas.vmware.com/service1: # can be empty, label presence is enough
>     servicetype.iaas.vmware.com/service2: # can be empty, label presence is enough
> ```
> 
> This way the same label key pattern is used across resource types and operators. It makes **_no_** sense for `contentlibrary` to be in the label name, and when I think about it, `imageregistry` probably should not be either. I vote we go with the above approach in VM Operator, and our downstream consumers can rally to this pattern.
> 
> cc @ericvmw @acharyasreej

As per the discussion with @akutz and @fstrudel, we have decided to change the label to the following:

```
metadata:
  labels:
    type.services.vmware.com/tkg:      # can be empty, label presence is enough
    type.services.vmware.com/service1: # can be empty, label presence is enough
    type.services.vmware.com/service2: # can be empty, label presence is enough
```

pkg/vmprovider/providers/vsphere/constants/constants.go
dilyar85, line 118: @zyiyi11 If this is only used in the ccl-item controller test code above, `controllers/contentlibrary/utils/test_utils.go` might be a better place to store it.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -119,8 +120,12 @@ func (r *Reconciler) ReconcileNormal(ctx goctx.Context, cclItem *imgregv1a1.Clus
 		logger.Error(nameErr, "Unsupported ClusterContentLibraryItem, skip reconciling")
 		return nil
 	}
-	cvmi := &vmopv1a1.ClusterVirtualMachineImage{}
-	cvmi.Name = cvmiName
+	cvmi := &vmopv1a1.ClusterVirtualMachineImage{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   cvmiName,
+			Labels: cclItem.Labels,
```
dilyar85, line 126: We have a function defined below `setUpCVMIFromCCLItem` to set up cvmi fields from cclItem. We should probably move the label setup there too to make everything related in one place.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller_unit_test.go
```diff
@@ -155,6 +155,9 @@ func unitTestsReconcile() {
 						Status: corev1.ConditionTrue,
 					},
 				}
+				labels := make(map[string]string)
+				labels[utils.ContentLibraryServiceTypeLabelKey] = ""
+				cclItem.ObjectMeta.Labels = labels
```
dilyar85, line 160: This label setup is irrelevant to the statement from the "When" block above. Can you instead add the label in the `DummyClusterContentLibraryItem` function and update all positive test cases to check the expected labels from cvmi?

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -202,6 +202,8 @@ func (r *Reconciler) setUpCVMIFromCCLItem(cvmi *vmopv1a1.ClusterVirtualMachineIm
 		return err
 	}
 
+	cvmi.Labels = cclItem.Labels
```
akutz, line 205: We don't want to just replace `labels` wholesale, right? What if there are labels on the images not on the CL items?

akutz (changes requested):

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
zyiyi11: makes sense. We only wanna watch for service type labels from cl items.

akutz (approved):
Lgtm, thank you!

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -202,6 +202,17 @@ func (r *Reconciler) setUpCVMIFromCCLItem(cvmi *vmopv1a1.ClusterVirtualMachineIm
 		return err
 	}
 
+	if cvmi.Labels == nil {
+		cvmi.Labels = make(map[string]string)
+	}
+
+	// Only watch for service type labels from ClusterContentLibraryItem
```
acharyasreej, line 209: You also  need to handle the case where the TKG library is changed  and the current TKG library is set as a cluster CL. In that case, the label  would be removed from the old CCLItems and hence should be removed from the CVMI resources as well.

zyiyi11: I think reconciliation will handle this situation. Once cclitem updates, controller will update the corresponding ClusterVirtualMachineImage resource.

PR #36, opened 2022-12-14
Best effort to update VirtualMachine

yi0909:
We should not completely fail the VM reconcile if the VM image goes away. This change includes:
- When update VM, best effort to get VM image status and do not return any error.
- Check if VM image status is empty when getting VM image.
- Update extra config based on VM image status. Return error if
  - image status is nil, while JSON_EXTRA_CONFIG exists in the global config but is missing from VM config.
  - image status is nil, VM metadata is not cloud-init and defer cloud-init config key exists in the VM config.

Test done:
- make test
- make test-integration
- WCP env: 
  Manually verified:
    - Deploy a VM using vmi vmi-8ba052935d29cd377:
    ```
    root@423fc941a1a0ea2ff656bc3b2a1ef962 [ ~ ]# k get vm vm-ubuntu-1 -n test-ns -ojsonpath={.spec.imageName}
    vmi-8ba052935d29cd377

    root@423fc941a1a0ea2ff656bc3b2a1ef962 [ ~ ]# k get vm -A
    NAMESPACE   NAME          POWER-STATE   AGE
    test-ns     vm-ubuntu-1   poweredOn     29m
    ```
    - Remove the VMI and power off the VM.
    ```
    root@423fc941a1a0ea2ff656bc3b2a1ef962 [ ~ ]# k get vm -A
    NAMESPACE   NAME          POWER-STATE   AGE
    test-ns     vm-ubuntu-1   poweredOff    31m

    root@423fc941a1a0ea2ff656bc3b2a1ef962 [ ~ ]# k get vmi -n test-ns | grep vmi-8ba052935d29cd377
    root@423fc941a1a0ea2ff656bc3b2a1ef962
    ```
    - Power on the VM
    ```
    root@423fc941a1a0ea2ff656bc3b2a1ef962 [ ~ ]# k get vm -A
    NAMESPACE   NAME          POWER-STATE   AGE
    test-ns     vm-ubuntu-1   poweredOn     40m
    ```

### DIFF
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_update.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_update.go
@@ -6,18 +6,15 @@ package session
 import (
 	"fmt"
 	"reflect"
-	"strings"
-	"text/template"
 
-	vimTypes "github.com/vmware/govmomi/vim25/types"
 	apiEquality "k8s.io/apimachinery/pkg/api/equality"
 
 	"github.com/vmware/govmomi/object"
+	vimTypes "github.com/vmware/govmomi/vim25/types"
 
 	"github.com/vmware-tanzu/vm-operator/api/v1alpha1"
 
 	"github.com/vmware-tanzu/vm-operator/pkg"
-	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/lib"
 	"github.com/vmware-tanzu/vm-operator/pkg/util"
@@ -30,6 +27,10 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/virtualmachine"
 )
 
+const (
+	FirstBootDoneAnnotation = "virtualmachine.vmoperator.vmware.com/first-boot-done"
+)
+
 type VMMetadata struct {
 	Data      map[string]string
 	Transport v1alpha1.VirtualMachineMetadataTransport
@@ -38,7 +39,6 @@ type VMMetadata struct {
 // VMUpdateArgs contains the arguments needed to update a VM on VC.
 type VMUpdateArgs struct {
 	VMClass        *v1alpha1.VirtualMachineClass
-	VMImageStatus  *v1alpha1.VirtualMachineImageStatus
 	ResourcePolicy *v1alpha1.VirtualMachineSetResourcePolicy
 	MinCPUFreq     uint64
 	ExtraConfig    map[string]string
@@ -49,6 +49,10 @@ type VMUpdateArgs struct {
 
 	NetIfList  network.InterfaceInfoList
 	DNSServers []string
+
+	// hack. Remove after VMSVC-1261.
+	// indicating if this VM image used is VM service v1alpha1 compatible.
+	VirtualMachineImageV1Alpha1Compatible bool
 }
 
 func ethCardMatch(newBaseEthCard, curBaseEthCard vimTypes.BaseVirtualEthernetCard) bool {
@@ -297,27 +301,14 @@ func UpdateConfigSpecExtraConfig(
 	config *vimTypes.VirtualMachineConfigInfo,
 	configSpec,
 	classConfigSpec *vimTypes.VirtualMachineConfigSpec,
-	vmImageStatus *v1alpha1.VirtualMachineImageStatus,
 	vmClassSpec *v1alpha1.VirtualMachineClassSpec,
 	vm *v1alpha1.VirtualMachine,
-	globalExtraConfig map[string]string) {
-
-	// The only use of this is for the global JSON_EXTRA_CONFIG to set the image name.
-	renderTemplateFn := func(name, text string) string {
-		t, err := template.New(name).Parse(text)
-		if err != nil {
-			return text
-		}
-		b := strings.Builder{}
-		if err := t.Execute(&b, vmImageStatus); err != nil {
-			return text
-		}
-		return b.String()
-	}
+	globalExtraConfig map[string]string,
+	imageV1Alpha1Compatible bool) {
 
 	extraConfig := make(map[string]string)
 	for k, v := range globalExtraConfig {
-		extraConfig[k] = renderTemplateFn(k, v)
+		extraConfig[k] = v
 	}
 
 	virtualDevices := vmClassSpec.Hardware.Devices
@@ -351,14 +342,16 @@ func UpdateConfigSpecExtraConfig(
 	// Enabling the defer-cloud-init extraConfig key for V1Alpha1Compatible images defers cloud-init from running on first boot
 	// and disables networking configurations by cloud-init. Therefore, only set the extraConfig key to enabled
 	// when the vmMetadata is nil or when the transport requested is not CloudInit.
+	// VMSVC-1261: we may always set this extra config key to remove image from VM customization.
+	// If a VM is deployed from an incompatible image,
+	// it will do nothing and won't cause any issues, but can introduce confusion.
 	if vm.Spec.VmMetadata == nil || vm.Spec.VmMetadata.Transport != v1alpha1.VirtualMachineMetadataCloudInitTransport {
-		if conditions.IsTrueFromConditions(vmImageStatus.Conditions, v1alpha1.VirtualMachineImageV1Alpha1CompatibleCondition) {
-			ecMap := ExtraConfigToMap(config.ExtraConfig)
-			if ecMap[constants.VMOperatorV1Alpha1ExtraConfigKey] == constants.VMOperatorV1Alpha1ConfigReady {
-				// Set VMOperatorV1Alpha1ExtraConfigKey for v1alpha1 VirtualMachineImage compatibility.
-				configSpec.ExtraConfig = append(configSpec.ExtraConfig,
-					&vimTypes.OptionValue{Key: constants.VMOperatorV1Alpha1ExtraConfigKey, Value: constants.VMOperatorV1Alpha1ConfigEnabled})
-			}
+		ecMap := ExtraConfigToMap(config.ExtraConfig)
+		if ecMap[constants.VMOperatorV1Alpha1ExtraConfigKey] == constants.VMOperatorV1Alpha1ConfigReady &&
+			imageV1Alpha1Compatible {
+			// Set VMOperatorV1Alpha1ExtraConfigKey for v1alpha1 VirtualMachineImage compatibility.
+			configSpec.ExtraConfig = append(configSpec.ExtraConfig,
+				&vimTypes.OptionValue{Key: constants.VMOperatorV1Alpha1ExtraConfigKey, Value: constants.VMOperatorV1Alpha1ConfigEnabled})
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


bryanv (changes requested):
@yi0909 There are a lot of lingering issues and cruft in the "update" path that I didn't get to before being peel off into other work, but this approach - by allowing the field to be nil and trying add nil checks - will too likely lead to NPEs. In fact, on a more recent main, we'd hit just that at https://github.com/vmware-tanzu/vm-operator/blob/c15e92d72d3a54942d35434dacc35ac4e4f85cc7/pkg/vmprovider/providers/vsphere/vmprovider_vm.go#L707, and adding another nil check for that case will cause a cascade of more now potential nil checks.

I've had a few ideas in the direction I was going to take this code to both fix this specific issue and address the cruft. Please ping me next week to discuss them.

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -349,6 +375,18 @@ func resolveVMImageStatus(
 		return nil, errors.Wrap(err, imageNotFoundMsg)
 	}
 
+	// Status is a subresource, return error if it is empty.
+	if imageStatus.ImageName == "" {
```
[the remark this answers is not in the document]
yi0909: This situation won't happen. If the image image is empty, the entire block is empty. Will update this empty check to use reflect.DeepEqual.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -669,11 +659,10 @@ func (s *Session) prepareVMForPowerOn(
 func (s *Session) poweredOnVMReconfigure(
 	vmCtx context.VirtualMachineContext,
 	resVM *res.VirtualMachine,
-	config *vimTypes.VirtualMachineConfigInfo,
-	classConfigSpec *vimTypes.VirtualMachineConfigSpec) error {
+	config *vimTypes.VirtualMachineConfigInfo) error {
 
 	configSpec := &vimTypes.VirtualMachineConfigSpec{}
-	UpdateConfigSpecChangeBlockTracking(config, configSpec, classConfigSpec, vmCtx.VM.Spec)
+	UpdateConfigSpecChangeBlockTracking(config, configSpec, nil, vmCtx.VM.Spec)
```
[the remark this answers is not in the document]
yi0909: Yep this is intentional, for a poweredOn -> poweredOn reconcile, I don't think we have to get VM class? In this case, the classConfigSpec can be nil and the vmCtx.VM.Spec.AdvancedOptions can update the `ChangeBlockTracking`. The first block will be used during VM powerOn reconfigure, so it's not redundant.

pkg/vmprovider/providers/vsphere/utils/vm_utils.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2022 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package vsphere
+package utils
```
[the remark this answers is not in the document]
yi0909: This file is moved to a new place, that's why this package name is changed. I'll rename it.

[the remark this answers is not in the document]
yi0909: I would keep this package as name `utils` for now? I kind of feel like other place or name doesn't make more sense. 😞  This change refactored some of VM update logic, so now there are two packages referencing these utility functions, one from vmprovider_vm (package vsphere) during VM creation, another from session_vm_update (package session) during VM update and package `vsphere` imports package `session`. Since we are going to remove session package, I don't want to add more functions to it. When we finally remove creation/update from session, we can find a better place for these functions.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -457,7 +451,7 @@ func updateConfigSpec(
 	UpdateHardwareConfigSpec(config, configSpec, &vmClassSpec)
 	UpdateConfigSpecCPUAllocation(config, configSpec, &vmClassSpec, updateArgs.MinCPUFreq)
 	UpdateConfigSpecMemoryAllocation(config, configSpec, &vmClassSpec)
-	UpdateConfigSpecExtraConfig(config, configSpec, updateArgs.ConfigSpec, updateArgs.VMImageStatus, &vmClassSpec, vmCtx.VM, updateArgs.ExtraConfig)
+	UpdateConfigSpecExtraConfig(config, configSpec, updateArgs.ConfigSpec, &vmClassSpec, vmCtx.VM, updateArgs.ExtraConfig)
 	UpdateConfigSpecChangeBlockTracking(config, configSpec, updateArgs.ConfigSpec, vmCtx.VM.Spec)
```
aruneshpa, line 455: Not your change, but should this be `vmClassSpec`?  It sound like the [function that populates ](https://github.com/vmware-tanzu/vm-operator/pull/36/files#diff-d5b7c2c460dc9978a628f2cb03390868152b8d5fdd4e2dd99c537ec36a2f6143R864)the `updateArgs.ConfigSpec` (`vmUpdateGetArgs`) doesn't take CBT into account, so IIUC, CBT will never get set?

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -669,11 +663,10 @@ func (s *Session) prepareVMForPowerOn(
 func (s *Session) poweredOnVMReconfigure(
 	vmCtx context.VirtualMachineContext,
 	resVM *res.VirtualMachine,
-	config *vimTypes.VirtualMachineConfigInfo,
-	classConfigSpec *vimTypes.VirtualMachineConfigSpec) error {
+	config *vimTypes.VirtualMachineConfigInfo) error {
 
 	configSpec := &vimTypes.VirtualMachineConfigSpec{}
-	UpdateConfigSpecChangeBlockTracking(config, configSpec, classConfigSpec, vmCtx.VM.Spec)
+	UpdateConfigSpecChangeBlockTracking(config, configSpec, nil, vmCtx.VM.Spec)
```
aruneshpa, line 669: Instead of passing the ClassSpec, we can just have this `UpdateConfigSpecChangeBlockTracking` take the desired CBT state (true/false) and update it.  That way, we don't have to split brain this method (passing non-nil and nil class spec).  Can be a followup.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -783,3 +783,98 @@ func (s *Session) UpdateVirtualMachine(
...
+		&updateArgs.VMClass.Spec,
+		updateArgs.MinCPUFreq,
+		imageFirmware,
+		updateArgs.ClassConfigSpec)
+
+	return updateArgs, nil
+}
+
+func isVMFirstBoot(vmCtx context.VirtualMachineContext) bool {
+	if _, ok := vmCtx.VM.Annotations[FirstBootDoneAnnotation]; ok {
```
aruneshpa, line 875: A devops user can potentially remove the annotation and that could cause the reconcile to break.  Can you file a JIRA story to have some prefix to denote if an annotation is a "system" one?  That way, we can reject any requests to modify the annotation using webhooks.

pkg/vmprovider/providers/vsphere/utils/vm_utils.go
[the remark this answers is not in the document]
aruneshpa: Don't like that. :( We are accumulating techdebt.  I am not sure when the session code will go away, likely it will stay for decent amount of time.
But considering the urgency of this change, it is okay.  Could you please file a JIRA story and link it to fix the tech debt?

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -783,3 +783,98 @@ func (s *Session) UpdateVirtualMachine(
...
+		return nil, err
+	}
+
+	updateArgs.VMClass = vmClass
+	updateArgs.VMMetadata = vmMD
+	if resourcePolicy != nil {
+		updateArgs.ResourcePolicy = resourcePolicy
+	} else {
+		// If resourcePolicy is not set, use default MinCPUFreq value.
+		updateArgs.MinCPUFreq = 0
```
aruneshpa, line 810: It is not trivial to infer how resource policy and CPU min frequency are related?  Can you add a comment here?  Ignore if you are just moving code around.

aruneshpa (approved):
LGTM barring a new comments.

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -85,10 +85,23 @@ func GetVirtualMachineClass(
...
+	isCreate bool) (imgStatus *vmopv1alpha1.VirtualMachineImageStatus, clUUID string, retErr error) {
 
 	imageName := vmCtx.VM.Spec.ImageName
+	defer func() {
+		if !isCreate {
+			// If this is called from vm update, VirtualMachinePrereqReadyCondition is always ready.
+			conditions.MarkTrue(vmCtx.VM, vmopv1alpha1.VirtualMachinePrereqReadyCondition)
+
+			if retErr != nil {
+				// Ignore error if this is called from UpdateVirtualMachine.
```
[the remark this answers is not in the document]
yi0909: This part of code has been refactored and the comment doesn't apply to the current implementation. Resolve this conversion.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
yi0909: if `vmClassConfigSpec` [is not nil](https://github.com/vmware-tanzu/vm-operator/blob/a96f5cf1732685251cd1d377c80fba5468587dc8/pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go#L27), it will contain in the `updateArgs.ConfigSpec` . Users can set CBT in the VM Class configSpec I think?

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
yi0909: Once this JIRA story https://jira.eng.vmware.com/browse/VMSVC-1261 is finished, this firstboot annotation can be removed. (We'll set image firmware as an annotation) I modified the story description to include webhook part.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
yi0909: For now there are two different places can set CBT, VM spec and Class spec. If we want to get the desired CBT state, we have to get from both VM spec and class spec, which leave a lot of work to the caller :( Even if the classConfigSpec is `nil`, we can't simply saying that the desired CBT state is false and still have to compare with the current VM config. This introduces code duplication if the check is done from caller.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
bryanv: Can we - at least with the regards to the firmware - ignore that at this point? The class ConfigSpec has gotten confusing in that's theoretically 2 FFS that's effectively consolidated into 1 but the firmware is something that we can or do set when creating the VM and let it be after than.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -783,3 +783,98 @@ func (s *Session) UpdateVirtualMachine(
 	// TODO: Find a better place for this?
 	return s.attachClusterModule(vmCtx, resVM, updateArgs.ResourcePolicy)
```
bryanv, line 784: `updateArgs.ResourcePolicy` is only set when L759 is hit so for like a TKG VM that has the cluster module annotation, we'd NPE when trying to find/attach the CM.

I understand that we're to do something quick here, but ay scope that we now have a partially initialized `updateArgs` would make me worried we have an NPE either now or later due to how messy & tangled the existing code is.

pkg/vmprovider/providers/vsphere/utils/vm_utils.go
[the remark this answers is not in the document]
yi0909: Fall back to vmprovider_vm_utils.go, this conservation didn't apply any more. Close this.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
yi0909: Refactor this, not apply to the current implementation any more.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
bryanv: Yes, I do think the existing behavior is suspect - the VM Spec should take priority when set. This is something we can sort out as we flesh out more of the ConfigSpec CL Deploy work (likely, the class value will only be used at create time, again only if the VM Spec is not set).

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
bryanv: Yes, the prior `UpdateConfigSpecFooBar()` pattern wasn't super well suited to adding yet another param for the class ConfigSpec; resolving the desired state as a `*T` at a higher level is something we'll likely end up doing, and may make the v1a2 work easier.

bryanv (approved):

aruneshpa (approved):

PR #38, opened 2022-12-15
Support for JSON-based ConfigSpec in VM Class

akutz:
This patch introduces support for a JSON-based ConfigSpec in a VirtualMachineClass. The latter's CRD has been updated to remove the XML bits, and now the ConfigSpec is stored as a RawExtension so it will be present in the VM Class's YAML.

This patch also updated the GoVmomi dependency, and thus pulled in a breaking change with the cluster placement types.


config/crd/bases/vmoperator.vmware.com_virtualmachineclasses.yaml
```diff
@@ -61,14 +61,12 @@ spec:
...
-                  xml:
-                    description: XML contains a vim.vm.ConfigSpec object that has
-                      been serialized to XML and base64-encoded.
-                    type: string
+                description: "ConfigSpec describes additional configuration information
+                  for a VirtualMachine. \n The contents of this field are the VirtualMachineConfigSpec
+                  data object (https://bit.ly/3HDtiRu) marshaled to JSON using the
+                  discriminator field \"_typeName\" to preserve type information."
                 type: object
+                x-kubernetes-preserve-unknown-fields: true
```
aruneshpa, line 69: Are we adding this field so the older Supervisor clusters that don't understand this field also have this in the VM Class custom resource?  An alternative approach can be to _not_ store this field and have the controller in WCP reconcile the cluster upon upgrade to update the VM Class with the new field.

In the current approach, we are "leaking" information to the users who can't use this capability anyway.  Thoughts?

pkg/util/configspec.go
```diff
@@ -80,6 +81,38 @@ func UnmarshalConfigSpecFromBase64XML(
 	return UnmarshalConfigSpecFromXML(data)
 }
 
+// MarshalConfigSpecToJSON returns a byte slice of the provided ConfigSpec
+// marshalled to a JSON string.
```
aruneshpa, line 85:
```suggestion
// marshaled to a JSON string.
```
marshalled -> marshaled.

aruneshpa (approved):
LGTM, with one minor comment and a nit.

config/crd/bases/vmoperator.vmware.com_virtualmachineclasses.yaml
akutz: To which field are you referring?

pkg/util/configspec.go
akutz: Hmm, I think it works either way. I've never been sure, and I know I've used it both ways. Since you picked the nit, I decided [to check](https://sapling.ai/usage/marshaled-vs-marshalled). It is not even close 😃

config/crd/bases/vmoperator.vmware.com_virtualmachineclasses.yaml
akutz: > Are we adding this field so the older Supervisor clusters that don't understand this field also have this in the VM Class custom resource?

I am not sure I understand the quesiton. In general, the purpose of this field is to avoid having to query VCSA every time we want to find out a VM Class's associated ConfigSpec. 

Are you asking specifically about `x-kubernetes-preserve-unknown-fields: true`? If so, then that's because the field is a `runtime.RawExtension` ([GoDoc](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#RawExtension), [Kube docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#rawextension)). Maybe we should be using `[]byte` instead? Hmm.

> so the older Supervisor clusters that don't understand this field

I keep coming back to this. I _think_ you are concerned about `x-kubernetes-preserve-unknown-fields: true` and think it means a vSphere will create a `VirtualMachineClass` with a non-empty `spec.ConfigSpec` on a Supervisor with an older VM Operator API? If that's the case then I think we are okay.

[Pruning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning) with `x-kubernetes-preserve-unknown-fields: true` is tricky. In this case ``x-kubernetes-preserve-unknown-fields: true` does not control whether the `spec.configSpec` field is pruned, but whether data _inside_ the `RawExtension` is pruned.

To illustrate what I mean:

1. Create a Kind cluster:

    ```shell
    kind create cluster
    ```

1. Use the Kind cluster:

    ```shell
    kubectl config use-context kind-kind
    ```

1. Create a `Fake` CRD:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    metadata:
      name: fakes.example.com
    spec:
      group: example.com
      names:
        kind: Fake
        listKind: FakeList
        plural: fakes
        singular: fake
      scope: Cluster
      versions:
      - name: v1alpha1
        schema:
          openAPIV3Schema:
            type: object
            properties:
              apiVersion:
                type: string
              kind:
                type: string
              metadata:
                type: object
              spec:
                type: object
                properties:
                  real:
                    type: string
              status:
                type: object
        served: true
        storage: true
        subresources:
          status: {}
    EOF
    ```

1. Create a new `fake`:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: example.com/v1alpha1
    kind: Fake
    metadata:
      name: fake
    spec:
      real: is it though?
    EOF
    ```

1. Print the new `fake`:

    ```shell
    kubectl get fake fake -oyaml
    ```


1. Try to update the `fake` object with `spec.fake`, an unknown field:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: example.com/v1alpha1
    kind: Fake
    metadata:
      name: fake
    spec:
      fake: fake all the way down!
    EOF
    ```

    The above command will return an error:

    ```
    error: error validating "STDIN": error validating data: ValidationError(Fake.spec): unknown field "fake" in com.example.v1alpha1.Fake.spec; if you choose to ignore these errors, turn validation off with --validate=false
    ```

1. To get the `Fake` API to accept `spec.fake`, short of defining `spec.properties.fake` in the CRD, we can set the `Fake` CRD's `spec.x-kubernetes-preserve-unknown-fields` field to `true`:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    metadata:
      name: fakes.example.com
    spec:
      group: example.com
      names:
        kind: Fake
        listKind: FakeList
        plural: fakes
        singular: fake
      scope: Cluster
      versions:
      - name: v1alpha1
        schema:
          openAPIV3Schema:
            type: object
            properties:
              apiVersion:
                type: string
              kind:
                type: string
              metadata:
                type: object
              spec:
                type: object
                x-kubernetes-preserve-unknown-fields: true
                properties:
                  real:
                    type: string
              status:
                type: object
        served: true
        storage: true
        subresources:
          status: {}
    EOF
    ```

1. Now the same command as before:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: example.com/v1alpha1
    kind: Fake
    metadata:
      name: fake
    spec:
      fake: fake all the way down!
    EOF
    ```

    results in:

    ```
    fake.example.com/fake configured
    ```

The change in the `VirtualMachineClass` CRD is equivalent _not_ to setting `spec.x-kubernetes-preserve-unknown-fields` to `true`, but rather:

* removing `spec.x-kubernetes-preserve-unknown-fields` 
* defining `spec.properties.fake.type` as an `object` that allows unknown properties

Let's try it:

1. Delete the existing `fake` object:

    ```shell
    kubectl delete fake fake
    ```

2. Update the `Fake` API so that `spec` no longer preserves unknown properties and `spec.properties.fake` is an `object` that preserves unknown properties:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    metadata:
      name: fakes.example.com
    spec:
      group: example.com
      names:
        kind: Fake
        listKind: FakeList
        plural: fakes
        singular: fake
      scope: Cluster
      versions:
      - name: v1alpha1
        schema:
          openAPIV3Schema:
            type: object
            properties:
              apiVersion:
                type: string
              kind:
                type: string
              metadata:
                type: object
              spec:
                type: object
                properties:
                  real:
                    type: string
                  fake:
                    type: object
                    x-kubernetes-preserve-unknown-fields: true
                    properties:
              status:
                type: object
        served: true
        storage: true
        subresources:
          status: {}
    EOF
    ```

1. The same command as before should now fail since `spec.fake` is now defined as an `object`:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: example.com/v1alpha1
    kind: Fake
    metadata:
      name: fake
    spec:
      fake: fake all the way down!
    EOF
    ```

    and yep, it does:

    ```
    The Fake "fake" is invalid: spec.fake: Invalid value: "string": spec.fake in body must be of type object: "string"
    ```

1. It should also no longer be possible to set unknown fields inside `spec`:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: example.com/v1alpha1
    kind: Fake
    metadata:
      name: fake
    spec:
      who-are-you: you you!
    EOF
    ```

    Indeed, `spec.who-are-you` is a no-go:

    ```
    error: error validating "STDIN": error validating data: ValidationError(Fake.spec): unknown field "who-are-you" in com.example.v1alpha1.Fake.spec; if you choose to ignore these errors, turn validation off with --validate=false
    ```

1. However, it will _now_ be possible to set `spec.fake.really-fake` since `spec.fake` is an `object` that allows unknown fields:

    ```shell
    cat <<EOF | kubectl apply -f -
    apiVersion: example.com/v1alpha1
    kind: Fake
    metadata:
      name: fake
    spec:
      fake:
        really-fake: fugazi
    EOF
    ```

    ```
    fake.example.com/fake created
    ```

    Bingo, bango, bongo -- Bob's your uncle.


So I hope this shows that setting `spec.properties.configSpec.x-kubernetes-preserve-unknown-fields` to `true` is not going to result in the behavior I _think_ about which you were concerned. It will not be possible for vSphere to create VM Classes on older Supervisors with `spec.configSpec` assigned.

pkg/util/configspec.go
aruneshpa: 😆  I dropped this because I am pretty sure the linter has complained about incorrect spelling for this in my changes.  Also, the golang JSON package also spells it as marshaled/unmarshaled (https://pkg.go.dev/encoding/json, search for marshaled/unmarshaled).  But anyway, I will leave it to you if you feel strongly about this.

config/crd/bases/vmoperator.vmware.com_virtualmachineclasses.yaml
aruneshpa: > I keep coming back to this. I think you are concerned about x-kubernetes-preserve-unknown-fields: true and think it means a vSphere will create a VirtualMachineClass with a non-empty spec.ConfigSpec on a Supervisor with an older VM Operator API? If that's the case then I think we are okay.

Yes.  I was referring to this.  And yeah, I missed that that the `preserve-unknown-fields` is a property of the `ConfigSpec` and not the `VirtualMachineClassSpec`.  Your change makes sense.

aruneshpa (approved):
LGTM.

pkg/util/configspec.go
akutz: Ah, good catch about what Golang uses. You'd think if know given I was just in that code :)

akutz: I actually used `Marshaled` on the field definition :)

PR #39, opened 2022-12-19
Update MaxDeployThreads to be configurable

akutz:
This patch updates the way VM Operator handles `MaxDeployThreads`. The value used to be hardcoded as `80%` of the `MaxConcurrentReconciles`, which is `20`. However, `MaxConcurrentReconciles` is configurable via the VM Operator deployment, so `80%` of `20`, which is `16`, should increase or decrease if `MaxConcurrentReconciles` does as well. This patch now derives `MaxDeployThreads` by looking at the environment variable `MAX_CREATE_VMS_ON_PROVIDER` which is expected to be a whole number that represents the percentage of the `MaxConcurrentReconciles` that equals `MaxDeployThreads`. `MaxDeployThreads` is then calculated by `MaxConcurrentReconciles / (100 / MaxDeployThreadsEnvVarValue)`. If the env var is not set then it defaults to `80`.

### DIFF
--- a/pkg/vmprovider/providers/vsphere/vmprovider_vm.go
+++ b/pkg/vmprovider/providers/vsphere/vmprovider_vm.go
@@ -38,9 +38,8 @@ type vmUpdateArgs = session.VMUpdateArgs
 type vmMetadata = session.VMMetadata
 
 var (
-	createCountLock          sync.Mutex
-	concurrentCreateCount    int
-	maxConcurrentCreateCount = 16 // TODO: If we don't get to CL DeployVOF task stuff, add back some % from max reconciles
+	createCountLock       sync.Mutex
+	concurrentCreateCount int
 )
 
 func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachine(
@@ -215,7 +214,12 @@ func (vs *vSphereVMProvider) createVirtualMachine(
 
 	// BMV: This is about where we used to do this check but it prb make more sense
 	// to do earlier, as to limit wasted work.
-	allowed, createDeferFn := vs.vmCreateConcurrentAllowed(vmCtx)
+	maxDeployThreads, ok := vmCtx.Value(context.MaxDeployThreadsContextKey).(int)
+	if !ok {
+		return nil, fmt.Errorf("MaxDeployThreadsContextKey missing from context")
+	}
+
+	allowed, createDeferFn := vs.vmCreateConcurrentAllowed(vmCtx, maxDeployThreads)
 	if !allowed {
 		return nil, nil
 	}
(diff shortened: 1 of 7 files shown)


pkg/vmprovider/providers/vsphere/vmprovider_vm.go
```diff
@@ -215,7 +214,12 @@ func (vs *vSphereVMProvider) createVirtualMachine(
 
 	// BMV: This is about where we used to do this check but it prb make more sense
 	// to do earlier, as to limit wasted work.
-	allowed, createDeferFn := vs.vmCreateConcurrentAllowed(vmCtx)
+	maxDeployThreads, ok := vmCtx.Value(context.MaxDeployThreadsContextKey).(int)
+	if !ok {
+		return nil, fmt.Errorf("MaxDeployThreadsContextKey missing from context")
```
aruneshpa, line 219: I am wondering if it is okay to just add a warning and still allow the VM to be created if there's an error in parsing the key from the `Context`.  Blocking reconciles because we can't parse a non-essential field seems off.

On the other hand, I don't see a scenario where the context will not have the key.  I will leave this to your judgement.

aruneshpa (approved):

pkg/vmprovider/providers/vsphere/vmprovider_vm.go
akutz: The reason for this is because, as you said, there should not be a case where it's missing. I honestly almost opted for a panic instead to ensure there's never a situation where it's missing. The logic has a default associated with it, so it should be fine. Plus, as BV mentioned, this will go away sooner than later once I or BV implements the task-based reconciles.

akutz: Oh, and also, the other reason was how the value is calculated. I elected to do the calculation at the point of injection instead of at use. Thus at this point we don't want to be calling the `lib` package to get the config. If anything I'll switch this to a panic, but the point is it shouldn't be defaulting here, but where it's injected.

akutz: This is also a side-effect of the provider model that tried to divorce the Kubernetes side of things from the backend when it's probably an unnecessary abstraciton. Once we reign that in a bit, we can start to move config back closer to implementation.

akutz: As it is, I don't see a sane way to inject a default here that makes sense. As we both said, it should always be injected at this point, and I think we rely on that than defaulting to something that could easily mask if the actual, configured value _isn't_ being passed through correctly.

PR #42, opened 2023-01-05
Update VMClass.Spec.ConfigSpec to be RawMessage

akutz:
This patch updates the `VirtualMachineClass.Spec.ConfigSpec` to be `json.RawMessage`, which is similar to `runtime.RawExtension`, but has no opinion about being a `runtime.Object`.
    
Per https://github.com/kubernetes-sigs/controller-tools/issues/637, it was necessary to decorate the field with the following kubebuilder annotations to force the wire protocol to be an object:

```go
// +kubebuilder:validation:Schemaless
// +kubebuilder:validation:Type=object
// +kubebuilder:pruning:PreserveUnknownFields
```

### DIFF
--- a/docs/user-guide/apis/v1alpha1.md
+++ b/docs/user-guide/apis/v1alpha1.md
@@ -679,7 +679,7 @@ _Appears in:_
 | `hardware` _[VirtualMachineClassHardware](#virtualmachineclasshardware)_ | Hardware describes the configuration of the VirtualMachineClass attributes related to virtual hardware.  The configuration specified in this field is used to customize the virtual hardware characteristics of any VirtualMachine associated with this VirtualMachineClass. |
 | `policies` _[VirtualMachineClassPolicies](#virtualmachineclasspolicies)_ | Policies describes the configuration of the VirtualMachineClass attributes related to virtual infrastructure policy.  The configuration specified in this field is used to customize various policies related to infrastructure resource consumption. |
 | `description` _string_ | Description describes the configuration of the VirtualMachineClass which is not related to virtual hardware or infrastructure policy. This field is used to address remaining specs about this VirtualMachineClass. |
-| `configSpec` _RawExtension_ | ConfigSpec describes additional configuration information for a VirtualMachine. 
+| `configSpec` _integer array_ | ConfigSpec describes additional configuration information for a VirtualMachine. 
  The contents of this field are the VirtualMachineConfigSpec data object (https://bit.ly/3HDtiRu) marshaled to JSON using the discriminator field "_typeName" to preserve type information. |
 
 
--- a/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
+++ b/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package vsphere_test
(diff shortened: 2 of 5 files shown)


docs/user-guide/apis/v1alpha1.md
```diff
@@ -679,7 +679,7 @@ _Appears in:_
 | `hardware` _[VirtualMachineClassHardware](#virtualmachineclasshardware)_ | Hardware describes the configuration of the VirtualMachineClass attributes related to virtual hardware.  The configuration specified in this field is used to customize the virtual hardware characteristics of any VirtualMachine associated with this VirtualMachineClass. |
 | `policies` _[VirtualMachineClassPolicies](#virtualmachineclasspolicies)_ | Policies describes the configuration of the VirtualMachineClass attributes related to virtual infrastructure policy.  The configuration specified in this field is used to customize various policies related to infrastructure resource consumption. |
 | `description` _string_ | Description describes the configuration of the VirtualMachineClass which is not related to virtual hardware or infrastructure policy. This field is used to address remaining specs about this VirtualMachineClass. |
-| `configSpec` _RawExtension_ | ConfigSpec describes additional configuration information for a VirtualMachine. 
+| `configSpec` _integer array_ | ConfigSpec describes additional configuration information for a VirtualMachine.
```
aruneshpa, line 682: Does this need to be updated?

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -168,9 +167,7 @@ func vmTests() {
 					// Update the VM Class with the XML.
 					vmClass := &vmopv1alpha1.VirtualMachineClass{}
 					Expect(ctx.Client.Get(ctx, client.ObjectKey{Name: vm.Spec.ClassName}, vmClass)).To(Succeed())
-					vmClass.Spec.ConfigSpec = &runtime.RawExtension{
-						Raw: w.Bytes(),
-					}
+					vmClass.Spec.ConfigSpec = w.Bytes()
```
aruneshpa, line 170: Would `json.RawMessage(obj_data)` be more readable and explicit here?

aruneshpa (approved):
LGTM with two minor comments.

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
akutz: The type conversion is implicit so this is unecessary.

docs/user-guide/apis/v1alpha1.md
akutz: It was. That's how it got generated. If it wasn't updated then the GH action linter check wouldn't pass. The API generator clearly has an issue. Anyway, there's never any need to comment on generated assets as the GH action that validates generated assets are up-to-date would fail if they were not updated.

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
akutz: > If `type T K` then you can `var t T; var k K; t = k`.

The above assertion doesn't not work for non-slice primitive types. Will dig into this more, but ([Go playground](https://go.dev/play/p/N98cmmW5Tuq):

```go
package main

import (
	"fmt"
)

type Int int

func main() {
	var I Int = 1
	var i int = I
	fmt.Println(i)
}
```

does not work, and will emit:

```
./prog.go:11:14: cannot use I (variable of type Int) as type int in variable declaration

Go build failed.
```

While ([Go playground](https://go.dev/play/p/LsTxA33zd6k)) _does_ work:

```go
package main

import (
	"fmt"
)

type SliceOfInts []int

func main() {
	var ints []int
	var sois SliceOfInts = SliceOfInts{1, 2, 3}
	ints = sois
	fmt.Println(ints)
}
```

and emits:

```
[1 2 3]

Program exited.
```

PR #45, opened 2023-01-17
Use APIReader client to Get() VirtualMachinePublishRequest

bryanv:
Unlike ConfigMaps and Secrets which we cannot cache because their quantity and size will exceed our container's memory limit, the pub req shouldn't be cached because of how the controller is implemented so make that explicit in the controller.


yi0909 (approved):

dilyar85 (approved):
LGTM.

PR #47, opened 2023-01-18
Implement caching for download of virtual machine images

dilyar85:
This patch introduces an in-memory cache to store and manage downloaded OVF for virtual machine images. It also cleans up stale cached items after a specified time (currently set to 30 minutes). This avoids unnecessary requests to vCenter when the same content library is associated with multiple namespaces within a cluster and speeds up vm image reconciliation time.

Testing Done:
- Internal kind-based e2e tests (both fast and full) passed
- Manually deployed this patch in a WCP testbed (with Image-Registry FSS enabled) and verified the expected workflow by associating a content library to multiple namespaces simultaneously. Logs from VM-Operator (in verbose mode):
```bash
I0207 09:15:21.266601       1 vmprovider.go:248] vsphere "msg"="Cache item miss, downloading OVF from vCenter" "contentVersion"="2" "itemID"="732f3e46-4699-4b66-aa6c-2740b0bfe6ed"
...
I0207 09:15:29.134953       1 vmprovider.go:264] vsphere "msg"="Cache item put" "contentVersion"="2" "itemID"="732f3e46-4699-4b66-aa6c-2740b0bfe6ed" "putResult"=1
I0207 09:15:29.135280       1 vmprovider.go:246] vsphere "msg"="Cache item hit, using cached OVF" "contentVersion"="2" "itemID"="732f3e46-4699-4b66-aa6c-2740b0bfe6ed"
I0207 09:15:29.136055       1 vmprovider.go:246] vsphere "msg"="Cache item hit, using cached OVF" "contentVersion"="2" "itemID"="732f3e46-4699-4b66-aa6c-2740b0bfe6ed"
```

### DIFF
--- a/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
+++ b/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package clustercontentlibraryitem
@@ -260,7 +260,7 @@ func (r *Reconciler) syncImageContent(ctx goctx.Context,
 		return nil
 	}
 
-	err := r.VMProvider.SyncVirtualMachineImage(ctx, cclItem.Spec.UUID, cvmi)
+	err := r.VMProvider.SyncVirtualMachineImage(ctx, cclItem, cvmi)
 	if err != nil {
 		conditions.MarkFalse(cvmi,
 			vmopv1a1.VirtualMachineImageSyncedCondition,
--- a/pkg/util/cache.go
+++ b/pkg/util/cache.go
@@ -1,0 +1,183 @@
+// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package util
+
+import (
+	"sync"
+	"time"
+)
+
+// CacheItem wraps T so it has a LastUpdated field and can be used with Cache.
+type CacheItem[T any] struct {
+	// Item is the cached item.
+	Item T
+
+	// LastUpdated is the last time the item was updated.
+	LastUpdated time.Time
+}
+
+// Cache is a generic implementation of a cache that can be configured with a
+// maximum number of items and can evict items after a certain amount of time.
+type Cache[T any] struct {
+	mu       sync.Mutex
+	items    map[string]CacheItem[T]
+	maxItems int
+	done     chan struct{}
+	donce    sync.Once
+
+	// expiredChan returns a channel on which the IDs of items that are expired
+	// are sent. This channel is closed when the cache is closed, but only after
+	// all, pending eviction notices are received.
+	expiredChan chan string
+	expiringNum sync.WaitGroup
+}
+
+// NewCache initializes a new cache with the provided expiration options.
+func NewCache[T any](
+	expireAfter, checkExpireInterval time.Duration,
+	maxItems int) *Cache[T] {
+
+	c := &Cache[T]{
+		expiredChan: make(chan string),
+		done:        make(chan struct{}),
+		items:       map[string]CacheItem[T]{},
+		maxItems:    maxItems,
+	}
+
+	// Start a goroutine to expire items from the cache.
+	go func() {
+		ticker := time.NewTicker(checkExpireInterval)
+
+		// evictFn iterates over the cached items and evicts expired items
+		evictFn := func() {
+			c.mu.Lock()
+			defer c.mu.Unlock()
+			for k, v := range c.items {
+				if time.Since(v.LastUpdated) > expireAfter {
+					// Delete the item from the cache.
+					delete(c.items, k)
+
+					// Notify subscribers that the item with the ID of k has
+					// been evicted if the cache is not closed. Otherwise we
+					// add one to the number of items being expired. This
+					// ensures the channel cannot be closed until all, pending
+					// notifications are sent.
+					select {
+					case <-c.done:
+					default:
+						c.expiringNum.Add(1)
+						go func(k string) {
+							c.expiredChan <- k
+							c.expiringNum.Done()
+						}(k)
+					}
+				}
+			}
+		}
+		for {
+			select {
+			case <-c.done:
+				ticker.Stop()
+				return
+			case <-ticker.C:
+				evictFn()
+			}
+		}
+	}()
+
+	return c
+}
+
+// ExpiredChan returns a channel on which the IDs of items that are expired are
+// sent. This channel is closed when the cache is closed, but only after all,
+// pending eviction notices are received.
+func (c *Cache[T]) ExpiredChan() <-chan string {
+	return c.expiredChan
+}
+
+// Close shuts down the cache. It is safe to call this function more than once.
+func (c *Cache[T]) Close() {
+	c.donce.Do(func() {
+		close(c.done)
+
+		// Wait until all pending, expiry notifications are received until the
+		// expired channel is closed.
+		go func() {
+			c.expiringNum.Wait()
+			close(c.expiredChan)
+		}()
+	})
+}
+
+// Delete removes the specified item from the cache.
+func (c *Cache[T]) Delete(id string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	delete(c.items, id)
+}
+
+// Get returns the cached item with the provided ID. The function isHit may be
+// optionally provided to further determine whether a cached item should hit
+// or miss.
+func (c *Cache[T]) Get(id string, isHit func(t T) bool) (T, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var emptyT T
+
+	v, ok := c.items[id]
+	if !ok {
+		return emptyT, false
+	}
+
+	if isHit != nil && !isHit(v.Item) {
+		return emptyT, false
+	}
+
+	return v.Item, true
+}
+
+// CachePutResult describes the result of putting an item into the cache.
+type CachePutResult uint
+
+const (
+	// CachePutResultCreate indicates the item was created in the cache.
+	CachePutResultCreate CachePutResult = iota + 1
+
+	// CachePutResultUpdate indicates an existing, cached item was updated.
+	CachePutResultUpdate
+
+	// CachePutResultMaxItemsExceeded indicates the item was not stored because
+	// the cache's maximum number of items would have been exceeded.
+	CachePutResultMaxItemsExceeded
+)
+
+// Put stores the provided item in the cache, updating existing items.
+func (c *Cache[T]) Put(id string, t T) CachePutResult {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	v, ok := c.items[id]
+
+	if ok {
+		// Update existing item.
+		v.Item = t
+		v.LastUpdated = time.Now()
+		c.items[id] = v
+		return CachePutResultUpdate
+	}
+
+	// Store the new item in the cache as long as the maximum number of items
+	// was note exceeded.
+	if len(c.items) < c.maxItems {
+		// Store new item in the cache.
+		c.items[id] = CacheItem[T]{
+			Item:        t,
+			LastUpdated: time.Now(),
+		}
+		return CachePutResultCreate
+	}
+
+	return CachePutResultMaxItemsExceeded
+}
--- a/pkg/util/lock_pool.go
+++ b/pkg/util/lock_pool.go
@@ -1,0 +1,47 @@
+// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package util
+
+import (
+	"reflect"
+	"sync"
+)
+
+// LockPool is a synchronized set of maps that can be obtained with a provided
+// ID.
+type LockPool[K any, V sync.Locker] struct {
+	ool sync.Map
+}
+
+// Get returns the lock for the provided ID.
+func (p *LockPool[K, V]) Get(id K) sync.Locker {
+	if obj, ok := p.ool.Load(id); ok {
+		return obj.(V)
+	}
+
+	var (
+		v          V
+		objToStore any
+	)
+
+	tv := reflect.TypeOf(v)
+	if tv.Kind() == reflect.Pointer {
+		// Instantiate the type to which V points, ex. if V is a *sync.Mutex
+		// then we want to instantiate a sync.Mutex and use its address.
+		objToStore = reflect.New(tv.Elem()).Interface()
+	}
+
+	// If objToStore does not implement V then &objToStore will.
+	if _, ok := objToStore.(V); !ok {
+		objToStore = &objToStore
+	}
+
+	obj, _ := p.ool.LoadOrStore(id, objToStore)
+	return obj.(V)
+}
+
+// Delete removes the specified lock from the pool.
+func (p *LockPool[K, V]) Delete(id K) {
+	p.ool.Delete(id)
+}
--- a/pkg/vmprovider/providers/vsphere/contentlibrary/content_library_provider.go
+++ b/pkg/vmprovider/providers/vsphere/contentlibrary/content_library_provider.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2019-2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package contentlibrary
@@ -13,8 +13,6 @@ import (
 
 	k8serrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/utils/pointer"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -36,7 +34,7 @@ type Provider interface {
 	ListLibraryItems(ctx context.Context, libraryUUID string) ([]string, error)
 	UpdateLibraryItem(ctx context.Context, itemID, newName string, newDescription *string) error
 	RetrieveOvfEnvelopeFromLibraryItem(ctx context.Context, item *library.Item) (*ovf.Envelope, error)
-	SyncVirtualMachineImage(ctx context.Context, itemID string, vmi client.Object) error
+	RetrieveOvfEnvelopeByLibraryItemID(ctx context.Context, itemID string) (*ovf.Envelope, error)
 
 	// TODO: Testing only. Remove these from this file.
 	CreateLibraryItem(ctx context.Context, libraryItem library.Item, path string) error
@@ -146,6 +144,22 @@ func (cs *provider) GetLibraryItem(ctx context.Context, libraryUUID, itemName st
 	}
 
 	return item, nil
+}
+
+// RetrieveOvfEnvelopeByLibraryItemID retrieves the OVF Envelope by the given library item ID.
+func (cs *provider) RetrieveOvfEnvelopeByLibraryItemID(ctx context.Context, itemID string) (*ovf.Envelope, error) {
+	libItem, err := cs.libMgr.GetLibraryItem(ctx, itemID)
+	if err != nil {
+		return nil, err
+	}
+
+	if libItem == nil || libItem.Type != library.ItemTypeOVF {
+		log.Error(nil, "empty or non OVF library item type, skipping", "itemID", itemID)
+		// No need to return the error here to avoid unnecessary reconciliation.
+		return nil, nil
+	}
+
+	return cs.RetrieveOvfEnvelopeFromLibraryItem(ctx, libItem)
 }
 
 // RetrieveOvfEnvelopeFromLibraryItem downloads the supported file from content library.
@@ -397,62 +411,4 @@ func (cs *provider) generateDownloadURLForLibraryItem(
 	}
 
 	return url.Parse(fileURL)
-}
-
-func (cs *provider) SyncVirtualMachineImage(ctx context.Context, itemID string, vmi client.Object) error {
-	logger := log.WithValues("itemID", itemID)
-	item, err := cs.libMgr.GetLibraryItem(ctx, itemID)
-	if err != nil {
-		return err
-	}
-
-	if item.Type != library.ItemTypeOVF {
-		logger.Info("Skip syncing up non-OVF library item for VirtualMachineImage/ClusterVirtualMachineImage",
-			"vmiName", vmi.GetName(), "itemType", item.Type)
-		return nil
-	}
-
-	ovfEnvelope, err := cs.RetrieveOvfEnvelopeFromLibraryItem(ctx, item)
-	if err != nil {
-		logger.Error(err, "error extracting the OVF envelope from the library item", "itemName", item.Name)
-		return err
-	}
-	if ovfEnvelope == nil {
-		logger.Error(err, "no valid OVF envelope found, skipping library item", "itemName", item.Name)
-		return nil
-	}
-
-	var spec *v1alpha1.VirtualMachineImageSpec
-	var status *v1alpha1.VirtualMachineImageStatus
-	switch vmi := vmi.(type) {
-	case *v1alpha1.VirtualMachineImage:
-		spec = &vmi.Spec
-		status = &vmi.Status
-	case *v1alpha1.ClusterVirtualMachineImage:
-		spec = &vmi.Spec
-		status = &vmi.Status
-	}
-
-	if ovfEnvelope.VirtualSystem != nil {
-		updateImageSpecWithOvfVirtualSystem(spec, ovfEnvelope.VirtualSystem)
-
-		ovfSystemProps := getVmwareSystemPropertiesFromOvf(ovfEnvelope.VirtualSystem)
-		annotations := vmi.GetAnnotations()
-		if annotations == nil {
-			annotations = make(map[string]string)
-			vmi.SetAnnotations(annotations)
-		}
-		for k, v := range ovfSystemProps {
-			annotations[k] = v
-		}
-		// Set Status.ImageSupported to combined compatibility of OVF compatibility or WCP_UNIFIED_TKG FSS state.
-		status.ImageSupported = pointer.BoolPtr(isImageSupported(vmi, ovfEnvelope.VirtualSystem, ovfSystemProps))
-
-		// Set Status Firmware from the envelope's virtual hardware section
-		if virtualHwSection := ovfEnvelope.VirtualSystem.VirtualHardware; len(virtualHwSection) > 0 {
-			status.Firmware = getFirmwareType(virtualHwSection[0])
-		}
-	}
-
-	return nil
 }
--- a/pkg/vmprovider/providers/vsphere/contentlibrary/content_library_utils.go
+++ b/pkg/vmprovider/providers/vsphere/contentlibrary/content_library_utils.go
@@ -103,6 +103,41 @@ func LibItemToVirtualMachineImage(
 	return image
 }
 
+// UpdateVmiWithOvfEnvelope updates the given vmi object with the content of given OVF envelope.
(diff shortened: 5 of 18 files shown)


akutz:
Instead of making the cache part of the provider, you could create a context key for the cache and store/retrieve the cache via the context. It would enable better testing as well as remove the need to pass around the cache everywhere outside of access to the provider struct.

image

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_utils.go
```diff
@@ -103,6 +103,41 @@ func LibItemToVirtualMachineImage(
...
+// UpdateVmiWithOvfEnvelope updates the given vmi object with the content of given OVF envelope.
+func UpdateVmiWithOvfEnvelope(vmi client.Object, ovfEnvelope ovf.Envelope) {
+	var spec *v1alpha1.VirtualMachineImageSpec
+	var status *v1alpha1.VirtualMachineImageStatus
+	switch vmi := vmi.(type) {
+	case *v1alpha1.VirtualMachineImage:
+		spec = &vmi.Spec
+		status = &vmi.Status
+	case *v1alpha1.ClusterVirtualMachineImage:
+		spec = &vmi.Spec
```
zyiyi11, line 115: what is the point of switching cases when setting it to the same value? Are you trying to guard against other types than vmi/cvmi?

zyiyi11 (commented):
LGTM

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_utils.go
akutz: Because he's casting it, and if you combine them into a single case, it won't let you access the fields of that type.

dilyar85: What Andrew said :) Btw, this is from the existing code and I am just moving it to a separate function in the `content_library_utils.go` file (as it doesn't contain any dependency on the client).

dilyar85:
> Instead of making the cache part of the provider, you could create a context key for the cache and store/retrieve the cache via the context. It would enable better testing as well as remove the need to pass around the cache everywhere outside of access to the provider struct.

@akutz As previously discussed offline, I am going to keep the current implementation since it doesn't introduce any global package variables. Could you please take another look when you get a chance? Thank you.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/vmware/govmomi/ovf"
+)
+
+type Cache struct {
+	mu    sync.Mutex
+	items map[string]*cacheItem
+}
+
+type cacheItem struct {
+	ovfEnvelope    ovf.Envelope
```
ericvmw, line 19: Consider setting a max number of items that can be cached, potentially the cache might take some good memory space. If there are 1000 items with 100kb OVF descriptor in the cluster, then the cache may take 100mb if they may be cached around the same time period.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
```diff
@@ -0,0 +1,76 @@
...
+
+// GetOvfEnvelope returns the ovf envelope if found in the cache and updates the last access time.
+func (c *Cache) GetOvfEnvelope(id, contentVersion string) (ovf.Envelope, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	logger := log.WithValues("itemID", id, "contentVersion", contentVersion)
+	item := c.items[id]
+	if item == nil || item.contentVersion != contentVersion {
+		logger.V(5).Info("ovf envelope cache miss")
```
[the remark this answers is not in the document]
dilyar85: Added a separate block to avoid referencing it when `item == nil`.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
dilyar85: That's a fair point. I applied a max size and set it to 100 in the new patch.

pkg/vmprovider/providers/vsphere/vmprovider.go
```diff
@@ -179,17 +186,30 @@ func (vs *vSphereVMProvider) GetVirtualMachineImageFromContentLibrary(
...
+		if err != nil {
+			return err
+		}
+
+		ovfEnvelope, err = client.ContentLibClient().RetrieveOvfEnvelopeFromLibraryItemID(ctx, itemID)
+		if err != nil {
+			return err
+		}
+
+		vs.clCache.SetOvfEnvelope(itemID, contentVersion, ovfEnvelope)
```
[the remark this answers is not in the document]
dilyar85: Thanks for pointing this out. I updated it to store the content library version directly from fetched item in the new patch.

bryanv (commented):
How effective is this cache on pod restart? Is this mostly effective when creating a new namespace with CL already for other namespaces? When there are more items than `maxSize` and during the re-sync, I'm not sure if updating `lastAccessTime` is really cache friend to how we'll access the cache.

Is there any practical benefit of caching on behalf of cluster scoped images? Sharing a CL between cluster and namespace scoped doesn't seem like a likely scenario, if not disallowed in VC.

Ultimately I think we need a friendlier API from CL, especially for when a CL is shared among multiple clusters.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
ericvmw: CL validates OVF descriptor to ensure less than 4mb per OVF spec, most OVF template size should be way less than that. Tracking the total size of all cache entries can be also helpful, so it can have the cache bounded by some fixed or configurable memory space as well.

pkg/vmprovider/providers/vsphere/vmprovider.go
```diff
@@ -179,17 +189,48 @@ func (vs *vSphereVMProvider) GetVirtualMachineImageFromContentLibrary(
...
+		if err != nil {
+			return err
+		}
+
+		item, err := client.ContentLibClient().GetLibraryItemByItemID(ctx, itemID)
+		if err != nil {
+			return err
+		}
+
+		if item != nil && item.Type != library.ItemTypeOVF {
```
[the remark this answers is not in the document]
dilyar85: Fair point. Changed it to `if item == nil || item.Type != library.ItemTypeOVF` with the error message updated.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
```diff
@@ -0,0 +1,89 @@
...
+func NewCache(itemExpiration, expirationCheckInterval time.Duration, maxSize int) *Cache {
+	c := &Cache{items: make(map[string]*itemCache), maxSize: maxSize}
+
+	// The following go routine cleans up expired cache items at the given check interval.
+	go func() {
+		itemExpirationSec := int64(itemExpiration.Seconds())
+		for range time.Tick(expirationCheckInterval) {
+			c.mu.Lock()
+			for key, item := range c.items {
+				if time.Now().Unix()-item.lastAccessTime > itemExpirationSec {
```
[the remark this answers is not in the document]
dilyar85: This is just for removing items that have been cached for a while (longer than the itemExpirationSec) so that we can free some space in the cache. Otherwise, there is no way to add more items after it reaches max size.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
dilyar85: In the worst case then, the cache can grow up to 4MB * 100 = 400MB which is around 381MiB, less than the current 500MiB limit on the container. I checked the TKG CL and it shows as 4.23 MB in total. In case we encounter any OOM from this cache, we can configure a smaller max size and shorter expiration time to clean up items more frequently.

akutz:
> How effective is this cache on pod restart? Is this mostly effective when creating a new namespace with CL already for other namespaces? When there are more items than `maxSize` and during the re-sync, I'm not sure if updating `lastAccessTime` is really cache friend to how we'll access the cache.
> 
> Is there any practical benefit of caching on behalf of cluster scoped images? Sharing a CL between cluster and namespace scoped doesn't seem like a likely scenario, if not disallowed in VC.
> 
> Ultimately I think we need a friendlier API from CL, especially for when a CL is shared among multiple clusters.

@dilyar85 , Please respond to @bryanv here. I know we've discussed answers to some of his questions.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -260,7 +260,7 @@ func (r *Reconciler) syncImageContent(ctx goctx.Context,
 		return nil
 	}
 
-	err := r.VMProvider.SyncVirtualMachineImage(ctx, cclItem.Spec.UUID, cvmi)
+	err := r.VMProvider.SyncVirtualMachineImage(ctx, cclItem.Spec.UUID, latestVersion, cvmi)
```
akutz, line 263: The moment we've added the 3rd param we should decide if an options struct is worth it. For example:

```go
type SyncVirtualMachineImageOpts struct {
  ContentLibraryItemID string
  LatestVersion        string
  VirtualMachineImage  client.Object
}
```

Something like that. And to be clear, the `context` should never go in an options struct.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
```
akutz, line 1: Big comment incoming on this.

akutz (changes requested):

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
```diff
@@ -0,0 +1,94 @@
...
+	c := &Cache{items: make(map[string]*itemCache), maxSize: maxSize}
+
+	// The following go routine cleans up expired cache items at the given check interval.
+	go func() {
+		itemExpirationSec := int64(itemExpiration.Seconds())
+		for range time.Tick(expirationCheckInterval) {
+			c.mu.Lock()
+			for key, item := range c.items {
+				if time.Now().Unix()-item.lastUpdatedTime > itemExpirationSec {
+					log.V(5).Info("Deleting expired cache item", "itemID", key)
```
akutz, line 37: Don't log here. Having a logger in a goroutine like this is a good way to hit an NPE when the system is shutting down or something else you didn't expect, this Goroutine is still running, and the logger is no longer initialized enough to work correctly. I'll show you a better way in a moment with the refactor I'm providing.

akutz (changes requested):
Hi @dilyar85 ,

Thank you for this. However, please refactor your cache to use this generic implementation I whipped up -- https://github.com/akutz/vm-operator/blob/feature/generic-cache/pkg/util/cache.go. I'm not gonig to submit this as a PR, you can just take it for your PR. Please note there's a series of simple tests too, but please feel free to expand on them -- https://github.com/akutz/vm-operator/blob/feature/generic-cache/pkg/util/cache_test.go.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
akutz: See this comment -- https://github.com/vmware-tanzu/vm-operator/pull/47#pullrequestreview-1277855425

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
bryanv: I wouldn't expect that to leave sufficient for the pod to not OOM.

It does seem that eventually that a combination of large OVF image(s) and/or mapped to many namespaces, we'll be forced to do MD-only watches and disable the ctrl-runtime client cache for NS-scoped image. That said, at present, if this cache doesn't contain cluster-scoped images, it is more or less another namespace worth of images.

bryanv:
> > How effective is this cache on pod restart? Is this mostly effective when creating a new namespace with CL already for other namespaces? When there are more items than `maxSize` and during the re-sync, I'm not sure if updating `lastAccessTime` is really cache friend to how we'll access the cache.
> 
> @bryanv The purpose of implementing this cache is to avoid downloading duplicate library item when same CL is attached to multiple namespaces in a cluster. Agree with your suggestion on `lastAccessTime`. I have updated the `get` function to avoid changing the item time (so it's pure read-only) and only the `set` function will touch it now (when creating or updating the existing item).

In the steady state, won't this cache be empty the controllers will noop when the ContentVersion matches, and the cache 
expires items 30 minutes after they're touched, correct? For attachments to multiple namespaces, what order reconcile request ultimately end up in the controller's queue seems to dictate effectiveness: if the items from different namespaces are interweaved, an item in one namespace may not get a cache hit because it is still being downloaded for the same CL item in another namespace.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
dilyar85: After a second thought, I think it might be better to make the cl/ccl-item itself as a param here so that we won't need to add more params or update the option struct if needed later.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
dilyar85: I actually thought about making this cache generic or using a 3rd party cache library, however, I made it specific to the content-library as I didn't see any obvious usecase for storing other objects atm. I am taking a deeper look at your implementation above and will try adopting it in this PR. Thanks again for sharing it.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
dilyar85: Ah, I didn't realize that, thanks for pointing it out.

dilyar85:
> In the steady state, won't this cache be empty the controllers will noop when the ContentVersion matches, and the cache expires items 30 minutes after they're touched, correct? For attachments to multiple namespaces, what order reconcile request ultimately end up in the controller's queue seems to dictate effectiveness: if the items from different namespaces are interweaved, an item in one namespace may not get a cache hit because it is still being downloaded for the same CL item in another namespace.

@bryanv Yes, you got it right on the steady state. The current implementation does need reconciliation requests to come in the right order for it to work effectively. But I believe we can improve this. Instead of having a lock on the whole cache, what about putting a lock on each cache item?

Here's how I am thinking it would work: let's say there are multiple namespaces tied to the same CL. This means multiple reconciliation requests will come in to sync the VMIs, some of which may involve downloading the same OVF file. If the first request results in a cache miss, that specific cache item will be locked while the OVF is being downloaded and saved in the cache. This way, other requests for that same OVF will be put on hold until the lock is released, avoiding cache misses during the OVF download. 

Let me know if this approach sounds viable to you, and if so I'll implement this change in Andrew's generic cache example above. Thanks.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
akutz: It doesn't matter if you need to update the option struct. The point of the option struct is to make it so the function signature doesn't have to change and break things. Put the CL/CCL item into the option struct.

akutz:
> > How effective is this cache on pod restart? Is this mostly effective when creating a new namespace with CL already for other namespaces? When there are more items than `maxSize` and during the re-sync, I'm not sure if updating `lastAccessTime` is really cache friend to how we'll access the cache.
> 
> @bryanv The purpose of implementing this cache is to avoid downloading duplicate library item when same CL is attached to multiple namespaces in a cluster. Agree with your suggestion on `lastAccessTime`. I have updated the `get` function to avoid changing the item time (so it's pure read-only) and only the `set` function will touch it now (when creating or updating the existing item).

There's a reason in the generic example that I changed `maxSize` to `maxItems`. The name `maxSize` is misleading and implies content size, not number of items.

akutz:
> > In the steady state, won't this cache be empty the controllers will noop when the ContentVersion matches, and the cache expires items 30 minutes after they're touched, correct? For attachments to multiple namespaces, what order reconcile request ultimately end up in the controller's queue seems to dictate effectiveness: if the items from different namespaces are interweaved, an item in one namespace may not get a cache hit because it is still being downloaded for the same CL item in another namespace.
> 
> @bryanv Yes, you got it right on the steady state. The current implementation does need reconciliation requests to come in the right order for it to work effectively. But I believe we can improve this. Instead of having a lock on the whole cache, what about putting a lock on each cache item?
> 
> Here's how I am thinking it would work: let's say there are multiple namespaces tied to the same CL. This means multiple reconciliation requests will come in to sync the VMIs, some of which may involve downloading the same OVF file. If the first request results in a cache miss, that specific cache item will be locked while the OVF is being downloaded and saved in the cache. This way, other requests for that same OVF will be put on hold until the lock is released, avoiding cache misses during the OVF download.
> 
> Let me know if this approach sounds viable to you, and if so I'll implement this change in Andrew's generic cache example above. Thanks.

So what you need is a `sync.Map` that follows the pattern of `LoadOrStore`. For example, please see what I just pushed to:

* https://github.com/akutz/vm-operator/blob/feature/generic-cache/pkg/util/lock_pool.go
* https://github.com/akutz/vm-operator/blob/feature/generic-cache/pkg/util/lock_pool_test.go

These make it super easy to create and manage the locks necessary. You could easily add a TTL to them as well if you wanted.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_provider.go
```diff
@@ -382,60 +380,29 @@ func (cs *provider) generateDownloadURLForLibraryItem(
...
+	logger := log.WithValues("itemID", itemID, "itemName", item.Name)
 	if item.Type != library.ItemTypeOVF {
-		logger.Info("Skip syncing up non-OVF library item for VirtualMachineImage/ClusterVirtualMachineImage",
-			"vmiName", vmi.GetName(), "itemType", item.Type)
-		return nil
+		logger.Info("Skip fetching non-OVF library item", "itemType", item.Type)
+		return ovf.Envelope{}, nil
 	}
 
 	ovfEnvelope, err := cs.RetrieveOvfEnvelopeFromLibraryItem(ctx, item)
```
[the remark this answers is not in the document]
dilyar85: @ericvmw I looked into it but the implementation would require a wider test framework change in the content library provider code. As this patch is focusing on the image cache I would address it hopefully in a future PR.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
dilyar85: I ended up having just clitem and vmi (both in `ctrlruntime.Object` type) for this method's params. I don't expect it would change again given that all the sync info would be included in these objects.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
dilyar85: Updated this PR with your generic type cache included.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_cache.go
dilyar85: Resolving this conversation as the `content_lirbary_cache` is replaced with the generic type cache.

pkg/util/lock_pool.go
```diff
@@ -0,0 +1,47 @@
...
+	if _, ok := objToStore.(V); !ok {
+		objToStore = &objToStore
+	}
+
+	obj, _ := p.ool.LoadOrStore(id, objToStore)
+	return obj.(V)
+}
+
+// Delete removes the specified lock from the pool.
+func (p *LockPool[K, V]) Delete(id K) {
```
bryanv, line 45: Maybe I'm missing it but is this being called anywhere?

pkg/util/cache.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}()
+
+	return c
+}
+
+// ExpiredChan returns a channel on which the IDs of items that are expired are
+// sent. This channel is closed when the cache is closed, but only after all,
+// pending eviction notices are received.
+func (c *Cache[T]) ExpiredChan() <-chan string {
```
bryanv, line 95: This is only called in the tests, right? Does that mean that the send done as part of the expire is going to block because there is nothing reading from the channel?

bryanv:
In an earlier comment, I asked if this made sense for cluster scoped VMI but I didn't see that addressed or answered. Looks like this still caches all images. Sorry if I missed the response in the comments.

I'm still a little concerned about effectiveness and the visible behavior around this, ie, from a CU perspective, why is the behavior - when looking at the VC UI - different when creating a second NS 15 vs 45 min after the first NS? In most env, NS churn is not likely terribly common but I think that means the cache will be empty in the common case anyways b/c of the 30 min expiration?

IDK what the lift is for CL to provide this but I infer that they're cracking & caching part of the OVF based what's shown in the UI.

Don't treat any of my comments necessarily as blockers but I hope there is a longer term CL solution planned. Maybe an intermediate solution, implemented in VMOP, would be to set aside a separate vmware-system- NS, and use k8s as a durable cache, copying the VMIs from there (tracking CL item deletion vs CL un-association vs CL deletion would be at least a moderate PITA).

pkg/util/lock_pool.go
```diff
@@ -0,0 +1,47 @@
...
+package util
+
+import (
+	"reflect"
+	"sync"
+)
+
+// LockPool is a synchronized set of maps that can be obtained with a provided
+// ID.
+type LockPool[K any, V sync.Locker] struct {
```
ericvmw, line 13: Does this pool need to be maintained outside of cache? the key is the same, and it looks the lifecycle of each lock pool entry needs to be in sync with the cache item, i.e. when item is cleaned, it needs to be removed from the map here.

pkg/util/lock_pool.go
dilyar85: I pushed a new commit that will delete the specific OVF item lock when it's expired in the cache, so this function is being used now (from `InitOvfCacheAndLockPool` within vmprovider.go file).

pkg/util/cache.go
dilyar85: I don't think it's blocking anything as the expire send is always done in a goroutine. With that being said, the newly added `InitOvfCacheAndLockPool` will now read from this channel to clean up the corresponding lock item.

pkg/util/cache.go
```diff
@@ -0,0 +1,183 @@
...
+// Cache is a generic implementation of a cache that can be configured with a
+// maximum number of items and can evict items after a certain amount of time.
+type Cache[T any] struct {
+	mu       sync.Mutex
+	items    map[string]CacheItem[T]
+	maxItems int
+	done     chan struct{}
+	donce    sync.Once
+
+	// expiredChan returns a channel on which the IDs of items that are expired
```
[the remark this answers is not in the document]
dilyar85: I pushed a new commit that will read from the expiredChan now.

pkg/util/cache.go
```diff
@@ -0,0 +1,183 @@
...
+	defer c.mu.Unlock()
+
+	var emptyT T
+
+	v, ok := c.items[id]
+	if !ok {
+		return emptyT, false
+	}
+
+	if isHit != nil && !isHit(v.Item) {
```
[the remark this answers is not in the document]
dilyar85: We are adding this cache as a generic type under the `pkg/util`, not specifically to the OVF content. In our current use case, we check the content version but it could be possible to have a simple string id for other types of cache implementation in the future.

pkg/util/lock_pool.go
dilyar85: Good point. I have added a new function `InitOvfCacheAndLockPool` under the `vmprovider.go` file to clean up the lock when the same item expires from the cache.

pkg/util/lock_pool.go
```diff
@@ -0,0 +1,47 @@
...
+	ool sync.Map
```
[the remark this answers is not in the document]
dilyar85: I believe it's named as this way so that we can call `p.ool` in its methods. I will let @akutz confirm this as the code is's originally from his branch at https://github.com/akutz/vm-operator/blob/feature/generic-cache/pkg/util/lock_pool.go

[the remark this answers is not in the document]
ericvmw: I see, it looks good for private variables.

pkg/util/cache.go
```diff
@@ -0,0 +1,183 @@
...
+	donce    sync.Once
+
+	// expiredChan returns a channel on which the IDs of items that are expired
+	// are sent. This channel is closed when the cache is closed, but only after
+	// all, pending eviction notices are received.
+	expiredChan chan string
+	expiringNum sync.WaitGroup
+}
+
+// NewCache initializes a new cache with the provided expiration options.
```
ericvmw, line 36: It would be helpful to document the cache behavior with regarding to the expireAfter and maxItems, also LastUpdated in CacheItem.

ericvmw (approved):

pkg/util/cache.go
dilyar85: I think those are already documented in the `Cache` and `CacheItem` structs. Will keep the current comments for now as they are all in the same file.

akutz (approved):

pkg/util/cache.go
bryanv: The goroutine will block because there is/was nothing reading from the channel.

pkg/vmprovider/providers/vsphere/vmprovider.go
```diff
@@ -60,14 +76,36 @@ type vSphereVMProvider struct {
...
+// to cache the ovf envelope and lock the ovf envelope when it is being downloaded.
+func InitOvfCacheAndLockPool(expireAfter, checkExpireInterval time.Duration, maxItems int) (
+	*util.Cache[VersionedOVFEnvelope], *util.LockPool[string, *sync.RWMutex]) {
+	ovfCache := util.NewCache[VersionedOVFEnvelope](expireAfter, checkExpireInterval, maxItems)
+	ovfLockPool := &util.LockPool[string, *sync.RWMutex]{}
+
+	// Clean up the lock pool when the ovf cache item expires.
+	expiredChan := ovfCache.ExpiredChan()
+	go func() {
+		for k := range expiredChan {
```
bryanv, line 101: Isn't this racey: before the message is handled here, another call comes into getOvfEnvelope() and then an in-use lock is removed from this cache?

dilyar85: Yeah, but that shouldn't block us because the in-use lock can still be safely executed after removal, although it won't prevent another download of the same item. I considered locking it during deletion, but it wouldn't prevent the lock from being used after removal. Adding another layer of protection for the lock pool isn't ideal either, as there could be at most 2 duplicate downloads in the worst-case scenario. And this only occurs in rare cases when the lock is removed while in use.

dilyar85: On second thoughts, maybe locking and unlocking during deletion could still be useful. If getOvfEnvelope() can lock it, the lock won't be deleted until the item is downloaded and put into the cache, avoiding duplicate downloads. However, if delete() locks it first, there's still a small chance of duplicate downloads. I'll add comments in the updated code to document our discussion here.

bryanv: Yes, it will break the guarantee of no duplicate OVF retrieves we're trying to achieve in the first place by adding this outer locking pool.

Ultimately I think we're trying to make something more complicated that what's needed: the outer lock should be pushed into the CacheItem[T], and the Cache[T] has a Load/Fill function that populates the T, so we have something like https://guava.dev/releases/19.0/api/docs/com/google/common/cache/LoadingCache.html, and there's no need for the eviction channel to try to maintain consistency of the other cache.

dilyar85: I believe that's what I had in my original content library specific cache, where the item has its own lock... However, I understand this util cache doesn't have item-level locks as it may not be necessary for common use cases, where the cache is populated or retrieved quickly. It may not completely eliminate duplicate OVF retrieves, but it only occurs in rare edge cases where an OVF expires right after starting to download, and only two duplicates could happen even with multiple namespaces associated. If you feel strongly about this, I'm happy to discuss it further offline and make a follow-up change to address it.

PR #48, opened 2023-01-20
Set CloudInit(Prep) transport to GuestInfo

akutz:
# Overview

This patch ensures that the vAppConfig transport is set to GuestInfo when a user selects the CloudInit transport with the CloudInit prep option.

This ensures that CloudInit(Prep) works even if a user selects a VM Class that does not have a CD-ROM and VM Image where the vAppConfig transport is set to ISO.

This patch basically apes https://github.com/vmware-tanzu/vm-operator/commit/92e73100916ec8a1ed22ce8ccbdd17aa3ffb29c8, which solved the same issue for the `OvfEnv` and `vAppConfig` transports.

# Testing

This patch was validated by:

1. Replacing VM Operator on a running Supervisor with an image built with this patch.
1. A new VM was deployed:

    ```yaml
    apiVersion: v1
    kind: Secret
    metadata:
      name: my-vm-bootstrap-data
      namespace: my-namespace
    stringData:
      user-data: |
        #cloud-config
        ssh_pwauth: true
        users:
          - name: vmware
            sudo: ALL=(ALL) NOPASSWD:ALL
            lock_passwd: false
            # Password set to Admin!23
            passwd: '$1$salt$SOC33fVbA/ZxeIwD5yw1u1'
            shell: /bin/bash
        write_files:
          - content: |
              VMSVC Says Hello World
            path: /helloworld
    ---
    kind: VirtualMachine
    metadata:
      name: my-vm
      namespace: my-namespace
      annotations:
        vmoperator.vmware.com/cloudinit-type: cloudinitprep
    spec:
      networkInterfaces:
      - networkName: primary
        networkType: vsphere-distributed
      className: best-effort-small
      imageName: ubuntu-2210-server
      powerState: poweredOn
      vmMetadata:
        secretName: my-vm-bootstrap-data
        transport: CloudInit
    ```

1. The VM Operator logs showed:

    ```shell
    I0121 20:30:18.890099       1 session_vm_customization.go:285] vsphere "msg"="Customization Reconfigure" "vmName"="my-namespace/my-vm" "configSpec"={"vAppConfig":{"ovfEnvironmentTransport":["com.vmware.guestInfo"]}}
    ```

1. The VM came online with an IP:

    image

1. And it's clear the VM:

    1. is not using the ISO transport (see previous image)
    1. does not have a connected CD-ROM device

        image

### DIFF
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_customization_test.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_customization_test.go
@@ -378,16 +378,27 @@ var _ = Describe("Cloud-Init Customization", func() {
 
 	Context("GetCloudInitPrepCustSpec", func() {
 		var (
-			err      error
 			custSpec *vimTypes.CustomizationSpec
 		)
 		JustBeforeEach(func() {
-			custSpec, err = session.GetCloudInitPrepCustSpec(cloudInitMetadata, updateArgs)
+			var (
+				err        error
+				configSpec *vimTypes.VirtualMachineConfigSpec
+			)
+			configSpec, custSpec, err = session.GetCloudInitPrepCustSpec(cloudInitMetadata, updateArgs)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Validate that Cloud-Init Prep always uses the GuestInfo transport
+			// for the CI Prep's meta and user data.
+			Expect(configSpec).ToNot(BeNil())
+			Expect(configSpec.VAppConfig).ToNot(BeNil())
+			Expect(configSpec.VAppConfig.GetVmConfigSpec()).ToNot(BeNil())
+			Expect(configSpec.VAppConfig.GetVmConfigSpec().OvfEnvironmentTransport).To(HaveLen(1))
+			Expect(configSpec.VAppConfig.GetVmConfigSpec().OvfEnvironmentTransport[0]).To(Equal(session.OvfEnvironmentTransportGuestInfo))
 		})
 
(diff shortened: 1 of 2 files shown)


pkg/vmprovider/providers/vsphere/session/session_vm_customization_test.go
```diff
@@ -378,16 +378,27 @@ var _ = Describe("Cloud-Init Customization", func() {
...
+				err        error
+				configSpec *vimTypes.VirtualMachineConfigSpec
+			)
+			configSpec, custSpec, err = session.GetCloudInitPrepCustSpec(cloudInitMetadata, updateArgs)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Validate that Cloud-Init Prep always uses the GuestInfo transport
+			// for the CI Prep's meta and user data.
+			Expect(configSpec).ToNot(BeNil())
+			Expect(configSpec.VAppConfig).ToNot(BeNil())
```
zyiyi11, line 394: tini nit： `vAppConfigSpec := configSpec.VAppConfig` for reuse
LGTM

zyiyi11 (approved):

pkg/vmprovider/providers/vsphere/session/session_vm_customization_test.go
akutz: Thanks! I thought of that, but opt'd for uninterrupted Expect clauses.

bryanv (approved):

PR #49, opened 2023-01-23
Update uses of WCP_VM_Class_as_Config to DND

akutz:
This patch updates instances of the feature state switch from `WCP_VM_Class_as_Config` to the DaynDate variation. The original feature state switch is not removed entirely as it will be reused for a follow-up release.

### DIFF
--- a/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
+++ b/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
@@ -131,7 +131,7 @@ func vmTests() {
 			return vcVM, nil
 		}
 
-		Context("VMClassAsConfig FSS is enabled", func() {
+		Context("VMClassAsConfigDaynDate FSS is enabled", func() {
 
 			var (
 				vcVM       *object.VirtualMachine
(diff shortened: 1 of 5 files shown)


pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -352,254 +370,152 @@ func vmTests() {
 					Expect(ethDevice.MacAddress).NotTo(Equal(ethCard.MacAddress))
 
 					p := devList.SelectByType(&types.VirtualPCIPassthrough{})
-					Expect(p).To(HaveLen(1))
+					Expect(p).To(HaveLen(2))
 					pciDev1 := p[0].GetVirtualDevice()
 					pciBacking, ok1 := pciDev1.Backing.(*types.VirtualPCIPassthroughVmiopBackingInfo)
```
bryanv, line 375: Nit: `pciBacking` -> `pciBacking1` here, and `pciBacking1` -> `pciBacking2` below to keep the number suffix consistent. Or scope these in separate blocks.

bryanv (approved):

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
akutz: That was the original code, I just moved it around. But sure, can take this.

PR #50, opened 2023-01-29
Add PR labeler based on LoC

dilyar85:
This patch adds a new GitHub action that automatically attaches labels to PRs based on their LoC.
It can also be served as a reminder for developers to conduct necessary testing for PRs with large LoC.

Testing Done:
- Updated my fork `main` branch with this change pushed
- Created a PR against my fork with different LoC changing
- Confirmed the labels are added/removed in the PR accordingly

Screenshot of the [above PR](https://github.com/dilyar85/vm-operator/pull/3) in my fork:
image

### DIFF
--- a/.github/configs/labeler.yml
+++ b/.github/configs/labeler.yml
@@ -1,17 +1,20 @@
 version: 1
 labels:
-  - label: "xsmall-diff"
+  - label: "size/XS"
     size-below: 10
-  - label: "small-diff"
+  - label: "size/S"
     size-above: 9
+    size-below: 30
+  - label: "size/M"
+    size-above: 29
     size-below: 100
-  - label: "medium-diff"
+  - label: "size/L"
     size-above: 99
     size-below: 500
-  - label: "large-diff"
+  - label: "size/XL"
     size-above: 499
     size-below: 1000
-  - label: "xlarge-diff"
+  - label: "size/XXL"
     size-above: 999
-  - label: "testing-required-internal-e2e"
+  - label: "testing-needed-e2e-fast"
     size-above: 99
--- a/.github/configs/labeler.yml
+++ b/.github/configs/labeler.yml
@@ -1,0 +1,17 @@
+version: 1
+labels:
+  - label: "xsmall-diff"
+    size-below: 10
+  - label: "small-diff"
+    size-above: 9
+    size-below: 100
+  - label: "medium-diff"
+    size-above: 99
+    size-below: 500
+  - label: "large-diff"
+    size-above: 499
+    size-below: 1000
+  - label: "xlarge-diff"
+    size-above: 999
+  - label: "testing-required-internal-e2e"
+    size-above: 99
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,3 +225,13 @@ jobs:
         path: |
           code-coverage-results.md
           pull_request_id
+
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      # https://github.com/marketplace/actions/pr-labeler-based-on-multiple-rules
+      - uses: srvaroa/labeler@v0.9
+        with:
+          config_path: .github/configs/labeler.yml
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,13 +225,3 @@ jobs:
         path: |
           code-coverage-results.md
           pull_request_id
-
-  label:
-    runs-on: ubuntu-latest
-    steps:
-      # https://github.com/marketplace/actions/pr-labeler-based-on-multiple-rules
-      - uses: srvaroa/labeler@v0.9
-        with:
-          config_path: .github/configs/labeler.yml
-        env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,20 @@
+name: "Pull Request Labeler"
+on:
+- pull_request_target
+
+jobs:
+  triage:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    # Source: https://github.com/srvaroa/labeler
+    # Using this instead of the GH labeler action as the former supports labeling based on LoC changed.
+    # However, it doesn't work with regex pattern to ignore generated files (CRDs, docs, etc.) in diff.
+    # An issue has been filed https://github.com/srvaroa/labeler/issues/33 to track this improvement.
+    - uses: srvaroa/labeler@v0.9
+      with:
+        config_path: .github/configs/labeler.yml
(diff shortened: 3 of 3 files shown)


dilyar85:
Hmm looks like there is some permission issue for the Action app to access this repo:
```
2023/01/29 12:19:07 Setting labels to vmware-tanzu/vm-operator#50: [small-diff]
2023/01/29 12:19:07 Unable to execute action: PUT https://api.github.com/repos/vmware-tanzu/vm-operator/issues/50/labels: 403 Resource not accessible by integration []
```
Will try resolving this on Monday.

.github/configs/labeler.yml
```diff
@@ -0,0 +1,17 @@
...
+    size-below: 100
+  - label: "medium-diff"
+    size-above: 99
+    size-below: 500
+  - label: "large-diff"
+    size-above: 499
+    size-below: 1000
+  - label: "xlarge-diff"
+    size-above: 999
+  - label: "testing-required-internal-e2e"
```
akutz, line 16: Please change to the agreed upon `testing-needed-e2e`

.github/configs/labeler.yml
```diff
@@ -0,0 +1,17 @@
+version: 1
+labels:
+  - label: "xsmall-diff"
```
akutz, line 3: Remove the `-diff` suffix from the labels. The size alone is enough to indicate that this is an N size change.

.github/configs/labeler.yml
akutz: Actually, please change it to `testing-needed-e2e-fast`. We will figure out how/if to apply `testing-needed-e2e-full` to indicate when a full run is required vs the Kind-based workflow. For now we can make it so admins of this repo can manually apply `testing-needed-e2e-full` when/if it's deemed required.

.github/configs/labeler.yml
akutz: Actually, please ensure these match the following examples from how Kubernetes repos do it:
image

Please see [this query](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/labels?q=size) from CAPV for a list of its labels and their descriptions. I will go ahead and precreate these labels in the VM Op repo.

akutz: Okay, I [added the labels](https://github.com/vmware-tanzu/vm-operator/labels?q=size) to VM Op:

image

.github/workflows/ci.yml
```diff
@@ -225,3 +225,13 @@ jobs:
         path: |
           code-coverage-results.md
           pull_request_id
+
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      # https://github.com/marketplace/actions/pr-labeler-based-on-multiple-rules
```
akutz, line 232: Why not use the [GitHub-provided labeler action](https://github.com/actions/labeler)?

.github/workflows/ci.yml
```diff
@@ -225,3 +225,13 @@ jobs:
...
+      - uses: srvaroa/labeler@v0.9
```
akutz, line 233: Why not use the [GitHub-provided labeler action](https://github.com/actions/labeler)?

akutz: Ah, I see, the action allows a richer set of conditions. Please update the above comment to indicate why the non-GH labeler action is used. Also, please change the link to https://github.com/srvaroa/labeler as it is shorter and takes us directly to the action's source in case we ever need to look into it.

.github/configs/labeler.yml
akutz: Please update the LoC to match the numbers from the above labels that I created. Thanks!

akutz (changes requested):
Hi @dilyar85,

Thank you for owning and driving this! I really appreciate it! In addition to the feedback I have already provided, I would also like you to update all of the size labels to exclude:

```yaml
- "*_generated.go"       # generated Go sources
- "config/crd/bases/.*"  # generated CRDs
- "docs/apis/v*.md"      # generated API documentation,
                         # ex. docs/apis/v1alpha1.md
```

The problem is that the labeler action does not currently support an exclude option. There is an issue from November 2021 requesting this feature -- https://github.com/srvaroa/labeler/issues/33. Apparently the following does _not_ work:

```yaml
files:
- "^*_generated.go"       # generated Go sources
- "^config/crd/bases/.*"  # generated CRDs
- "^docs/apis/v*.md"      # generated API documentation,
                         # ex. docs/apis/v1alpha1.md
```

The standard GitHub labeler action _does_ support exclusions -- https://github.com/actions/labeler/pull/22. However, it does not support labeling based on the size of the diff, although an enhancement request for this _was_ filed last week -- https://github.com/actions/labeler/issues/486.

For now I think we go with srvaroa/labeler@v0.9 and acknowledge that the generated files cannot be easily ignored. I'm not sure if it's worth maintaining a list of all the files to include -- that will inevitably miss _something_ as files are added/removed.

Please update the labeler action with some notes regarding the concerns in this feedback so that others can track that we had this discussion. Thanks!

akutz:
> Hmm looks like there is some permission issue for the Action app to access this repo:
> 
> ```
> 2023/01/29 12:19:07 Setting labels to vmware-tanzu/vm-operator#50: [small-diff]
> 2023/01/29 12:19:07 Unable to execute action: PUT https://api.github.com/repos/vmware-tanzu/vm-operator/issues/50/labels: 403 Resource not accessible by integration []
> ```
> 
> After a quick research, it looks like GitHub Action's token will be downgraded to read-only in PRs from forked repos. Will try resolving this on Monday.

Hi @dilyar85,

Let's chat about this before you do anything. I already mentioned that you'll need to validate this in your fork. The reason is the perms. There is a reason https://github.com/vmware-tanzu/vm-operator/blob/main/.github/workflows/post-coverage-to-pr.yml exists the way it does -- actions spawned as a result of a PR from a fork do not have permission to write to the PR on the primary repo. Thus the `workflow_action` pattern documented at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run.

However, in the case of _this_ workflow, you should merge it to main _in your **fork**_. That's what I did to get the workflows correct for this repo in the first place. I kept rebasing main in my fork to validate the configurations. Then I would open PRs against my fork to validate PR workflows, etc. Once I was happy with everything, I would then open a PR against the primary repo from a branch on my fork, ensuring my fork's `main` branch was once again reset to the same as the primary repo.

.github/configs/labeler.yml
dilyar85: Updated in the latest patch as suggested.

.github/configs/labeler.yml
dilyar85: Thanks for adding these labels. I've updated the labeler config to match the above. I think we should remove the "ignoring generated files" as it's not supported yet?

.github/workflows/ci.yml
dilyar85: Added comments in the latest patch.

dilyar85:
@akutz Thanks for the review. I have updated this change to address your comments above.  
Also verified this workflow using my fork (see the testing done section in PR's description).

akutz (approved):
Lgtm!

PR #54, opened 2023-01-30
tkg: Propogate Supervisor DNS Search Suffixes for TKGs nodes

akutz:
This patch enables the automatic propogation of Supervisor DNS search suffixes for VMs created on behalf of Cluster API on Supervisor. If a VM has the label "capw.vmware.com/cluster.role" or "capv.vmware.com/cluster.role", then VM Operator will read the DNS search suffix information from the VM Op ConfigMap on Supervisor and use that information when constructing the netplan payload used to configure the VM's networking with Cloud-Init.

A few notes:

* This only works for TKGs nodes deployed with VM Op that set their bootstrap provider type to CloudInit in the VM Op VM resource.
* This is a band-aid until VM Op API is updated to express the full netplan configuration, at which point CAPV can use that to set DNS search suffixes when deploying VMs with VM Op.

### DIFF
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_update.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_update.go
@@ -8,6 +8,8 @@ import (
 	"reflect"
 
 	apiEquality "k8s.io/apimachinery/pkg/api/equality"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/vmware/govmomi/object"
 	vimTypes "github.com/vmware/govmomi/vim25/types"
@@ -47,8 +49,9 @@ type VMUpdateArgs struct {
 	ConfigSpec      *vimTypes.VirtualMachineConfigSpec
 	ClassConfigSpec *vimTypes.VirtualMachineConfigSpec
 
-	NetIfList  network.InterfaceInfoList
-	DNSServers []string
+	NetIfList      network.InterfaceInfoList
+	DNSServers     []string
+	SearchSuffixes []string
 
 	// hack. Remove after VMSVC-1261.
 	// indicating if this VM image used is VM service v1alpha1 compatible.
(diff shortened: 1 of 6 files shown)


pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -663,6 +673,42 @@ func (s *Session) prepareVMForPowerOn(
...
+	_, ok := labels[CAPWClusterRoleLabelKey]
+	if !ok {
+		_, ok = labels[CAPVClusterRoleLabelKey]
+	}
+	if !ok {
+		return nil
+	}
+
+	searchSuffixes, err := config.GetSearchSuffixesFromConfigMap(client)
+	if err != nil {
```
bryanv, line 704: Unless something has changed in the last year, gce2e doesn't create this CM - that's why the `GetNameserversFromConfigMap()` error is swallowed 😞 .  There's a whole lot of cruft left in this code - I think having `GetNameserversFromConfigMap()` (renaming to taste) return this value too would make sense so the CM is fetched in a single place, but all this stuff needs to be rewritten sooner than later anyways.

bryanv (approved):

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
akutz: Yeah, I was really torn between keeping to the patterns or refactoring. I would love to decompose this to divorce the testing responsibility a bit. But because I'm doing the logic to check if the label exists in the function, it means I should not read the ConfigMap outside the function. I thought about passing in `func() (*ConfigMap, error)`?

akutz: Thoughts on this?

akutz: Also, will refactor so that the other function returns both.

akutz:
@sidharthsurana / @dougm Ahead of updating Supervisor to inject the DNS search suffixes into the ConfigMap, we can validate this in a live environment by manually updating the same ConfigMap and adding the search suffixes. After which, a deployed TKC should have that information propogated into it.

akutz:
Internal, e2e testing timed out, but not in a way that invalidates this patch.

PR #55, opened 2023-01-31
Fix class ConfigSpec NIC with NSX-T network

bryanv:
When zipping the class ConfigSpec and the NIC device from the VM Spec the AddressType=manual needs to be added in order for the NCP MAC address to be preserved. Otherwise, VC will silently ignore the MacAddress field, generating a new one. While here, assert that the class ConfigSpec ResourceAllocation is preserved.

This is just the minimal change here to fix the immediate issue. The prior code didn't make a lot of sense to mix in the class ConfigSpec but that sorted out as a larger effect to provide the NICs at VM create time.


pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -157,6 +157,9 @@ func vmTests() {
 					},
 					AddressType: string(types.VirtualEthernetCardMacTypeGenerated),
 					MacAddress:  "00:0c:29:93:d7:27",
+					ResourceAllocation: &types.VirtualEthernetCardResourceAllocation{
```
akutz, line 160: This is in the test, but are we copying it from the config spec in the implementation?

bryanv: I don't fully follow your question. But here `ethCard` is later serialized into the class ConfigSpec, and later in the implementation, `networkDevices[]` is the devices from that ConfigSpec, in which some fields - like the backing, and now the MAC address from NCP - are then overwritten. NCP/NetOP don't have a `ResourceAllocation` or `WOL` so whatever is set in the class, if anything, is just blindly used.

akutz: Ahhh. I see what you mean. I had the relationship backwards. I thought this was copying the info FROM the ConfigSpec INTO the NIC list we already created. Ack. Makes sense. Approved.

akutz (approved):
lgtm!

aruneshpa (approved):
LGTM

PR #57, opened 2023-02-03
Do not consider hardware spec and RLS settings in Update path

aruneshpa:
In update path of a VM reconcile, we reconfigure the VM to match the desired config specified in the VM Class.  This means that if a VM class is edited _after_ a VM has already been created from it, we will end up Reconfiguring the VM during next power cycle.  This was never a supported workflow and was just a by product of the original implementation.

With VM Class as Config, we are creating the VM with hardware spec and RLS settings specified form the VM class.  So, we don't need to update these during update.  This change preserves the behavior for clusters that do not have the feature enabled.

Testing Done:
- [x] Existing unit and integration tests
- [x] New tests added
- [x] end to end tests

### DIFF
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_update.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_update.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2020 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2018-2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package session
@@ -402,20 +402,29 @@ func UpdateHardwareConfigSpec(
 	configSpec *vimTypes.VirtualMachineConfigSpec,
 	vmClassSpec *v1alpha1.VirtualMachineClassSpec) {
 
-	// TODO: Looks like a different default annotation gets set by VC.
-	if config.Annotation != constants.VCVMAnnotation {
-		configSpec.Annotation = constants.VCVMAnnotation
-	}
 	if nCPUs := int32(vmClassSpec.Hardware.Cpus); config.Hardware.NumCPU != nCPUs {
 		configSpec.NumCPUs = nCPUs
 	}
 	if memMB := virtualmachine.MemoryQuantityToMb(vmClassSpec.Hardware.Memory); int64(config.Hardware.MemoryMB) != memMB {
 		configSpec.MemoryMB = memMB
 	}
+}
+
+func UpdateConfigSpecAnnotation(
+	config *vimTypes.VirtualMachineConfigInfo,
+	configSpec *vimTypes.VirtualMachineConfigSpec) {
+	if config.Annotation != constants.VCVMAnnotation {
+		configSpec.Annotation = constants.VCVMAnnotation
+	}
+}
+
+func UpdateConfigSpecManagedBy(
+	config *vimTypes.VirtualMachineConfigInfo,
+	configSpec *vimTypes.VirtualMachineConfigSpec) {
 	if config.ManagedBy == nil {
 		configSpec.ManagedBy = &vimTypes.ManagedByInfo{
-			ExtensionKey: "com.vmware.vcenter.wcp",
-			Type:         "VirtualMachine",
+			ExtensionKey: constants.ManagedByExtensionKey,
+			Type:         constants.ManagedByExtensionType,
 		}
 	}
 }
@@ -445,6 +454,8 @@ func UpdateConfigSpecDeviceGroups(
 	}
 }
 
+// updateConfigSpec overlays the VM Class spec with the provided ConfigSpec to form a desired
+// ConfigSpec that will be used to reconfigure the VM.
 func updateConfigSpec(
 	vmCtx context.VirtualMachineContext,
 	config *vimTypes.VirtualMachineConfigInfo,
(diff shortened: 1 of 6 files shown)


dilyar85 (approved):
LGTM. And thanks for adding all the detailed code comments.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -402,20 +402,29 @@ func UpdateHardwareConfigSpec(
...
+
+func UpdateConfigSpecAnnotation(
+	config *vimTypes.VirtualMachineConfigInfo,
+	configSpec *vimTypes.VirtualMachineConfigSpec) {
+	if config.Annotation != constants.VCVMAnnotation {
+		configSpec.Annotation = constants.VCVMAnnotation
+	}
+}
+
+func UpdateConfigSpecManagedBy(
```
akutz, line 421: A few notes:

* Let's switch the order of the params, and
* use `dst` and `src` as they make sense in this instance.
* Let's also introduce `defaultValue string` in case `dst.Annotation` is `nil`. The value of `defaultValue` can be the value of the VM Class's ConfigSpec managedBy value.

```
func UpdateConfigSpecManagedBy(
	dst *vimTypes.VirtualMachineConfigSpec,
	src vimTypes.VirtualMachineConfigInfo,
	defaultValue *vimTypes.ManagedByInfo) {

	// If the VM was created the constant value, then do not
	// ever update the value again.
	if src.ManagedBy != nil && 
		src.ManagedBy.ExtensionKey == constants.ManagedByExtensionKey && 
		src.Type == constants.ManagedByExtensionType {

		return
	}

	// Update the destination's managedBy if it is not already set.
	if dst.ManagedBy == nil {
		if defaultValue != nil {
			dst.ManagedBy = defaultValue
		} else {
			dst.ManagedBy =  &vimTypes.ManagedByInfo{
				ExtensionKey: constants.ManagedByExtensionKey,
				Type:         constants.ManagedByExtensionType,
			}
		}
	}
}
```

akutz (changes requested):

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -402,20 +402,29 @@ func UpdateHardwareConfigSpec(
...
-	}
 	if nCPUs := int32(vmClassSpec.Hardware.Cpus); config.Hardware.NumCPU != nCPUs {
 		configSpec.NumCPUs = nCPUs
 	}
 	if memMB := virtualmachine.MemoryQuantityToMb(vmClassSpec.Hardware.Memory); int64(config.Hardware.MemoryMB) != memMB {
 		configSpec.MemoryMB = memMB
 	}
+}
+
+func UpdateConfigSpecAnnotation(
```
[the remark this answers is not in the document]
aruneshpa: Probably.  But as it stands right now, this file houses _all_ the methods that operate on ConfigSpec.  So I'd like to keep this here if you don't feel strongly about that.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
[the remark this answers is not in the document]
aruneshpa: Same as above.  Prefer to keep it here unless you feel strongly about this.

akutz: @aruneshpa and I discussed, we cannot allow VM Class to own ManagedBy due to iconography and telemetry.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -402,20 +402,29 @@ func UpdateHardwareConfigSpec(
...
+	config *vimTypes.VirtualMachineConfigInfo,
```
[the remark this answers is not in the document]
aruneshpa: Per our conversation, since we want to reserve the Annotations, I am leaving it as is.  Resolving both your comments.

sreyasn (commented):
Hi Arunesh, 
Change LGTM to handle CPU and Memory related hardware changes to the VMClass. But won't the same happen with device related changes to the VM Class, (ie) PCI devices, on power cycle?. 

Is the plan to prevent any hardware related reconfigures after VM Class as Config is enabled?

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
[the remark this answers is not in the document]
aruneshpa: Marked this as resolved.  We can open a new issue for the refactor if you feel strongly.

aruneshpa:
> Is the plan to prevent any hardware related reconfigures after VM Class as Config is enabled?
@sreyasn , that's the plan regardless of the FSS.  Since we don't _really_ support changing the hardware of the VM, we might as well explicitly stop supporting that.

akutz (approved):

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -212,6 +214,97 @@ func vmTests() {
...
+					var o mo.VirtualMachine
+					Expect(vcVM.Properties(ctx, vcVM.Reference(), nil, &o)).To(Succeed())
+
+					// Freq used by vcsim
+					const freq = 2294
+
+					reservation := o.Config.CpuAllocation.Reservation
+					vmClassCPURes := virtualmachine.CPUQuantityToMhz(vmClass.Spec.Policies.Resources.Requests.Memory, freq)
+					Expect(reservation).ToNot(BeNil())
+					Expect(reservation).To(Not(BeEquivalentTo(vmClassCPURes)))
```
bryanv, line 271: Nit here & a few other places: our usual pattern is `ToNot()` instead

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -453,9 +464,19 @@ func updateConfigSpec(
...
 
-	UpdateHardwareConfigSpec(config, configSpec, &vmClassSpec)
-	UpdateConfigSpecCPUAllocation(config, configSpec, &vmClassSpec, updateArgs.MinCPUFreq)
-	UpdateConfigSpecMemoryAllocation(config, configSpec, &vmClassSpec)
+	// Before VM Class as Config, VMs were deployed from the OVA, and are then
+	// reconfigured to match the desired CPU and memory reservation.  Maintain that
+	// behavior.  With the FSS enabled, VMs will be _created_ with desired HW spec, and we
+	// will not modify the hardware of the VM post creation.  So, don't populate the
+	// Hardware config and CPU/Memory reservation.
+	if !lib.IsVMClassAsConfigFSSDaynDateEnabled() {
```
bryanv, line 472: I believe that would leave us with a window where a CL Deploy is running and then a upgrade is kicked off, and the first VM "update" reconcile runs with the FFS enabled.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -402,20 +402,29 @@ func UpdateHardwareConfigSpec(
 	configSpec *vimTypes.VirtualMachineConfigSpec,
 	vmClassSpec *v1alpha1.VirtualMachineClassSpec) {
 
-	// TODO: Looks like a different default annotation gets set by VC.
```
bryanv, line 405: I believe this is still the case

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
aruneshpa: The chances of this happening are so slim that it is okay to ignore in favor of _not_ reconciling each VM with update hardware spec/reservations during power cycle.  Would you agree?

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
aruneshpa: Missed it.  Will take care of it in a folllowup.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
aruneshpa: Do you mean I should add the comment back?  The previous is maintained for non FSS enabled clusters because we are calling `UpdateConfigSpecAnnotations` method -- which will update the annotation to our default value from the one set by VC.

PR #58, opened 2023-02-07
Only add PCI passthrough devices from ConfigSpec if FSS is enabled

aruneshpa:
After VM Class as Config is enabled, a ConfigSpec is always expected to (mostly) accurately represent the desired set of devices.  This is done in part, by duplicating the vGPU and DirectPath devices from VM Class spec into the ConfigSpec.

Currently, we look at ConfigSpec /and/ the VM Class `spec.devices` field to determine what devices do we need to add to the VM.  This leads to the vGPU and DirectPath devices being added twice to the VM (since they are present in VM Class `spec.devices` field and are duplicated in ConfigSpec).

This change fixes this behavior by only looking at the ConfigSpec when the FSS is enabled.

Testing Done:
- [X] Existing Unit and Integration tests
- [X] Added new tests
- [x] e2e tests

### DIFF
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_update_test.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_update_test.go
@@ -177,8 +177,10 @@ var _ = Describe("Mutate update args with DNS information", func() {
 
 var _ = Describe("Update ConfigSpec", func() {
 
-	var config *vimTypes.VirtualMachineConfigInfo
-	var configSpec *vimTypes.VirtualMachineConfigSpec
+	var (
+		config     *vimTypes.VirtualMachineConfigInfo
+		configSpec *vimTypes.VirtualMachineConfigSpec
+	)
 
 	BeforeEach(func() {
 		config = &vimTypes.VirtualMachineConfigInfo{}
@@ -1121,28 +1123,28 @@ var _ = Describe("Update ConfigSpec", func() {
 			},
 		}
 		var pciDevices vmopv1alpha1.VirtualDevices
-		Context("For vGPU device", func() {
+		Context("For VM Class Spec vGPU device", func() {
 			BeforeEach(func() {
 				pciDevices = vmopv1alpha1.VirtualDevices{
 					VGPUDevices: vgpuDevices,
 				}
 			})
 			It("should create vSphere device with VmiopBackingInfo", func() {
-				vSphereDevices := virtualmachine.CreatePCIDevices(pciDevices, nil)
+				vSphereDevices := virtualmachine.CreatePCIDevicesFromVMClass(pciDevices)
 				Expect(vSphereDevices).To(HaveLen(1))
 				virtualDevice := vSphereDevices[0].GetVirtualDevice()
 				backing := virtualDevice.Backing.(*vimTypes.VirtualPCIPassthroughVmiopBackingInfo)
 				Expect(backing.Vgpu).To(Equal(pciDevices.VGPUDevices[0].ProfileName))
 			})
 		})
-		Context("For Dynamic DirectPath I/O device", func() {
+		Context("For VM Class Spec Dynamic DirectPath I/O device", func() {
 			BeforeEach(func() {
 				pciDevices = vmopv1alpha1.VirtualDevices{
 					DynamicDirectPathIODevices: ddpioDevices,
 				}
 			})
 			It("should create vSphere device with DynamicBackingInfo", func() {
-				vSphereDevices := virtualmachine.CreatePCIDevices(pciDevices, nil)
+				vSphereDevices := virtualmachine.CreatePCIDevicesFromVMClass(pciDevices)
 				Expect(vSphereDevices).To(HaveLen(1))
 				virtualDevice := vSphereDevices[0].GetVirtualDevice()
 				backing := virtualDevice.Backing.(*vimTypes.VirtualPCIPassthroughDynamicBackingInfo)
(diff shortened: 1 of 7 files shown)


sreyasn (approved):
LGTM

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -235,9 +235,9 @@ func vmTests() {
 					var o mo.VirtualMachine
 					Expect(vcVM.Properties(ctx, vcVM.Reference(), nil, &o)).To(Succeed())
 					Expect(o.Summary.Config.NumCpu).To(BeEquivalentTo(vmClass.Spec.Hardware.Cpus))
-					Expect(o.Summary.Config.NumCpu).To(Not(BeEquivalentTo(configSpec.NumCPUs)))
+					Expect(o.Summary.Config.NumCpu).ToNot(BeEquivalentTo(configSpec.NumCPUs))
```
dilyar85, line 238: I really wish Gomega only supports one matcher to keep things consistent instead of 3 for the same case (e.g. `ToNot`, `NotTo`, `To(Not)`)...

pkg/vmprovider/providers/vsphere/virtualmachine/devices.go
```diff
@@ -32,42 +33,43 @@ func CreatePCIDevices(
 	pciDevicesFromVMClass v1alpha1.VirtualDevices,
```
bryanv, line 33: Can we handle this higher in the call stack? IMO this change is poor ergonomics here because the caller still needs to provide both arguments but only one will ever be used. In the earlier class ConfigSpec changes I worked to @sreyasn to do these FFS checks early on so most things can just depend on nil vs non-nil pointers & empty vs nonempty collections.

If we can't do that, maybe time to split this function into two?

akutz: There's a push/pull between re-using code and FSS split-logic. I tend to err on the side of keeping the FSS check as early as possible to prevent things like possible NPEs in the call stack.

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -396,6 +396,87 @@ func vmTests() {
...
+											CustomLabel: "SampleLabel2",
+										},
+									},
+								},
+							},
+						},
+					}
+				})
+
+				It("GPU and DirectPath devices from VM Class are ignored", func() {
```
sreyasn, line 453: Nit: "from VM Class spec" just to make the description more explicit

pkg/vmprovider/providers/vsphere/session/session_vm_update_test.go
```diff
@@ -1151,44 +1153,67 @@ var _ = Describe("Update ConfigSpec", func() {
...
+						{
+							VirtualDevice: vimTypes.VirtualDevice{
+								Backing: &vimTypes.VirtualPCIPassthroughVmiopBackingInfo{
+									Vgpu: "configspec-profile",
+								},
 							},
 						},
 					}
+				})
+				It("should create vSphere device with DynamicBackingInfo", func() {
```
sreyasn, line 1173: "VmiopBackingInfo" instead of Dynamic for vGPUs ?.

sreyasn (approved):
LGTM otherwise.

pkg/vmprovider/providers/vsphere/virtualmachine/devices.go
```diff
@@ -62,15 +73,8 @@ func CreatePCIDevices(
...
-	// Append PCI devices from the VM Class ConfigSpec.
-	for i := range pciDevsFromConfigSpec {
-		dev := pciDevsFromConfigSpec[i]
-		dev.Key = deviceKey
-		devices = append(devices, dev)
-		deviceKey--
-	}
-
 	return devices
+
```
bryanv, line 77: Remove stray empty line

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -396,6 +396,87 @@ func vmTests() {
...
+					Expect(ok1).Should(BeTrue())
+					Expect(pciBacking1.Vgpu).To(Equal("SampleProfile2"))
+					Expect(pciBacking1.Vgpu).ToNot(Equal(vmClass.Spec.Hardware.Devices.VGPUDevices[0].ProfileName))
+
+					pciDev2 := p[1].GetVirtualDevice()
+					pciBacking2, ok2 := pciDev2.Backing.(*types.VirtualPCIPassthroughDynamicBackingInfo)
+					Expect(ok2).Should(BeTrue())
+					Expect(pciBacking2.AllowedDevice).To(HaveLen(1))
+					Expect(pciBacking2.AllowedDevice[0].VendorId).To(Equal(int32(52)))
+					Expect(pciBacking1.Vgpu).ToNot(Equal(vmClass.Spec.Hardware.Devices.DynamicDirectPathIODevices[0].VendorID))
```
bryanv, line 472: Are the `pciBacking1.Vgpu` assertions here a copy-pasta? This is already asserted above, and a little odd to be asserting the vGPU to not equals`vmClass...DynamicDirectPathIODevices[0]` fields.

bryanv (changes requested):

pkg/vmprovider/providers/vsphere/virtualmachine/devices.go
aruneshpa: Fixed.  Moved the FSS check up the call stack.

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -396,6 +396,81 @@ func vmTests() {
...
+					Expect(pciBacking1.Vgpu).To(Equal("profile-from-configspec"))
+
+					pciDev2 := p[1].GetVirtualDevice()
+					pciBacking2, ok2 := pciDev2.Backing.(*types.VirtualPCIPassthroughDynamicBackingInfo)
+					Expect(ok2).Should(BeTrue())
+					Expect(pciBacking2.AllowedDevice).To(HaveLen(1))
+					Expect(pciBacking2.AllowedDevice[0].VendorId).To(Equal(int32(52)))
+					Expect(pciBacking2.AllowedDevice[0].DeviceId).To(Equal(int32(53)))
+					Expect(pciBacking2.CustomLabel).To(Equal("label-from-configspec"))
+
```
bryanv, line 470: One more stray line

bryanv (approved):

PR #64, opened 2023-02-23
 Fix race in VirtualMachineSetResourcePolicy integration test

bryanv:
There is a data race with the called var. Switch it to an atomic until we revise how we're testing the whole provider interface, and the interface itself.


aruneshpa:
Nit: could you reduce a few chars from the title GH doesn't break the title halfway through a word?

aruneshpa (approved):
LGTM.

dilyar85 (approved):

PR #65, opened 2023-02-23
Stop creating TKG content source related resources.

yi0909:
This change stops watching TKG content source in the provider configmap and creating/updating/deleting related resources, such as ContentLibraryProvider and ContentSourceBindings when WCP_VM_Image_Registry FSS is enabled.

Test done:
- make test
- make test-integration
- manually deployed to a WCP cluster
    - passed internal gce2e tests
    - no TKG related ContentSource/ContentSourceBindings are created.

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -33,9 +33,6 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
 	if err := infraprovider.AddToManager(ctx, mgr); err != nil {
 		return errors.Wrap(err, "failed to initialize InfraProvider controller")
 	}
-	if err := providerconfigmap.AddToManager(ctx, mgr); err != nil {
-		return errors.Wrap(err, "failed to initialize ProviderConfigMap controller")
-	}
 	if err := virtualmachine.AddToManager(ctx, mgr); err != nil {
 		return errors.Wrap(err, "failed to initialize VirtualMachine controller")
 	}
@@ -67,6 +64,12 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
 	} else {
 		if err := contentsource.AddToManager(ctx, mgr); err != nil {
 			return errors.Wrap(err, "failed to initialize ContentSource controller")
+		}
+		// We only update TKG related ContentSource/ContentLibraryProvider/ContentSourceBinding resources
+		// in provider configmap reconcile. These resources will be removed when the FSS is enabled,
+		// add provider configmap controller to the manager only when the FSS is disabled.
+		if err := providerconfigmap.AddToManager(ctx, mgr); err != nil {
+			return errors.Wrap(err, "failed to initialize ProviderConfigMap controller")
 		}
 	}
 	return nil


controllers/providerconfigmap/provider_configmap_controller.go
```diff
@@ -287,6 +288,21 @@ func (r *ConfigMapReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (ct
 func (r *ConfigMapReconciler) ReconcileNormal(ctx goctx.Context, cm *corev1.ConfigMap) error {
 	r.Logger.Info("Reconciling VM provider ConfigMap", "name", cm.Name, "namespace", cm.Namespace)
 
+	// We only update TKG related ContentSource/ContentLibraryProvider/ContentSourceBinding resources
+	// in provider configmap reconcile loop for now.
+	if lib.IsWCPVMImageRegistryEnabled() {
```
bryanv, line 293: I'm confused: shouldn't this be `!lib.IsWCPVMImageRegistryEnabled()` to _stop_ creating these when the FFS is enabled? My initial thought was that this whole controller isn't needed anymore when the FFS is enabled. What still needs the source or binding now because in the image lookup path AFAICT we don't depend on those CRs anymore when the FFS is enabled.

yi0909: Oh yes!!! I forgot the `!`

aruneshpa:
@yi0909 , Could you also add a test for this change?

controllers/providerconfigmap/provider_configmap_controller.go
yi0909: I was thinking about adding the entire controller to the manager only if the FSS is disabled, but ended up keeping this controller, so that if we don't have to add it back if we want to watch the provider configMap for other resources in the future? But yah, it's not needed when the FSS is enabled. Will update this.

controllers/controllers.go
```diff
@@ -65,6 +62,12 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
 			return errors.Wrap(err, "failed to initialize VirtualMachinePublishRequest controller")
 		}
 	} else {
+		// We only update TKG related ContentSource/ContentLibraryProvider/ContentSourceBinding resources
+		// in provider configmap reconcile. These resources will be removed when the FSS is enabled,
+		// add provider configmap controller to the manager only when the FSS is disabled.
+		if err := providerconfigmap.AddToManager(ctx, mgr); err != nil {
```
dilyar85, line 68: Nit: keep the controllers in alphabetical order, even in the conditional statement ;)

dilyar85 (approved):

controllers/providerconfigmap/provider_configmap_controller.go
bryanv: Yes let's disable this whole controller if it isn't needed.

bryanv (approved):
Do we have anything - or the work planned - to remove the unused CRs post FSS enabled upgrade? I glanced quick at our upgrade hooks and didn't see anything. Some of those now unneeded CR have finalizers so we if we're disabling the controllers we need to manually clean those up so we don't significantly delay namespace removal.

aruneshpa:
LGTM.

/approve

PR #66, opened 2023-02-23
Fix NSX-T LogicalSwitchUUID to DVPG lookup

bryanv:
A supported but very, very rare NSX-T configuration is for the same LogicalSwitchUUID to be shared between multiple DPVGs in a Datacenter. To support this, shortly before WCP 1.0 release, a change was made in WCP and then ported over here that used a "all the hosts in the CCR" check to determine which DVPG is the correct one to use. Ever since, this check in WCP and here have been peculiar and the details as to why mostly lost.

This is all host check strict in the sense that it will fail if a host is in MM or otherwise disconnected. Fortunately, within a CCR, the LogicalSwitchUUID to DVPG mapping is unique, so use that to simplify the lookup logic.

The possibility of multiple DVPGs for each CCR will be a pain when we start to do network placement.

Remove the NSX-T OpaqueNetwork support since that was only used pre-1.0 and never in a shipped release.

### DIFF
--- a/pkg/vmprovider/providers/vsphere/network/network_provider.go
+++ b/pkg/vmprovider/providers/vsphere/network/network_provider.go
@@ -19,6 +19,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/property"
 	"github.com/vmware/govmomi/vim25"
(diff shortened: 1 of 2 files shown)


dilyar85 (approved):
LGTM.

aruneshpa (approved):
LGTM.  Can you look through gce2e and make sure there are no tests for this that might break?

PR #67, opened 2023-02-23
Set default envtest asset directory when not provided

bryanv:
This saves a step - having to set KUBEBUILDER_ASSETS envar - when running an integration test suite on directly on the CLI or an IDE. Top level Makefile still sets KUBEBUILDER_ASSETS so that behavior is unchanged.

### DIFF
--- a/test/builder/test_suite.go
+++ b/test/builder/test_suite.go
@@ -29,7 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/rest"
-	klog "k8s.io/klog/v2"
+	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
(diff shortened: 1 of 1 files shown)


test/builder/test_suite.go
```diff
@@ -223,8 +223,14 @@ func (s *TestSuite) init(crdPaths []string) {
...
 	if s.flags.IntegrationTestsEnabled {
+		var assetsDir string
+		if _, exists := os.LookupEnv("KUBEBUILDER_ASSETS"); !exists {
+			assetsDir = filepath.Join(testutil.GetRootDirOrDie(), "hack/tools/bin/")
+		}
+
 		s.envTest = envtest.Environment{
-			CRDDirectoryPaths: crdPaths,
+			CRDDirectoryPaths:     crdPaths,
+			BinaryAssetsDirectory: assetsDir,
```
dilyar85, line 233: Given that this field can be overridden by setting KUBEBUILDER_ASSETS, perhaps just set it here without checking the env var existence above?

dilyar85 (approved):

PR #68, opened 2023-02-24
Update servers TLS config to use approved cipher suites

dilyar85:
This patch updates the controller-manager webhook and kube-rbac-proxy servers to use the approved TLS cipher suites. It upgrades the controller-runtime package to [v0.12.2](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.12.2) (along with its dependent packages), which supports TLS configuration on the webhook server.  
The TLS min version set has been moved under that same func as the `server.TLSMinVersion` is now [deprecated](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/server.go#L74).

Testing Done:
- Internal fast kind-based e2e tests
- Manually deployed this patch to a WCP testbed and verified the cipher suites using [sslscan](https://github.com/rbsec/sslscan)
```console
$ kubectl get pod -n vmware-system-vmop -o wide
NAME                                                        READY   STATUS    RESTARTS   AGE     IP              
vmware-system-vmop-controller-manager-66fddb8fb-6g6jl       2/2     Running   0          5m10s   192.168.128.0
vmware-system-vmop-controller-manager-66fddb8fb-tm4sh       2/2     Running   0          5m13s   192.168.128.1
vmware-system-vmop-web-console-validator-55bdcbfc44-wfpkh   1/1     Running   0          5m14s   192.168.128.1

$ ./sslscan https://192.168.128.1:9848
...
  Supported Server Cipher(s):
Preferred TLSv1.3  128 bits  TLS_AES_128_GCM_SHA256        Curve 25519 DHE 253
Accepted  TLSv1.3  256 bits  TLS_AES_256_GCM_SHA384        Curve 25519 DHE 253
Accepted  TLSv1.3  256 bits  TLS_CHACHA20_POLY1305_SHA256  Curve 25519 DHE 253
Preferred TLSv1.2  128 bits  ECDHE-RSA-AES128-GCM-SHA256   Curve 25519 DHE 253
Accepted  TLSv1.2  256 bits  ECDHE-RSA-AES256-GCM-SHA384   Curve 25519 DHE 253

$ ./sslscan https://192.168.128.2:9878
...
  Supported Server Cipher(s):
Preferred TLSv1.3  128 bits  TLS_AES_128_GCM_SHA256        Curve 25519 DHE 253
Accepted  TLSv1.3  256 bits  TLS_AES_256_GCM_SHA384        Curve 25519 DHE 253
Accepted  TLSv1.3  256 bits  TLS_CHACHA20_POLY1305_SHA256  Curve 25519 DHE 253
Preferred TLSv1.2  128 bits  ECDHE-RSA-AES128-GCM-SHA256   Curve 25519 DHE 253
Accepted  TLSv1.2  256 bits  ECDHE-RSA-AES256-GCM-SHA384   Curve 25519 DHE 253
```


main.go
```diff
@@ -283,8 +280,16 @@ func main() {
 		os.Exit(1)
 	}
 
+	setupLog.Info("setting up webhook server TLS config")
+	webhookServer := mgr.GetWebhookServer()
+	if webhookServer == nil {
+		setupLog.Error(err, "unable to get controller manager webhook server")
```
bryanv, line 286: May be clearer to pass an explicit `nil` here instead. That said, I think we can safely depend on `mgr. GetWebhookServer()` not ever returning `nil` 'cause the manager implementation will panic, and if it would return nil effect is the ultimately the same

bryanv (approved):

main.go
dilyar85: Makes sense. Updated to remove the nil check.

yi0909 (approved):

PR #69, opened 2023-02-27
Add finalizer to VirtualMachinePublishRequest

yi0909:
This change adds finalizer to VirtualMachinePublishRequest, so that the DeleteMetrics is guaranteed to be called.

Test done:
- make test
- make test-integration
- manually deployed to a WCP cluster and verified:
```root@42158b2f7e39152b667d27bcedfd360f [ ~ ]# k describe vmpub vmpub-3 -n yijiang-ns
Name:         vmpub-3
Namespace:    yijiang-ns
Labels:       <none>
Annotations:  <none>
API Version:  vmoperator.vmware.com/v1alpha1
Kind:         VirtualMachinePublishRequest
Metadata:
  Creation Timestamp:  2023-02-27T22:07:19Z
  Finalizers:
    virtualmachinepublishrequest.vmoperator.vmware.com
  Generation:  1
  ...
  Resource Version:  21655992
  UID:               c9f637b0-4f4d-46ca-8d0f-cef4146cc568
Spec:
 ...
  Ttl Seconds After Finished:  60
Status:
  Attempts:  1
  ...
Events:             <none>
```
  - the finalizer is applied to the object.
  - the object can be removed.
 ```After some time
root@42158b2f7e39152b667d27bcedfd360f [ ~ ]# k describe vmpub vmpub-3 -n yijiang-ns
Error from server (NotFound): virtualmachinepublishrequests.vmoperator.vmware.com "vmpub-3" not found
```
- internal gce2e job

### DIFF
--- a/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
+++ b/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
@@ -20,6 +20,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -41,6 +42,7 @@ import (
 )
 
 const (
+	finalizerName     = "virtualmachinepublishrequest.vmoperator.vmware.com"
 	TaskDescriptionID = "com.vmware.ovfs.LibraryItem.capture"
 
 	// waitForTaskTimeout represents the timeout to wait for task existence in task manager.
@@ -203,10 +205,6 @@ func (r *Reconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Resu
 	// Update() of stale object will be rejected by API server and result in unnecessary reconciles.
 	err := r.apiReader.Get(ctx, req.NamespacedName, vmPublishReq)
 	if err != nil {
-		// Delete registered metrics if the resource is not found.
-		if apiErrors.IsNotFound(err) {
-			r.Metrics.DeleteMetrics(r.Logger, req.Name, req.Namespace)
-		}
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
@@ -215,6 +213,26 @@ func (r *Reconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Resu
 		Logger:           ctrl.Log.WithName("VirtualMachinePublishRequest").WithValues("name", req.NamespacedName),
 		VMPublishRequest: vmPublishReq,
 	}
+
+	patchHelper, err := patch.NewHelper(vmPublishReq, r.Client)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "failed to init patch helper for %s", fmt.Sprintf("%s/%s", vmPublishReq.Namespace, vmPublishReq.Name))
+	}
+
+	// the patch is skipped when the VirtualMachinePublishRequest.Status().Update
+	// is called during publishVirtualMachine().
+	defer func() {
+		if vmPublishCtx.SkipPatch {
+			return
+		}
+
+		if err := patchHelper.Patch(ctx, vmPublishReq); err != nil {
+			if reterr == nil {
+				reterr = err
+			}
+			vmPublishCtx.Logger.Error(err, "patch failed")
+		}
+	}()
 
 	if !vmPublishReq.DeletionTimestamp.IsZero() {
 		return r.ReconcileDelete(vmPublishCtx)
@@ -253,25 +271,22 @@ func (r *Reconciler) updateSourceAndTargetRef(ctx *context.VirtualMachinePublish
 }
 
 // publishVirtualMachine checks if source VM and target is valid. Publish a VM if all requirements are met.
-// Returns
-// - A boolean value represents if the following Patch should be skipped.
-// - error.
-func (r *Reconciler) publishVirtualMachine(ctx *context.VirtualMachinePublishRequestContext) (bool, error) {
+func (r *Reconciler) publishVirtualMachine(ctx *context.VirtualMachinePublishRequestContext) error {
 	vmPublishReq := ctx.VMPublishRequest
 	// Check if the source and target is valid
 	if err := r.checkIsSourceValid(ctx); err != nil {
 		ctx.Logger.Error(err, "failed to check if source is valid")
-		return false, err
+		return err
 	}
 
 	if err := r.checkIsTargetValid(ctx); err != nil {
 		ctx.Logger.Error(err, "failed to check if target is valid")
-		return false, err
+		return err
 	}
 
 	// In case we mark Upload condition to true when checking target, return early.
 	if conditions.IsTrue(ctx.VMPublishRequest, vmopv1alpha1.VirtualMachinePublishRequestConditionUploaded) {
-		return false, nil
+		return nil
 	}
 
 	if conditions.IsTrue(ctx.VMPublishRequest, vmopv1alpha1.VirtualMachinePublishRequestConditionSourceValid) &&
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go [hunk 1]
```diff
@@ -41,6 +42,16 @@ import (
...
+	clItemPrefix = "clibitem-"
+
+	finalizerName = "virtualmachinepublishrequest.vmoperator.vmware.com"
+
+	ItemParseErrorMessage = "Failed to get the uploaded item ID. This error is unrecoverable." +
+		" Please create a new VirtualMachinePublishRequest to retry VM publishing."
+
+	// ItemDescriptionRegexString is used to filter the VMPub UID from the content library item description.
+	ItemDescriptionRegexString = "virtualmachinepublishrequest\\.vmoperator\\.vmware\\.com: ([a-z0-9-]*)"
+
```
aruneshpa, line 54: Are you trying to keep these consts var names in alphabetical order?  I am not aware if that is necessarily a recommended/best practice.  I would err on the side of keeping the diff as small as possible.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -216,11 +220,32 @@ func (r *Reconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Resu
 		VMPublishRequest: vmPublishReq,
 	}
 
+	skipPatch := false
```
aruneshpa, line 223: Move this to L228 (close to where this is referenced).

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -810,15 +835,25 @@ func parseItemIDFromTaskResult(result vimtypes.AnyType) (string, error) {
...
 }
 
-func (r *Reconciler) ReconcileNormal(ctx *context.VirtualMachinePublishRequestContext) (_ ctrl.Result, reterr error) {
+func (r *Reconciler) ReconcileNormal(ctx *context.VirtualMachinePublishRequestContext) (_ ctrl.Result,
+	skipPatch bool, reterr error) {
 	ctx.Logger.Info("Reconciling VirtualMachinePublishRequest")
 	vmPublishReq := ctx.VMPublishRequest
 
+	if !controllerutil.ContainsFinalizer(vmPublishReq, finalizerName) {
+		// The finalizer must be present before proceeding in order to ensure that the VM will
```
aruneshpa, line 844: copy paste? 
`VM` -> `VirtualMachinePublishRequest`

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
akutz: I would also redefine this as:

```golang
var skipPatch bool
```

It's just preference, but since the default value is `false` anyway, it looks cleaner (to me). While at it, I would love to see some documentation:

```golang
var skipPatch bool // the patch may be skipped when the following
                   // conditions occur:
                   //
                   //     * it is rainy outside
                   //     * the patch just is not feeling it today
                   //     * the application failed due to gremlins in the system
                   //     * etc.
```

akutz (changes requested):

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
akutz: I prefer things in alphabetical order, and I may have shared that with her. However, I also agree it's not necessary to do that in an unrelated patch. We can do that later in a dedicated change.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -840,28 +875,9 @@ func (r *Reconciler) ReconcileNormal(ctx *context.VirtualMachinePublishRequestCo
 		vmopv1alpha1.VirtualMachinePublishRequestConditionComplete); isComplete {
 		requeueAfter, deleted, err := r.removeVMPubResourceFromCluster(ctx)
 		isDeleted = deleted
-		return ctrl.Result{RequeueAfter: requeueAfter}, err
+		return ctrl.Result{RequeueAfter: requeueAfter}, false, err
```
dilyar85, line 878: Should we return true here to skip the patch since there is no update of the resource in `removeVMPubResourceFromCluster`?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -900,15 +916,17 @@ func (r *Reconciler) ReconcileNormal(ctx *context.VirtualMachinePublishRequestCo
 		if deleted {
 			skipPatch = true
 		}
-		isDeleted = deleted
-		return ctrl.Result{RequeueAfter: requeueAfter}, err
+		return ctrl.Result{RequeueAfter: requeueAfter}, false, err
```
dilyar85, line 919: Two questions here:
1) why do you remove `isDeleted = deleted`? Shouldn't this still be used when registering metrics in defer?
2) I don't think your overhead `skipPath = true` is applied as you return hard-coded false at the end.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -900,15 +916,17 @@ func (r *Reconciler) ReconcileNormal(ctx *context.VirtualMachinePublishRequestCo
...
 	}
 
-	return requeueResult(ctx), nil
+	return requeueResult(ctx), false, nil
 }
 
 func (r *Reconciler) ReconcileDelete(ctx *context.VirtualMachinePublishRequestContext) (ctrl.Result, error) {
-	r.Metrics.DeleteMetrics(ctx.Logger, ctx.VMPublishRequest.Name, ctx.VMPublishRequest.Namespace)
-	// no op
+	if controllerutil.ContainsFinalizer(ctx.VMPublishRequest, finalizerName) {
```
dilyar85, line 926: Should we also return a `skipPatch` from this function to avoid patching when the resource doesn't contain the finalizer?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
```diff
@@ -95,17 +97,39 @@ func unitTestsReconcile() {
...
-	getVirtualMachinePublishRequest := func() *vmopv1alpha1.VirtualMachinePublishRequest {
-		newVMPub := &vmopv1alpha1.VirtualMachinePublishRequest{}
-		Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(vmpub), newVMPub)).To(Succeed())
-		return newVMPub
-	}
+	// getVirtualMachinePublishRequest := func() *vmopv1alpha1.VirtualMachinePublishRequest {
+	// 	newVMPub := &vmopv1alpha1.VirtualMachinePublishRequest{}
+	// 	Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(vmpub), newVMPub)).To(Succeed())
+	// 	return newVMPub
+	// }
```
dilyar85, line 104: Unclean test code?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
yi0909: That line was removed by accident. Added back.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
yi0909: It will be a noop, or even the patch failed, it is no harm. I'll leave it as it is. `SkipPatch` is needed due to CL API limitations. We have to ensure that each attempt, we only trigger one `CreateOVF` API, the .status().Update() will deny requests if the resourceVersion doesn't match.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
yi0909: Will keep it as `false` based on below comments.

bryanv (approved):

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -216,11 +218,34 @@ func (r *Reconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Resu
 		VMPublishRequest: vmPublishReq,
 	}
 
+	patchHelper, err := patch.NewHelper(vmPublishReq, r.Client)
```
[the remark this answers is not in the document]
dilyar85: I don't think it's needed now with the addition of the finalizer.

dilyar85 (approved):
LGTM.

akutz (approved):

PR #70, opened 2023-03-01
Bump golangci-lint and remove unused tools

bryanv:
Our prior version of golangci-lint is not compatible with go 1.20 [1] so update that now to front run our go version bump. Fix some low hanging test/ linter excludes.

Switch to a unique tools build tag to work around golangci-lint dependency compile failure [2].

While here remove some old Jenkins code coverage tools that are no longer needed.

[1] - https://github.com/golangci/golangci-lint/issues/3565
[2] - https://github.com/ryancurrah/gomodguard/issues/33


hack/tools/tools.go
```diff
@@ -1,5 +1,5 @@
-//go:build tools
-// +build tools
+//go:build vmop_tools
+// +build vmop_tools
```
aruneshpa, line 2: Out of curiosity, is there a reason behind this rename?

aruneshpa (approved):
LGTM.

dilyar85 (approved):
Thanks for upgrading this!

hack/tools/tools.go
bryanv: See [2] in the commit message: one of golangci-lint's dependencies recently added a file with the "tools" build tag, which is fine when just building that dependency alone, but breaks when being pulled into the golangci-lint build. This will likely end up breaking various CAPI-inspired hack/tools.

PR #71, opened 2023-03-01
Remove image supported VM validation check

bryanv:
With UnifiedTKG FFS being enabled for some time now, this image check has been a noop but we still had tests for the non-FFS path: I recently hit an integration test race where the just created VM Image hadn't yet made it into the ctrl-runtime client cache. The validation webhooks should really just worry about internal object consistency, so here we only care that the imageName field isn't empty.

There is another problematic VM Image check in here w.r.t. the PVC volumes and OVA HW version that will addressed in a later MR.

### DIFF
--- a/pkg/vmprovider/providers/vsphere/constants/constants.go
+++ b/pkg/vmprovider/providers/vsphere/constants/constants.go
@@ -21,11 +21,6 @@ const (
 	ManagedByExtensionKey  = "com.vmware.vcenter.wcp"
 	ManagedByExtensionType = "VirtualMachine"
 
-	// VMOperatorImageSupportedCheckKey Annotation key to skip validation checks of GuestOS Type
-	// TODO: Rename and move to vmoperator-api.
-	VMOperatorImageSupportedCheckKey     = pkg.VMOperatorKey + "/image-supported-check"
-	VMOperatorImageSupportedCheckDisable = "disable"
-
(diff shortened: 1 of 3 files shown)


zyiyi11 (approved):
LGTM

akutz:
This will break the marketplace images and TKGs. We need to make sure at least Unified TKG is no longer depending on the bit set in the extra config to work. As far as the market place images, even though they're marked deprecated, if we remove this, they _will_ fail going forward.

Shruthi?

akutz (changes requested):
Putting a hold on this until we resolve the conversation I left in the general comments.

bryanv:
> This will break the marketplace images and TKGs.

I don't see what this is breaking, at least if nothing was already broken before. The Unified TKG FFS has been enabled for awhile everywhere AFAIK - and we've made other changes assuming & implying as such - so the validation check was returning early before checking ImageSupported field any real env. Or are you saying the existing logic was wrong: the early return for Unified TKG FFS should not have been here in the first place?

akutz:
Thanks @srajashekar-vmw , you're right, I reviewed too quickly. I thought we were removing the bits that preserved that flag even if uTKG was enabled. This is just the webhook validator and will have no impact.

akutz (approved):
Lgtm, thank you for being patient!

PR #72, opened 2023-03-03
Bump up controller-runtime version

aruneshpa:
Bump up controller-runtime to support 1.25 and newer Supervisor.  Most of the changes are around the `pointer` package which removes the `Ptr` suffix in method names to get a pointer variable.  Another change to use Generics sets instead of typed ones.


akutz (approved):
Lgtm!

bryanv (approved):

PR #74, opened 2023-03-03
Remove Scheme from ControllerManagerContext and Reconciler

bryanv:
The controller-runtime Client added a Scheme() getter a while back so consistently use that for the Scheme so we don't have multiple ways to get the same thing.


akutz (approved):
Lgtm, thanks!

aruneshpa (approved):
LGTM.

PR #75, opened 2023-03-06
Update pr-labeler workflow to use a new GITHUB_TOKEN

### LINKED ISSUE #73: ci: The testing-needed labels appear to not always work
I've noticed the `testing-needed` labels are not working as intended. For example, on https://github.com/vmware-tanzu/vm-operator/pull/72 the label did not cause the PR check to fail until I removed and then re-added the label manually. PR https://github.com/vmware-tanzu/vm-operator/pull/71 is a good example of where the label is present, but it is not blocking the PR. 

@dilyar8, can you PTAL at this? Thanks!

dilyar85:
**NOTE: This PR should be merged AFTER adding the new GitHub token as a secret to this repo.**

This patch updates the `pr-labeler` workflow to use a new (to-be-added) GitHub token, so that it can trigger the `testing-needed` workflow after setting the corresponding labels  (_testing-needed-e2e-fast/full_).

It also removes the "synchronize", "opened", and "reopened" event types from the `testing-needed` workflow to avoid triggering from a stale pull request change (e.g. when the _testing-needed-e2e-fast/full_ labels were not set).

Testing Done:
- Merged this patch in my fork repo and verified the expected status check from its PR: https://github.com/dilyar85/vm-operator/pull/17

Reference doc: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

Closes #73

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -35,4 +35,6 @@ jobs:
       with:
         config_path: .github/configs/labeler.yml
       env:
-        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        # A non GITHUB_TOKEN is used to enable triggering the testing-needed
+        # workflow when related labels are added from the current workflow.
+        GITHUB_TOKEN: "${{ secrets.PR_LABELER_TOKEN }}"
--- a/.github/workflows/testing-needed.yml
+++ b/.github/workflows/testing-needed.yml
@@ -3,9 +3,6 @@ name: testing-needed
 on:
   pull_request:
     types:
-    - synchronize
-    - opened
-    - reopened
     - labeled
     - unlabeled
 


zyiyi11 (approved):
@akutz a new GitHub token needs to be added by maintainer.

akutz:
What permissions are required by the new token?

akutz:
> @akutz a new GitHub token needs to be added by maintainer.

Thanks @zyiyi11 , I'm aware, but there is insufficient information in the original issue discussion / this PR to determine what to add. 

@dilyar85 , also, `GITHUB_TOKEN` is a special token for every repo. I think you actually want to use a different env var. Please see https://docs.github.com/en/actions/security-guides/automatic-token-authentication for more information on the special var `GITHUB_TOKEN`.

dilyar85:
> @dilyar85 What permissions are required by the new token? Is this supposed to be personal token from a maintainer's account? Please provide more information related to the token that should be added to the repo. Thanks.

Yes, it needs to be a personal token according to GitHub's [triggering-a-workflow-from-a-workflow](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) doc:
> If you do want to trigger a workflow from within a workflow run, you can use a personal access token instead of GITHUB_TOKEN to trigger events that require a token. You'll need to create a personal access token and store it as a secret.

For this workflow, "Actions" and "Pull requests" with read-and-write access should be sufficient to add and remove labels on PRs. I was previously using a full-access token when testing this change in my personal fork. I will try generating a more limited token to test and let you know the exact permission needed.

PR #78, opened 2023-03-06
Account for GPU devices when the DaynDate FSS is disabled

### LINKED ISSUE #62: pcidevices: Refactor PCI device code in CreateConfigSpecForPlacement
I believe [this code](https://github.com/vmware-tanzu/vm-operator/blob/c4b0bf4c8062810b54c486a1b4d61c513119dff4/pkg/vmprovider/providers/vsphere/virtualmachine/configspec.go#L133-L138) needs to updated for the feature "VM Class as Config". Otherwise, _duplicate_ vGPUs and/or DDPIO PCI devices will be added to the placement spec, since said devices will exist in both the VM Class VAPI _and_ the ConfigSpec once the feature is enabled.

I believe the following change is needed:

```golang
	// Do not add the PCI devices from the VAPI if the following FSS is enabled, because
	// the devices will have already been copied to the ConfigSpec by Supervisor.
	if !lib.IsVMClassAsConfigFSSDaynDateEnabled() {
		for _, dev := range CreatePCIDevicesFromVMClass(vmClassSpec.Hardware.Devices) {
			configSpec.DeviceChange = append(configSpec.DeviceChange, &vimtypes.VirtualDeviceConfigSpec{
				Operation: vimtypes.VirtualDeviceConfigSpecOperationAdd,
				Device:    dev,
			})
		}
	}
```

aruneshpa:
Due to a bug introduced by 44cd693, we are not dropping the PCI passthrough devices when the DaynDate FSS is disabled.  This change fixes that.

# Testing Done:
- [x] Added a UT for this
- [x] Exiting Unit and Integration tests

Closes #62

### DIFF
--- a/pkg/vmprovider/providers/vsphere/virtualmachine/configspec_test.go
+++ b/pkg/vmprovider/providers/vsphere/virtualmachine/configspec_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package virtualmachine_test
@@ -6,15 +6,11 @@ package virtualmachine_test
 import (
 	goctx "context"
 
-	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
-
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/lib"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/virtualmachine"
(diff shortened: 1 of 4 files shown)


sreyasn (approved):
Thanks for the fix @aruneshpa

dilyar85 (approved):
LGTM.

bryanv (approved):
LGTM. Can you also look at https://github.com/vmware-tanzu/vm-operator/issues/62 since it is a similar isuse?

If we're going to have to live/support the non-FFS path here for awhile, I'd prefer that in that case, we convert the VMClass Spec.Hardware to the appropriate ConfigSpec early on in `{Create,Update}VirtualMachine` so we don't have this proliferation of FFS checks and dual code paths down in the call stack.

pkg/vmprovider/providers/vsphere/virtualmachine/configspec_test.go
```diff
@@ -1,24 +1,23 @@
-// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package virtualmachine_test
 
 import (
 	goctx "context"
 
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
```
bryanv, line 9: While you're here, instead of shuffling the import around, you can replace where the sink logger is used here with `suite.GetLogger().WithValues("vmName", vm.GetName()),` like we do in other tests so any log output is shown in the tests.

bryanv (approved):

pkg/vmprovider/providers/vsphere/virtualmachine/configspec_test.go
aruneshpa: Ah.  Thanks so much for pointing that out.  Fixed.

sreyasn (approved):
Thanks for handling issue #62

PR #79, opened 2023-03-07
Remove VM image hardware version check in VM validation webhook.

yi0909:
This change removes VM image hardware version check in VM validation webhook. After we enable WCP_VM_Image_Registry FSS, VMI name format has been changed. When PVC is added to an existing VM, which is created from the old format VMI, the validation webhook rejects the request since it fails to find the VMI and check the image hardware version.

To solve this issue, this change introduces a new condition PVCHardwareVersion in VM status. When the volume controller reconciles, it will first check if the HW version is supported when the PVC is sepcified in the VM spec.

Test done:
- [X] make test
- [X] make test-integration
- [X] manually test in WCP cluster:
    ```
    Deployed a VM with PVC:
    root@421eb72411625ac28f829640dd47abe7 [ ~ ]# k describe vm ubuntu-volume -n yijiang-ns
    ...
    Spec:
      Class Name:  best-effort-small
      Image Name:  vmi-8fae851a896a59a92
      Network Interfaces:
        Network Type:  vsphere-distributed
      Power State:     poweredOn
      Storage Class:   wcpglobal-storage-profile
      Volumes:
        Name:  my-pvc
        Persistent Volume Claim:
          Claim Name:  my-pvc
          Read Only:   false
    ...
    Events:
      Type     Reason                   Age                 From

                 Message
      ----     ------                   ----                ----

                 -------
      Warning  VolumeAttachmentFailure  36s (x14 over 77s)  vmware-system-vmop/vmware-system-vmop-controller-manager-7c9d847f57-ckppw/volume-controller          VirtualMachine has an unsupported hardware version 10 for PersistentVolumes. Minimum supported hardware version 13
      Warning  CreateOrUpdateFailure    35s (x14 over 77s)  vmware-system-vmop/vmware-system-vmop-controller-manager-7c9d847f57-ckppw/virtualmachine-controller  status update pending for persistent volume: my-pvc on VM
    ```
- [X] Upgrade test:
    ```
    - Deploy a VM when FSS is disabled.
    - VM is created and powered on.
    - Enable the FSS.
    - Add volume in VM .Spec.
    ...
    Events:
    ...
      Warning  VolumeAttachmentFailure  4s (x27 over 12m)     vmware-system-vmop/vmware-system-vmop-controller-manager-7c69df85b6-fwp8t/volume-controller  VirtualMachine has an unsupported hardware version 10 for PersistentVolumes. Minimum supported hardware version 13
  ```

### DIFF
--- a/controllers/volume/volume_controller.go
+++ b/controllers/volume/volume_controller.go
@@ -34,6 +34,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/lib"
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
+	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/constants"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/instancestorage"
 )
@@ -55,6 +56,7 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
 		ctrl.Log.WithName("controllers").WithName("volume"),
 		record.New(mgr.GetEventRecorderFor(controllerNameLong)),
 		mgr.GetScheme(),
+		ctx.VMProvider,
 	)
 
 	c, err := controller.New(controllerName, mgr, controller.Options{
@@ -96,12 +98,14 @@ func NewReconciler(
 	client client.Client,
 	logger logr.Logger,
 	recorder record.Recorder,
-	scheme *runtime.Scheme) *Reconciler {
+	scheme *runtime.Scheme,
+	vmProvider vmprovider.VirtualMachineProviderInterface) *Reconciler {
 	return &Reconciler{
-		Client:   client,
-		logger:   logger,
-		recorder: recorder,
-		scheme:   scheme,
+		Client:     client,
+		logger:     logger,
+		recorder:   recorder,
+		scheme:     scheme,
+		VMProvider: vmProvider,
 	}
 }
 
@@ -109,9 +113,10 @@ var _ reconcile.Reconciler = &Reconciler{}
 
 type Reconciler struct {
 	client.Client
-	logger   logr.Logger
-	recorder record.Recorder
-	scheme   *runtime.Scheme
+	logger     logr.Logger
+	recorder   record.Recorder
+	scheme     *runtime.Scheme
+	VMProvider vmprovider.VirtualMachineProviderInterface
 }
 
 // +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines,verbs=get;list;watch;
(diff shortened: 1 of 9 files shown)


api/v1alpha1/condition_consts.go
```diff
@@ -41,6 +41,16 @@ const (
 	VirtualMachineImageNotReadyReason = "VirtualMachineImageNotReady"
 )
 
+const (
+	// PVCHardwareVersionCondition exposes the information of the VirtualMachineImage hardware version
+	// specified in the VirtualMachine.
+	PVCHardwareVersionCondition ConditionType = "PVCHardwareVersion"
```
dilyar85, line 47: `PVCHardwareVersionSupported` might be a clearer value for this condition (e.g. when you call `conditions.IsTrue(ctx.VM, vmopv1alpha1.PVCHardwareVersionSupportedCondition)`).

controllers/volume/volume_controller.go
```diff
@@ -532,6 +533,14 @@ func (r *Reconciler) processAttachments(
 			continue
 		}
 
+		// Skip process attachments if the image hardware version is not at least minimal version.
+		if !conditions.IsTrue(ctx.VM, vmopv1alpha1.PVCHardwareVersionCondition) {
+			ctx.Logger.Info("VirtualMachineImage has an unsupported hardware version for PersistentVolumes. " +
+				"Deferring volume attachment")
+			// Do not return error here since this error is unrecoverable until the image hardware version is updated.
+			return nil
+		}
```
dilyar85, line 542: Can we also emit an event to the `ctx.VM` to make it more transparent for users?

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -89,117 +94,159 @@ func GetVMImageStatusAndContentLibraryUUID(
...
-	matchingContentSourceBinding := false
-	for _, csBinding := range csBindingList.Items {
-		if csBinding.ContentSourceRef.Kind == "ContentSource" && csBinding.ContentSourceRef.Name == contentSourceName {
-			matchingContentSourceBinding = true
-			break
+	return vmImageStatus, clUUID, nil
+}
+
+// checkVMImageHWVersion checks if the VirtualMachineImage's HW version is at least the minimum supported virtual hw version.
+func checkVMImageHWVersion(vmCtx context.VirtualMachineContext, imageSpec *vmopv1alpha1.VirtualMachineImageSpec) error {
```
dilyar85, line 222: Instead of having this as a nested function, I suggest making it public and calling it directly from vm_provider after getting the imageSpec. This will help make it easier for someone to locate the code responsible for setting the VM spec condition.

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -89,117 +94,159 @@ func GetVMImageStatusAndContentLibraryUUID(
...
+	imageHardwareVersion := imageSpec.HardwareVersion
+	var hasPVC bool
+
+	for _, vol := range vmCtx.VM.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil {
+			hasPVC = true
```
dilyar85, line 228: break earlier if `hasPVC=true` already?

bryanv:
Do you happen to have an older HW version image handy to try out and see what error is returned by CSI in the CNSNodeVMattachment? By removing the check in the validation webhook - a good thing in itself! - we've accepted the VM CR in the system, and since volume attachment can fail for a myriad of other reasons, it is really worth it for us to try to precheck just this _specific_ reason? Our lives become simpler lump this under the other reasons this can fail, and report the error if it does fail, and we then wouldn't have potentially outdated knowledge baked in of what something else supports or not.

Yes, we need to do a better job to report provision status & any errors overall, and we can use this as an additional reason to prioritize that work (here, I'd imagine we'd rollup the CNSNodeVMattachments' Status into a single Condition)

api/v1alpha1/condition_consts.go
yi0909: Refactored the code, the comment doesn't apply to the current version. Resolve conversion.

controllers/volume/volume_controller.go
```diff
@@ -42,6 +45,10 @@ const (
 	AttributeFirstClassDiskUUID = "diskUUID"
 )
 
+var (
+	hardwareVersionReg = regexp.MustCompile("([0-9]+)")
```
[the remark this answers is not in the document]
yi0909: Used `ParseVirtualHardwareVersion` from contentlibrary package.

controllers/volume/volume_controller.go
```diff
@@ -553,6 +558,31 @@ func (r *Reconciler) processAttachments(
...
 		}
 
+		// If VM hardware version doesn't meet minimal requirement, don't create CNS attachment.
+		// We only fetch the hardware version when this is the first time to create the CNS attachment.
+		//
+		// This change is along with removing VMI hardware version check from VM validation webhook.
+		// We used to deny requests if a PVC is specified in the VM spec while VMI hardware version is not supported.
+		// In this case, no attachments can be created if VM hardware version doesn't meet the requirement.
+		// So, at this point, we can safely assume that all VMs with attachments created have passed the
+		// hardware version check.
```
dilyar85, line 568: Consider placing the highlighted comment after this `if` statement for better clarity.

dilyar85 (approved):

bryanv (approved):

PR #81, opened 2023-03-10
Fix a few ContentLibraryItem controllers issues

bryanv:
Move the RegisterVMIContentSync() into syncImageContent(): an error in CreateOrPatch prior to calling syncImageContent() means syncErr would be nil, so it could count as a success despite never being called. Similarly, in the steady state, the content versions will match so this was basically counting reconciles.

In the cluster item controller, RegisterVMIResourceResolve() would only be called after the createOrPatchErr failure check, so it would never count failures.

Reduce potential log spam around the SecurityCompliance checks, although the behavior of deleting the VMI seems rather odd to me and source of potential confusion.

Don't add the namespace to the NS-scoped item logger twice under different keys

### DIFF
--- a/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
+++ b/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -178,23 +170,22 @@ func (r *Reconciler) ReconcileNormal(ctx goctx.Context, cclItem *imgregv1a1.Clus
...
-		syncErr = r.syncImageContent(ctx, cvmi, cclItem)
+		syncErr = r.syncImageContent(ctx, logger, cvmi, cclItem)
 		// Do not return syncErr here as we still want to patch the updated fields we get above.
 		return nil
 	})
 
 	logger = logger.WithValues("operationResult", opRes)
+	defer func() {
+		// createOrPatchErr can be updated below.
+		r.Metrics.RegisterVMIResourceResolve(logger, cvmi.Name, "", createOrPatchErr == nil)
```
bryanv, line 181: This seems somewhat odd to me to count: in the steady state it is basically just counting reconciles which I don't view as that interesting or useful. Maybe instead, only do it when `OpRes != None` so we actually did a Create/Patch/Update?

dilyar85: In the steady state, this is just a no-op as we set the metrics value to either 1 or 0, indicating the success or failure of a CVMI reconciliation, rather than counting the number of reconciliations. While we could avoid the no-op with a `OpRes != None`, it wouldn't alter the metrics values.

dilyar85 (approved):
Thanks for fixing the RegisterVMIContentSync metrics and improving the reconciliation logic.

dilyar85:
Combined the changes here in https://github.com/vmware-tanzu/vm-operator/pull/92 (with addressing the security compliance check too).

PR #86, opened 2023-03-15
sec: Disable Named Net Provider in Prod

akutz:
This patch disallows the use of the named network provider in production deployments of VM Operator. While used for testing, the named network provider will be rejected unless the VM Operator ConfigMap is explicitly configured to support the named network provider.

This is an important fix as the current state of things violates the security model of Supervisor and Namespace-isolated resources.


dilyar85 (approved):
LGTM.

bryanv (approved):

PR #88, opened 2023-03-15
Add new deploy-wcp script

bryanv:
With the move to Github, the Makefile driver for the previous script was dropped, and that script requires an external script that is not committed to our repo.

With the move to unique creds on our ephemeral testbeds, this script can take an testbedInfo.json - either locally or an URL - to determine the Supervisor info.

This script only updates the image; the yaml changes that the prior script did was too often an annoyance. Just testing a new image is the very, very common case anyways.

I've been carrying this around locally for a while, and is only mildly polished with lots of room for improvements, both in speed and ergonomics, but sample usage:

  $ make docker-build
  $ docker save docker.io/library/vmoperator-controller:latest > vmop.tar
  $ hack/deploy-wcp.sh -T testbedInfo.json vmop.tar


dilyar85 (approved):
Thanks for adding this!

PR #89, opened 2023-03-16
Standardized VM Operator API import alias

bryanv:
In preparation for v1a2 API work use a common import alias - vmopv1 - for the VM Operator API definitions. This is a large, annoying change but will help reduce the diff and work in keeping v1a2 dev work in sync, and will avoid having to do another mass rename for v1a3 and later versions. In addition, during and after the v1a2 work, we'll try to reduce the proliferation of our API types throughout the code base.

Add linter rule to enforce a common alias but still allow unaliased import.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,6 +19,8 @@ linters-settings:
       # Kubernetes
       - pkg: k8s.io/api/core/v1
         alias: corev1
+      - pkg: github.com/vmware-tanzu/vm-operator/api/v1alpha1
+        alias: vmopv1
 
 linters:
   disable-all: true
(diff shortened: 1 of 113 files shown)


aruneshpa:
Is `vmopv1a1` or `vmopv1alpha1` more appropriate?  `vmopv1` might cause a conflict when we introduce `v1`.

bryanv:
> Is `vmopv1a1` or `vmopv1alpha1` more appropriate? `vmopv1` might cause a conflict when we introduce `v1`.

For most of the code whatever the stored version is will be what is needs imported. That's v1a1 today, v1a2 next, and then v1a3, a4, v1beta1, etc. My thought is we'll all be dead before we'd get to a v2 so I'm not worried about that, but we could use like `vmopapi` instead.

aruneshpa (approved):

akutz (approved):
/lgtm, thank you!

sreyasn (approved):

PR #92, opened 2023-03-22
Add finalizer for vmi and cvmi resources

dilyar85:
This patch updates the cl-item & ccl-item controllers by adding/removing a finalizer during their reconciliation process. This ensures that when these resources are deleted from a K8s cluster, the corresponding image metrics are also deleted by calling our ReconcileDelete function. Two `contentlibraryitem_context` and `clustercontentlibraryitem_context` files are added to streamline the reconciling function params in both controllers.

Additionally, the security compliance check of item resources is also updated to avoid deleting the image resources. Instead, a new condition is introduced and added to images, which is verified during the VM deployment process (in the `vmprovider_vm_utils.go` file).

Testing: 
- [x] Add and pass unit tests
- [x] Add and pass integration tests 
- [x] Internal end-to-end tests
- [x] Manually applied this patch to an internal testbed and verified the expected workflow:
```console
# CCL-Item & CVMI
$ kubectl get cclitem clitem-d71052e4b82561b37 -o jsonpath={.metadata.finalizers}
["clustercontentlibraryitem.vmoperator.vmware.com","clustercontentlibraryitem.imageregistry.vmware.com"]
$ curl -s localhost:8083/metrics | grep vmi-d71052e4b82561b37
vmservice_vmi_content_sync{vmi_name="vmi-d71052e4b82561b37",vmi_namespace=""} 1
vmservice_vmi_resource_resolve{vmi_name="vmi-d71052e4b82561b37",vmi_namespace=""} 1
$ kubectl delete cclitem clitem-d71052e4b82561b37
clustercontentlibraryitem.imageregistry.vmware.com "clitem-d71052e4b82561b37" deleted
$ curl -s localhost:8083/metrics | grep vmi-d71052e4b82561b37
$

# CL-Item & VMI
$ kubectl get clitem -n sdiliyaer-test-2 clitem-018c0d3a6ebdd07ca -o jsonpath={.metadata.finalizers}
["contentlibraryitem.imageregistry.vmware.com","contentlibraryitem.vmoperator.vmware.com"]
$ kubectl get vmi -n sdiliyaer-test-2 vmi-018c0d3a6ebdd07ca -o json | jq .status.conditions
[
  {
    "lastTransitionTime": "2023-03-29T03:16:19Z",
    "status": "True",
    "type": "VirtualMachineImageProviderReady"
  },
  {
    "lastTransitionTime": "2023-03-29T15:56:04Z",
    "status": "True",
    "type": "VirtualMachineImageProviderSecurityCompliance"
  },
  {
    "lastTransitionTime": "2023-03-29T04:05:15Z",
    "status": "True",
    "type": "VirtualMachineImageSynced"
  },
  {
    "lastTransitionTime": "2023-03-29T04:05:15Z",
    "status": "True",
    "type": "VirtualMachineImageV1Alpha1Compatible"
  }
]

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -104,6 +104,9 @@ const (
 
 	// VirtualMachineImageProviderReadyCondition denotes readiness of the VirtualMachineImage provider.
 	VirtualMachineImageProviderReadyCondition ConditionType = "VirtualMachineImageProviderReady"
+
+	// VirtualMachineImageProviderSecurityComplianceCondition denotes security compliance of the library item provider.
+	VirtualMachineImageProviderSecurityComplianceCondition ConditionType = "VirtualMachineImageProviderSecurityCompliance"
 )
 
 // Condition.Reason for Conditions related to VirtualMachineImages.
@@ -124,4 +127,8 @@ const (
 	// VirtualMachineImageProviderNotReadyReason (Severity=Error) documents that the VirtualMachineImage provider
 	// is not in ready state.
 	VirtualMachineImageProviderNotReadyReason = "VirtualMachineImageProviderNotReady"
+
+	// VirtualMachineImageProviderSecurityNotCompliantReason (Severity=Error) documents that the
+	// VirtualMachineImage provider doesn't meet security compliance requirements.
+	VirtualMachineImageProviderSecurityNotCompliantReason = "VirtualMachineImageProviderSecurityNotCompliant"
 )
--- a/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
+++ b/controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

**Review by aruneshpa — APPROVED** [review]

File: controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go [hunk 1]
```diff
@@ -159,9 +155,12 @@ func (r *Reconciler) ReconcileNormal(ctx goctx.Context, cclItem *imgregv1a1.Clus
 			savedStatus = cvmi.Status.DeepCopy()
 		}()
 
-		if utils.CheckItemReadyCondition(cclItem.Status.Conditions) {
-			conditions.MarkTrue(cvmi, vmopv1.VirtualMachineImageProviderReadyCondition)
-		} else {
+		if !controllerutil.ContainsFinalizer(cvmi, utils.ClusterVirtualMachineImageFinalizer) {
```
bryanv, line 158: This doesn't look correct to me: this controller watches the ClusterContentLibraryItem, and that's what needs our finalizer if we need to ensure that the ClusterContentLibraryItem is not removed before we do our cleanup. Here, by adding the finalizer on the ClusterVirtualMachineImage, the CCLI Owner could go away but the image is orphaned b/c the finalizer is still present.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller_intg_test.go
```diff
@@ -94,27 +87,6 @@ func cclItemReconcile() {
...
-			Expect(ctx.Client.DeleteAllOf(ctx, &imgregv1a1.ClusterContentLibraryItem{})).Should(Succeed())
-			Eventually(func() bool {
-				cclItemList := &imgregv1a1.ClusterContentLibraryItemList{}
-				if err := ctx.Client.List(ctx, cclItemList); err != nil {
-					return false
-				}
-				return len(cclItemList.Items) == 0
-			}).Should(BeTrue())
-
-			// Manually delete CVMI as envTest doesn't delete it even if the OwnerReference is set up.
```
zyiyi11, line 107: So we don't cleanup CVMI after the test?

zyiyi11 (approved):

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller_intg_test.go
dilyar85: It's not needed anymore because we are deploying a new item with a unique name in each spec.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller_intg_test.go
```diff
@@ -40,25 +41,17 @@ func cclItemReconcile() {
 			}
 			utils.PopulateRuntimeFieldsTo(expectedCVMI, &curCVMI)
 			return reflect.DeepEqual(curCVMI.OwnerReferences, expectedCVMI.OwnerReferences) &&
```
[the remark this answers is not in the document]
dilyar85: Ah, I wished I had known this earlier to save a lot of debugging time... Updated both test files as suggested.

controllers/contentlibrary/contentlibraryitem/contentlibraryitem_controller.go
```diff
@@ -114,9 +114,18 @@ func (r *Reconciler) ReconcileDelete(clItem *imgregv1a1.ContentLibraryItem) erro
...
 	}
 
-	r.Metrics.DeleteMetrics(logger, vmiName, clItem.Namespace)
+	vmi := &vmopv1.VirtualMachineImage{}
+	namespace := clItem.Namespace
+	if err := r.Get(ctx, types.NamespacedName{Name: vmiName, Namespace: namespace}, vmi); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	r.Metrics.DeleteMetrics(logger, vmiName, namespace)
```
yi0909, line 123: Should this be moved inside the `if controllerutil.ContainsFinalizer(vmi, utils.VirtualMachineImageFinalizer)` blocker?

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -111,9 +112,17 @@ func (r *Reconciler) ReconcileDelete(cclItem *imgregv1a1.ClusterContentLibraryIt
 		return nil
 	}
 
+	cvmi := &vmopv1.ClusterVirtualMachineImage{}
+	if err := r.Get(ctx, types.NamespacedName{Name: cvmiName}, cvmi); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
 	r.Metrics.DeleteMetrics(logger, cvmiName, "")
```
yi0909, line 120: Should this be moved inside finalizer check if block?

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
bryanv: `ReconcileDelete()` is only guaranteed to be reached if the `Get()` in `Reconcile()` succeeds. A controller can ensure that succeeds - and is able to observe `!DeletionTimestamp.IsZero()` - by adding its own finalizer on the object being watched. Here, yes, the OwnerRef would cause GC to set DeletionTimestamp on the image when the item is marked for delete, but then you're racing with the image registry controller to observe the delete timestamp before it updates the item, removing its finalizer, and then causing the item removal from etcd. You can experiment with this behavior by creating a parent/child CMs and using `kubectl delete/edit` to play the role of controller (the OwnerRef BlockOwnerDeletion only applies to foreground deletes).

> To address this, I updated the PR to set the controller reference first before adding the finalizer

I don't get why this order would matter. Both fields are a part of the ObjectMeta, and would be sent to the api-server together regardless.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller_intg_test.go
```diff
@@ -33,32 +33,24 @@ func cclItemReconcile() {
 	waitForClusterVirtualMachineImageReady := func(ctx *builder.IntegrationTestContext) {
 		curCVMI := vmopv1.ClusterVirtualMachineImage{}
 		expectedCVMI := utils.GetExpectedCVMIFrom(*cclItem, intgFakeVMProvider.SyncVirtualMachineImageFn)
-		Eventually(func() bool {
+		Eventually(func(g Gomega) error {
 			cvmiName := utils.GetTestVMINameFrom(cclItem.Name)
 			if err := ctx.Client.Get(ctx, client.ObjectKey{Name: cvmiName}, &curCVMI); err != nil {
```
bryanv, line 38: You can simplify this more by `g.Expect(err).ToNot(HaveOccurred())` here, and then change the func's return type to unit/void. Gomega will treat a non-assertion function as succeeding then.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -111,9 +112,17 @@ func (r *Reconciler) ReconcileDelete(cclItem *imgregv1a1.ClusterContentLibraryIt
 		return nil
 	}
 
+	cvmi := &vmopv1.ClusterVirtualMachineImage{}
+	if err := r.Get(ctx, types.NamespacedName{Name: cvmiName}, cvmi); err != nil {
```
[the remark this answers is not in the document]
dilyar85: Is there any preference between them? I am asking this because I noticed internally it has `type ObjectKey = types.NamespacedName` defined in the client library. Looks like we could have one less package (types) to import so I am going to update as suggested.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
dilyar85: Ditto.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
dilyar85: Ah, I didn't realize that K8s API uses background cascading deletion by default. What you said above totally makes sense now and I have updated the PR to add the finalizer in cl/ccl-item resources as suggested.

Regarding the image deletion case I mentioned earlier (where the item itself is not deleted), I figured it should be fine since the controller will recreate the image once it reconciles the item resource.

> I don't get why this order would matter.

In the previous implementation, the owner reference wouldn't be set if the item condition wasn't ready while leaving the image with a finalizer. This should no longer be a problem as the finalizer is added to item resources now. However, I still kept this change so that we always set up the image spec/status based on existing item resources, and update the image condition afterwards.

controllers/contentlibrary/contentlibraryitem/contentlibraryitem_controller.go
dilyar85: I ended up putting it inside with moving the finalizer to item resources.

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -159,63 +154,82 @@ func (r *Reconciler) ReconcileNormal(ctx goctx.Context, cclItem *imgregv1a1.Clus
...
+			ctx.Logger.Info("ClusterContentLibraryItem is not ready yet, skipping further reconciliation")
 			return nil
 		}
+		conditions.MarkTrue(cvmi, vmopv1.VirtualMachineImageProviderReadyCondition)
 
-		if err := r.setUpCVMIFromCCLItem(cvmi, cclItem); err != nil {
-			logger.Error(err, "Failed to set up ClusterVirtualMachineImage from ClusterContentLibraryItem")
-			return err
+		// Check if the item is security compliant.
+		cclItemSecurityCompliance := ctx.CCLItem.Status.SecurityCompliance
```
bryanv, line 177: Can this change on an existing item? Should the `syncImageContent` be done so the rest of the image status is updated if there are other item changes?

api/v1alpha1/condition_consts.go
```diff
@@ -104,6 +104,9 @@ const (
 
 	// VirtualMachineImageProviderReadyCondition denotes readiness of the VirtualMachineImage provider.
 	VirtualMachineImageProviderReadyCondition ConditionType = "VirtualMachineImageProviderReady"
+
+	// VirtualMachineImageProviderSecurityComplianceCondition denotes security compliance of the library item provider.
+	VirtualMachineImageProviderSecurityComplianceCondition ConditionType = "VirtualMachineImageProviderSecurityCompliance"
```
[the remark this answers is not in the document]
bryanv: This was probably just when the v1a2 branch was first cut. I can sync these over later, since there are other image fields missing in v1a2 like `ContentVersion` that we'll require.

bryanv (approved):

controllers/contentlibrary/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
dilyar85: Yep. Moved it inside the `setUpCVMIFromCCLItem` function to make things clearer here.

dilyar85:
Updated the testing-done section with the changes of moving finalizers to cl/ccl-item resources.

PR #127, opened 2023-03-22
Fix VirtualMachinePublishRequest integration test race.

yi0909:
We were using Update in VMPublishRequest integration tests to force trigger a reconcile. This may cause race since vmPub controller is also trying to update the resource. This change uses Patch instead of Update to fix such race.

### DIFF
--- a/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_intg_test.go
+++ b/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_intg_test.go
@@ -15,6 +15,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/google/uuid"
 	"github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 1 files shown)


controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_intg_test.go
```diff
@@ -164,8 +164,9 @@ func virtualMachinePublishRequestReconcile() {
 
 					// Force trigger a reconcile.
 					vmPubObj := getVirtualMachinePublishRequest(ctx, client.ObjectKeyFromObject(vmpub))
+					beforeVMPubObj := vmPubObj.DeepCopy()
 					vmPubObj.Annotations = map[string]string{"dummy": "dummy-1"}
-					Expect(ctx.Client.Update(ctx, vmPubObj)).To(Succeed())
+					Expect(ctx.Client.Patch(ctx, vmPubObj, client.MergeFrom(beforeVMPubObj))).To(Succeed())
```
akutz, line 169: Why not use https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrPatch? Same for below?

yi0909: Updated.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_intg_test.go
```diff
@@ -6,6 +6,7 @@ package virtualmachinepublishrequest_test
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
```
aruneshpa, line 9: Order these?

aruneshpa (approved):
LGTM.

PR #128, opened 2023-03-24
Add initial v1alpha2 conversion

bryanv:
Much of this is autogenerated by conversion-gen, with the remaining manual conversions at least stubbed out so it compiles. There are many TODOs remaining, mainly around the VirtualMachine Spec/Status since that has changed the most in v1a2. We'll likely have to resort to the CAPI trick of sorting the prior version deser to JSON in the object's annotations to provide full v1a1<->v1a2. Note that from some of our types, based on how they are used and who - us - creates and updates them we can likely get by with lossy conversions.

Import conversion fuzz testing from CAPI. Include fuzz tests for VirtualMachineService and VirtualMachineSetResourcePolicy since they are able to be converted by just the autogenerated code.


aruneshpa (approved):
LGTM

zyiyi11 (approved):
LGTM

PR #129, opened 2023-03-25
Remove VirtualMachineClassBinding logic

yi0909:
When WCP_Namespaced_Class_And_Windows_Support is enabled, VM Class will be namespace scoped. This change removes VirtualMachineClassBinding logic from VM Operator repo.

Test done:
- [X] make test
- [X] make test-integration
- [X] verified on a testbed

### DIFF
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2019-2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package virtualmachine
@@ -62,13 +62,19 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
 
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(controlledType).
-		WithOptions(controller.Options{MaxConcurrentReconciles: ctx.MaxConcurrentReconciles}).
-		Watches(&source.Kind{Type: &vmopv1.VirtualMachineClassBinding{}},
-			handler.EnqueueRequestsFromMapFunc(classBindingToVMMapperFn(ctx, r.Client)))
+		WithOptions(controller.Options{MaxConcurrentReconciles: ctx.MaxConcurrentReconciles})
 
 	if !lib.IsWCPVMImageRegistryEnabled() {
 		builder = builder.Watches(&source.Kind{Type: &vmopv1.ContentSourceBinding{}},
 			handler.EnqueueRequestsFromMapFunc(csBindingToVMMapperFn(ctx, r.Client)))
+	}
+
+	if !lib.IsNamespacedClassAndWindowsFSSEnabled() {
+		builder = builder.Watches(&source.Kind{Type: &vmopv1.VirtualMachineClassBinding{}},
+			handler.EnqueueRequestsFromMapFunc(classBindingToVMMapperFn(ctx, r.Client)))
+	} else {
+		builder = builder.Watches(&source.Kind{Type: &vmopv1.VirtualMachineClass{}},
+			handler.EnqueueRequestsFromMapFunc(classToVMMapperFn(ctx, r.Client)))
 	}
 
 	return builder.Complete(r)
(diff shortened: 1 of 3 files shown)


controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -165,6 +171,39 @@ func classBindingToVMMapperFn(ctx *context.ControllerManagerContext, c client.Cl
...
+		// Find all vms that match this vmclass
+		vmList := &vmopv1.VirtualMachineList{}
+		if err := c.List(ctx, vmList, client.InNamespace(class.Namespace)); err != nil {
+			logger.Error(err, "Failed to list VirtualMachines for reconciliation due to VirtualMachineClassBinding watch")
+			return nil
+		}
+
+		// Populate reconcile requests for vms matching the classbinding reference
+		var reconcileRequests []reconcile.Request
+		for _, vm := range vmList.Items {
```
aruneshpa, line 195: Hmm.  With this rework, it will be much easier to only watch for Create and Delete events for a VM Class and only queue reconcile for those events.  That might need more discussion since it will be a behavior change (although a desirable one).  Can you either file a story, or start a conversation on this?

controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -165,6 +171,39 @@ func classBindingToVMMapperFn(ctx *context.ControllerManagerContext, c client.Cl
...
+// for the VirtualMachines in response to an event on the VirtualMachineClass resource when
+// WCP_Namespaced_Class_And_Windows_Support is enabled.
+func classToVMMapperFn(ctx *context.ControllerManagerContext, c client.Client) func(o client.Object) []reconcile.Request {
+	// For a given VirtualMachineClass, return reconcile requests
+	// for those VirtualMachines with corresponding VirtualMachinesClasses referenced
+	return func(o client.Object) []reconcile.Request {
+		class := o.(*vmopv1.VirtualMachineClass)
+		logger := ctx.Logger.WithValues("name", class.Name, "namespace", class.Namespace)
+
+		logger.V(4).Info("Reconciling all VMs referencing a VM class because of a VirtualMachineClass watch")
```
aruneshpa, line 184: If the mapper function is also called after resync period (every 10 minutes?) then on scale systems, it would result in log spew for _all_ VM classes even though they did not result in any reconciles.  What do you think about bumping the log level even further down?  

Separately, L202 is critical so I would vote for bumping that up.  It might help knowing the what VM class change resulted in a reconcile.

controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -165,6 +171,39 @@ func classBindingToVMMapperFn(ctx *context.ControllerManagerContext, c client.Cl
...
+
+		// Find all vms that match this vmclass
+		vmList := &vmopv1.VirtualMachineList{}
+		if err := c.List(ctx, vmList, client.InNamespace(class.Namespace)); err != nil {
+			logger.Error(err, "Failed to list VirtualMachines for reconciliation due to VirtualMachineClassBinding watch")
```
aruneshpa, line 189: copy/pasta

```suggestion
			logger.Error(err, "Failed to list VirtualMachines for reconciliation due to VirtualMachineClass watch")
```

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -37,6 +37,9 @@ func GetVirtualMachineClass(
 	k8sClient ctrlclient.Client) (*vmopv1.VirtualMachineClass, error) {
 
 	className := vmCtx.VM.Spec.ClassName
+	if lib.IsNamespacedClassAndWindowsFSSEnabled() {
```
aruneshpa, line 40: You can dedupe some of this and the following block so that you don't need the separate `resolveVMClass` func?

```golang
key := client.ObjectKey{Name: className}

if lib.IsNamespacedClassAndWindowsFSSEnabled() {
        // namespace scoped VM classes
        key.Namespace = ctx.VM.Namespace
}
// and then go on to fetch the class
vmClass := &vmopv1.VirtualMachineClass{}
if err := k8sClient.Get(vmCtx, ctrlclient.ObjectKey{Name: className}, vmClass); err != nil {
        msg := fmt.Sprintf("Failed to get VirtualMachineClass: %s", className)
	conditions.MarkFalse(vmCtx.VM,
	        vmopv1.VirtualMachinePrereqReadyCondition,
	        vmopv1.VirtualMachineClassNotFoundReason,
	        vmopv1.ConditionSeverityError,
	        msg)
        return nil, errors.Wrap(err, msg)
}

Comment (yi0909) [reply pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go:40]: I have thought about this before, but since we need to return immediately after getting the VM Class when the FSS is enabled, I feel like have a separate function may be more readable? Otherwise, we need to check the FSS twice, one for the key, another for return immediately check.

File: controllers/virtualmachine/virtualmachine_controller.go (hunk 2 above)
Comment (yi0909) [reply controllers/virtualmachine/virtualmachine_controller.go:184]: Yeah, I bumped up to level `Info`, which is the default level.

File: pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go [hunk 5]
```diff
@@ -49,6 +54,10 @@ func GetVirtualMachineClass(
 		return nil, errors.Wrap(err, msg)
 	}
 
+	if lib.IsNamespacedClassAndWindowsFSSEnabled() {
```
aruneshpa, line 57: Nit: maybe add a comment here indicating why we don't need to check the bindings etc.

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -328,6 +337,23 @@ func GetVMClassConfigSpec(raw json.RawMessage) (*types.VirtualMachineConfigSpec,
 	return classConfigSpec, nil
 }
 
+func resolveVMClass(ctx context.VirtualMachineContext,
```
aruneshpa, line 340: Nuke this?

aruneshpa (approved):

dilyar85 (approved):
LGTM.

bryanv (approved):

PR #133, opened 2023-04-04
Fix CRD webhook configuration

bryanv:
CRD webhook patches had become either stale or incorrect over time as we don't use them for our webhooks. These however are needed for the ctrl-runtime version conversion webhooks.

Add the SetupWebhookWithManager() methods for each type that's either a Hub() or Spoke() to enable version conversion. Note that the ctrl-runtime managed webhooks also allow for validation and mutation callbacks instead of our webhooks/ builder stuff.

Remove CRD patches for deprecated content library related CRDs.


zyiyi11 (approved):
/LGTM
Just wonder what WEBHOOK_CERTIFICATE_NAME are we using?

bryanv:
> /LGTM Just wonder what WEBHOOK_CERTIFICATE_NAME are we using?

It will be vmware-system-vmop. That annotation is used by cert-manager to populate the caBundle field.

PR #137, opened 2023-04-11
Fix VirtualMachineService selectorless behavior

bryanv:
Previously a selectorless VirtualMachineService would end up selecting every VM in the namespace, which is not the expected behavior.

Improve the VM EnqueueRequestsFromMapFunc handler by not repeating listing all the VMs that match each service. Instead, list the services once and see if the changed VM would have matched.

### DIFF
--- a/controllers/virtualmachineservice/virtualmachineservice_controller_unit_test.go
+++ b/controllers/virtualmachineservice/virtualmachineservice_controller_unit_test.go
@@ -408,12 +408,13 @@ func unitTestsReconcile() {
 
 		Context("Creates expected Endpoints", func() {
 			var endpoints *corev1.Endpoints
-			var labelSelector map[string]string
+			var labelSelector, vmLabels map[string]string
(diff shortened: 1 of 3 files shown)


controllers/virtualmachineservice/virtualmachineservice_controller_unit_test.go
```diff
@@ -651,6 +652,69 @@ func unitTestsReconcile() {
...
+				Expect(err).To(HaveOccurred())
+				Expect(errors.IsNotFound(err)).To(BeTrue())
+			})
+
+			Context("When VirtualMachineService becomes selectorless", func() {
+				BeforeEach(func() {
+					vmService.Spec.Selector = labelSelector
+				})
+
+				/* This matches the k8s Service behavior when it changes to selectorless. */
```
dilyar85, line 706: Thanks for clarifying this.

dilyar85 (approved):
LGTM!

PR #139, opened 2023-04-13
Fix up computeCPUMinFrequency for the createVM path and add some CPU/Mem Reservation tests

sreyasn:
This change to use VM Class spec for resource reservation and limits before computing min frequency was done earlier for the update path but missed out for create.

Adding some cpu/mem reservation tests to validate the same.

### DIFF
--- a/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
+++ b/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
@@ -247,6 +247,43 @@ func vmTests() {
 				})
 			})
 
+			Context("VM Class spec CPU reservation & limits are non-zero and ConfigSpec specifies CPU reservation", func() {
+				BeforeEach(func() {
+					vmClass.Spec.Policies.Resources.Requests.Cpu = resource.MustParse("2")
+					vmClass.Spec.Policies.Resources.Limits.Cpu = resource.MustParse("2")
+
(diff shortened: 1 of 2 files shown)


pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -247,6 +247,44 @@ func vmTests() {
...
 
+			Context("VM Class spec CPU reservation & limits are non-zero and ConfigSpec specifies CPU reservation", func() {
+				BeforeEach(func() {
+					vmClass.Spec.Policies.Resources.Requests.Cpu = resource.MustParse("2")
+					vmClass.Spec.Policies.Resources.Limits.Cpu = resource.MustParse("2")
+
+					// Specify a CPU reservation via ConfigSpec
+					rsv := int64(6)
+					configSpec = &types.VirtualMachineConfigSpec{
+						Name: "dummy-VM",
```
bryanv, line 258: It is set on other `VirtualMachineConfigSpec` in this file but `Name` isn't actually required right?

bryanv (approved):

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
sreyasn: yeah. it's not. it's always overwritten by the VM spec name.

PR #142, opened 2023-05-02
Various v1a2 VirtualMachine CRD updates

bryanv:
Add InstanceStorage support back in the same manner that it is done in v1a1

Support default provisioning options. Remove unused v1a2 VirtualMachineVolumeProvisioningOptions struct

Convert PowerState enum values

Full version conversion fidelity for VirtualMachines is still some ways off


akutz (approved):
lgtm, thank you!

PR #143, opened 2023-05-03
Initial pass at v1a2 VirtualMachine CRD doc updates

bryanv:
Update shorten CloudInit docs link to chase new docs URL

Update VM IP printcolumn path to new IP4 field since we do not have just one "primary" IP field anymore

Remove Severity from Conditions comment since metav1.Conditions do not have that field

Add various description boilerplate


dilyar85 (approved):
LGTM!

PR #152, opened 2023-05-24
Power State Enhancements

akutz:

This patch updates a VM's power lifecycle.

* Now it is possible to shutdown a VM via a graceful, guest shutdown instead of a halt.
* It is also possible to suspend and resume a VM.
* Finally, it is no longer necessary to specify the power state when creating a new VM. If omitted on creation, the power state defaults to poweredOn. This default behavior only occurs upon a VM's creation.
* The GoVmomi dependency was updated to the latest version that has support for the `StandbyGuest` API both in vC Sim and `object.VirtualMachine`.
* ~~The exported symbols for the power state constants in v1a1 were renamed to match those in v1a2. This does not impact anything on the wire, only external projects that may be linking to this one at compile time and referencing those symbols. Those projects will just need to update to new symbols when linking to a new version of this module at compile time.~~ (Skipping this for now to avoid complicating the outstanding v1a2 work BV is doing)

Fixes `NA`

`NA`

```release-note
Support for gracefully shutting down & suspending VMs as well as treating a VM's powerState as optional when creating a new VM, defaulting to poweredOn.
```

----
:books: Documentation preview :books:: https://vm-operator--152.org.readthedocs.build/en/152/


pkg/vmprovider/providers/vsphere/resources/vm.go
```diff
@@ -158,22 +162,27 @@ func (vm *VirtualMachine) SetPowerState(ctx context.Context, desiredPowerState v
...
+	if te, ok := err.(task.Error); ok {
+		// Ignore error if VM was already in desired state.
+		if ips, ok := te.Fault().(*types.InvalidPowerStateFault); ok && ips.ExistingState == ips.RequestedState {
+			return nil
+		}
+	}
+	if taskInfo != nil {
+		err = errors.Wrapf(err, "%s failed", taskInfo.Name)
+	}
+	vm.logger.Error(err, "VM change power state task failed", "state", desiredPowerState)
```
srm09, line 184: nit: It will be helpful to know what is the power state transition by just looking at the logs instead of having to look 
```suggestion
	vm.logger.Error(err, "VM change power state task failed", "existing-state", existingPowerState, "desired-state", desiredPowerState)
```

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -126,6 +128,20 @@ func (m mutator) vmFromUnstructured(obj runtime.Unstructured) (*vmopv1.VirtualMa
 	return vm, nil
 }
 
+// SetDefaultPowerState sets the default power state for a new VM.
+// Return true if default power state is set, otherwise false.
```
srm09, line 132: nit: To make it explicit that the function returns true only if the power state was modified during the course of the function execution.
```suggestion
// Return true if the default power state was set, otherwise false.
```

srm09 (commented):
Coupe of minor nits, otherwise lgtm

srm09:
@akutz Are the API Enum changes omitted from the v1a2 types for a particular reason?

akutz:
> @akutz Are the API Enum changes omitted from the v1a2 types for a particular reason?

The enum is [already there](https://github.com/vmware-tanzu/vm-operator/blob/e8822bff22aa38e37ab5855925c784af370c8e08/api/v1alpha2/virtualmachine_types.go#L85-L106) in v1a2. They are not camel cased there due to K8s enum casing rules.

srm09 (approved):

akutz:
* nit: Need to update validating webhook to refuse transition from suspended to poweredOff or guestPoweredOff.
* check: Verify nothing needs to be updated in the logic that surfaces the VM's power state in the VM status

akutz:
:wave: @aruneshpa  / @srm09,

What are y'all's thoughts on changing `guestPoweredOff` to simply `shutdown`?

api/v1alpha1/virtualmachine_types.go
```diff
@@ -320,8 +320,14 @@ type VirtualMachineSpec struct {
...
-	// PowerState describes the desired power state of a VirtualMachine.  Valid power states are "poweredOff" and "poweredOn".
-	PowerState VirtualMachinePowerState `json:"powerState"`
+	// PowerState describes the desired power state of a VirtualMachine.
+	//
+	// Please note this field may be omitted when creating a new VM and will
+	// default to "poweredOn." However, once the field is set to a non-empty
+	// value, it may no longer be set to an empty value.
+	//
+	// +optional
+	PowerState VirtualMachinePowerState `json:"powerState,omitempty"`
```
aruneshpa, line 330: From a backward compatibility POV, changing a mandatory field to an optional one should be allowed (since we are making the API more relaxed).  But can this be considered a behavior change -- since this technically breaks any automation that relies on the webhook returning error for a spec that did not contain `powerState`?

akutz:
> * nit: Need to update validating webhook to refuse transition from suspended to poweredOff or guestPoweredOff.
This has been handled in the latest change to the PR.
 
>     * check: Verify nothing needs to be updated in the logic that surfaces the VM's power state in the VM status
Verified that the code that sets the status is fine:

```go
vm.Status.PowerState = vmopv1.VirtualMachinePowerState(summary.Runtime.PowerState)
```

This will reflect `poweredOn`, `poweredOff`, or `suspended`. This _will_ need to be updated in the v1a2 variant since K8s enum values for power state are no longer camel cased.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -475,6 +470,111 @@ func (v validator) validateReadinessProbe(ctx *context.WebhookRequestContext, vm
...
+	ctx *context.WebhookRequestContext,
+	newVM *vmopv1.VirtualMachine) field.ErrorList {
+
+	var (
+		allErrs       field.ErrorList
+		newPowerState = newVM.Spec.PowerState
+	)
+
+	switch newPowerState {
+	case vmopv1.VirtualMachineGuestPoweredOff, vmopv1.VirtualMachineSuspended:
```
aruneshpa, line 483: Should this also be part of the API commentary on the `powerState` field that a VM can only be created in powered on and powered off state (self explanatory, but still)?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -611,3 +711,12 @@ func (v validator) isNetworkRestrictedForReadinessProbe(ctx *context.WebhookRequ
 
 	return configMap.Data[isRestrictedNetworkKey] == "true", nil
 }
+
+func equalsOneOf[T comparable](a T, set ...T) bool {
```
aruneshpa, line 715: I don't see this being referenced anywhere in the diff?

api/v1alpha1/virtualmachine_types.go
akutz: Fair point. But, realistically, the response to such an error would be to correct it, right? The only impact is if it was corrected to poweredOff in someone's logic. I think the likelyhood for impact is minimal to none and worth the value it adds to the overall UX. Thoughts?

pkg/vmprovider/providers/vsphere/resources/vm.go
```diff
@@ -138,42 +138,64 @@ func (vm *VirtualMachine) UniqueID(ctx context.Context) (string, error) {
...
-		return err
+	taskInfo, err := powerTask.WaitForResult(ctx, nil)
+	if err == nil {
+		return nil
 	}
 
-	return nil
+	if te, ok := err.(task.Error); ok {
+		// Ignore error if VM was already in desired state.
+		if ips, ok := te.Fault().(*types.InvalidPowerStateFault); ok && ips.ExistingState == ips.RequestedState {
```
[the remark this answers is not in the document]
akutz: Maybe? I did not attempt to optimize this function, but it's a fair point. Let me take a look.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -801,15 +801,25 @@ func (s *Session) UpdateVirtualMachine(
 		}
 	}()
 
-	isOff := moVM.Runtime.PowerState == vimTypes.VirtualMachinePowerStatePoweredOff
+	// Translate the VM's current power state into the VM Op power state value.
+	var existingPowerState vmopv1.VirtualMachinePowerState
```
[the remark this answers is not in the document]
akutz: There is no equivalent power state on the backend for `vmopv1.VirtualMachineGuestPoweredOff` It's just `vimTypes.VirtualMachinePoweredOff`. Thus `existingPowerState` will never be empty. Maybe I am missing something?

pkg/vmprovider/providers/vsphere/resources/vm.go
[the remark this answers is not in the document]
akutz: Ah, and the answer is no, we should not. After looking at it again, remember that we only know the power state when we read it -- it could have changed. The best thing to do is to take action and handle the result. This ensures we attempt to implement the desired state regardless of what the observed state was at one time.

webhooks/virtualmachine/validation/virtualmachine_validator.go
akutz: Yeah, I need to remove it. I did used it once, and then stopped.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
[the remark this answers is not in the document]
aruneshpa: Okay.  I missed that.  I thought each of these mapped to a distinct state on vCenter.  But it doesn't appear to be so.
https://vdc-repo.vmware.com/vmwb-repository/dcr-public/1ef6c336-7bef-477d-b9bb-caa1767d7e30/82521f49-9d9a-42b7-b19b-9e6cd9b30db1/vim.VirtualMachine.PowerState.html.

Resolving.

api/v1alpha1/virtualmachine_types.go
aruneshpa: Yeah, fair enough.  The value add here is definitely substantial compared to the off chance that someone has built an automation around the behavior of our webhooks.

pkg/vmprovider/providers/vsphere/resources/vm.go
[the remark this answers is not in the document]
aruneshpa: Ah.  Okay.  I did not consider that the state of the VM could have changed on vCenter out-of-band of VM operator (though unsupported).  Good point.

aruneshpa (approved):
LGTM.

akutz:
@bryanv Can you please verify this does not break the way you are handling deletes? I notice it has its own power off function logic, separate from the normal reconcile loop.

pkg/vmprovider/providers/vsphere/virtualmachine/delete.go
```diff
@@ -23,7 +24,11 @@ func DeleteVirtualMachine(
 	// Only a powered off VM can be destroyed.
 	if state != types.VirtualMachinePowerStatePoweredOff {
 		vmCtx.Logger.Info("Powering off VM prior to destroy", "currentState", state)
-		if err := ChangePowerState(vmCtx, vcVM, types.VirtualMachinePowerStatePoweredOff); err != nil {
+		if err := ChangePowerState(
+			vmCtx,
+			vcVM,
+			types.VirtualMachinePowerStatePoweredOff,
+			vmCtx.VM.Spec.PowerOffMode == vmopv1.VirtualMachineShutdown); err != nil {
```
bryanv, line 31: I think is reasonable in the case of a volume that is going to be reused to let it sync.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -475,6 +471,98 @@ func (v validator) validateReadinessProbe(ctx *context.WebhookRequestContext, vm
...
+					powerStatePath,
+					newPowerState,
+					fmt.Sprintf(invalidPowerStateOnUpdateFmt, msg, "powered off")))
+		}
+
+	case vmopv1.VirtualMachineSuspended:
+		// Mark the power state as invalid if the request is attempting to
+		// halt or shutdown a VM that is suspended.
+		var msg string
+		if newPowerState == vmopv1.VirtualMachinePoweredOff {
```
bryanv, line 550: I believe Suspended -> PowerOff is a supported transition (my local ESXI allows it), and it seems reasonable if you cannot power on the VM due to like resource constraints (or just don't want to) power off should still be an option to free up the suspend state.

pkg/vmprovider/providers/vsphere/virtualmachine/powerstate.go
```diff
@@ -32,6 +37,10 @@ func ChangePowerState(
 		return errors.Wrapf(err, "failed task creation to change power state to %s", ps)
```
bryanv, line 37: Nit: if `ShutdownGuest()` fails this error is now a little misleading. Mostly implementation detail so ignore comment if you don't feel like shuffle code around or whatnot.

api/v1alpha2/virtualmachine_types.go
```diff
@@ -173,10 +181,30 @@ type VirtualMachineSpec struct {
 
 	// PowerState describes the desired power state of a VirtualMachine.
 	//
+	// Please note this field may be omitted when creating a new VM and will
+	// default to "PoweredOn." However, once the field is set to a non-empty
+	// value, it may no longer be set to an empty value.
+	//
+	// Additionally, setting this value to "Shutdown" or "Suspended" is not
```
bryanv, line 188: "Shutdown" is not one of the allowed `VirtualMachinePowerState` values. Maybe was in a prior revision before `VirtualMachinePowerOffMode` was introduced?

One could quickly edit the PowerState before the VM has been reconciled/created on VC so I believe our code still needs to handle a just created-never powered on VM being suspended (I presume by just reporting it as powered off - would want to double check that we don't get into a reconcile error loop).

api/v1alpha1/virtualmachine_types.go
```diff
@@ -320,8 +328,31 @@ type VirtualMachineSpec struct {
...
+	// creating a new VM. The valid values when creating a new
+	// VM are "poweredOn" and "poweredOff." An empty value is also allowed on
+	// create since this value defaults to "poweredOn" for new VMs.
+	//
+	// +optional
+	PowerState VirtualMachinePowerState `json:"powerState,omitempty"`
+
+	// PowerOffMode describes the desired behavior when powering off a VM.
+	//
+	// By default a VM is powered off with the equivalent of a halt operation,
```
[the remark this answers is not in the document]
akutz: I had thought of TryGraceful as well, but the problem with changing the default now is it would greatly impact TKG without them changing something. A graceful shutdown of a Kube node takes a rather long time.

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -65,15 +63,39 @@ func convert_v1alpha2_VirtualMachinePowerState_To_v1alpha1_VirtualMachinePowerSt
...
-		return "guestPoweredOff"
 	case v1alpha2.VirtualMachinePowerStateSuspended:
-		return "suspended"
+		return VirtualMachineSuspended
 	}
 
 	return VirtualMachinePowerState(in)
 }
 
+func convert_v1alpha1_VirtualMachinePowerOffMode_To_v1alpha2_VirtualMachinePowerOffMode(
```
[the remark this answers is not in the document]
akutz: Ack. I manually created the funcs since it was just two.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -819,43 +829,44 @@ func (s *Session) UpdateVirtualMachine(
...
+		if err != nil {
+			return err
+		}
 
-			err = s.prepareVMForPowerOn(vmCtx, resVM, config, updateArgs)
-			if err != nil {
-				return err
-			}
+		// TODO: Find a better place for this?
+		if err := s.attachClusterModule(vmCtx, resVM, updateArgs.ResourcePolicy); err != nil {
```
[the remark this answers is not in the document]
akutz: I think it's still there but in a higher diff? I'll double check.

webhooks/virtualmachine/validation/virtualmachine_validator.go
akutz: Ack. I guess it just acts like a halt on a suspended physical system.

api/v1alpha2/virtualmachine_types.go
akutz: Yes, sorry, I meant to remove these. In an earlier revision there was no PowerOffMode but rather `Shutdown` as one of the CRD power states. I changed it so `kubectl delete vm` could also take advantage of the specified power off mode.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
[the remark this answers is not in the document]
akutz: Looking at the diff it wasn't there prior to this change. Want me to reintroduce it?

api/v1alpha1/virtualmachine_types.go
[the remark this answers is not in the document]
akutz: @bryanv How about we change the default behavior for v1a2 then? That way as long as TKG is still pointing to v1a1, they're using halt by default?

pkg/vmprovider/providers/vsphere/virtualmachine/delete.go
akutz: I am not sure I understand your comment. Is this about the default behavior? Again, I agree that in v1a2 it should default to graceful by default, but for v1a1 we should preserve the existing behavior.

api/v1alpha1/virtualmachine_conversion.go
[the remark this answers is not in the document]
akutz: I re-ran the target -- nothing changed.

pkg/vmprovider/providers/vsphere/virtualmachine/powerstate.go
akutz: Resolved by returning a different message for ShutdownGuest.

webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
```diff
@@ -106,5 +106,60 @@ func intgTestsMutating() {
...
+					Expect(ctx.Client.Create(ctx, vm)).To(Succeed())
+				})
+				When("When VM PowerState is empty", func() {
+					It("Should not mutate PowerState", func() {
+						modified := &vmopv1.VirtualMachine{}
+						Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(vm), modified)).Should(Succeed())
+						modified.Spec.PowerState = ""
+						Expect(ctx.Client.Update(ctx, modified)).To(Succeed())
+						Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(vm), modified)).Should(Succeed())
+						Expect(modified.Spec.PowerState).Should(BeEmpty())
```
zyiyi11, line 147: if it's not allowed to mutate powerstate when vm powerstate is empty, shouldn't this line to `Equal(vmopv1.VirtualMachinePoweredOff)`?

webhooks/virtualmachine/mutation/virtualmachine_mutator_unit_test.go
```diff
@@ -214,4 +214,31 @@ func unitTestsMutating() {
...
+
+		Context("When VM PowerState is not empty", func() {
+			BeforeEach(func() {
+				ctx.vm.Spec.PowerState = vmopv1.VirtualMachinePoweredOff
+			})
+
+			It("Should not mutate PowerState", func() {
+				oldVM := ctx.vm.DeepCopy()
+				Expect(mutation.SetDefaultPowerState(&ctx.WebhookRequestContext, ctx.Client, ctx.vm)).To(BeFalse())
+				Expect(ctx.vm.Spec.PowerState).Should(Equal(oldVM.Spec.PowerState))
```
zyiyi11, line 240: Is there a reason for introducing oldVM instead of using `vmopv1.VirtualMachinePoweredOff`?

akutz: Because this way it is not necessary to know the original value of the power state and the code could be copied elsewhere and re-used.

webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
akutz: Good catch! I'll add a code comment here. Basically there's no better way to validate this. This is specifically to validate that on update the value is not _automatically_ set to `poweredOn` if it is empty. Technically the original state is not possible, but I forced it to validate the update logic.

akutz:
This PR passed the appropriate test pipelines internally. Going to merge and can revert later if necessary.

PR #155, opened 2023-06-01
✨ Support Restart

akutz:

This patch introduces support for restarting a VM by setting `spec.nextRestartTime` to one or more values that end up getting compared to the last time a VM was restarted.

A special :tada: shoutout to @hickeng for suggesting the idea of a time-based means for expressing a declarative action in a desired state API model.

Fixes `NA`

None.

```release-note
Support restarting a VM using spec.nextRestartTime.
```

----
:books: Documentation preview :books:: https://vm-operator--155.org.readthedocs.build/en/155/


pkg/util/vsphere/vm/power_state.go
```diff
@@ -370,35 +400,341 @@ func doAndWaitOnSoftPowerOp(
...
+		// Do not restart if there is a recorded, last restart time that occurs
+		// after the desired last restart time.
+		if currentLastRestartTime.After(desiredLastRestartTime) {
+			log.V(5).Info(
+				"will not restart VM as desiredLastRestartTime < currentLastRestartTime",
+				"currentPowerState", currentPowerState,
+				"currentLastRestartTime", currentLastRestartTimeString,
+				"desiredLastRestartTime", desiredLastRestartTimeString,
+				"desiredPowerOpBehavior", powerOpBehavior.String())
+			return PowerOpResultNone, nil
```
dilyar85, line 537: I understand that an error isn't returned here to avoid unnecessary reconciling. However, this means that users won't know if they specify `vm.spec.lastRestart` before `currentLastRestartTime`. I wonder if we could submit a K8s event or have a `vm.status.lastRestart` field to display this information for a better user experience?

akutz: > vm.status.lastRestart

I was going to add this prior to requesting a review, so good catch! :)

pkg/util/vsphere/vm/power_state.go
```diff
@@ -370,35 +400,341 @@ func doAndWaitOnSoftPowerOp(
...
+		}
+	}
+
+	log.V(5).Info(
+		"restarting VM",
```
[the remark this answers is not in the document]
akutz: @srm09 and I discussed this, and we think the margin of error is acceptable. I rewrote the API description to account for this:

```go
	// NextRestartTime indicates the VM should be restarted at or after the time
	// specified.
	//
	// The value may be expressed as an RFC3339 formatted timestamp string,
	// ex. 2006-01-02T15:04:05-07:00, the string "now" (case-insensitive), or
	// a string that can be parsed by time.ParseDuration, ex. 5m or 1h30m, which
	// is then added to the current time.
	//
	// The VM is not guaranteed to be restarted exactly at the specified time.
	// Rather the VM is restarted when the specified value occurs after the,
	// time the VM was last restarted. Once this condition is met, the
	// VM will be restarted the next time it is reconciled, which should never
	// be more than N time, where N is the controller's resync period.
	//
	// Please note that regardless of any specified timezone, the value will be
	// mutated to reflect UTC.
	//
	// +optional
	NextRestartTime string `json:"nextRestartTime,omitempty"`
```

pkg/util/vsphere/vm/power_state.go
```diff
@@ -59,16 +60,16 @@ func (b PowerOpBehavior) String() string {
...
+// PowerOpResult or error.
+type ErrInvalidPowerOpResult struct {
+	PowerOpResult any
 }
 
 // Error enables this type to be returned as a Golang error object.
-func (e ErrInvalidSetPowerStateResult) Error() string {
-	return fmt.Sprintf("invalid result: %q", e.SetPowerStateResult)
+func (e ErrInvalidPowerOpResult) Error() string {
+	return fmt.Sprintf("invalid result: %q", e.PowerOpResult)
```
srm09, line 72: Am conflicted on the `PowerOpBehavior` naming convention, it specifically handles the behavior when the VM is not running i.e. Suspend/Power Off. It does not describe the intent correctly.

Something like `PowerDownOpBehavior`, just an example, might be closer to the intent of the enum.

srm09 (commented):
minor nits, and waiting on the changes to allow setting the `nestRestartTime` to `now` (only) when the VM is powered on.

api/v1alpha1/virtualmachine_types.go
```diff
@@ -401,6 +401,41 @@ type VirtualMachineSpec struct {
...
+	// interaction inside of the guest. The soft mode requires the VM's guest to
+	// have VM Tools installed and asks the guest to restart the VM. Its
+	// variant, trySoft, first attempts a soft restart, and if that fails or
+	// does not complete within five minutes, the VM is hard reset.
+	//
+	// If omitted, the mode defaults to hard.
+	//
+	// +optional
+	// +kubebuilder:default=hard
+	RestartMode VirtualMachinePowerOpMode `json:"restartMode,omitempty"`
```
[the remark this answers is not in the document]
akutz: I have no idea what the suggested change actually is.

pkg/util/vsphere/vm/power_state.go
akutz: It's a superset of the existing VMODL type [`VirtualMachinePowerOpType`](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/1ef6c336-7bef-477d-b9bb-caa1767d7e30/82521f49-9d9a-42b7-b19b-9e6cd9b30db1/vim.vm.DefaultPowerOpInfo.PowerOpType.html) to include the value `trySoft`.

api/v1alpha1/virtualmachine_types.go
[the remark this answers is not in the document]
akutz: Marking resolved since I took the suggestion from the first sentence. If there is more, please let me know.

srm09 (approved):

PR #160, opened 2023-06-08
 🐛 Set Hardware versions for VMs created with PVC/PCI devices

sreyasn:
Set a minimum hardware version for VMs created with PVC/PCI devices when no hardware version is specified in the VM Class
- PVCs == vmx-15
- PVCs and/or PCI (vGPU/DDPIO) devices == vmx-17
- If the VM Image has a higher version than the minimum mentioned above, the image version takes precedence

This makes sure that going forward brownfield VMs/workloads with PVCs/PCI devices still have a supported hardware version specified when the VM Class doesn't mention one.

Fixes #
NA

None

```
Set minimum hardware version for VMs created with Persistent Volume Claims or PCI devices (eg Nvidia GPUs) when
the VM Class doesn't mention a hardware version
```

**Testing**:
VM with PCI device (vGPU) gets vmx-17
Screenshot 2023-06-08 at 6 24 11 PM

VM with PVC gets vmx-15
Screenshot 2023-06-08 at 6 25 53 PM

### DIFF
--- a/pkg/util/devices.go
+++ b/pkg/util/devices.go
@@ -135,6 +135,17 @@ func IsDeviceDynamicDirectPathIO(dev vimTypes.BaseVirtualDevice) bool {
 	return false
 }
 
+// SelectVGPU returns a slice of vGPU devices.
+func SelectVGPU(
+	devices []vimTypes.BaseVirtualDevice,
+) []*vimTypes.VirtualPCIPassthrough {
+
+	return SelectDevicesByDeviceAndBackingType[
+		*vimTypes.VirtualPCIPassthrough,
+		*vimTypes.VirtualPCIPassthroughVmiopBackingInfo,
+	](devices)
+}
+
 // SelectDynamicDirectPathIO returns a slice of dynamic direct path I/O devices.
 func SelectDynamicDirectPathIO(
 	devices []vimTypes.BaseVirtualDevice,
--- a/pkg/util/devices_test.go
+++ b/pkg/util/devices_test.go
@@ -6,7 +6,6 @@ package util_test
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
 	vimTypes "github.com/vmware/govmomi/vim25/types"
 
 	"github.com/vmware-tanzu/vm-operator/pkg/util"
@@ -138,6 +137,29 @@ var _ = Describe("SelectDynamicDirectPathIO", func() {
 			Expect(devOut[0]).To(BeEquivalentTo(newPCIPassthroughDevice("")))
 			Expect(devOut[1].Backing).To(BeAssignableToTypeOf(&vimTypes.VirtualPCIPassthroughDynamicBackingInfo{}))
 			Expect(devOut[1]).To(BeEquivalentTo(newPCIPassthroughDevice("")))
+		})
+	})
+})
+
+var _ = Describe("SelectVGPU", func() {
+	Context("selecting a vGPU device", func() {
+		It("will return only the selected device type", func() {
+			devOut := util.SelectVGPU(
+				[]vimTypes.BaseVirtualDevice{
+					newPCIPassthroughDevice(""),
+					&vimTypes.VirtualVmxnet3{},
+					newPCIPassthroughDevice("profile1"),
+					&vimTypes.VirtualSriovEthernetCard{},
+					newPCIPassthroughDevice(""),
+					newPCIPassthroughDevice("profile2"),
+				},
+			)
+			Expect(devOut).To(BeAssignableToTypeOf([]*vimTypes.VirtualPCIPassthrough{}))
+			Expect(devOut).To(HaveLen(2))
+			Expect(devOut[0].Backing).To(BeAssignableToTypeOf(&vimTypes.VirtualPCIPassthroughVmiopBackingInfo{}))
+			Expect(devOut[0]).To(BeEquivalentTo(newPCIPassthroughDevice("profile1")))
+			Expect(devOut[1].Backing).To(BeAssignableToTypeOf(&vimTypes.VirtualPCIPassthroughVmiopBackingInfo{}))
+			Expect(devOut[1]).To(BeEquivalentTo(newPCIPassthroughDevice("profile2")))
 		})
 	})
 })
--- a/pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
+++ b/pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
@@ -15,11 +15,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/vmware/govmomi/vim25/types"
 
-	imgregv1a1 "github.com/vmware-tanzu/vm-operator/external/image-registry/api/v1alpha1"
-
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
-
 	clutils "github.com/vmware-tanzu/vm-operator/controllers/contentlibrary/utils"
+	imgregv1a1 "github.com/vmware-tanzu/vm-operator/external/image-registry/api/v1alpha1"
 	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/lib"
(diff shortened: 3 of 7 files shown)


pkg/util/devices_test.go
```diff
@@ -142,6 +144,56 @@ var _ = Describe("SelectDynamicDirectPathIO", func() {
...
+					newPCIPassthroughDevice(""),
+					&vimTypes.VirtualVmxnet3{},
+					newPCIPassthroughDevice("profile1"),
+					&vimTypes.VirtualSriovEthernetCard{},
+					newPCIPassthroughDevice(""),
+					newPCIPassthroughDevice("profile2"),
+				},
+			)
+			Expect(devOut).To(BeAssignableToTypeOf([]*vimTypes.VirtualPCIPassthrough{}))
+			Expect(devOut).To(HaveLen(2))
```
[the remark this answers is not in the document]
sreyasn: Because of how newPCIPassthroughDevice() util function is working, It returns back a DDPIO device when the profile is "" and a vGPU when the profile is non-empty.

pkg/util/devices.go
```diff
@@ -146,6 +159,16 @@ func SelectDynamicDirectPathIO(
 	](devices)
 }
 
+// HasPVC returns true if the VirtualMachine spec has a Persistent Volume claim.
```
akutz, line 162: Please remove this function. This package should not link to *any* VM Op API packages. If this function is required, move it to where it is called as an anon function.

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -99,21 +97,21 @@ func GetVirtualMachineClass(
...
-func GetVMImageStatusAndContentLibraryUUID(
+func GetVMImageAndContentLibraryUUID(
 	vmCtx context.VirtualMachineContext,
-	k8sClient ctrlclient.Client) (*vmopv1.VirtualMachineImageStatus, string, error) {
+	k8sClient ctrlclient.Client) (*vmopv1.VirtualMachineImageSpec, *vmopv1.VirtualMachineImageStatus, string, error) {
 
 	imageName := vmCtx.VM.Spec.ImageName
 	if lib.IsWCPVMImageRegistryEnabled() {
-		vmImageStatus, err := resolveVMImageStatus(vmCtx, k8sClient, imageName)
+		vmImageSpec, vmImageStatus, err := resolveVMImage(vmCtx, k8sClient, imageName)
```
srm09, line 106: Should we just go ahead and return the `vmImage` object instead of returning the two separately now that we use both?

pkg/util/devices_test.go
```diff
@@ -4,11 +4,13 @@
...
 
 import (
+	corev1 "k8s.io/api/core/v1"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
 	vimTypes "github.com/vmware/govmomi/vim25/types"
 
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
```
akutz, line 13: Please remove any VM Op API import from `pkg/util/...` and its descendants.

srm09 (commented):
Logic seems good to me, minor nit about multiple return values v/s a single one.

akutz (changes requested):
Looks good, thank you! Most of my comments are preference, but for the one about not linking to `./api/...` from `./pkg/util/...`. That is a no-no. I do think you should consider the other feedback as well, but the one thing that must be addressed prior to merge is the import issue. Thanks again!

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
sreyasn: I thought about this but 
-We don't need the vmImageSpec for the Update path at all. Just for the Create path. That's the addition of imageSpec to only the createArgs and not updateArgs
- Making resolveVMImage return the entire vmImage can be done but It's not very simple since we support a Cluster scoped and a namespace scoped VMImage resources (two different vmoperator api types) after the ImageRegistryFSS. Both resource share the same spec/status fields so we return just the spec and status internally..
- The Object metadata for this Images isn't need for us atm.  

We can just keep the diff minimal for this change and not have to go down that path.

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -440,3 +438,28 @@ func resolveContentLibraryUUID(
...
 	return libraryUUID, nil
 }
+
+// HardwareVersionForPVCandPCIDevices returns a hardware version for VMs with PVCs and PCI devices(vGPUs/DDPIO devices)
+// The hardware version is determined based on the below criteria: VMs with
+// - Persistent Volume Claim (PVC) get the max(the image hardware version, minimum supported virtual hardware version for persistent volumes)
+// - vGPUs/DDPIO devices get the max(the image hardware version, minimum supported virtual hardware version for PCI devices)
+// - Both vGPU/DDPIO devices and PVCs get the max(the image hardware version, minimum supported virtual hardware version for PCI devices)
+// - none of the above returns 0.
+func HardwareVersionForPVCandPCIDevices(vmSpec vmopv1.VirtualMachineSpec, imageHWVersion int32, configSpec *types.VirtualMachineConfigSpec) int32 {
```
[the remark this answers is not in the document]
sreyasn: This MR handles `only` VMs that have devices like PVCs/PCI devices when a class config spec hardware version is unset.
All other VMs when a class config spec hardware version isn't set, will still default to the cluster default which we will need to document as a behavior change after this FSS.

So, my question was this floorHardwareVersion logic will not apply to VMs that don't have PVCs and PCI devices (vGPU && DDPIO bespoke) because we decided not to default everything to the image hardware version with this MR.

Planning to take this is up as a follow up if required?. does that sound good. I'll make all the other requested changes for now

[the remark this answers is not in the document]
akutz: nit: `s/This MR/This PR/g`

Your GitLab is showing :smile:

> Planning to take this is up as a follow up if required?. does that sound good. I'll make all the other requested changes for now

Sure, please file a GH issue with an explanation of the change you will handle in the future.

akutz (approved):
Approved, but please file a GH issue to track the work that is being punted. Going forward I would like our floor to be higher than "minimum required."

pkg/util/devices.go
```diff
@@ -135,6 +135,17 @@ func IsDeviceDynamicDirectPathIO(dev vimTypes.BaseVirtualDevice) bool {
 	return false
 }
 
+// SelectVGPU returns a slice of vGPU devices.
```
akutz, line 138: @sreyasn Please file a GH issue to track updating this function name to `SelectVGPUWithVMIOPBacking`.

sreyasn: can vGPUs have other Backing types? I thought they could only have VMIOP?

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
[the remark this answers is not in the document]
sreyasn: Will do. Need some more context before filing an issue. 

> if the VM Op VM does not have PVCs then floorHardwareVersion is imageHWVersion
Do we plan on flooring all VMs with `no PVC` and `no vGPUs/DDPIO` with `no class config spec hardware version` set to the image's hardwareVersion?.  I thought we decided against it and said they will default to the cluster's version because the Image's version doesn't necessarily describe the class's devices anymore.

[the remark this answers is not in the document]
akutz: > Do we plan on flooring all VMs with no PVC and no vGPUs/DDPIO with no class config spec hardware version set to the image's hardwareVersion?. I thought we decided against it and said they will default to the cluster's version because the Image's version doesn't necessarily describe the class's devices anymore.

No clue. That's why I suggest the `GetMinimumCompatibleHardwareVersion` function take floor as an input param. The logic for what consitutes the floor value we send in is determined in our business logic. The logic in the `GetMinimumCompatibleHardwareVersion` function itself is based purely on what vSphere requires.

pkg/util/devices.go
akutz: nVidia GPUs use VMIOP. Intel GPUs use PCI Passthrough with DVX backings. If we wanted to rename the function to `SelectNVidiaVGPU`, that would also be fine, but I did not want to be so specific.

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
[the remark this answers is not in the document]
akutz: FWIW, the `GetMinimumCompatibleHardwareVersion` function as I imagined it would be:

```go
func GetMinimumCompatibleHardwareVersion(
  ctx context.Context,
  floorHardwareVersion int32,
  configSpec types.VirtualMachineConfigSpec) int32
```

The purpose of the `floorHardwareVersion` as an input parameter is so the function can basically return the authoritative hardware version for the provided `configSpec` without needing logic after this function is called. However, I'm also fine if we wanted to do this:

```go
func GetMinimumCompatibleHardwareVersion(
  ctx context.Context,
  configSpec types.VirtualMachineConfigSpec) int32
```

The above _only_ looks at the provided `configSpec` to determine the hardware version and expects all other logic to occur before or after the function is called. Honestly, this probably makes more sense and keeps things isolated in a neater manner.

PR #163, opened 2023-06-13
✨ Support ControllerName for VM Class

akutz:

This patch adds support for specifying the field `spec.controllerName` in a `VirtualMachineClass` resource. If this field is set to `vmoperator.vmware.com/vsphere`, then the VM Operator controller for `VirtualMachine` reconciles all VM resources. If the field is empty, then the environment variable `DEFAULT_VM_CLASS_CONTROLLER_NAME` is used to determine the default controller name. If this environment variable is empty, it defaults to `vmoperator.vmware.com/vsphere`. Otherwise the controller responsible for reconciling VM resources is determined by the field `spec.controllerName` in the VM Class referenced by a VM.

Fixes `NA`

* @srm09 / @aruneshpa -- Should we also skip webhooks for VM resources based on this logic? I opted for `no`, but I wanted your thoughts.
* @bryanv -- I just cannot get `make generate-go-conversions` to work. The field is added in both v1alpha1 and v1alpha2, and it's a `string`. I'm not sure how it gets any simpler.

```release-note
Support spec.controllerName in VirtualMachineClass to enable concurrent controllers for the VM Operator API.
```

----
:books: Documentation preview :books:: https://vm-operator--163.org.readthedocs.build/en/163/

### DIFF
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -31,10 +31,28 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/prober"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
+	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider"
 )
 
-const finalizerName = "virtualmachine.vmoperator.vmware.com"
+const (
+	finalizerName = "virtualmachine.vmoperator.vmware.com"
+
+	// vmClassControllerName is the name of the controller specified in a
+	// VirtualMachineClass resource's field spec.controllerName field that
+	// indicates a VM pointing to that VM Class should be reconciled by this
+	// controller.
+	vmClassControllerName = "vmoperator.vmware.com/vsphere"
+)
+
+var (
+	// isDefaultVMClassController is used to configure watches and predicates
+	// for the controller. If a VirtualMachineClass resource's
+	// spec.controllerName field is missing or empty, this controller will
+	// consider that VM Class as long as isDefaultVMClassController is true.
+	isDefaultVMClassController = vmClassControllerName ==
+		lib.GetDefaultVirtualMachineClassControllerName()
+)
 
 // AddToManager adds this package's controller to the provided manager.
 func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) error {
@@ -76,6 +94,25 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
 		builder = builder.Watches(&source.Kind{Type: &vmopv1.VirtualMachineClass{}},
 			handler.EnqueueRequestsFromMapFunc(classToVMMapperFn(ctx, r.Client)))
 	}
+
+	// Filter any VMs that reference a VM Class with a spec.controllerName set
+	// to a non-empty value other than "vmoperator.vmware.com/vsphere". Please
+	// note that a VM will *not* be filtered if the VM Class does not exist. A
+	// VM is also not filtered if the field spec.controllerName is missing or
+	// empty as long as the default VM Class controller is
+	// "vmoperator.vmware.com/vsphere".
+	builder = builder.WithEventFilter(
+		kubeutil.VMForControllerPredicate(
+			r.Client,
+			// These events can be very verbose, so be careful to not log them
+			// at too high of a log level.
+			r.Logger.WithName("VMForControllerPredicate").V(8),
+			vmClassControllerName,
+			kubeutil.VMForControllerPredicateOptions{
+				MatchIfVMClassNotFound:            true,
+				MatchIfControllerNameFieldEmpty:   isDefaultVMClassController,
+				MatchIfControllerNameFieldMissing: isDefaultVMClassController,
+			}))
 
 	return builder.Complete(r)
 }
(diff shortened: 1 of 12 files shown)


srm09:
Webhooks are responsible for gatekeeping the values for each of the fields, which is a contract that is placed on the API type itself. +1 to keeping the webhook logic intact irrespective of the controller responsible for reconciling it.

srm09 (commented):
Since we are exposing the field `spec.ControllerName` why not default it to vmoperator.vmware.com unless explicitly set to a different value during creation.

akutz:
> Since we are exposing the field `spec.ControllerName` why not default it to vmoperator.vmware.com unless explicitly set to a different value during creation.

Because of clients that use older versions of the `v1alpha1` schema, such as the CSI images that run on TKC worker nodes. They are using client bindings for `v1alpha1` that do not have this field, and so when they do a `GET` and then later an `UPDATE` (they do not patch), this field will not be part of their updated payload and thus be emptied. It is a version of the problem outlined at https://github.com/kubernetes/kubernetes/issues/111703.

bryanv:
> * @bryanv -- I just cannot get `make generate-go-conversions` to work. The field is added in both v1alpha1 and v1alpha2, and it's a `string`. I'm not sure how it gets any simple

If you remove zz_generated.conversion.go, does it get regenerated? For me, it is working locally: on main it updates the committed zz_generated.conversion.go with the recent power state VM spec changes.

srm09 (approved):
LGTM

akutz:
> > * @bryanv -- I just cannot get `make generate-go-conversions` to work. The field is added in both v1alpha1 and v1alpha2, and it's a `string`. I'm not sure how it gets any simple
> 
> If you remove zz_generated.conversion.go, does it get regenerated? For me, it is working locally: on main it updates the committed zz_generated.conversion.go with the recent power state VM spec changes.

```shell
$ rm ./api/v1alpha1/zz_generated.conversion.go
rm: ./api/v1alpha1/zz_generated.conversion.go: No such file or directory

$ make generate-go-conversions
make -C hack/tools conversion-gen
make[1]: Entering directory '/Users/akutz/Projects/vmop/vmop/hack/tools'
go build -tags=vmop_tools -o bin/conversion-gen k8s.io/code-generator/cmd/conversion-gen
make[1]: Leaving directory '/Users/akutz/Projects/vmop/vmop/hack/tools'
hack/tools/bin/conversion-gen \
	--input-dirs=./api/v1alpha1 \
	--output-file-base=zz_generated.conversion \
	--go-header-file=./hack/boilerplate/boilerplate.generatego.txt

$ git status
On branch feature/controller-name
Your branch is up to date with 'origin/feature/controller-name'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    api/v1alpha1/zz_generated.conversion.go
```

Still no joy

aruneshpa:
One thing that is confusing to me is, why does a field in the VM Class determine who reconciles the _VirtualMachine_?  I would have expected this field to go on the `VirtualMachine` resource.  Is it because we do not want to change the VirtualMachine CRD?

I can imagine a use case where two VMs from the same class are reconciled by different controllers.

pkg/util/kube/predicates.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+type vmForVMClassPredicate struct {
+	client         client.Client
+	log            logr.Logger
+	controllerName string
+	opts           VMForControllerPredicateOptions
+}
+
+// Create returns true if the Create event should be processed
```
srm09, line 56:
```suggestion
// Create returns true if the Create event should be processed.
```

controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -77,6 +81,24 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
...
+		kubeutil.VMForControllerPredicate(
+			r.Client,
+			// These events can be very verbose, so be careful to not log them
+			// at too high of a log level.
+			r.Logger.WithName("VMForControllerPredicate").V(8),
+			vmopv1.GroupName,
+			kubeutil.VMForControllerPredicateOptions{
+				MatchIfVMClassNotFound:            true,
+				MatchIfControllerNameFieldEmpty:   true,
+				MatchIfControllerNameFieldMissing: true,
```
srm09, line 99: I like this pattern, it keeps the controller builder logic saner and enables unit testing of the predicate logic if need be. 💯

pkg/util/kube/predicates.go
```diff
@@ -0,0 +1,211 @@
...
+			r.log.Error(
+				err,
+				"Failed to match object for VMForVMClassPredicate.DeleteEvent")
+		}
+		return false
+	}
+	return true
+}
+
+// Update returns true if the Update event should be processed
```
srm09, line 82:
```suggestion
// Update returns true if the Update event should be processed.
```

pkg/util/kube/predicates.go
```diff
@@ -0,0 +1,211 @@
...
+			r.log.Error(
+				err,
+				"Failed to match object for VMForVMClassPredicate.UpdateEvent")
+		}
+		return false
+	}
+	return true
+}
+
+// Generic returns true if the Generic event should be processed
```
srm09, line 95:
```suggestion
// Generic returns true if the Generic event should be processed.
```

srm09 (commented):
Minor nits to the comments for the linter to pass. Otherwise LGTM.

controllers/virtualmachine/virtualmachine_controller.go
akutz: This is also so the current `VirtualMachine` controller can use this predicate in _default_ mode where other VM controllers can re-use this predicate and force filter out classes that have no or an empty `spec.controllerName` field.

akutz:
> One thing that is confusing to me is, why does a field in the VM Class determine who reconciles the _VirtualMachine_? I would have expected this field to go on the `VirtualMachine` resource. Is it because we do not want to change the VirtualMachine CRD?
> 
> I can imagine a use case where two VMs from the same class are reconciled by different controllers.

Because an end-user should not have to know who reconciles a VM resource to create said resource, nor should they necesarily be allowed to make this decision. I view it as an admin decision, hence part of the class. I also eventually intend to use a VM Class to also reference information about how to connect to the platform on which to deploy a VM realized from said class (even if all the classes on Supervisor point to the same information).

PR #167, opened 2023-06-23
✨ Adds new condition to VirtualMachine to the expose state of metadata+userdata

### LINKED ISSUE #148: feature: Use condition to indicate when Cloud-Init metadata/userdata exceeds allowed size

If a user supplies Cloud-Init userdata that exceeds 64KiB once it is gzipped and base64 encoded, the VM should have a condition added to it indicating the userdata is invalid due to size restrictions. The [maximum size of any guestinfo value is 64KiB](https://github.com/vmware/open-vm-tools/blob/42ce53f39be45b7795a9f1adf892af84629b4a19/open-vm-tools/lib/include/guest_msg_def.h#L100-L104).

And because guestinfo values can be stored in the vmx file, the vmx file has a maximum size (by default) of 1MiB, although this is configurable via the property [`tools.setInfo.sizeLimit`](https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-security/GUID-91BF834E-CB92-4014-8CF7-29CE40F3E8A3.html).

The condition added to a VM should also apply to the generated Cloud-Init metadata, both combined with the userdata or separately.

* If userdata exceeds 64KiB, error
* If metdata exceeds 64KiB, error

`NA`

<table>
<tr>
<td></td>
<th>Value</th>
<th>How to Obtain</th>
</tr>
<tr>
<th>Supervisor version</th>
<td><code>NA</code></td>
</tr>
<tr>
<td><code>NA</code></td>
</tr>
<tr>
<th>Kubernetes version</th>
<td><code>NA</code></td>
</tr>
<tr>
<th>VM Operator version</th>
<td><code>NA</code></td>
<td>

```shell
  sort -u | \
  grep vmop | \
  awk -F'/' '{print $3}' | \
  awk -F: '{print $2}'
```

</td>
</tr>
</table>

srm09:
Adds a new condition to the `VirtualMachine` CR to expose the state of the metadata+userdata.

Fixes #148 

```release-note
Adds new condition to VirtualMachine to the expose state of metadata+userdata
```

### DIFF
--- a/api/v1alpha1/condition_consts.go
+++ b/api/v1alpha1/condition_consts.go
@@ -12,6 +12,30 @@ const (
 // Conditions and condition Reasons for the VirtualMachine object.
 
 const (
+	// VirtualMachineMetadataReadyCondition documents the state of the metadata passed to the VirtualMachine.
+	VirtualMachineMetadataReadyCondition ConditionType = "VirtualMachineMetadataReady"
+
+	// VirtualMachineMetadataDuplicatedReason (Severity=Error) documents that the mutually exclusive Secret and ConfigMap are
+	// specified in the VirtualMachineMetadata.
+	//
+	// This validation is performed by the validation webhook, defensively adding the reason.
+	VirtualMachineMetadataDuplicatedReason = "VirtualMachineMetadataDuplicated"
+
+	// VirtualMachineMetadataNotFoundReason (Severity=Error) documents that the Secret or ConfigMap specified in the VirtualMachineSpec
+	// cannot be found.
+	VirtualMachineMetadataNotFoundReason = "VirtualMachineMetadataNotFound"
+
+	// VirtualMachineMetadataFormatInvalidReason (Severity=Error) documents that the format of the supplied metadata is invalid.
+	//
+	// It is used to indicate issues with the formatting of the metadata irrespective of the Transport in use.
+	VirtualMachineMetadataFormatInvalidReason = "VirtualMachineMetadataFormatInvalid"
+
+	// VirtualMachineGuestInfoSizeExceededReason (Severity=Error) documents that the supplied Cloud Init metadata exceeds the
+	// maximum allowed capacity.
+	VirtualMachineGuestInfoSizeExceededReason = "VirtualMachineGuestInfoSizeExceeded"
+)
+
+const (
 	// VirtualMachinePrereqReadyCondition documents that all of a VirtualMachine's prerequisites declared in the spec
 	// (e.g. VirtualMachineClass) are satisfied.
 	VirtualMachinePrereqReadyCondition ConditionType = "VirtualMachinePrereqReady"
--- a/api/v1alpha2/virtualmachine_types.go
+++ b/api/v1alpha2/virtualmachine_types.go
@@ -10,6 +10,32 @@ import (
 	"github.com/vmware-tanzu/vm-operator/api/v1alpha2/common"
 )
 
+// TODO: Since this API version replaces the concept of metadata to BootstrapSpec, rename the types once
+// these reasons before using them in Conditions. Keeping the reasons the same for the sake of consistency.
+const (
+	// VirtualMachineMetadataReadyCondition documents the state of the metadata passed to the VirtualMachine.
+	VirtualMachineMetadataReadyCondition = "VirtualMachineMetadataReady"
+
+	// VirtualMachineMetadataDuplicatedReason (Severity=Error) documents that the mutually exclusive Secret and ConfigMap are
+	// specified in the VirtualMachineMetadata.
+	//
+	// This validation is performed by the validation webhook, defensively adding the reason.
+	VirtualMachineMetadataDuplicatedReason = "VirtualMachineMetadataDuplicated"
+
+	// VirtualMachineMetadataNotFoundReason (Severity=Error) documents that the Secret or ConfigMap specified in the VirtualMachineSpec
+	// cannot be found.
+	VirtualMachineMetadataNotFoundReason = "VirtualMachineMetadataNotFound"
+
+	// VirtualMachineMetadataFormatInvalidReason (Severity=Error) documents that the format of the supplied metadata is invalid.
+	//
+	// It is used to indicate issues with the formatting of the metadata irrespective of the Transport in use.
+	VirtualMachineMetadataFormatInvalidReason = "VirtualMachineMetadataFormatInvalid"
+
+	// VirtualMachineGuestInfoSizeExceededReason (Severity=Error) documents that the supplied Cloud Init metadata exceeds the
+	// maximum allowed capacity.
+	VirtualMachineGuestInfoSizeExceededReason = "VirtualMachineGuestInfoSizeExceeded"
+)
+
 const (
 	// VirtualMachineConditionClassReady indicates that a referenced
 	// VirtualMachineClass is ready.
--- a/pkg/vmprovider/providers/vsphere/constants/constants.go
+++ b/pkg/vmprovider/providers/vsphere/constants/constants.go
@@ -4,6 +4,8 @@
 package constants
 
 import (
+	"github.com/alecthomas/units"
+
 	"github.com/vmware-tanzu/vm-operator/pkg"
 )
 
@@ -71,6 +73,12 @@ const (
 	CloudInitGuestInfoMetadataEncoding = "guestinfo.metadata.encoding"
 	CloudInitGuestInfoUserdata         = "guestinfo.userdata"
 	CloudInitGuestInfoUserdataEncoding = "guestinfo.userdata.encoding"
+
+	// CloudInitGuestInfoMaxSize the maximum allowed size for CloudInit metadata.
+	// As defined in the https://github.com/vmware/open-vm-tools/blob/42ce53f39be45b7795a9f1adf892af84629b4a19/open-vm-tools/lib/include/guest_msg_def.h#L100-L104
+	//
+	// The human-readable value for the constant equates to 64 Kibibytes, or 64KiB.
+	CloudInitGuestInfoMaxSize = 64 * units.KiB
 
 	// InstanceStoragePVCNamePrefix prefix of auto-generated PVC names.
 	InstanceStoragePVCNamePrefix = "instance-pvc-"
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_customization.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_customization.go
@@ -17,6 +17,7 @@ import (
 	apiEquality "k8s.io/apimachinery/pkg/api/equality"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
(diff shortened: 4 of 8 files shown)


pkg/vmprovider/providers/vsphere/session/session_vm_customization.go
```diff
@@ -245,22 +278,33 @@ func customizeCloudInit(
...
 	}
 
-	if err != nil {
+	if errWithReason != nil {
+		conditions.MarkFalse(vmCtx.VM,
+			vmopv1.VirtualMachineMetadataReadyCondition,
+			errWithReason.conditionReason,
+			vmopv1.ConditionSeverityError,
+			errWithReason.Error())
 		return nil, nil, err
```
bryanv, line 308: This is no longer the correct `err`. The pattern we're introducing here - an error wrapper that doesn't implement Error() - makes me concerned that similar bugs will continue to crop up.

In v1a2 - BTW please add the appropriate types to v1a2 too - we have have finer grain conditions, and I've gone a little back and forward with a similar approach here and setting the conditions closer to were the error occurs. I settled the later for now; in various other projects I've seen a `func errWithCondition(...) error` so the usage becomes like `return nil, errWithCondition(obj, MyCondition, "Reason", err)` but I haven't had the time to really think about the ergonomics. I do want us to settle on a consistent pattern of whose responsibility it is to set the condition.

srm09: Since the new type `ErrWithReason` contains an embedded `error` type, the function call `customErr.Error()` is valid and would work. [Here](https://go.dev/play/p/CB6qPRQk1tD) is a golfing playground snippet I tried out to confirm this.

💯 To have a consistent behavior for all conditions, I think we should rally behind a consistent pattern when making the move onto the v1a2 types.

bryanv: Oh, yes, I missed that error was embedded this morning. My comment about the usage is still applicable though: the code is `return nil, nil, err`, which `err` will now always be `nil` at this point because the change introduced `errWithReason`. Having multiple err-vars in scope is too likely for the wrong one to be returned that we should really avoid that.

api/v1alpha1/condition_consts.go
```diff
@@ -11,6 +11,28 @@ const (
 
 // Conditions and condition Reasons for the VirtualMachine object.
```
zyiyi11, line 13: nit: empty line after comment

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -245,7 +256,11 @@ func GetVMMetadata(
 		secret := &corev1.Secret{}
 		err := k8sClient.Get(vmCtx, ctrlclient.ObjectKey{Name: metadata.SecretName, Namespace: vmCtx.VM.Namespace}, secret)
 		if err != nil {
-			// TODO: Condition
+			conditions.MarkFalse(vmCtx.VM,
+				vmopv1.VirtualMachineMetadataReadyCondition,
+				vmopv1.VirtualMachineMetadataNotFoundReason,
+				vmopv1.ConditionSeverityError,
+				"Unable to get metadata Secret %s/%s", metadata.ConfigMapName, vmCtx.VM.Namespace)
```
zyiyi11, line 263: copy/pasta: `SecretName` instead of `ConfigMapName`

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -228,14 +228,25 @@ func GetVMMetadata(
...
+			"Both ConfigMapName %s/%s and SecretName %s/%s are specified",
+			metadata.ConfigMapName, vmCtx.VM.Namespace,
+			metadata.SecretName, vmCtx.VM.Namespace)
 		return vmMD, errors.New("invalid VM Metadata: both ConfigMapName and SecretName are specified")
 	}
 
 	if metadata.ConfigMapName != "" {
 		cm := &corev1.ConfigMap{}
 		err := k8sClient.Get(vmCtx, ctrlclient.ObjectKey{Name: metadata.ConfigMapName, Namespace: vmCtx.VM.Namespace}, cm)
 		if err != nil {
```
zyiyi11, line 244: do we need to check err here to be `IsNotFound` in order to give `VirtualMachineMetadataNotFoundReason`?

zyiyi11 (approved):
Overall LGTM. I like the ErrWithReason implementation. Thanks

pkg/vmprovider/providers/vsphere/vmprovider_vm_utils.go
srm09: Even in the case of another error (say API server not being available), the Reason still holds.

sreyasn (approved):
Change LGTM!. Thanks @srm09.

api/v1alpha1/condition_consts.go
```diff
@@ -11,6 +11,28 @@ const (
...
 
+const (
+	// VirtualMachineMetadataReadyCondition documents the state of the metadata passed to the VirtualMachine.
+	VirtualMachineMetadataReadyCondition ConditionType = "VirtualMachineMetadataReady"
+
+	// VirtualMachineMetadataDuplicatedReason (Severity=Error) documents that the mutually exclusive Secret and ConfigMap are
+	// specified in the VirtualMachineMetadata.
+	//
+	// This validation is performed by the validation webhook, defensively adding the reason.
+	VirtualMachineMetadataDuplicatedReason = "VirtualMachineMetadataDuplicated"
```
aruneshpa, line 22: Hmm.  Is there _any_ scenario where this will be hit though?  IIRC, we have liveness probes that fail the resource creation if the webhooks are down.

api/v1alpha2/virtualmachine_types.go
```diff
@@ -10,6 +10,28 @@ import (
 	"github.com/vmware-tanzu/vm-operator/api/v1alpha2/common"
 )
 
+const (
+	// VirtualMachineMetadataReadyCondition documents the state of the metadata passed to the VirtualMachine.
+	VirtualMachineMetadataReadyCondition = "VirtualMachineMetadataReady"
```
aruneshpa, line 15: I don't think we have a concept of "metadata" in v1a2?  It is called `VirtualMachineBootstrapSpec`.  Or maybe this is more nuanced and I am missing something?

api/v1alpha1/condition_consts.go
```diff
@@ -11,6 +11,28 @@ const (
...
+
+	// VirtualMachineMetadataNotFoundReason (Severity=Error) documents that the Secret or ConfigMap specified in the VirtualMachineSpec
+	// cannot be found.
+	VirtualMachineMetadataNotFoundReason = "VirtualMachineMetadataNotFound"
+
+	// VirtualMachineMetadataInvalidReason (Severity=Error) documents that the supplied Cloud Init metadata is invalid.
```
aruneshpa, line 28: > specified Cloud Init metadata

This is slightly confusing, right?  We _chose_ to call it metadata -- but really, it is userdata that is specified by a field called `vmMetadata`.

I would reword this a little bit to make it clear that this is talking about the vmMetadata resource specified -- and that this is only applicable if the transport type is CloudInit.  Same for the type below.  What do you think?

pkg/vmprovider/providers/vsphere/constants/constants.go
```diff
@@ -72,6 +72,9 @@ const (
 	CloudInitGuestInfoUserdata         = "guestinfo.userdata"
 	CloudInitGuestInfoUserdataEncoding = "guestinfo.userdata.encoding"
 
+	// CloudInitGuestInfoMaxSize the maximum allowed size for cloud init metadata.
+	CloudInitGuestInfoMaxSize = 64 * 1024
```
aruneshpa, line 76: Curious: is there a way to fetch this at a cluster level?  That way, we don't have to maintain this bespoke const in our code and worry about it falling out of sync with what is enforced at the Guest level.

If not, we should at least call out in comments that this is enforced by the infra (ESXi) (and maybe link a public KB) so readers are not wondering how we arrived at this const.

pkg/vmprovider/providers/vsphere/session/session_vm_customization.go
```diff
@@ -147,17 +158,27 @@ func GetCloudInitPrepCustSpec(
...
+			error:           fmt.Errorf("encoding cloud-init metadata failed %v", err),
+			conditionReason: vmopv1.VirtualMachineMetadataInvalidReason,
+		}
 	}
 	extraConfig[constants.CloudInitGuestInfoMetadata] = encodedMetadata
 	extraConfig[constants.CloudInitGuestInfoMetadataEncoding] = "gzip+base64"
 
+	if len(encodedMetadata) > constants.CloudInitGuestInfoMaxSize {
+		return nil, &ErrWithReason{
+			error:           fmt.Errorf("size of cloud-init guest info exceeds the maximum allowed size of 64KiB"),
```
aruneshpa, line 177: Use the const here with the format specifier to plug in the value instead of hardcoding it? 

Same for L217

pkg/vmprovider/providers/vsphere/session/session_vm_customization_test.go
```diff
@@ -402,6 +403,59 @@ var _ = Describe("Cloud-Init Customization", func() {
...
+					Expect(err).ToNot(HaveOccurred())
+					cloudInitMetadata = data
+				})
+				It("will return a limit exceeded error", func() {
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring("exceeds the maximum allowed size"))
+				})
+			})
+
+			Context("With cloud init metadata + userdata exceeding the maximum size", func() {
```
aruneshpa, line 446: Is this a valid scenario though?  IIRC, the limit is enforced by the size of the value of each GuestInfo key.  If both of the keys individually are less than the limit, then we should be fine?

api/v1alpha1/condition_consts.go
```diff
@@ -11,6 +11,28 @@ const (
...
+	VirtualMachineMetadataInvalidReason = "VirtualMachineMetadataInvalid"
```
aruneshpa, line 29: Part of me wants to reserve the "Invalid" key here because it feels too broad.  It might be helpful in future when we decide to validating the specified user-data itself (e.g., that it's format is correct).
WDYT about renaming it `VirtualMachineMetadataFormatInvalid` (or something similar that indicates that the encode/decode fails -- because that's what we are really validating here)?

srm09 (commented):
@aruneshpa made the changes you asked for.

aruneshpa (approved):
LGTM.

pkg/vmprovider/providers/vsphere/session/session_vm_customization.go
```diff
@@ -261,6 +301,12 @@ func customizeCloudInit(
 	}
 
 	if err != nil {
+		errWithReason := err.(ErrWithReason)
```
srm09, line 304: This was supposed to be `errWithReason := err.(*ErrWithReason)`
Will try to add unit tests for this one.

PR #169, opened 2023-06-28
🌱 Only sanitize virtual disks that are non-RDM from class config spec

sreyasn:
When VMClassAsConfig is enabled, this change removes only disks that are not raw device mapping backed. All disk controllers can be added post this FSS.

Testing:
make test

N/A

Fixes #
N/A

N/A

```
Remove only virtual disks without RDM backings from the VM Class when VMClassAsConfig FSS is enabled 
```

### DIFF
--- a/pkg/util/devices.go
+++ b/pkg/util/devices.go
@@ -174,3 +174,13 @@ func isDiskOrDiskController(dev vimTypes.BaseVirtualDevice) bool {
 		return false
 	}
 }
+
+// isNonRDMDisk returns true for all virtual disk devices excluding disks with a raw device mapping backing.
+func isNonRDMDisk(dev vimTypes.BaseVirtualDevice) bool {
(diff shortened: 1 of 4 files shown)


pkg/util/devices.go
```diff
@@ -174,3 +174,13 @@ func isDiskOrDiskController(dev vimTypes.BaseVirtualDevice) bool {
 		return false
 	}
 }
+
+// isNonRDMDisk returns true for all virtual disk devices excluding disks with a raw device mapping backing and other device types.
```
zyiyi11, line 178: what are included in  `and other device types`?

zyiyi11 (approved):
LGTM

pkg/util/devices.go
sreyasn: agreed. That comment is a little confusing. Fixed it.

PR #170, opened 2023-06-30
Pass the none VirtualMachine object to Controller for further processing

yixingjia:
1. For any objects that are not VirtualMachine but watched by VirtualMachine Controller, pass the object to controller for further process instead of report an error.
2. If the controllerName mismatch happens, just return (false, nil) instead of error, as if two controller watch VirtualMachine at the same time, controllerName mismatch will always happen. We shouldn't report an error.

### DIFF
--- a/pkg/util/kube/predicates.go
+++ b/pkg/util/kube/predicates.go
@@ -108,25 +108,31 @@ func (r *vmForVMClassPredicate) Generic(e event.GenericEvent) bool {
 func (r *vmForVMClassPredicate) matches(obj runtime.Object) (bool, error) {
 	obj = obj.DeepCopyObject() // do not mutate the original
 
-	vm, err := r.getVM(obj)
+	object, kind, err := r.getWatchedObj(obj)
 	if err != nil {
 		return false, err
+	}
+
+	// If the object is not VirutalMachine, then pass to the xxxToVMMapperFn
+	// for further processing.
+	if kind != "VirtualMachine" {
+		return true, nil
 	}
 
 	// Get name of the VM Class used by the VM.
 	className, ok, err := unstructured.NestedString(
-		vm.Object, "spec", "className")
+		object.Object, "spec", "className")
 	if err != nil {
 		return false, err
 	}
 	if !ok {
-		return false, fmt.Errorf("spec.className not found for %s", vm)
+		return false, fmt.Errorf("spec.className not found for %s", object)
 	}
 
 	// Get the VM Class referenced by the VM.
 	vmClass, err := r.getVMClass(
-		vm.GroupVersionKind().GroupVersion(),
-		vm.GetNamespace(),
+		object.GroupVersionKind().GroupVersion(),
+		object.GetNamespace(),
 		className)
 	if err != nil {
 		if apierrors.IsNotFound(err) && r.opts.MatchIfVMClassNotFound {
@@ -158,16 +164,14 @@ func (r *vmForVMClassPredicate) matches(obj runtime.Object) (bool, error) {
 	}
 
 	if controllerName != r.controllerName {
-		return false, fmt.Errorf(
-			"spec.controllerName=%q, expected=%q",
-			controllerName, r.controllerName)
+		return false, nil
 	}
 
 	return true, nil
 }
 
-func (r *vmForVMClassPredicate) getVM(
-	inObj runtime.Object) (*unstructured.Unstructured, error) {
+func (r *vmForVMClassPredicate) getWatchedObj(
+	inObj runtime.Object) (*unstructured.Unstructured, string, error) {
 
 	// Controller-runtime does not assign an object's TypeMeta prior to sending
 	// the object into a predicate. In order to ascertain the object's group,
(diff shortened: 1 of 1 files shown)


yixingjia:
@akutz Andrew, Please help review this MR.

srm09 (approved):
Looks good to me, nice catch

pkg/util/kube/predicates.go
```diff
@@ -108,25 +108,30 @@ func (r *vmForVMClassPredicate) Generic(e event.GenericEvent) bool {
 func (r *vmForVMClassPredicate) matches(obj runtime.Object) (bool, error) {
 	obj = obj.DeepCopyObject() // do not mutate the original
 
-	vm, err := r.getVM(obj)
+	object, kind, err := r.getWatchedObj(obj)
```
akutz, line 111: Please rename `object` to `obj` as the latter is clear and more succint.

pkg/util/kube/predicates.go
```diff
@@ -108,25 +108,30 @@ func (r *vmForVMClassPredicate) Generic(e event.GenericEvent) bool {
...
 	if err != nil {
 		return false, err
 	}
 
+	// if the object is not VirutalMachine, then pass to the xxxToVMMapperFn for further processing.
```
akutz, line 116: Please ensure comments wrap at 80chars. Also, please capitalize `If the object` so sentences are grammatically correct.

akutz (changes requested):
Lgtm, just the two small changes please.

pkg/util/kube/predicates.go
yixingjia: Let me know if you prefer other var name.

pkg/util/kube/predicates.go
```diff
@@ -108,25 +108,31 @@ func (r *vmForVMClassPredicate) Generic(e event.GenericEvent) bool {
...
 
-	vm, err := r.getVM(obj)
+	object, kind, err := r.getWatchedObj(obj)
 	if err != nil {
 		return false, err
 	}
 
+	// If the object is not VirutalMachine, then pass to the xxxToVMMapperFn
+	// for further processing.
+	if kind != "VirtualMachine" {
```
bryanv, line 118: Is there a reason this predicate is being used as a EventFilter (global predicate) instead of a predicate just on the specific VirtualMachine For()/Owns()/Watches() in the controller?

pkg/util/kube/predicates.go
```diff
@@ -191,15 +195,13 @@ func (r *vmForVMClassPredicate) getVM(
...
-	if kind := outObj.GetKind(); kind != "VirtualMachine" {
-		return nil, fmt.Errorf("kind=%q, expected kind=VirtualMachine", kind)
-	}
+	kind := outObj.GetKind()
 
-	return &outObj, nil
+	return &outObj, kind, nil
 }
 
 func (r *vmForVMClassPredicate) getVMClass(
```
bryanv, line 207: Closest GH will let me comment: I don't believe the ordering here is quite right. We should check the namespace-scoped first as a I'm pretty sure that a NS-scoped request w/o a namespace will be defaulted to "default" NS.

yixingjia: @akutz  please comment on @bryanv question.

pkg/util/kube/predicates.go
yixingjia: This predicate mainly used for filter out Virtualmachine with a className , and the virtualmachineclass with controllerName that match the specific controller. 

There will be multiple controllers that will watch on virtualMachine in the feature, one example is we are going to use vmoperator to manage VMs on ESXi directly. The existing virtual machine controller will not work and there are a lot of limition on ESXi compare with vcenter.

srm09: @yixingjia The way the event filter is setup, it gets called on every object the controller watches. In this case, it's `VirtualMachineClassBinding`, `ContentSourceBinding` and so on. Rather than excluding all the other types in the EventFilter predicate as done in the MR, we can get rid of these extra checks by just moving the predicate filter to the VirtualMachine objects only. The way to do that would be:
```golang
import (
  ...
  ctrlbuilder "sigs.k8s.io/controller-runtime/pkg/builder"
  ...
)

  builder := ctrl.NewControllerManagedBy(mgr).
  For(controlledType, ctrlbuilder.WithPredicates(kubeutil.VMForControllerPredicate(
			r.Client,
			// These events can be very verbose, so be careful to not log them
			// at too high of a log level.
			r.Logger.WithName("VMForControllerPredicate").V(8),
			vmClassControllerName,
			kubeutil.VMForControllerPredicateOptions{
				MatchIfVMClassNotFound:            true,
				MatchIfControllerNameFieldEmpty:   isDefaultVMClassController,
				MatchIfControllerNameFieldMissing: isDefaultVMClassController,
			}))).
```

zyiyi11:
https://github.com/vmware-tanzu/vm-operator/pull/207 addressed this issue. Closing this MR.

PR #172, opened 2023-07-06
Add v1a2 webhooks

bryanv:
This is the first round of v1a2 webhooks. At present, only unit tests are hooked up on main as integration tests require manifest generation changes that we're not ready for since the integration tests use the same artifacts that are shipped.

Fix golangci-lint config to allow aliased imports

### DIFF
--- a/test/builder/utila2.go
+++ b/test/builder/utila2.go
@@ -1,0 +1,174 @@
+// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package builder
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+)
+
+func DummyVirtualMachineSetResourcePolicyA2() *vmopv1.VirtualMachineSetResourcePolicy {
+	return &vmopv1.VirtualMachineSetResourcePolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "test-",
+		},
+		Spec: vmopv1.VirtualMachineSetResourcePolicySpec{
+			ResourcePool: vmopv1.ResourcePoolSpec{
+				Name: "dummy-resource-pool",
+				Reservations: vmopv1.VirtualMachineResourceSpec{
+					Cpu:    resource.MustParse("1Gi"),
+					Memory: resource.MustParse("2Gi"),
+				},
+				Limits: vmopv1.VirtualMachineResourceSpec{
+					Cpu:    resource.MustParse("2Gi"),
+					Memory: resource.MustParse("4Gi"),
+				},
+			},
+			Folder:              "dummy-folder",
+			ClusterModuleGroups: []string{"dummy-cluster-modules"},
+		},
+	}
+}
+
+func DummyVirtualMachineServiceA2() *vmopv1.VirtualMachineService {
+	return &vmopv1.VirtualMachineService{
+		ObjectMeta: metav1.ObjectMeta{
+			// Using image.GenerateName causes problems with unit tests
+			Name: fmt.Sprintf("test-%s", uuid.New()),
+		},
+		Spec: vmopv1.VirtualMachineServiceSpec{
+			Type: vmopv1.VirtualMachineServiceTypeLoadBalancer,
+			Ports: []vmopv1.VirtualMachineServicePort{
+				{
+					Name:       "dummy-port",
+					Protocol:   "TCP",
+					Port:       42,
+					TargetPort: 4242,
+				},
+			},
+			Selector: map[string]string{
+				"foo": "bar",
+			},
+		},
+	}
+}
+
+func DummyVirtualMachineA2() *vmopv1.VirtualMachine {
+	return &vmopv1.VirtualMachine{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "test-",
+			Labels:       map[string]string{},
+			Annotations:  map[string]string{},
+		},
+		Spec: vmopv1.VirtualMachineSpec{
+			ImageName:  DummyImageName,
+			ClassName:  DummyClassName,
+			PowerState: vmopv1.VirtualMachinePowerStateOn,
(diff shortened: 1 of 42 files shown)


srm09 (commented):
Went through some of them, will go through the rest in another pass.

test/builder/utila2.go
```diff
@@ -0,0 +1,174 @@
...
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
```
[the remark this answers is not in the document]
bryanv: > ... one way to future proof this is ...

I don't see the future proofing here. The transition to v1a2 is made more difficult & disruptive since it addresses years of hacks and lessons learned for v1a1 but for future v1a3+ and later versions, I expect the schema changes to be incremental & relatively small. Maybe I am misunderstand what you're proposing here, but using `vmopv2` would make a later `vmopv3` diff much larger since numerous lines need to change instead of the just import in most cases. The use of a common import alias was an intentional choice at the start of the v1a2 work.

webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/pkg/errors"
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
```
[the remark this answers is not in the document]
bryanv: The current alias was an intentional choice made in https://github.com/vmware-tanzu/vm-operator/pull/89. What we settled on was inspired by CAPI, for better or worse, and changing it now is disruptive and significant churn with all the other v1a2 changes in flight.

webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator_intg_test.go
```diff
@@ -0,0 +1,151 @@
...
+			)
+			BeforeEach(func() {
+				lastRestartTime = time.Now().UTC()
+				modified = &vmopv1.VirtualMachine{}
+			})
+			JustBeforeEach(func() {
+				Expect(ctx.Client.Create(ctx, ctx.vm)).To(Succeed())
+				Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(ctx.vm), modified)).Should(Succeed())
+			})
+			When("spec.nextRestartTime is empty", func() {
```
sreyasn, line 88:
```suggestion
			When("existing spec.nextRestartTime is empty", func() {
```

webhooks/virtualmachineclass/v1alpha2/mutation/virtualmachineclass_mutator.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// vmClassFromUnstructured returns the VirtualMachineClass from the unstructured object.
+func (m mutator) vmClassFromUnstructured(obj runtime.Unstructured) (*vmopv1.VirtualMachineClass, error) {
+	vmClass := &vmopv1.VirtualMachineClass{}
+	if err := m.converter.FromUnstructured(obj.UnstructuredContent(), vmClass); err != nil {
+		return nil, err
+	}
+	return vmClass, nil
+}
```
zyiyi11, line 92: Missing https://github.com/vmware-tanzu/vm-operator/blob/main/webhooks/virtualmachineclass/v1alpha1/mutation/virtualmachineclass_mutator.go#L108. Might need to rebase.

zyiyi11 (commented):
Will keep reviewing tmr

webhooks/virtualmachineclass/v1alpha2/mutation/virtualmachineclass_mutator_unit_test.go
```diff
@@ -0,0 +1,58 @@
...
+		Context("when update request comes in while deletion in progress ", func() {
+			It("should admit update operation", func() {
+				t := metav1.Now()
+				ctx.WebhookRequestContext.Obj.SetDeletionTimestamp(&t)
+				response := ctx.Mutate(&ctx.WebhookRequestContext)
+				Expect(response.Allowed).To(BeTrue())
+			})
+		})
+	})
+}
```
sreyasn, line 58: I noticed the ControllerName related tests have been dropped. Intentional?.

sreyasn (approved):

sreyasn:
BV, Noticed that the [virtualmachinewebconsolerequest](https://github.com/vmware-tanzu/vm-operator/tree/main/webhooks/virtualmachinewebconsolerequest) v1alpha2 webhooks aren't part of this PR?. Intentional?. Is it part of a different change?

webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator_intg_test.go
bryanv: I'm keeping this identical to v1a1 tests

webhooks/virtualmachineclass/v1alpha2/mutation/virtualmachineclass_mutator_unit_test.go
bryanv: Just missed. Thanks.

bryanv:
> BV, Noticed that the [virtualmachinewebconsolerequest](https://github.com/vmware-tanzu/vm-operator/tree/main/webhooks/virtualmachinewebconsolerequest) v1alpha2 webhooks aren't part of this PR?. Intentional?. Is it part of a different change?

There is a separate story tracking this type. In v1a2 the CRD name has changed so conversion isn't possible, meaning we'll have to keep around the old v1a1 type, webhook, and controller. My immediate goal is to get VMOP switched to v1a2  internally while the existing clients/users still use & work correctly with the v1a1 types. As a practical matter the v1a2 VirtualMachineWebconsoleRequest type doesn't really matter until we update the kubectl plugin.

PR #174, opened 2023-07-07
Import v1a2 pkg changes

bryanv:
This contains the non provider pkg/ changes for v1a2, which is most of the changes required to get the v1a2 controllers working.

With the change to metav1.Condition types, the Reason field is now required which previously we wouldn't provide when setting the condition to true. To keep the later changes smaller we reuse the type field value but will need to refactor our conditions API later to add a reason argument and meaningful reason messages where we want to.

Contains selective changes from #172 and #173 required to compile that will rebase out

### DIFF
--- a/pkg/conditions2/getter.go
+++ b/pkg/conditions2/getter.go
@@ -1,0 +1,257 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package conditions2
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+)
+
+const (
+	ReadyConditionType = "Ready"
+)
+
+// Getter interface defines methods that an object should implement in order to
+// use the conditions package for getting conditions.
+type Getter interface {
+	client.Object
+
+	// GetConditions returns the list of conditions for a cluster API object.
+	GetConditions() []metav1.Condition
+}
+
+// Get returns the condition with the given type, if the condition does not exists,
+// it returns nil.
+func Get(from Getter, t string) *metav1.Condition {
+	for _, condition := range from.GetConditions() {
+		if condition.Type == t {
+			return &condition
+		}
+	}
+	return nil
+}
+
+// Has returns true if a condition with the given type exists.
+func Has(from Getter, t string) bool {
+	return Get(from, t) != nil
+}
+
+// IsTrueFromConditions returns true if the condition with the given type is True.
+// This is helpful to check the condition without passing a specific resource object.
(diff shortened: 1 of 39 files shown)


pkg/conditions2/getter.go
```diff
@@ -0,0 +1,257 @@
...
+	fallbackReason  string
+	fallbackMessage string
+}
+
+// MirrorOptions defines an option for mirroring conditions.
+type MirrorOptions func(*mirrorOptions)
+
+// WithFallbackValue specify a fallback value to use in case the mirrored condition does not exists;
+// in case the fallbackValue is false, given values for reason, severity and message will be used.
+func WithFallbackValue(fallbackValue bool, reason string, message string) MirrorOptions {
```
zyiyi11, line 201: Why do we have `conditions2` folder instead of reusing `conditions`?
Why does upstream cluster-api has severity and here we don't?
https://github.com/kubernetes-sigs/cluster-api/blob/main/util/conditions/getter.go#L206C1-L207C1

zyiyi11 (commented):
IIRC, it's only change Reason field to required. Why don't we change in v1alpha1 `pkg/conditions`?  So we don't need this big chunck of `patch2`, `prober2`

pkg/conditions2/getter.go
bryanv: The type is changing from our own (that was more or less copied from CAPI) to k8s `metav1.Conditions`, which do not  have a `severity`

pkg/patch2/patch_test.go
```diff
@@ -0,0 +1,738 @@
...
+					}
+
+					return obj.Status.Host == objAfter.Status.Host &&
+						conditions.IsTrue(objAfter, readyConditionType) &&
+						reflect.DeepEqual(obj.Spec, objAfter.Spec)
+				}, timeout).Should(BeTrue())
+			})
+		})
+
+		/*
```
sreyasn, line 605: remove the commented out tests? . what plans for this?

sreyasn (approved):
Looks good to me.

zyiyi11 (commented):
So to summarize this change, pls correct me if i am wrong:
`conditions2` and `patch2` are using k8s metav1 Conditions,
`prober2` is referring vmopv1 as "github.com/vmware-tanzu/vm-operator/api/v1alpha2"

pkg/conditions2/patch_test.go
```diff
@@ -0,0 +1,285 @@
...
+			name:    "Change: When a condition exists with conflicts but there is agreement on the final state, it should change",
+			before:  getterWithConditions(fooFalse),
+			after:   getterWithConditions(fooTrue),
+			latest:  setterWithConditions(fooTrue),
+			want:    conditionList(fooTrue),
+			wantErr: false,
+		},
+		/*
+				{
+					name:    "Change: When a condition exists with conflicts but there is no agreement on the final state, it should error",
```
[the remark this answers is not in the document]
bryanv: I had skipped these tests on my first pass since they depended on the Severity field, which metav1.Conditions do not have. I updated the tests. Thanks.

pkg/patch2/patch_test.go
bryanv: These are commented out in existing `pkg/patch` tests from CAPI, likely because our types didn't have an observed generation field. I'm trying to keep these diffs small at the moment to just get us close to v1a1 feature parity w/ v1a2. metav1.Conditions do have a generation field so these tests are something that we can work on reenabling. I've added this work to my internal P2 tracking stories.

bryanv:
> So to summarize this change, pls correct me if i am wrong:
> `conditions2` and `patch2` are using k8s metav1 Conditions,
> `prober2` is referring vmopv1 as "github.com/vmware-tanzu/vm-operator/api/v1alpha2"

Yes. In general I'm aiming for the minimal changes to convert these to work with v1a2 while keeping the API the same. It is still rather frustrating though since we have deal with nearly identical code as we try to support both in main.

PR #178, opened 2023-07-13
Move controllers under v1alpha1 sub packages

bryanv:
Like was done with the existing webhooks in 3d5d8dea9, move the existing v1a1 CRD controllers into v1a1 sub directory in preparation of adding the v1a2 controllers. This is unfortunately a large, painful, annoying, disruptive change based on how we're trying to support multiple storageVersions in more or less the same code base.


aruneshpa:
LGTM.

aruneshpa (approved):

PR #180, opened 2023-07-17
Do not allocate NodePorts for VirtualMachineService Services

bryanv:
For VirtualMachineServices we create a selectorless Service and manage
the Endpoints to point to the underlying VirtualMachines so NodePorts are
not required because we don't proxy via the Supervisor CP VMs. Set field
added in k8s 1.24 to disable allocating NodePorts. Make this automatic
for LB-type services. This field is ignored/reset for non-LB type Services.

This addresses a scaling issue where we can exhaust the WCP SV cluster
 out of NodePorts, usually via many TKG Services..

### DIFF
--- a/controllers/virtualmachineservice/virtualmachineservice_controller.go
+++ b/controllers/virtualmachineservice/virtualmachineservice_controller.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -392,6 +393,11 @@ func (r *ReconcileVirtualMachineService) createOrUpdateService(ctx *context.Virt
 		service.Spec.ExternalName = vmService.Spec.ExternalName
 		service.Spec.LoadBalancerIP = vmService.Spec.LoadBalancerIP
 		service.Spec.LoadBalancerSourceRanges = vmService.Spec.LoadBalancerSourceRanges
+		if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
+			service.Spec.AllocateLoadBalancerNodePorts = pointer.Bool(false)
+		} else {
+			service.Spec.AllocateLoadBalancerNodePorts = nil
+		}
 
 		// Parts of the Service.Spec can be updated by k8s after creation, and we need to
 		// preserve those fields.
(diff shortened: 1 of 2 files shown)


controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -392,6 +393,11 @@ func (r *ReconcileVirtualMachineService) createOrUpdateService(ctx *context.Virt
 		service.Spec.ExternalName = vmService.Spec.ExternalName
 		service.Spec.LoadBalancerIP = vmService.Spec.LoadBalancerIP
 		service.Spec.LoadBalancerSourceRanges = vmService.Spec.LoadBalancerSourceRanges
+		if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
```
aruneshpa, line 396: Hmm.  It loos like the nodeports already allocated for the service (before this change) will not get cleaned up automatically (from https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation).  How do we handle that?  Do we have a release note (or some other way) to tell customers to clean up those nodeports manually?  Do customers even have RBAC perms to delete those nodeports?

aruneshpa (commented):
Other than one comment, code looks good.

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -392,6 +393,11 @@ func (r *ReconcileVirtualMachineService) createOrUpdateService(ctx *context.Virt
...
+			service.Spec.AllocateLoadBalancerNodePorts = pointer.Bool(false)
+		} else {
+			service.Spec.AllocateLoadBalancerNodePorts = nil
```
zyiyi11, line 399: Do we need to specify the default behavior?

akutz (approved):
Hey @bryanv, looks good, thank you! I am going to add a label to this PR to indicate that cross/back porting it would be problematic given the field `spec.AllocateLoadBalancerNodePorts` was only marked stable in Kube 1.24. Do you happen to know when it was introduced and available as a gated feature?

controllers/virtualmachineservice/virtualmachineservice_controller.go
bryanv: To be explicit & the CreateOrUpdate() doesn't have an otherwise unexpected change during the steady-state reconcile.

bryanv:
> Hey @bryanv, looks good, thank you! I am going to add a label to this PR to indicate that cross/back porting it would be problematic given the field `spec.AllocateLoadBalancerNodePorts` was only marked stable in Kube 1.24. Do you happen to know when it was introduced and available as a gated feature?

The field was first added during v1.20: https://github.com/kubernetes/kubernetes/commit/1f4d852f2f46ab203ca37f0059506d6f993ae3ea. Recent WCP SV is on 1.26. I did have an concern of our ancient internal Kind based pipeline but that appears to be gone 👏

PR #184, opened 2023-07-25
:sparkles: Ensure lossless conversion of VirtualMachineClass new fields

srm09:
This patch extends the conversion function for the VirtualMachineClass API type to ensure lossless conversion of the new fields in the Status of the v1alpha2 version of the CR.
When down-converting to v1alpha1 version, the extra fields get added to the CR's annotation and this data can be read during up-conversion to v1alpha2 version.

The v1alpha2 version of the `VirtualMachineClass` CR has new fields introduced in the `VirtualMachineClassStatus` type, namely `Ready`, `Conditions` and `Capabilities` which are not present in the v1alpha1 version. When down-converting a v1alpha2 API version, say, via the v1alpha1 versioned client, a straight up conversion will cause these new fields to be lost.

Instead, we store the fields and their JSON marshaled values as an annotation under a special key `vmoperator.vmware.com/conversion` to ensure lossless conversion.  

```release-note
:sparkles: Ensure lossless conversion of VirtualMachineClass new fields
```

### DIFF
--- a/api/utilconversion/conversion.go
+++ b/api/utilconversion/conversion.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// NOTE: These functions are borrowed from upstream cluster-api repo.
+
+package utilconversion
+
+import (
+	"encoding/json"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// AnnotationKey is the annotation key to ensure lossless
+// conversion of objects between API versions.
+const AnnotationKey = "vmoperator.vmware.com/conversion"
+
+// MarshalData stores the source object as json data in the destination object annotations map.
+// It ignores the metadata of the source object.
+func MarshalData(src metav1.Object, dst metav1.Object) error {
+	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(src)
+	if err != nil {
+		return err
+	}
+	delete(u, "metadata")
+
+	data, err := json.Marshal(u)
+	if err != nil {
+		return err
+	}
+	annotations := dst.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[AnnotationKey] = string(data)
+	dst.SetAnnotations(annotations)
+	return nil
+}
+
+// UnmarshalData tries to retrieve the data from the annotation and
+// unmarshals it into the object passed as input.
+func UnmarshalData(from metav1.Object, to interface{}) (bool, error) {
+	annotations := from.GetAnnotations()
+	data, ok := annotations[AnnotationKey]
+	if !ok {
+		return false, nil
+	}
+	if err := json.Unmarshal([]byte(data), to); err != nil {
+		return false, err
+	}
+	delete(annotations, AnnotationKey)
+	from.SetAnnotations(annotations)
+	return true, nil
+}
--- a/api/v1alpha1/conversion_test.go
+++ b/api/v1alpha1/conversion_test.go
@@ -15,6 +15,7 @@ import (
 	runtimeserializer "k8s.io/apimachinery/pkg/runtime/serializer"
 
 	"github.com/vmware-tanzu/vm-operator/api/utilconversion"
+
 	"github.com/vmware-tanzu/vm-operator/api/v1alpha1"
 	nextver "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 )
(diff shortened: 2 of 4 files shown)


api/v1alpha1/conversion_utils.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"encoding/json"
+	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ConversionAnnotationKey is the annotation key to ensure lossless
+// conversion of objects between API versions.
+const ConversionAnnotationKey = "vmoperator.vmware.com/conversion"
```
aruneshpa, line 15: Since this is just an annotation, it can be changed by users (accidentally, or deliberately).  I am curious what the upstream pattern on this is?  Should we have some kind of webhook prevent users from setting this annotation?

api/v1alpha1/conversion_utils.go
```diff
@@ -0,0 +1,38 @@
...
+
+func splitFieldsPath(input string) []string {
+	return strings.Split(input, ".")
+}
+
+// addConversionAnnotation appends the marshalled data to Object as a special
```
aruneshpa, line 21:
```suggestion
// addConversionAnnotation appends the JSON marshaled data to Object as a special
```

api/v1alpha1/conversion_utils.go
```diff
@@ -0,0 +1,38 @@
...
+// prefixed annotation.
+func addConversionAnnotation(obj client.Object, data map[string]interface{}) error {
```
aruneshpa, line 23: Minor Nit: IIRC, the idiomatic way is to return early to avoid indentation.  So maybe:
```go
func addConversionAnnotation(obj client.Object, data map[string]interface{}) error {
	if len(data) == 0 {
		return nil
	}
 	data, err := json.Marshal(data)
 	if err != nil {
 		return err
 	}
 	annotations := obj.GetAnnotations()
 	if annotations == nil {
 		annotations = map[string]string{}
 	}
 	annotations[ConversionAnnotationKey] = string(data)
 	obj.SetAnnotations(annotations)
 	return nil
 }

File: api/v1alpha1/virtualmachineclass_conversion.go [hunk 4]
```diff
@@ -23,13 +27,57 @@ func Convert_v1alpha2_VirtualMachineClassStatus_To_v1alpha1_VirtualMachineClassS
...
+
+	status := dst.Status
+	rawMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&status)
+	if err != nil {
+		return err
+	}
+
+	// Manually restore data
+	var valErrs []error
+	conversionData, ok := src.GetAnnotations()[ConversionAnnotationKey]
```
aruneshpa, line 43: Consider combining the two lines?
```go
if conversionData, ok := ...; ok {

}
```

aruneshpa (approved):
LGTM with nits.

api/v1alpha1/conversion_utils.go
srm09: [Upstream kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) has certain reserved prefixes such as `kubernetes.io/` and `k8s.io` which are reserved for use by core components.
CAPI has a similar prefix `node.cluster.x-k8s.io` which is owned and operated by their components. 

Following convention and similar usages in the codebase, used the annotation prefixed with `vmoperator.vmware.com` to declare the intent that it is used internally by VMOp controllers.

srm09: As for the suggestion of preventing users from setting these annotations, webhooks will intercept all requests and would not differentiate between user generated request v/s conversion function update requests.

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -11,7 +11,7 @@ package v1alpha1
 import (
 	"encoding/json"
 	"k8s.io/api/core/v1"
-	runtime "k8s.io/apimachinery/pkg/runtime"
```
bryanv, line 14: I think you have a slightly stale controller-gen in hack/tools/bin.

api/v1alpha1/conversion_utils.go
```diff
@@ -0,0 +1,40 @@
...
+
+// ConversionAnnotationKey is the annotation key to ensure lossless
+// conversion of objects between API versions.
+const ConversionAnnotationKey = "vmoperator.vmware.com/conversion"
+
+func splitFieldsPath(input string) []string {
+	return strings.Split(input, ".")
+}
+
+// addConversionAnnotation appends the JSON marshalled data to Object as a special
```
bryanv, line 21: Is there a reason we're not cribbing closer from CAPI: https://github.com/kubernetes-sigs/cluster-api/blob/main/util/conversion/conversion.go#L101? Related, why not crib their `UnmarshalData()` instead of doing that in the type conversion, which I think means we'll end up with basically duplicated code once we do the VirtualMachine type.

api/v1alpha1/conversion_test.go
```diff
@@ -145,7 +145,6 @@ func overrideVirtualMachineClassFieldsFuncs(codecs runtimeserializer.CodecFactor
 			// TODO: Need to save serialized object to support lossless conversions.
 			classStatus.Capabilities = nil
```
bryanv, line 146: While here, can we just enable conversion of these fields too? Sure, TBD if we'll end to supporting these in time but doing it now should be pretty low cost w.r.t. the rest of the change.

srm09: Done

Makefile
```diff
@@ -255,6 +255,7 @@ modules-download: ## Downloads and caches the modules
 .PHONY: generate
 generate: ## Generate code
 	$(MAKE) generate-go
+	$(MAKE) generate-go-conversions
```
bryanv, line 258: This had been purposely omitted from here for the time being because it is just so slow (~75sec on my fast laptop), and I didn't want to impact people's existing workflow. I'd prefer we make that target not PHONY before we enable it by default.

api/v1alpha1/zz_generated.deepcopy.go
bryanv: Double check that your MR drops this change before merging :)

bryanv (approved):

Makefile
srm09: Removed this one.

api/utilconversion/conversion.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
srm09: Updated the copyright and added a note that this has been borrowed from upstream CAPI repo.

PR #187, opened 2023-07-26
Add relative binary path to -only .PHONY targets

bryanv:
Without this the presence of an existing binary file, regardless of age w.r.t the rest of the project, won't result in a recompile. This can cause confusion around a using stale binary or whether the current state of the project actually compiles.


srm09:
Thanks for this one, I was facing the exact same issue the other day and I changed the Makefile to **always** remove the binary before building it. This is way neater than my hack 💯

srm09 (approved):

PR #188, opened 2023-07-27
🌱 Rename SelectVGPU taking BackingType into account

### LINKED ISSUE #162: Rename SelectVGPU to SelectVGPUWithVMIOPBacking

[`SelectVGPU`](https://github.com/vmware-tanzu/vm-operator/blob/b12f76d4655ce1f0c169ea7ae880fd1eaa2938f2/pkg/util/devices.go#L139) is ambiguous and needs to be renamed to `SelectVGPUWithVMIOPBacking` or `SelectNvidiaVGPU`
since Intel vGPU use a DVX backing and not VMIOP

NA

sreyasn:
This makes the method SelectVGPU more explicit since vGPUs can have DVX backings as well.

This makes the method more explicit since vGPUs can have VMIOP and DVX backing.

Fixes #162

### DIFF
--- a/pkg/util/devices.go
+++ b/pkg/util/devices.go
@@ -116,8 +116,8 @@ func SelectVirtualPCIPassthrough(
 	return SelectDevicesByType[*vimTypes.VirtualPCIPassthrough](devices)
 }
 
-// IsDeviceVGPU returns true if the provided device is a vGPU.
-func IsDeviceVGPU(dev vimTypes.BaseVirtualDevice) bool {
+// IsDeviceNvidiaVgpu returns true if the provided device is an Nvidia vGPU.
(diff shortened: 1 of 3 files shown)


akutz:
Should we call this `SelectNvidiaVgpu`? It is my understanding that other GPU vendors do not use the VmIop backing, and one could imagine eventually we would want a `SelectIntelVgpu` version of this function, right?

Besides, _technically_ this should be `SelectVirtualPCIPassthroughWithVmIopBacking`, right? So we're already kind of fudging things by calling it `SelectVgpu...` in the first place...

akutz (approved):
lgtm, thanks!

srm09 (approved):

PR #189, opened 2023-07-28
📖  Documentation for VM's resources (CPU, Memory, Storage and Network)

### LINKED ISSUE #105: doc: Update docs/concepts/workloads/vm.md#resources
This ticket tracks updating [`./docs/concepts/workloads/vm.md#resources`](https://github.com/vmware-tanzu/vm-operator/tree/main/docs/concepts/workloads/vm.md#resources). Please refer to https://vm-operator.rtfd.io/en/latest/ref/proj/docs/#project-docs for guidelines on writing VM Operator project documentation.

aruneshpa:
Adds documentation to explain the various resources of a VM such as CPU, memory, storage and network.

Fixes #105 

N/A

```release-note
NONE
```

----
:books: Documentation preview :books:: https://vm-operator--189.org.readthedocs.build/en/189/

### DIFF
--- a/docs/concepts/workloads/vm.md
+++ b/docs/concepts/workloads/vm.md
@@ -84,15 +84,98 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
 
 ## Resources
 
-Some of a VM's hardware resources are derived, and there are some that may be influenced directly by a user.
-
-// TODO ([github.com/vmware-tanzu/vm-operator#105](https://github.com/vmware-tanzu/vm-operator/issues/105))
+Some of a VM's hardware resources are derived from the policies defined by your infrastructure administrator, others may be influenced directly by a user.
 
 ### CPU and Memory
+CPU and memory of a VM are derived form the `VirtualMachineClass` that is used to create the VM. Specifically, the `memoryMB` and `numCPUs` properties in the `configSpec` field of the `VirtualMachineClass` resource dictate the exact number of virtual CPUs, and the virtual memory that the VM will be created with.
+
+Additionally, your administrator might also define certain policies in your VM class in form of resource [reservations](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.resmgmt.doc/GUID-8B88D3D8-E9D9-4C05-A065-B3DE1FFFB401.html) or [limits](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.resmgmt.doc/GUID-117972E3-F5D3-4641-9EAC-F9DD2B0761C3.html). These are vSphere constructs which are used to reserve, or set a ceiling on resources consumed by a VM. Users can view these policies by inspecting the `ConfigSpec.cpuAllocation` and `ConfigSpec.memoryAllocation` fields of the VM Class for CPU and memory reservation/limits information respectively.
+
+For an example, consider the following VM Class:
+```
+apiVersion: vmoperator.vmware.com/v1alpha1
+kind: VirtualMachineClass
+metadata:
+  name: my-vm-class
+spec:
+  configSpec:
+    _typeName: VirtualMachineConfigSpec
+    cpuAllocation:
+      _typeName: ResourceAllocationInfo
+      limit: 400
+      reservation: 200
+    memoryAllocation:
+      _typeName: ResourceAllocationInfo
+      limit: 2048
+      reservation: 1024
+  hardware:
+    cpus: 2
+    memory: 4Gi
+```
+A VM created with this VM class will have 2 vCPUs, 4 GiB of memory. Additionally, the VM will be created with a guaranteed CPU bandwidth of 200MHz and 1024 MB of memory. The CPU bandwidth of the VM will never be allowed to exceed 400MHz and the memory will not exceed 2GB.
+
+!!! note "Units for CPU reservations/limits are in MHz"
+
+    Please note that the units for CPU are different in hardware and reservations/limits.  Virtual hardware is specified in units of vCPUs, whereas CPU reservations/limits are in MHz.  Memory can be specified in MiB, GiB, whereas memory reservations/limits are always in MB.
 
 ### Networking
+Users can optionally specify a list of network interfaces for their VM in the `spec.NetworkInterfaces` field. Each interface must specify the network type which can be either `nsx-t` or `vsphere-distributed`, depending on the network provider being used. Users can further customize each interface with optional fields such as the Ethernet card type, network name etc. If a VM does not specify any network interface, VM operator creates an interface that is connected to the default network available in the namespace. The IP address of the network interface is handled by the available network provider (configured by the administrator).
+
+#### How are a VM's Network Interfaces Configured?
+VM operator uses the network devices specified in the VM Class to configure the network interfaces specified in the VM spec. If it cannot find sufficient number of devices in the VM class, a set of defaults are used to configure the network interfaces.
+
+As an example, consider the following VM Class that specifies two network devices, each with a different Ethernet card type.
+
+```yaml
+apiVersion: vmoperator.vmware.com/v1alpha1
+kind: VirtualMachineClass
+metadata:
+  name: class-with-two-nics
+spec:
+  configSpec:
+    _typeName: VirtualMachineConfigSpec
+    deviceChange:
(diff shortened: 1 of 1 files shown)


docs/concepts/workloads/vm.md
```diff
@@ -84,15 +84,45 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
 
 ## Resources
 
-Some of a VM's hardware resources are derived, and there are some that may be influenced directly by a user.
-
-// TODO ([github.com/vmware-tanzu/vm-operator#105](https://github.com/vmware-tanzu/vm-operator/issues/105))
+Some of a VM's hardware resources are derived from the policies defined by your infrastructure administrator, others may be influenced directly by a user.
 
 ### CPU and Memory
+CPU and memory of a VM are derived form the `VirtualMachineClass` that is used to create the VM. Specifically, the `memoryMB` and `numCPUs` properties in the `configSpec` field of the `VirtualMachineClass` resource dictate the exact number of virtual CPUs, and the virtual memory (in MegaBytes) that the VM will be created with.
```
zyiyi11, line 90: What do you think moving `(in MegaBytes)` to line 117 in `!!! note "Units for CPU reservations/limits are in MHz"`? So the units are in the same place and reader doesn't get distracted here.

docs/concepts/workloads/vm.md
```diff
@@ -84,15 +84,45 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
...
+      reservation: 200
+    memoryAllocation:
+      _typeName: ResourceAllocationInfo
+      limit: 2048
+      reservation: 1024
+  hardware:
+    cpus: 2
+    memory: 4Gi
+```
+A VM created with this VM class will have 2 vCPUs, 4 GiB of memory. Additionally, the VM will be created with a guaranteed 200MHz of CPU bandwidth and 1024 MB of memory. The CPU bandwidth of the VM will never be allowed to exceed 400MHz and the memory will not exceed 4Gi.
```
zyiyi11, line 115: isn't memory limit 2048MB instead of 4Gi?

docs/concepts/workloads/vm.md
```diff
@@ -84,15 +84,45 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
...
+
+!!! note "Units for CPU reservations/limits are in MHz"
+
+Please note that the units for CPU are different in hardware and reservations/limits.  Virtual hardware is specified in units of vCPUs, whereas CPU reservations/limits are in MHz.  Memory can be specified in MiB, GiB, whereas memory reservations/limits are always in MB.
```
zyiyi11, line 119: Maybe highlight `**Note**` here instead of `Please note`

docs/concepts/workloads/vm.md
```diff
@@ -84,15 +84,45 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
...
 
 ### Networking
+Users can specify a list of network interfaces for their VM in the `spec.NetworkInterfaces` field. Each interface must specify the network type (one of `nsx-t`, or `vsphere-distributed` depending on the network provider being used). Users can further customize each interface with optional fields such as the Ethernet card, network name etc. For convenience, even if a VM does not specify any network interface, VM operator creates one connected to the default network available in the namespace. The IP address of the network interface is handled by the available network provider (configured by the administrator).
```
zyiyi11, line 123: Do we want to callout `spec.NetworkInterfaces` can be optional in the first sentence? 
Also maybe add a code snippet will help here?
```
kind: VirtualMachine
metadata:
 name: ubuntu-vm
 namespace: test-ns
spec:
 networkInterfaces:
 - networkName: "primary"
  networkType: vsphere-distributed
  ```

**Review by zyiyi11 — APPROVED** [review]
minor suggestions. LGTM. Thanks!

File: docs/concepts/workloads/vm.md [hunk 5]
```diff
@@ -84,15 +84,45 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
...
```
sreyasn, line 122: also `Ethernet card -> ethernet card type`?.

I'm also thinking, we should mention that the `configSpec` from the `VirtualMachineClass` takes precedence with selecting the Ethernet card type when specified because of code ref : [link](https://github.com/vmware-tanzu/vm-operator/blob/3a1bf614d9f15015f44d80c723d2809a6f0032a7/pkg/vmprovider/providers/vsphere/session/session_vm_update.go#L611) ?

docs/concepts/workloads/vm.md
```diff
@@ -84,15 +84,45 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
...
 
 ### Storage
+A VM deployed using VM operator inherits the storage defined in the `VirtualMachineImage`.  However, developers can also provision and manage additional storage dynamically by leveraging `PersistentVolumes`. To do this, a user would create a `PersistentVolumeClaim` resource by picking a `StorageClass` associated with their namespace, along with other properties such as the disk size, mode etc. VM operator then dynamically provisions a first class disks which is exposed to the guest as a block volume. Users can start using the disk after formatting and mounting it at a mountpoint. VM operator also supports resizing these volumes to increase their size.
```
sreyasn, line 125: Do we wanna reference https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims here ?.

aruneshpa: Fair point.  Yeah, I will add that link there.

docs/concepts/workloads/vm.md
aruneshpa: Thanks for spotting this.

docs/concepts/workloads/vm.md
aruneshpa: I changed the text to indicate that the network interface is optional.  I also modified the last line to make it more clear that the default network interface is created only when nothing is specified.  I don't think adding an example here is required (because we are not adding an example for any other fields such as volumes).

docs/concepts/workloads/vm.md
aruneshpa: Yeah, makes sense.  Done.

docs/concepts/workloads/vm.md
aruneshpa: Check out this section: https://vm-operator--189.org.readthedocs.build/en/189/concepts/workloads/vm/#cpu-and-memory.  The `Note` section comes out okay?

docs/concepts/workloads/vm.md
aruneshpa: https://vm-operator--189.org.readthedocs.build/en/189/concepts/workloads/vm/#networking

docs/concepts/workloads/vm.md
```diff
@@ -84,15 +84,98 @@ It is possible to update parts of an existing `VirtualMachine` resource. Some fi
...
+```
+The first two network interfaces of the VM are configured using the VM class. So, even though they specify a card type, they inherit the `VirtualE1000` and `VirtualVmxnet3` types respectively. The third interface is configured using the card type it specifies - `VirtualVmxnet2`. The fourth interface does not specify any type, so the default Ethernet card type of `VirtualVmxnet3` is used.
+
 
 ### Storage
+A VM deployed using VM operator inherits the storage defined in the `VirtualMachineImage`.  However, developers can also provision and manage additional storage dynamically by leveraging `PersistentVolumes`. To do this, a user would create a `[PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)` resource by picking a `StorageClass` associated with their namespace, along with other properties such as the disk size, mode etc. VM operator then dynamically provisions a first class disks which is exposed to the guest as a block volume. Users can start using the disk after formatting and mounting it at a mountpoint. VM operator also supports resizing these volumes to increase their size.
```
sreyasn, line 178: Remove the `` from [PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims). It's ignoring the hyperlink otherwise.

sreyasn (approved):
Thanks for this change @aruneshpa !. Looking good.

docs/concepts/workloads/vm.md
aruneshpa: Done.  Thanks for spotting this.

PR #195, opened 2023-08-03
✨ Make infra related controllers v1a2 aware

bryanv:

There is nothing inherently v1a2 about these controllers, but the way we have the fat VMProvider interface makes them so. We've wanted to and should get away from our current not Go'ish interface but that is out of scope for my current work. So just for these controllers, define the interface they need and pick the appropriate provider based on the FSS

```release-note
NONE
```


dilyar85 (approved):
LGTM.

PR #199, opened 2023-08-09
✨Split v1a1 and v1a2 CRD generations

sreyasn:

- This change splits the CRD generations for v1a1 and v1a1+v1a2 combined into separate directories. We do this for FSS based yaml generations for when v1alpha2 FSS is enabled.
- It makes sure the output infrastructure yaml from kustomize target 'config/wcp' is retained the same.
- This change also introduces a config/wcp-v1a2 directory for generating a kustomize v1a2 infrastructure yaml.
   - wcp-v1a2 will still need to include v1a2 webhook related manifests. This will come in a follow-up PR.

Fixes # N/A

N/A

**Testing Done**:
- Diff of config/wcp from earlier is retained with this new change. 
- gce2e vmservice tests
- deploying a wcp testbed with overriden vm-operator component

```release-note
Split v1a1 and v1a2 CRD yamls in separate directories
```


akutz:
I want to dig into this a bit.

sreyasn:
This PR is to be closed in favour of not creating different targets for v1a1 and v1a2 CRDs.

The integration test code harness that would different CRD versions and storage versions needs to be changed to support v1a1 and v1a2 on the same branch by fetching the right CRDs (v1a1/v1a2) depending on the tests. 

Suggestion from @akutz discussed offline: 
"I would update the test harness to be aware of the FSS and instead of using `CRDDirectoryPaths`, load the `CRDs` as objects, and then use the FSS to set served/storage where appropriate. FWIW, this would also be a nice enhancement for working with the other FSSs I mentioned. I could see a utility function that ingests the FSS states, a directory for CRDs, and returns the CRD objects based on ingested files after they are mutated where necessary based on the FSS states. "

PR #202, opened 2023-08-15
 🐛  Better handle when CnsNodeVmAttachment unexpectedly already exists

bryanv:
If a CnsNodeVmAttachment already exists when we were expecting to create it, inspect the existing one to see if we're the owners of it but it is otherwise stale. The usual cause of this will be that VC VM was deleted from underneath, and the replacement VM will end up with a new BiosUUID.

This addresses the immediate issue we have, but still leaves some lingering issues around the whole attachment workflow, like w.r.t. volume and PVC name changes, and how we determine if the volume is actually attached, that we'll have to keep throwing code at until we're able to revisit this workflow, and likely bring the attachment and detachment in-house.

```release-note
NONE
```

### DIFF
--- a/controllers/volume/v1alpha1/volume_controller.go
+++ b/controllers/volume/v1alpha1/volume_controller.go
@@ -624,16 +624,66 @@ func (r *Reconciler) createCNSAttachment(
 	}
 
 	if err := r.Create(ctx, attachment); err != nil {
-		// An IsAlreadyExists error is not really expected because the only way we'll do a Create()
-		// is if the attachment wasn't returned in the List() done at the start of the reconcile.
-		// The Client cache might have been stale, so ignore the error, but like the comment in the
-		// caller, we might want to perform an Update() in this case too.
-		if !apiErrors.IsAlreadyExists(err) {
-			return errors.Wrap(err, "Error creating CnsNodeVmAttachment")
+		if apiErrors.IsAlreadyExists(err) {
+			return r.createCNSAttachmentButAlreadyExists(ctx, attachmentName)
 		}
+		return errors.Wrap(err, "Cannot create CnsNodeVmAttachment")
 	}
 
 	return nil
+}
+
+// createCNSAttachmentButAlreadyExists tries to handle various conditions when a CnsNodeVmAttachment
+// unexpected already exists. Usually the existing attachment is for a prior VC VM has been deleted
+// from underneath us, and the replacement will have a different BiosUUID. The CnsNodeVmAttachment
+// workflow was from CNS and has been very cumbersome to us to use in practice, with this controller
+// being more complicated than it really should be for us to try to work around all the edge cases.
+// Note that CNS doesn't reevaluate an attachment at all once Status.Attached=true. We could Update
+// it instead - clearing Attached atomically since it doesn't have a Status subresource - but it
+// should not be treating the Status as the SoT.
+func (r *Reconciler) createCNSAttachmentButAlreadyExists(
+	ctx *context.VolumeContext,
+	attachmentName string) error {
+
+	attachment := &cnsv1alpha1.CnsNodeVmAttachment{}
+	err := r.Client.Get(ctx, client.ObjectKey{Name: attachmentName, Namespace: ctx.VM.Namespace}, attachment)
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			// Create failed but now the attachment does not exist. Most likely the attachment got GC'd.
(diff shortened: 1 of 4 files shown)


controllers/volume/v1alpha2/volume_controller_unit_test.go
```diff
@@ -570,6 +570,82 @@ func unitTestsReconcile() {
...
+						Expect(vm.Status.Volumes).To(BeEmpty())
+					})
+
+					By("attachment still exists", func() {
+						Expect(getCNSAttachmentForVolumeName(vm, vmVol.Name)).ToNot(BeNil())
+					})
+				})
+			})
+
+			When("CnsNodeVmAttachment owned by different VM", func() {
```
sreyasn, line 606: can we change this to maybe `owned by a VM with a different BiosUUID` to different from the previous test that's more about the ownerRef being different?

sreyasn (approved):
LGTM!

controllers/volume/v1alpha2/volume_controller_unit_test.go
bryanv: Oops copy-pasta here. I'll fix it. Thanks.

controllers/volume/v1alpha1/volume_controller.go
```diff
@@ -624,18 +624,66 @@ func (r *Reconciler) createCNSAttachment(
...
+		// happened is the VC VM was deleted, and then the VC VM is being recreated, generating a new
+		// BiosUUID. Since this attachment is ours, delete it to let CNS remove the attachment from
+		// the old VM - if it still exists - so that on a later reconcile we re-create the attachment
+		// with our new BiosUUID.
+		// We could check the attachment's DeletionTimestamp to avoid hitting Delete on it again but
+		// this window should be short.
+		if err := r.Client.Delete(ctx, attachment); err != nil {
+			// The attachment may have been GC'd since the Get() above and that's fine.
+			// Return an error to force another reconcile.
+			return errors.Wrap(err, "Failed to delete existing CnsNodeVmAttachment with stale BiosUUID")
```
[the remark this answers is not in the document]
bryanv: Yes. We could - and I thought about it - pass in the attachment to be created do that here but this whole function is just a for edge case, and it is easier to always have this return an error and we'll reevaluate the next time through reconcile.

controllers/volume/v1alpha1/volume_controller.go
```diff
@@ -624,18 +624,68 @@ func (r *Reconciler) createCNSAttachment(
...
 }
 
+// createCNSAttachmentButAlreadyExists tries to handle various conditions when a CnsNodeVmAttachment
+// unexpected already exists. Usually the existing attachment is for a prior VC VM has been deleted
+// from underneath us, and the replacement will have a different BiosUUID. The CnsNodeVmAttachment
+// workflow was from CNS and has been very cumbersome to us to use in practice, with this controller
+// being more complicated than it really should be for us to try to work around all the edge cases.
+// Note that CNS doesn't reevaluate an attachment at all once Status.Attached=true. We could Update
+// it instead - clearing Attached atomically since it doesn't have a Status subresource - but it
+// should not be treating the Status as the SoT.
```
aruneshpa, line 643: Can we open a bug/ticket to get this fixed from the CNS team?  Status should not be treated as SoT and should be recalculated during each reconcile.

controllers/volume/v1alpha1/volume_controller.go
```diff
@@ -624,18 +624,68 @@ func (r *Reconciler) createCNSAttachment(
...
+		}
+
+		return err
+	}
+
+	if !metav1.IsControlledBy(attachment, ctx.VM) {
+		// This attachment has our expected name but is not owned by us. Most likely, the owning VM
+		// is in the process of being deleted, and the attachment will be GC after the owner is gone.
+		// We just have to wait it out here: we cannot delete an attachment that isn't ours.
+		return errors.Errorf("CnsNodeVmAttachment %s has a different controlling owner", attachmentName)
```
aruneshpa, line 664: Should we consider having a requeue delay to avoid the controller being swarmed by repeated reconciles in these error scenarios (L664 and L657)?

bryanv: In the most common case here, our k8s VM is still the owner of the attachment so our Watch will result in prompt re-reconcile. Otherwise, we already get an exponential backoff for free from ctrl-runtime when returning an error; I'm not going to do anything additional here until that proves to not be sufficient.

PR #204, opened 2023-08-17
:seedling: Limit template rendering to only vAppConfig and Sysprep transport

### LINKED ISSUE #146: Only perform template substitutions for bootstrap type vAppConfig

VM operator can accept Golang template strings in bootstrap user-data and perform any substitutions before passing it down to the guest.  However, as of today, the only bootstrap method that supports templating is `vAppConfig`.  This should be enforced in code.

This can cause issues in situations where users want to send cloud-config user-data to the guest using the `CloudInit` boostrap type.  If the user-data contains any template markers (e.g., `{{`), the current code will try to evaluate that -- which can result in inconsistent and unwarranted behavior.  

Often times, guests use Jinja templating in cloud-config user-data where we want markers to be replaced by the instance metadata file in guest during cloud-init.
For example, the following snippet from the[ cloud-init docs](https://cloudinit.readthedocs.io/en/22.4.1/topics/instancedata.html) demonstrates using Jinja templating to replace the hostname of the EC2 instance type:
```
## template: jinja
#cloud-config
runcmd:
    - echo 'EC2 public hostname allocated to instance: {{
      ds.meta_data.public_hostname }}' > /tmp/instance_metadata
    - echo 'EC2 availability zone: {{ v1.availability_zone }}' >>
      /tmp/instance_metadata
    - curl -X POST -d '{"hostname": "{{ds.meta_data.public_hostname }}",
      "availability-zone": "{{ v1.availability_zone }}"}'
      https://example.com
```

With this bug, VM operator will try to evaluate the `{{ ds.meta_data.public_hostname }}` and emit an error.  This will cause confusion for users.

For any other bootstrap type except `vAppConfig`, templating in bootstrap data should not be performed.
More specifically, the following code should be modified to only perform template substitutions when the bootstrap type is `vAppConfig`:
https://github.com/vmware-tanzu/vm-operator/blob/4c4f22a0bcb0ce67aa91cbfc9b4dd5360b8cdb11/pkg/vmprovider/providers/vsphere/session/session_vm_customization.go#L256

N/A

This issue has existed as far as I can go back in the history of VM operator.

zyiyi11:
VM operator can accept Golang template strings in bootstrap user-data and perform any substitutions before passing it down to the guest. However, as of today, the only bootstrap method that supports templating is vAppConfig and Sysprep. This change enforce template rendering to only  vAppConfig and Sysprep transport.

Fixes #146 

Testing done: gce2e vmservice test passed with this change.

```release-note
Limit template rendering to only vAppConfig and Sysprep transport
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_customization.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_customization.go
@@ -273,8 +273,6 @@ func (s *Session) customize(
 	config *vimTypes.VirtualMachineConfigInfo,
 	updateArgs VMUpdateArgs) error {
 
-	TemplateVMMetadata(vmCtx, resVM, updateArgs)
-
 	transport := updateArgs.VMMetadata.Transport
 
 	var configSpec *vimTypes.VirtualMachineConfigSpec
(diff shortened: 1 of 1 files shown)


dilyar85 (approved):
LGTM. I feel like we should also update our webhook to fail VM creation when using templated expressions in non-supported transports. @zyiyi11, could you please file an issue to track it as a follow-up to this change?

PR #205, opened 2023-08-18
✨ Adds minimum hardware version to VirtualMachine CRD

### LINKED ISSUE #151: Expose hardware version configuration on VirtualMachine resource
Enable the `VirtualMachine` CRDs to allow configuring the hardware version for the VM to be created.

Using the `VirtualMachineConfigSpec` [attribute](https://github.com/vmware-tanzu/vm-operator/blob/3cda5fee2ddeed2fa32bb5134a827a5e52220de9/api/v1alpha2/virtualmachineclass_types.go#L57) users can use the hardware version set in the `VirtualMachineClass` CR to create the VM. For special cases, when the hardware version of the VM needs to be configured on a per VM basis (from a set of VMs using the same class), we need a configurable option which takes precedence over the value set in the `VirtualMachineClass.ConfigSpec` field.

<table>
<tr>
<td></td>
<th>Value</th>
<th>How to Obtain</th>
</tr>
<tr>
<th>Supervisor version</th>
<td><code></code></td>
</tr>
<tr>
<td><code></code></td>
</tr>
<tr>
<th>Kubernetes version</th>
<td><code></code></td>
</tr>
<tr>
<th>VM Operator version</th>
<td><code></code></td>
<td>

```shell
  sort -u | \
  grep vmop | \
  awk -F'/' '{print $3}' | \
  awk -F: '{print $2}'
```

</td>
</tr>
</table>

srm09:

Introduces a new field on the VirtualMachine CRD which allows the user to define the minimum hardware version for the provisioned VM. The behavior of the field is as follows:
- If the VMClassAsConfig FSS is enabled, this new field will override the hardware version derived from the VMClass's ConfigSpec if that value is lower than the minimum version field. Otherwise, the value from the ConfigSpec is honored.
- If the VMClassAsConfig FSS is disabled, it will reconfigure the VM post creation before the power on operation to ensure that the h/w version is atleast set to the minimum version. This means, it can override the hardware version of the OVA with a hardware version value lower than this field.

**Which issue(s) is/are addressed by this PR?** :
Fixes #151 

Testing performed:
- VM with min hardware version = 14 and VMClass h/w version = 13, resulting VM version = 14
- VM with min hardware version = 15 and VMClass h/w version = 16, resulting VM version = 16
- VM with no min hardware version and VMClass h/w version = 17, resulting VM version = 17
- VM with no min hardware version and no VMClass version, resulting VM version = 21 (host default)

```release-note
✨Adds minimum hardware version to VirtualMachine CRD
```

### DIFF
--- a/api/v1alpha2/virtualmachine_types.go
+++ b/api/v1alpha2/virtualmachine_types.go
@@ -174,7 +174,7 @@ type VirtualMachineSpec struct {
 	// +optional
 	ImageName string `json:"imageName,omitempty"`
 
-	// Class describes the name of the VirtualMachineClass resource used to
+	// ClassName describes the name of the VirtualMachineClass resource used to
 	// deploy this VM.
 	//
 	// This field is optional in the cases where there exists a sensible
@@ -335,6 +335,44 @@ type VirtualMachineSpec struct {
 	//
 	// +optional
 	Reserved VirtualMachineReservedSpec `json:"reserved,omitempty"`
+
+	// MinHardwareVersion specifies the desired minimum hardware version
+	// for this VM.
+	//
+	// Usually the VM's hardware version is derived from:
+	// 1. the VirtualMachineClass used to deploy the VM provided by the ClassName field
+	// 2. the datacenter/cluster/host default hardware version
+	// Setting this field will ensure that the hardware version of the VM
+	// is at least set to the specified value. To enforce this, it will override
+	// the value from the VirtualMachineClass.
+	//
+	// This field is never updated to reflect the derived hardware version.
+	// Instead, VirtualMachineStatus.HardwareVersion surfaces
+	// the observed hardware version.
+	//
+	// Please note, setting this field's value to N ensures a VM's hardware
+	// version is equal to or greater than N. For example, if a VM's observed
+	// hardware version is 10 and this field's value is 13, then the VM will be
+	// upgraded to hardware version 13. However, if the observed hardware
+	// version is 17 and this field's value is 13, no change will occur.
+	//
+	// Several features are hardware version dependent, for example:
+	//
+	// * NVMe Controllers        		 >= 14
+	// * Dynamic Direct Path I/O devices >= 17
+	//
+	// Please refer to https://kb.vmware.com/s/article/1003746 for a list of VM
+	// hardware versions.
+	//
+	// It is important to remember that a VM's hardware version may not be
+	// downgraded and upgrading a VM deployed from an image based on an older
+	// hardware version to a more recent one may result in unpredictable
+	// behavior. In other words, please be careful when choosing to upgrade a
+	// VM to a newer hardware version.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=13
+	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
 }
 
 // VirtualMachineReservedSpec describes a set of VM configuration options
@@ -455,6 +493,15 @@ type VirtualMachineStatus struct {
 	//
 	// +optional
 	LastRestartTime *metav1.Time `json:"lastRestartTime,omitempty"`
+
+	// HardwareVersion describes the VirtualMachine resource's observed
+	// hardware version.
+	//
+	// Please refer to VirtualMachineSpec.MinHardwareVersion for more
+	// information on the topic of a VM's hardware version.
+	//
+	// +optional
+	HardwareVersion int32 `json:"hardwareVersion,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/pkg/vmprovider/providers/vsphere/session/session_vm_status.go
+++ b/pkg/vmprovider/providers/vsphere/session/session_vm_status.go
@@ -16,6 +16,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/lib"
 	"github.com/vmware-tanzu/vm-operator/pkg/topology"
+	"github.com/vmware-tanzu/vm-operator/pkg/util"
 	res "github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/resources"
 )
 
@@ -108,6 +109,7 @@ func (s *Session) updateVMStatus(
 	vm.Status.UniqueID = resVM.MoRef().Value
 	vm.Status.BiosUUID = summary.Config.Uuid
 	vm.Status.InstanceUUID = summary.Config.InstanceUuid
+	vm.Status.HardwareVersion = util.ParseVirtualHardwareVersion(summary.Config.HwVersion)
 
 	if host := summary.Runtime.Host; host != nil {
 		hostSystem := object.NewHostSystem(s.Client.VimClient(), *host)
--- a/pkg/vmprovider/providers/vsphere/vmprovider_vm.go
+++ b/pkg/vmprovider/providers/vsphere/vmprovider_vm.go
@@ -26,7 +26,6 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/util"
 	vcclient "github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/client"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/constants"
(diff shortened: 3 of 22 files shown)


pkg/vmprovider/providers/vsphere/session/session_vm_update_test.go
```diff
@@ -386,6 +386,37 @@ var _ = Describe("Update ConfigSpec", func() {
...
+		})
+
+		Context("Existing hardware version is greater than min version", func() {
+			BeforeEach(func() {
+				config.Version = "vmx-17"
+				minHardwareVersion = 16
+			})
+
+			It("does not update the hardware version", func() {
+				Expect(configSpec.Version).NotTo(Equal("vmx-16"))
```
bryanv, line 414: Perhaps `.To(Equal("vmx-17")` instead to assert that is is really unchanged.

pkg/vmprovider/providers/vsphere/session/session_vm_status.go
```diff
@@ -165,5 +166,7 @@ func (s *Session) updateVMStatus(
 		}
 	}
 
+	vm.Status.HardwareVersion = util.ParseVirtualHardwareVersion(summary.Config.HwVersion)
```
bryanv, line 169: Maybe more near L111 so all the `summary.Config...` onelines are together?

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -457,6 +457,19 @@ func UpdateConfigSpecDeviceGroups(
...
+// UpdateConfigSpecHardwareVersion sets the hardware version in the ConfigSpec
+// if the current hardware version is lower than the specified minimum version.
+func UpdateConfigSpecHardwareVersion(
+	config *vimTypes.VirtualMachineConfigInfo,
+	configSpec *vimTypes.VirtualMachineConfigSpec,
+	minHardwareVersion int32) {
+
+	existingVersion := util.ParseVirtualHardwareVersion(config.Version)
+	if existingVersion < minHardwareVersion {
+		util.EnsureMinHardwareVersionInConfigSpec(configSpec, minHardwareVersion)
```
bryanv, line 469: Maybe just `configSpec.Version = fmt.Sprintf()` here instead, since the `EnsureMinHardwareVersionInConfigSpec` kind of ends up doing the same check, and to make it clearer here that the field is only set if really needs to be changed so we don't otherwise Reconfigure.

pkg/vmprovider/providers/vsphere/session/session_vm_update.go
```diff
@@ -476,6 +476,9 @@ func updateConfigSpec(
 		UpdateHardwareConfigSpec(config, configSpec, &vmClassSpec)
 		UpdateConfigSpecCPUAllocation(config, configSpec, &vmClassSpec, updateArgs.MinCPUFreq)
 		UpdateConfigSpecMemoryAllocation(config, configSpec, &vmClassSpec)
+		// Only in this case, the VM's hardware version is not set to the
+		// minimum version from the VM.Spec, if any.
+		util.EnsureMinHardwareVersionInConfigSpec(configSpec, vmCtx.VM.Spec.MinHardwareVersion)
```
sreyasn, line 481: We probably don't/can't do this in the update path even when VM Class as Config FSS is disabled.  Any changes of HardwareVersion on update would require the UpgradeAPI() right?

srm09:
Waiting on the v1a2 work to be brought to a stable state before introducing more API changes, work has been completed for now.

pkg/vmprovider/providers/vsphere/vmprovider_vm.go
```diff
@@ -659,6 +658,8 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpec(
 		createArgs.VMImageStatus.Firmware,
 		vmClassConfigSpec)
 
+	util.EnsureMinHardwareVersionInConfigSpec(createArgs.ConfigSpec, vmCtx.VM.Spec.MinHardwareVersion)
```
sreyasn, line 661: We should probably move this after _the check for empty hardware version and PCI/PVC devices to set the min supported hw version._  in 673.. thoughts?. 

I see that it's handled right in the v1a2 implementation.

srm09: Sure works either way, moving it after will ensure that the min h/w version for PCI devices is set, and if not needed, then the min version does get set.

sreyasn: Yeah, Better to keep it after the PCI/PVCDevices block if the intent is to not override the block completely. 
Because 
- The PCI/PVC devices block will not get evaluated when the configSpec.hardwareVersion is set earlier by EnsureMinHardwareVersionInConfigSpec. 
- This also opens up the possibility for the VM's with PVC/PCI devices to have a hardware version set by EnsureMinHardwareVersionInConfigSpec that's lesser than 'min supported' for the device

srm09: Makes sense, fixed

api/v1alpha2/virtualmachine_types.go
```diff
@@ -455,6 +493,15 @@ type VirtualMachineStatus struct {
...
 	LastRestartTime *metav1.Time `json:"lastRestartTime,omitempty"`
+
+	// HardwareVersion describes the VirtualMachine resource's observed
+	// hardware version.
+	//
+	// Please refer to VirtualMachineSpec.MinHardwareVersion for more
+	// information on the topic of a VM's hardware version.
+	//
+	// +optional
+	HardwareVersion int32 `json:"hardwareVersion,omitempty"`
```
sreyasn, line 504: I think we're missing setting the status field for v1a2 (vsphere2)

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -738,13 +739,14 @@ func (v validator) validateUpdatesWhenPoweredOn(ctx *context.WebhookRequestConte
...
 
-func (v validator) validateImmutableFields(ctx *context.WebhookRequestContext, vm, oldVM *vmopv1.VirtualMachine) field.ErrorList {
+func (v validator) validateImmutableFields(_ *context.WebhookRequestContext, vm, oldVM *vmopv1.VirtualMachine) field.ErrorList {
 	var allErrs field.ErrorList
 	specPath := field.NewPath("spec")
 
 	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.ImageName, oldVM.Spec.ImageName, specPath.Child("imageName"))...)
 	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.ClassName, oldVM.Spec.ClassName, specPath.Child("className"))...)
 	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.StorageClass, oldVM.Spec.StorageClass, specPath.Child("storageClass"))...)
+	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.MinHardwareVersion, oldVM.Spec.MinHardwareVersion, specPath.Child("minHardwareVersion"))...)
```
sreyasn, line 749: worth a adding similar to https://github.com/vmware-tanzu/vm-operator/blob/71c707d5edb2b50ccea9e6984bb599b6586e0b9a/webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator_intg_test.go#L107 for both v1a1 and v1a2 validations

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -881,6 +881,36 @@ func vmTests() {
 				})
 			})
 
+			Context("VM Class Config specifies a hardware version", func() {
```
sreyasn, line 884: Thanks for this test. 
- We'll need to also do the same unit test in vsphere2/vmprovider_vm_test.go. 
- Also, another thought, we could also cover the PVC/PCI device case by adding onto this test. Just by mentioning a vm.Spec.MinHardwareVersion in a BeforeEach() here (https://github.com/vmware-tanzu/vm-operator/blob/71c707d5edb2b50ccea9e6984bb599b6586e0b9a/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go#L848)

sreyasn (approved):
Thanks for working on this, Sagar. Looks good.

PR #207, opened 2023-08-21
🐛 Adds event filter to incoming VirtualMachine objects only

srm09:
This ensures that the controllerName filter is applied to the incoming VirtualMachine objects only instead of all the watched objects by the controller.

This PR replaces #170 

```release-note
🐛 Adds event filter to incoming VirtualMachine objects only
```


dilyar85 (approved):
LGTM!

sreyasn (approved):

aruneshpa (approved):

PR #211, opened 2023-08-30
:bug: fix bug when using CloudInit without userdata to bootstrap VM

zyiyi11:
This change fixes the bug when using CloudInit transport to bootstrap a VM, not passing user data actually results VM customized with GOSC (LinuxPrep). This patch introduced an annotation to set Cloudinit metadata instance-id and also introduced a webhook to prevent SSO users from setting the annotation manually.

Fixes N/A

Testing done: 
1. gce2e vmservice passed.
2. Greenfield: Manually deployed a VM using Cloud-init sans userdata. `guestinfo.metadata` got set successfully.
3. SSO User setting annotation is forbidden while VI Admin can annotate.
```
$ k apply -f vm1.yaml
Error from server (metadata.annotations.vmoperator.vmware.com/cloud-init-instance-id: Forbidden: adding this annotation is not allowed): error when creating "vm1.yaml": admission webhook "default.validating.virtualmachine.vmoperator.vmware.com" denied the request: metadata.annotations.vmoperator.vmware.com/cloud-init-instance-id: Forbidden: adding this annotation is not allowed

$ k annotate virtualmachine devops-test-vm-3 -n test-ns vmoperator.vmware.com/cloud-init-instance-id=iid-datasource-none
Error from server (metadata.annotations.vmoperator.vmware.com/cloud-init-instance-id: Forbidden: adding this annotation is not allowed): admission webhook "default.validating.virtualmachine.vmoperator.vmware.com" denied the request: metadata.annotations.vmoperator.vmware.com/cloud-init-instance-id: Forbidden: adding this annotation is not allowed

$ root@42001ff0a8e16406567b1f085e82d958 [ ~ ]# k annotate virtualmachine vm-sans-payload -n test-ns vmoperator.vmware.com/cloud-init-instance-id=iid-datasource-none
virtualmachine.vmoperator.vmware.com/vm-sans-payload annotated
```
brownfield scenario change in cayman_vm-operator:
upgrade logic to:
- apply the annotation with the value iid-datasource-none to all VMs that use Cloud-Init sans payload
- only apply the above if the source version being upgraded is less than the first version that has the patch to correct empty Cloud-Init payloads
- this will ensure this is only ever applied to brownfield VMs that meet these conditions prior to the patch being introduced

```release-note
Bootstrap VM using CloudInit without specifying user data now gets customized with CloudInit instead of LinuxPrep
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
+++ b/pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
@@ -1494,6 +1494,98 @@ func vmTests() {
 						})
 					})
 				})
+
+				Context("Cloudinint Transport without userdata", func() {
+					var ec map[string]interface{}
+
+					JustBeforeEach(func() {
+						vm.Spec.VmMetadata = &vmopv1.VirtualMachineMetadata{
+							Transport: vmopv1.VirtualMachineMetadataCloudInitTransport,
+						}
+						vcVM, err := createOrUpdateAndGetVcVM(ctx, vm)
+						Expect(err).ToNot(HaveOccurred())
+
+						var o mo.VirtualMachine
+						Expect(vcVM.Properties(ctx, vcVM.Reference(), nil, &o)).To(Succeed())
+
+						ec = map[string]interface{}{}
+						for _, option := range o.Config.ExtraConfig {
+							if val := option.GetOptionValue(); val != nil {
+								ec[val.Key] = val.Value.(string)
+							}
+						}
+					})
+
+					AfterEach(func() {
+						ec = nil
+					})
+
+					It("Metadata data is included in ExtraConfig", func() {
+						By("Should include default keys and values", func() {
+							Expect(ec).To(HaveKeyWithValue("disk.enableUUID", "TRUE"))
+							Expect(ec).To(HaveKeyWithValue("vmware.tools.gosc.ignoretoolscheck", "TRUE"))
+						})
+
+						By("Should include guestinfo metadata and no guestinfo userdata", func() {
+							Expect(ec).To(HaveKey("guestinfo.metadata"))
+							Expect(ec).To(HaveKey("guestinfo.metadata.encoding"))
+							Expect(ec).ToNot(HaveKey("guestinfo.userdata"))
+						})
+					})
+				})
+
+				Context("Cloudinint Transport with userdata", func() {
+					var ec map[string]interface{}
+
+					JustBeforeEach(func() {
+						configMap := &corev1.ConfigMap{
+							ObjectMeta: metav1.ObjectMeta{
+								GenerateName: "md-configmap-",
+								Namespace:    vm.Namespace,
+							},
+							Data: map[string]string{
+								"user-data": "data",
+							},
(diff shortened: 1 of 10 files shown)


webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator.go
```diff
@@ -64,6 +64,7 @@ const (
 	invalidNextRestartTimeOnCreate            = "cannot restart VM on create"
 	invalidNextRestartTimeOnUpdate            = "must be formatted as RFC3339Nano"
 	invalidNextRestartTimeOnUpdateNow         = "mutation webhooks are required to restart VM"
+	settingInstanceIDAnnotationNotAllowed     = "adding instance-id annotation is not allowed"
```
dilyar85, line 67: This error message seems redundant since you use `field.Forbidden` which includes the annotation key path already.

webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator.go
```diff
@@ -738,3 +741,22 @@ func (v validator) isNetworkRestrictedForReadinessProbe(ctx *context.WebhookRequ
...
+
+func (v validator) validateAnnotation(ctx *context.WebhookRequestContext, vm *vmopv1.VirtualMachine) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if ctx.IsPrivilegedAccount {
+		return allErrs
+	}
+
+	annotationPath := field.NewPath("metadata", "annotations")
+	forbiddenAnnotationKey := vmopv1.InstanceIDAnnotation
```
dilyar85, line 753: Nie: Move these declarations inside the if statement (as they are only used if the annotation is non-empty).

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -1494,6 +1494,98 @@ func vmTests() {
...
+
+						var o mo.VirtualMachine
+						Expect(vcVM.Properties(ctx, vcVM.Reference(), nil, &o)).To(Succeed())
+
+						ec = map[string]interface{}{}
+						for _, option := range o.Config.ExtraConfig {
+							if val := option.GetOptionValue(); val != nil {
+								ec[val.Key] = val.Value.(string)
+							}
+						}
```
dilyar85, line 1516: I feel like these should be executed inside the "It" block rather than "JustBeforeEach" here, so that you don't need to have the "AfterEach" block at all. Additionally, consider having a function to collect the `ec` (we might already have one defined as a  utility function) to avoid duplicate code.

dilyar85 (commented):
Left some comments mostly for suggestions. Overall looks good to me.

pkg/vmprovider/providers/vsphere/vmprovider_vm_test.go
zyiyi11: Moved inside "It" block. Thanks for the suggestion.

dilyar85 (approved):

PR #215, opened 2023-09-06
🌱 Rename IMAGE-NAME to HUMAN-READABLE-NAME in VM Image printer column

dilyar85:
This PR updates the printer column of VirtualMachine Image (in both v1a1 and v1a2) to rename "Image Name" to "Human Readable Name". The previous "Image Name" can be confusing with the `spec.imageName` in the VirtualMachine type.

Manually applied the updated CRD in a testbed and verified the output printer column:

```console
$ kubectl get cvmi
NAME                    HUMAN-READABLE-NAME                                          VERSION                  OS-TYPE               FORMAT   AGE
vmi-9b0796bae7b11ea63   photon-3-amd64-vmi-k8s-v1.23.8---vmware.2-tkg.1-zshippable   v1.23.8+vmware.1-tkg.1   vmwarePhoton64Guest   OVF      151m

$ kubectl get vmi -n sdiliyaer-test
NAME                    HUMAN-READABLE-NAME                                     VERSION         OS-TYPE                 FORMAT   AGE
vmi-4321c937fd22e514a   photon-ova_uefi-4.0                                     4.0             vmwarePhoton64Guest     OVF      150m
```

```release-note
Rename IMAGE-NAME to HUMAN-READABLE-NAME in the VM Image printer column.
```

### DIFF
--- a/api/v1alpha1/virtualmachineimage_types.go
+++ b/api/v1alpha1/virtualmachineimage_types.go
@@ -156,7 +156,7 @@ func (vmImage *VirtualMachineImage) SetConditions(conditions Conditions) {
 // +kubebuilder:resource:scope=Cluster,shortName=vmi;vmimage
 // +kubebuilder:storageversion:false
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="Image-Name",type="string",JSONPath=".status.imageName"
+// +kubebuilder:printcolumn:name="Human-Readable-Name",type="string",JSONPath=".status.imageName"
 // +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".spec.productInfo.version"
 // +kubebuilder:printcolumn:name="Os-Type",type="string",JSONPath=".spec.osInfo.type"
 // +kubebuilder:printcolumn:name="Format",type="string",JSONPath=".spec.type"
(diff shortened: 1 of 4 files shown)


aruneshpa:
From users' point of view, every name should be friendly name.  The fact that our image names are generated is an implementation detail.  We need to think of some other alias than "friendly name".  
"vSphere Name", "infra name" or something alike.  Ellen can be of help here as well.

dilyar85:
> From users' point of view, every name should be friendly name. The fact that our image names are generated is an implementation detail. We need to think of some other alias than "friendly name". "vSphere Name", "infra name" or something alike. Ellen can be of help here as well.

I kind of lean towards using the **vSphere Name** to ensure consistency with the `kubectl get clitem` output column: https://github.com/vmware-tanzu/image-registry-operator-api/blob/main/api/v1alpha1/contentlibraryitem_types.go#L164
I used "Friendly Name" in this PR based on Andrew's suggestion and saw it's mentioned in the v1a2 API as well: https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha2/virtualmachineimage_types.go#L135

zyiyi11 (approved):
LGTM

bryanv (approved):

akutz:
I'm not a fan of Human-Readable-Name. I do not understand why we veered from Friendly-Name. Not all IDs are friendly, so I kind of disagree with that. Some are friendlier than others. Friendly-Name is also much nicer sounding and more concise.

PR #217, opened 2023-09-07
🐛 Allow non user-configurable properties in VM deployment

### LINKED ISSUE #181: Consider passing non-user configurable properties into vAppConfig
We have a user trying to use Nested ESXi OVA, which includes one hidden property (used for debugging purposes), but it had no value set after deployed. The only way to resolve this right now was to import the OVA into vSphere, change the hidden property to `configurable` AND then specify that in `vAppConfig` spec and VM Service was able to power it on. 

For basic OVF/OVA, this is not an issue but for more complex ones that leverage hidden properties. other VMware appliances like VCSA also uses hidden properties for various reasons

Right now, any non user configurable properties are ignored. We can consider passing these into vAppConfig.
ref: https://github.com/vmware-tanzu/vm-operator/blob/8f2552faf4dce6a44d6c50d2c72163034d25796d/pkg/vmprovider/providers/vsphere/session/session_util.go#L52-L55

<table>
<tr>
<td></td>
<th>Value</th>
<th>How to Obtain</th>
</tr>
<tr>
<th>Supervisor version</th>
<td><code></code></td>
</tr>
<tr>
<td><code></code></td>
</tr>
<tr>
<th>Kubernetes version</th>
<td><code></code></td>
</tr>
<tr>
<th>VM Operator version</th>
<td><code></code></td>
<td>

```shell
  sort -u | \
  grep vmop | \
  awk -F'/' '{print $3}' | \
  awk -F: '{print $2}'
```

</td>
</tr>
</table>

aruneshpa:
Currently, VM operator ignores any vApp properties that are not user configurable when patching with values specified in the bootstrap resources (VM metadata ConfigMap/Secret). There are some valid use-cases where users might want to patch hidden properties defined in an image. This change fixes this behavior.

Fixes #181.

```release-note
Allow users to configure all properties defined in the OVF
```


aruneshpa:
TODO: We also need to update the official documentation to indicate that non user-configurable properties are also allowed now.
https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-with-tanzu-services-workloads/GUID-FB31603E-EE1D-4413-8567-DD2A5834929C.html

Excerpt from the doc:
```
Procedure
Make sure that for all properties to be configured, your OVF file includes the ovf:userConfigurable="true" entry.
```

PR #218, opened 2023-09-08
✨ Add v1a2 webconsole request controllers

sreyasn:

This change adds the v1a2 virtualmachinewebconsolerequest controllers and tests. The controller is guarded by the Feature state switch. The webhook changes will be a follow-up of this PR.

Fixes # N/A

N/A

```
Add v1a2 virtualmachinewebconsolerequest controllers and tests
```

### DIFF
--- a/controllers/virtualmachinewebconsolerequest/controllers.go
+++ b/controllers/virtualmachinewebconsolerequest/controllers.go
@@ -7,10 +7,17 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinewebconsolerequest/v1alpha1"
+	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinewebconsolerequest/v1alpha2"
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/lib"
 )
 
 // AddToManager adds the controller to the provided manager.
 func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) error {
+	if lib.IsVMServiceV1Alpha2FSSEnabled() {
+		if err := v1alpha2.AddToManager(ctx, mgr); err != nil {
+			return err
+		}
+	}
 	return v1alpha1.AddToManager(ctx, mgr)
 }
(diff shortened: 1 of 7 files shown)


controllers/virtualmachinewebconsolerequest/controllers.go
```diff
@@ -7,10 +7,15 @@ import (
...
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/lib"
 )
 
 // AddToManager adds the controller to the provided manager.
 func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) error {
+	if lib.IsVMServiceV1Alpha2FSSEnabled() {
+		return v1alpha2.AddToManager(ctx, mgr)
+	}
 	return v1alpha1.AddToManager(ctx, mgr)
```
bryanv, line 20: This controller needs to be added even in the v1a2 enabled world: these are separate types - since the object name has changed - so no version conversion is possible like we do with the other types.

controllers/virtualmachinewebconsolerequest/v1alpha2/webconsolerequest_controller.go
```diff
@@ -0,0 +1,216 @@
...
+
+// Reconciler reconciles a WebConsoleRequest object.
+type Reconciler struct {
+	client.Client
+	Logger     logr.Logger
+	Recorder   record.Recorder
+	VMProvider vmprovider.VirtualMachineProviderInterfaceA2
+}
+
+// -kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinewebconsolerequests,verbs=get;list;watch;create;update;patch;delete
```
bryanv, line 82: We should be able to enable these (`+kubebuilder:...`). I wouldn't expect the generated manifests to change since these don't include the version.

bryanv (approved):
It is a relative drop in the bucket for the rest of the tech debt we're saddling ourselves with for v1a2 but since the v1a1 controller/webhooks have to be kept around until v1a1 is completely removed, there is an opportunity now or later to share code here, either by writing our own ConverTo/From() or by defining our own generic intermediary struct.

controllers/virtualmachinewebconsolerequest/controllers.go
sreyasn: good catch. I missed the early return there. I'll return only when the first one(v1a2) errors out.

controllers/virtualmachinewebconsolerequest/v1alpha2/webconsolerequest_controller.go
sreyasn: Yeah. Just `rbac.yaml` changes for this type..

PR #219, opened 2023-09-08
✨ v1a2 provider & api changes

bryanv:
Snapshot of the current v1a2 work. This overall effort still has a lot of things to sort out but this works well enough to create working TKGs, with everything else will using v1a1 so we have a baseline of version conversion and functionality.

```release-note
NONE
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/network/network.go
+++ b/pkg/vmprovider/providers/vsphere2/network/network.go
@@ -1,0 +1,630 @@
+// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//nolint:revive
+package network
+
+import (
+	goctx "context"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	ctrlruntime "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	ncpv1alpha1 "github.com/vmware-tanzu/vm-operator/external/ncp/api/v1alpha1"
+	netopv1alpha1 "github.com/vmware-tanzu/vm-operator/external/net-operator/api/v1alpha1"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+	"github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/lib"
+)
+
+type NetworkInterfaceResults struct {
+	Results []NetworkInterfaceResult
+}
+
+type NetworkInterfaceResult struct {
+	IPConfigs  []NetworkInterfaceIPConfig
+	MacAddress string
+	ExternalID string
+	NetworkID  string
+	Backing    object.NetworkReference
+
+	Device vimtypes.BaseVirtualDevice
+
+	// Fields from the InterfaceSpec used later during customization.
+	Name          string
+	DHCP4         bool
+	DHCP6         bool
+	MTU           int64
+	Nameservers   []string
+	SearchDomains []string
+	Routes        []NetworkInterfaceRoute
+}
+
+type NetworkInterfaceIPConfig struct {
+	IPCIDR  string // IP address in CIDR notation e.g. 192.168.10.42/24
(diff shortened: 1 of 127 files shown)


pkg/vmprovider/providers/vsphere2/network/network.go
```diff
@@ -0,0 +1,631 @@
...
+
+	netIf, err = waitForReadyNetworkInterface(vmCtx, client, netIf.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	return netIfToResult(vimClient, netIf), nil
+}
+
+func netIfToResult(
```
sreyasn, line 297: How about netOpNetIfToResult() and ncpNetIfToResult() for the other one below?.

pkg/vmprovider/providers/vsphere2/network/network.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+
+	return netIf, nil
+}
+
+// NCPCRName returns the name to be used for the NCP VirtualNetworkInterface CR.
+func NCPCRName(vmName, networkName, interfaceName string, isV1A1 bool) string {
+	var name string
+
+	if isV1A1 {
```
sreyasn, line 377: Is this referring to the vmop v1a1/v1a2?. If so, should we pass in the FSS val here to NCPCRName(..,.., lib.Isv1a2enabled()) in line 411.

pkg/vmprovider/providers/vsphere2/network/network.go
```diff
@@ -0,0 +1,631 @@
...
+	if kind := interfaceSpec.Network.Kind; kind != "" && kind != "Network" {
+		return nil, fmt.Errorf("network kind %q is not supported for VDS", kind)
+	}
+
+	// If empty, NetOP will try to select a namespace default.
+	networkName := interfaceSpec.Network.Name
+
+	netIf := &netopv1alpha1.NetworkInterface{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      NetOPCRName(vmCtx.VM.Name, networkName, interfaceSpec.Name, false),
```
sreyasn, line 268: Should we pass in the FSS val for v1a2 here to the last argument?. or Is it referring to a different v1a1.

pkg/vmprovider/providers/vsphere2/network/network.go
```diff
@@ -0,0 +1,631 @@
...
+
+	vnetIf, err = waitForReadyNCPNetworkInterface(vmCtx, client, vnetIf.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	return vnetIfToResult(vmCtx, vimClient, clusterMoRef, vnetIf)
+}
+
+func vnetIfToResult(
```
sreyasn, line 437: maybe ncpNetIfToResult() ?

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_cloudinit.go
```diff
@@ -0,0 +1,186 @@
...
+				if userdata != "" {
+					break
+				}
+			}
+		}
+
+		// NOTE: The old code didn't error out if userdata wasn't found, so keep going.
+
+	} else {
+		return nil, nil, fmt.Errorf("TODO: inlined CloudConfig")
```
sreyasn, line 68: How do plan on supporting this?. 
do we need a JIRA/issue to track this somewhere?.

sreyasn:
Also JFYI @bryanv , Keep an eye out on the [BackupVirtualMachine()](https://github.com/vmware-tanzu/vm-operator/blob/87313a56d3718b670ef6235560e3be33943e5d15/pkg/vmprovider/interface_a2.go#L24) method added to the v1a2 interface. Without the provider for this the cayman_vm-operator build will fail.

aruneshpa (approved):

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_cloudinit.go
bryanv: Yes there is a series of lower priority BL items for these. My immediate goal is to get v1a2 to v1a1 parity, and then worry more about new functionality like the inline bootstrap configs here.

pkg/vmprovider/providers/vsphere2/network/network.go
bryanv: No. This is referring to if the VM was a prior a v1a1 VM which means we'll netIF CRDs with the prior naming convention. The existing v1a1 naming convention was ad-hoc and limiting (like only allow one per network) that I want to get away from, especially now that we have a interface name in the spec which was a prototype oversight. If the VM already has netIF CRDs we need to continue to use those and not create new ones.

This is part of the larger issue to ensure existing VMs continue to be reconciled properly post the switch to v1a2.

pkg/vmprovider/providers/vsphere2/network/network.go
bryanv: Sure

pkg/vmprovider/providers/vsphere2/network/network.go
bryanv: Sure

sreyasn (approved):
Thank you for this change, BV..

PR #222, opened 2023-09-13
🌱 Bump up controller-runtime to v0.15.0

aruneshpa:

Major changes include:
- Setting up a Watch for a type has a different signature now; Builder also has changes in how Watches are set up
- Handler methods for enqueuing reconciles are interfaces now
- Handler functions also take a context as an argument now
- Several other changes that hide types behind interfaces
- Webhook server is now an interface
- Cache and manager options related changes
- Polling functions use contexts now
- Client builder now exposes WithStatusSubresource to initialize objects with status sub-resources.  Patching status of objects without this resource will fail.

N/A

Reviewers might find the v0.15.0 controller-runtime [release notes](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0) helpful.

```release-note
NONE
```


dilyar85 (approved):
LGTM. Thanks for the update!

srm09:
This looks good, I have some follow up thoughts I will post in a PR once this merges.
/lgtm

PR #223, opened 2023-09-14
✨ Support VM deployment from image status (friendly) name

### LINKED ISSUE #214: Support user friendlier VM Spec.ImageName
With the image registry work, {Cluster}VirtualMachineImages CR are now named cvmi-$SUFFIX or vmi-$SUFFIX, where the $SUFFIX is from the corresponding {Cluster}ContentLibraryItem, which is hashed from the CL item UUID.

 - the image name doesn't provide any info to the user as to what image their VM is actually using, requiring an additional lookup step

 - since it looks like the $SUFFIX is derived from the CL item UUID that will vary from env to env, requiring bespoke VM yaml, or forcing complexity on users/automation tools to a per-env lookup. I believe we have some lookup code in gce2e.

We should support resolving the image by the friendlier image/item name as a fallback lookup, and use that as the image when there is one unambiguous match. We could support a "Content Library Name / Friendly Image Name" format to help resolve any ambiguity too, with a hopefully not-too-optimistic expectation that CO use common CL naming across their envs.

In the long term, we can consider having an `ImageSelector` field - kind of like what TKG/TKR do - to express something like "use the latest Ubuntu" image.

None.

N/A.

dilyar85:
This PR updates the VirtualMachine mutation webhook to replace the `vm.spec.imageName` if it's a friendly image name that matches a **single** namespace or cluster scope image. If multiple images are found, the mutator denies the VM creation request. The mutator also indexes the VirtualMachineImage and ClusterVirtualMachineImage objects by their `.status.imageName` fields to enable efficient querying.

Fixes #214

This patch formats the `controllers/contentlibrary/v1alpha1/utils/utils.go` file to not exceed 80 characters. It also replaces the FSS mock calls with `os.Setenv/Unsetenv` in the test code.

```release-note
Support VM deployment from image status (friendly) name by mutating `vm.spec.ImageName`.
```

**Testing Done**:
- Added test code to cover VM deployment with image status name.
- Got a successful end-to-end run internally with this change deployed.
- Manually deployed this patch to a WCP testbed and verified VM deployment with all supported `vm.spec.imageName`:

```console
$ kubectl get vmi -n sdiliyaer-test vmi-67264b93813ffcbb4
NAME                    PROVIDER-NAME              CONTENT-LIBRARY-NAME   IMAGE-NAME                      VERSION   OS-TYPE         FORMAT   AGE
vmi-67264b93813ffcbb4   clitem-67264b93813ffcbb4   cl-6e9719d7e81f4b756   centos-stream8-cloudinit-22.1             centos64Guest   OVF      10m

$ kubectl get cvmi -n sdiliyaer-test vmi-b5eef5a79bbee1451
NAME                    PROVIDER-NAME              CONTENT-LIBRARY-NAME   IMAGE-NAME                     VERSION   OS-TYPE         FORMAT   AGE
vmi-b5eef5a79bbee1451   clitem-b5eef5a79bbee1451   cl-6e9719d7e81f4b756   ubuntu-impish-21.10-cloudimg             ubuntu64Guest   OVF      13h

$ kubectl get vm -n sdiliyaer-test -o wide
NAME                                            POWER-STATE   CLASS               IMAGE                   PRIMARY-IP       AGE
vm-from-vmi-b5eef5a79bbee1451                   poweredOn     best-effort-small   vmi-b5eef5a79bbee1451   192.168.128.36   6m23s
vm-from-ubuntu-impish-21.10-cloudimg            poweredOn     best-effort-small   vmi-b5eef5a79bbee1451   192.168.128.37   5m48s
vm-from-cluster-vmi-67264b93813ffcbb4           poweredOn     best-effort-small   vmi-67264b93813ffcbb4   192.168.128.38   4m12s
vm-from-cluster-centos-stream8-cloudinit-22.1   poweredOn     best-effort-small   vmi-67264b93813ffcbb4   192.168.128.39   3m50s
``` 
- The webhook denies the request when multiple images are found with the friendly image name:

```console
$ kubectl get vmi -n sdiliyaer-test | {head -1; grep test-image}
NAME                    PROVIDER-NAME              CONTENT-LIBRARY-NAME   IMAGE-NAME                                              VERSION         OS-TYPE                 FORMAT   AGE
vmi-c23e87019ed276a63   clitem-c23e87019ed276a63   cl-00939f93d26f0624b   test-image                                                              centos8_64Guest         OVF      12m
vmi-c3b4c562afe996007   clitem-c3b4c562afe996007   cl-4def6cd2666ec1b87   test-image                                                              centos8_64Guest         OVF      13m

$ kubectl apply -f /tmp/vm-dup-image-name.yaml
Error from server (no single VM image exists for "test-image"): error when creating "/tmp/vm-dup-image-name.yaml": admission webhook "default.mutating.virtualmachine.vmoperator.vmware.com" denied the request: no single VM image exists for "test-image"
```

### DIFF
--- a/controllers/contentlibrary/v1alpha1/utils/utils.go
+++ b/controllers/contentlibrary/v1alpha1/utils/utils.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2023 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package utils
@@ -17,22 +17,24 @@ import (
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
 )
 
-// GetImageFieldNameFromItem returns the Image field name in format of "vmi-<uuid>"
-// by using the same identifier from the given Item name in "clitem-<uuid>".
+// GetImageFieldNameFromItem returns the image field name in the format of
+// "vmi-<uuid>" by using the identifier from the given item name "clitem-<uuid>".
 func GetImageFieldNameFromItem(itemName string) (string, error) {
 	if !strings.HasPrefix(itemName, ItemFieldNamePrefix) {
-		return "", fmt.Errorf("item name doesn't start with %q", ItemFieldNamePrefix)
+		return "", fmt.Errorf("item name %q doesn't start with %q",
+			itemName, ItemFieldNamePrefix)
 	}
+
 	itemNameSplit := strings.Split(itemName, "-")
 	if len(itemNameSplit) < 2 {
-		return "", fmt.Errorf("item name doesn't have an identifier after %s-", ItemFieldNamePrefix)
+		return "", fmt.Errorf("item name %q doesn't contain a uuid", itemName)
 	}
 
 	uuid := strings.Join(itemNameSplit[1:], "-")
 	return fmt.Sprintf("%s-%s", ImageFieldNamePrefix, uuid), nil
 }
 
-// IsItemReady returns if the given item conditions contain a Ready condition with status True.
+// IsItemReady checks if an item is ready by iterating through its conditions.
 func IsItemReady(itemConditions imgregv1a1.Conditions) bool {
 	var isReady bool
 	for _, condition := range itemConditions {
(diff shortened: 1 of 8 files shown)


controllers/contentlibrary/v1alpha1/utils/utils.go
```diff
@@ -45,23 +47,68 @@ func IsItemReady(itemConditions imgregv1a1.Conditions) bool {
...
 
-// GetVMImageSpecStatus returns the VirtualMachineImage Spec and Status fields in an image-registry service enabled cluster.
-// It first tries to get the namespace scope VM Image; if not found, it looks up the cluster scope VM Image.
-func GetVMImageSpecStatus(ctx context.Context, ctrlClient client.Client, imageName, namespace string) (
-	spec *vmopv1.VirtualMachineImageSpec, status *vmopv1.VirtualMachineImageStatus, err error) {
-
+// GetVMImageSpecStatus retrieves the VirtualMachineImageSpec and
+// VirtualMachineImageStatus for a given image name and namespace.
+// It first checks if a namespace scope image exists by the resource name,
+// and if not, checks if a namespace scope image exists by the status name.
```
zyiyi11, line 53: suggest: `checks if one and only one namespace scope image exists by the status image name(friendly name).` ditto below line 73

controllers/contentlibrary/v1alpha1/utils/utils.go
```diff
@@ -45,23 +47,68 @@ func IsItemReady(itemConditions imgregv1a1.Conditions) bool {
...
+	if !apierrors.IsNotFound(err) {
+		return nil, nil, err
+	}
+
+	// Check if a namespace scope image exists by the image status name.
+	vmiList := vmopv1.VirtualMachineImageList{}
+	err = ctrlClient.List(ctx, &vmiList, client.InNamespace(namespace))
+	if err == nil {
+		for _, vmi := range vmiList.Items {
+			if vmi.Status.ImageName == imageName {
```
zyiyi11, line 78: I think we need to add a check to see this vmi is the only one with this image name. Because for one image that is uploaded to multiple content library in the same ns, we cannot just return the first one found.

dilyar85: Good point. I was trying to best effort the image lookup process earlier. After further reflection, I think it would be better to throw an error if duplicate image status names exist as you suggested. This avoids unexpected behaviors, especially if those images are significantly different, but happen to have the same status name.

webhooks/virtualmachine/v1alpha1/mutation/virtualmachine_mutator.go
```diff
@@ -154,6 +186,48 @@ func SetDefaultPowerState(
 	return false
 }
 
+// ResolveImageName mutates the vm.spec.imageName if it's not set to a vmi name
```
zyiyi11, line 189: I think you meant `if it's set to a vmi name`

webhooks/virtualmachine/v1alpha1/mutation/virtualmachine_mutator.go
```diff
@@ -154,6 +186,48 @@ func SetDefaultPowerState(
...
+	// Check if a single namespace scope image exists by the status name.
+	vmiList := &vmopv1.VirtualMachineImageList{}
+	if err := c.List(ctx, vmiList, client.InNamespace(vm.Namespace),
+		client.MatchingFields{
+			"status.imageName": imageName,
+		},
+	); err != nil {
+		return false, err
+	}
+	if len(vmiList.Items) == 1 {
```
zyiyi11, line 211: in the case where 2 vmi and 1 cvmi has the same friendly name, we return cvmi. Is this the behavior we want?

webhooks/virtualmachine/v1alpha1/mutation/virtualmachine_mutator.go
dilyar85: No, the webhook only mutates `vm.spec.imageName` if it is not already set to a determined vmi name (i.e. a resource name with the "vmi-" prefix).

webhooks/virtualmachine/v1alpha1/mutation/virtualmachine_mutator.go
dilyar85: It's a rare case where the same image needs to be pushed into two CLs associated with the namespace, as well as another CL associated with the cluster. However, if that happens, I think it's acceptable to return the cvmi as it's determined in the cluster scope.

zyiyi11 (approved):
LGTM! Thanks for the change.

PR #228, opened 2023-09-22
:seedling: Bump up min go version to 1.20

aruneshpa:
`controller-runtime` v15 has a min Golang version requirement that breaks the container builds.  This change bumps up the Go version used to build the container.  Starting 1.21 Golang, go.mod requires the Go version and treats that as min required go version to compile your binary.  So, also bump up the Go version for local builds.

N/A

N/A

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: ci
 
 env:
-  GO_VERSION: 1.19.3
+  GO_VERSION: 1.20.0
 
 on:
   pull_request:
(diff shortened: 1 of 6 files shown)


dilyar85 (commented):
I think you may wanna update the [ci.yaml](https://github.com/vmware-tanzu/vm-operator/blob/main/.github/workflows/ci.yml#L4) as well to keep the go version consistent.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: ci
 
 env:
-  GO_VERSION: 1.19.3
+  GO_VERSION: 1.20
```
dilyar85, line 4: Looks like the `verify-go-modules` action failed because YAML trims this down to "1.2". 
Try wrapping it in a single quote?

dilyar85 (approved):

bryanv (approved):

PR #229, opened 2023-09-25
✨ Persist VM disk path and type info in ExtraConfig

dilyar85:
This PR persists a VM's disk path and type (determined by the existence of `VDiskID`) in ExtraConfig. This is required to register any disks as FCDs and expose them as PVCs during a restore operation.

N/A

Currently, the `vm.Reconfigure()` task will not be called because the corresponding FSS is not enabled. A follow-up patch will be submitted to optimize that VC call during VM reconciliation.

```release-note
Persist VM disk path and type info in ExtraConfig.
```

**Testing Done**:
- Added unit tests to cover the disk info check in vcSim VM's ExtraConfig
- Manually deployed this change to a WCP testbed and verified the expected backup data

```console
$ govc vm.info -e "test-vm" | grep vmservice.virtualmachine.diskdata
    vmservice.virtualmachine.diskdata:       H4sIAAAAAAAA/4quVgpzySzO9nRRslJS0lFyy8xJ9UvMTVWyUoouzkgsSk0Jy00r1jWIVShJLS7RLcvVh9J6Zbkp2Uq1sQAAAAD//wEAAP//LVdiuEEAAAA=

# VM without any FCD attached
$ echo "H4sIAAAAAAAA/4quVgpzySzO9nRRslJS0lFyy8xJ9UvMTVWyUoouzkgsSk0Jy00r1jWIVShJLS7RLcvVh9J6Zbkp2Uq1sQAAAAD//wEAAP//LVdiuEEAAAA=" \ 
| base64 -d | gunzip | jq .
[
  {
    "VDiskID": "",
    "FileName": "[sharedVmfs-0] test-vm/test-vm.vmdk"
  }
]

# VM with an FCD attached
$ echo "H4sIAAAAAAAA/3zNsQrCMBCA4VeRmxvbq8kl7VwEF8cupcjlkmCpAbGQRXx3Bx2c3P7l55ueMA7Ltp4G6AEqOC63eOYcoYdpu/IjhjGnTTXzrmR1L3LB+hP7ksMKr+r3R8/JU9cplCYo7cgottIqa8m1pDGRM3+NJKEmQSSvGyspajHMNqB3Wh/YSWDDX3h+AwAA//8BAAD//z6Cb9q6AAAA" \
| base64 -d | gunzip | jq .
[
  {
    "VDiskID": "",
    "FileName": "[sharedVmfs-0] vm-pvc_1/vm-pvc.vmdk"
  },
  {
    "VDiskID": "1bafb699-1c0d-4865-a7c2-77682641f685",
    "FileName": "[sharedVmfs-0] fcd/6c116b407cfe4c5aa7d1b8443a8cda5a.vmdk"
  }
]
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere/virtualmachine/backup.go
+++ b/pkg/vmprovider/providers/vsphere/virtualmachine/backup.go
@@ -4,6 +4,9 @@
 package virtualmachine
 
 import (
+	goctx "context"
+	"encoding/json"
+
 	"sigs.k8s.io/yaml"
 
 	"github.com/vmware/govmomi/object"
@@ -13,11 +16,21 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/util"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/constants"
+	res "github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere/resources"
 )
 
-// BackupVirtualMachine backs up the VM's kube data and bootstrap data into the
-// vSphere VM's ExtraConfig. The status fields will not be backed up, as we expect
-// to recreate and reconcile these resources during the restore process.
+type VMDiskData struct {
+	// ID of the virtual disk object (only set for FCDs).
+	VDiskID string
+	// Filename contains the datastore path to the virtual disk.
+	FileName string
+}
+
+// BackupVirtualMachine backs up the required data of a VM into its ExtraConfig.
+// Currently, the following data is backed up:
+// - Kubernetes VirtualMachine object in YAML format (without its .status field).
+// - VM bootstrap data in JSON (if provided).
+// - List of VM disk data in JSON (including FCDs attached to the VM).
 func BackupVirtualMachine(
 	vmCtx context.VirtualMachineContext,
 	vcVM *object.VirtualMachine,
@@ -37,6 +50,12 @@ func BackupVirtualMachine(
 		vmCtx.Logger.Info("No VM bootstrap data is provided for backup")
 	}
 
+	vmDiskData, err := getEncodedVMDiskData(vmCtx, vcVM)
+	if err != nil {
+		vmCtx.Logger.Error(err, "Failed to get encoded VM disk data")
+		return err
+	}
+
 	_, err = vcVM.Reconfigure(vmCtx, types.VirtualMachineConfigSpec{
 		ExtraConfig: []types.BaseOptionValue{
 			&types.OptionValue{
@@ -46,6 +65,10 @@ func BackupVirtualMachine(
 			&types.OptionValue{
 				Key:   constants.BackupVMBootstrapDataExtraConfigKey,
 				Value: vmBootstrapData,
+			},
+			&types.OptionValue{
+				Key:   constants.BackupVMDiskDataExtraConfigKey,
+				Value: vmDiskData,
 			},
 		}})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


akutz:
@bryanv I really want to get to a point where we can basically track a ConfigSpec throughout the reconcile and do one reconfigure at the end of the loop. Same for the property collector. A prop get at the front of the loop and pass that throughout the reconciler.

bryanv:
> @bryanv I really want to get to a point where we can basically track a ConfigSpec throughout the reconcile and do one reconfigure at the end of the loop. Same for the property collector. A prop get at the front of the loop and pass that throughout the reconciler.

Yah we've kind of, sort of have those pieces in place but that flow is the last remnant of the crummy prototype code that has been a constant battle of one step forward, two steps back in any effort to improve it. I think I took a decent step on the "create" path in my v1a2 work but dropped doing the "update" path due the pain of keeping a long running branch in sync.

Here though they way is being done is fine to experiment with but it will have to be improved (we cannot Reconfigure on every VM reconcile) before enablement. This needs to be folded into the "update" path before enablement, and maybe we can use it as a forcing function and line in the sand to work the existing code in to shape to support that.

dilyar85:
> Here though they way is being done is fine to experiment with but it will have to be improved (we cannot Reconfigure on every VM reconcile) before enablement. This needs to be folded into the "update" path before enablement, and maybe we can use it as a forcing function and line in the sand to work the existing code in to shape to support that.

Yep, I am currently working on reducing the VC call when storing backup data in the VM's ExtraConfig. I will also explore optimizing the "update" path to combine all reconfiguration calls at the end.

pkg/vmprovider/providers/vsphere/virtualmachine/backup.go
```diff
@@ -72,10 +92,41 @@ func getEncodedVMBootstrapData(bootstrapData map[string]string) (string, error)
...
+		if disk, ok := device.(*types.VirtualDisk); ok {
+			vmDiskData := VMDiskData{}
+			if disk.VDiskId != nil {
+				vmDiskData.VDiskID = disk.VDiskId.Id
+			}
+			if b, ok := disk.Backing.(*types.VirtualDiskFlatVer2BackingInfo); ok {
+				vmDiskData.FileName = b.FileName
+			}
+
+			diskData = append(diskData, vmDiskData)
```
bryanv, line 122: Does this need to be added if both fields are empty in the `vmDiskData`?

pkg/vmprovider/providers/vsphere/virtualmachine/backup_test.go
```diff
@@ -46,54 +47,67 @@ func backupTests() {
...
-		It("Should back up both VM kube data and bootstrap data in ExtraConfig", func() {
-			Expect(virtualmachine.BackupVirtualMachine(vmCtx, vcVM, bootstrapData)).To(Succeed())
-			verifyBackupDataInExtraConfig(ctx, vcVM, vmCtx, bootstrapData)
+	Context("VM Disk data", func() {
+
+		It("Should backup VM disk data as JSON in ExtraConfig", func() {
+			Expect(virtualmachine.BackupVirtualMachine(vmCtx, vcVM, nil)).To(Succeed())
+
+			// Use the default disk info from the vcSim VM for testing.
+			diskData := []virtualmachine.VMDiskData{
```
bryanv, line 90: Can we flip how this assertion is done around:

- assert the expected key is in the EC
- decode/uncompress the value
- assert expected size/value in the array

so it is clearer when something doesn't match what part that is. We have some helpers to convert the EC array to a map for easier lookup.

bryanv:
I haven't fully catch up from being out for a few weeks but why do we need to duplicate these from the VM's config into the VM's config ExtraConfig?

pkg/vmprovider/providers/vsphere/virtualmachine/backup.go
dilyar85: Updated to skip empty `vmDiskData` element.

pkg/vmprovider/providers/vsphere/virtualmachine/backup_test.go
dilyar85: Updated the `verifyBackupDataInExtraConfig()` using the suggested assertion.

dilyar85:
> I haven't fully catch up from being out for a few weeks but why do we need to duplicate these from the VM's config into the VM's config ExtraConfig?

@bryanv Vendors currently do not support backup/restore of VMs with FCDs attached. When a VM is restored, all of its disks are recreated as regular static disks. We're persisting the paths and types of disks in the VM's ExtraConfig, so that we can later register and expose those FCDs as PVCs during a restore operation.

pkg/vmprovider/providers/vsphere/virtualmachine/backup.go
```diff
@@ -37,6 +48,11 @@ func BackupVirtualMachine(
 		vmCtx.Logger.Info("No VM bootstrap data is provided for backup")
 	}
 
+	vmDiskData, err := getEncodedVMDiskData(vmCtx.Context, vcVM)
+	if err != nil {
+		vmCtx.Logger.Error(err, "Failed to get encoded VM disk data")
```
bryanv, line 53: Missing `return err`? Or is this purposely keep going to use an empty value (I'd `vmDiskData = ""` here to make that really clear then)

pkg/vmprovider/providers/vsphere/virtualmachine/backup.go
```diff
@@ -37,6 +48,11 @@ func BackupVirtualMachine(
 		vmCtx.Logger.Info("No VM bootstrap data is provided for backup")
 	}
 
+	vmDiskData, err := getEncodedVMDiskData(vmCtx.Context, vcVM)
```
bryanv, line 51: Can just use `vmCtx` here instead. Ignore comment if you intentionally did this for readability or another reason.

bryanv (approved):
Approved but prb needs a minor change

pkg/vmprovider/providers/vsphere/virtualmachine/backup.go
dilyar85: Added the missing `return err` statement. For now we only proceed with calling VM reconfiguration if all the required backup data is available. It may be changed as a best effort in the follow-up optimization.

PR #231, opened 2023-09-26
:seedling: Emit a webhook warning when OvfEnv or ExtraConfig transport is specified

aruneshpa:
This change emits a warning when a user specifies OvfEnv transport in a VM's spec.  In order to support warnings, modify the BuildValidationResponse so it can now take warnings from any validations.
At some point, we should move our webhooks to controller-runtime generated webhooks which will remove the need to synthesize the response from errors and warnings ourselves.

None

```release-note
Emit a webhook warning when OvfEnv or ExtraConfig transport is specified
```

### DIFF
--- a/webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator.go
@@ -65,6 +65,8 @@ const (
 	invalidNextRestartTimeOnUpdate            = "must be formatted as RFC3339Nano"
 	invalidNextRestartTimeOnUpdateNow         = "mutation webhooks are required to restart VM"
 	modifyAnnotationNotAllowedForNonAdmin     = "modifying this annotation is not allowed for non-admin users"
+	OVFEnvTransportDeprecated                 = "OvfEnv transport is deprecated. Please use CloudInit or vAppConfig transport instead."
+	ExtraConfigTransportDeprecated            = "ExtraConfig transport is deprecated. Please use CloudInit or vAppConfig transport instead."
 )
 
 // +kubebuilder:webhook:verbs=create;update,path=/default-validate-vmoperator-vmware-com-v1alpha1-virtualmachine,mutating=false,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachines,versions=v1alpha1,name=default.validating.virtualmachine.v1alpha1.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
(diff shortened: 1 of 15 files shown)


webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator_unit_test.go
```diff
@@ -422,99 +427,100 @@ func unitTestsValidateCreate() {
 	annotationPath := field.NewPath("metadata", "annotations")
 
 	DescribeTable("create table", validateCreate,
```
bryanv, line 429: From the v1a2 work there is a pending item to start to move away from the table driven tests here because they have become IMO untenable and unreadable for a lot of stuff (there are a lot more network and bootstrap stuff in v1a2 that I have no desire to express with bools).

How about for this MR we leave the table stuff as-is (and add an assert in validateCreate that warning is always empty) to keep this diff smaller & focused, and lay down the start of individual Context/It?

bryanv:
Can we also add an integration test the warning parts of the response? We just need one to validate it going forward - there are a lot of duplicate unit & integration tests there that otherwise don't add much value.

There should be an item in the BL to move BuildValidationResponse() "up" one caller to reduce some duplication but interia has made it difficult to really prioritize that.

Like I showed you yesterday, with v1a2 conversion we now have the pieces in place to use the ctrl-runtime webhooks but it is a fair amount of work - going the CustomValidator route is the prb least miserable way - and we don't get a configured as-we expect k8s client plumb down from above (but we really need to move towards the webhooks dealing with just internal consistency and defaulting instead of sometimes needed external objects). I had briefly considered converting some of the "easy" type webhooks as part of the v1a2 work but dropped it due to the churn involved. I'd prefer we wait until v1a2 has landed, and the dust has settled.

webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator.go
```diff
@@ -187,18 +196,25 @@ func (v validator) ValidateUpdate(ctx *context.WebhookRequestContext) admission.
...
+		return allWarnings, allErrs
 	}
 
 	mdPath := field.NewPath("spec", "vmMetadata")
 
+	// OvfEnv transport is marked as deprecated.  Emit a warnings to indicate that to users.
+	if vm.Spec.VmMetadata.Transport == vmopv1.VirtualMachineMetadataOvfEnvTransport {
+		allWarnings = append(allWarnings,
+			fmt.Sprint(ovfEnvTransportDeprecated))
+	}
```
dilyar85, line 216: Should we warn `ExtraConfig` here as well since it's deprecated according to our docs at https://vm-operator.readthedocs.io/en/latest/concepts/workloads/guest/#extraconfig?

webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator.go
```diff
@@ -65,6 +65,7 @@ const (
 	invalidNextRestartTimeOnUpdate            = "must be formatted as RFC3339Nano"
 	invalidNextRestartTimeOnUpdateNow         = "mutation webhooks are required to restart VM"
 	settingAnnotationNotAllowed               = "adding this annotation is not allowed"
+	ovfEnvTransportDeprecated                 = "OvfEnv transport is deprecated. Please use CloudInit or VAppConfig transport instead."
```
zyiyi11, line 68: `vAppConfig` see: https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha1/virtualmachine_types.go#L192

bryanv (approved):

dilyar85 (approved):
LGTM.

PR #234, opened 2023-09-29
✨ Make v1a1 webconsole controller v1a2 VMProvider aware

sreyasn:

When v1a2 FSS gets enabled, only the v1a2 VM Provider will be available for use. 
Since v1a1 webconsole CRs are still supported in v1a2 world, the v1a1 VM Provider that it uses will NPE.
This requires the v1a1 controller to be v1a2 Provider aware.

Fixes #
N/A

N/A

```
None
```

### DIFF
--- a/controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_intg_test.go
+++ b/controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_intg_test.go
@@ -5,6 +5,7 @@ package v1alpha1_test
 
 import (
 	"context"
+	"os"
 	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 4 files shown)


controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_intg_test.go
```diff
@@ -110,27 +126,59 @@ func webConsoleRequestReconcile() {
...
 			err = ctx.Client.Delete(ctx, vm)
 			Expect(err == nil || k8serrors.IsNotFound(err)).To(BeTrue())
+			err = ctx.Client.Delete(ctx, proxySvc)
+			Expect(err == nil || k8serrors.IsNotFound(err)).To(BeTrue())
+		})
+
+		When("v1a1 provider", func() {
+			It("resource successfully created", func() {
+				Eventually(func() bool {
+					wcr = getWebConsoleRequest(ctx, types.NamespacedName{Name: wcr.Name, Namespace: wcr.Namespace})
```
bryanv, line 145: I know this was the prior code, but in case you're not aware there is a nicer way to do assertions in the `Eventually` block: https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode - so the failure reason is more obvious. I've been trying to clean these up as I'm making other changes in the neighborhood. Up to you if you want to do that here now or not.

bryanv (approved):
Thank you. Once we're done with v1a2 we can start to tackle the fat provider interface and code so we can do this a little cleaner.

controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_intg_test.go
sreyasn: Thanks. I'll fix it.

PR #235, opened 2023-09-29
:bug: Return err during unmarshaling ovf file to trigger reconcile

zyiyi11:
With image service enabled, we now reconcile individual Content Library Items. So we don't want to skip error handling anymore. This change will allow retrying when encountered errors during `RetrieveOvfEnvelopeFromLibraryItem`.

Fixes N/A

```release-note
Return errors to trigger reconcilation when retrieve Ovf envelope from library item
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere/contentlibrary/content_library_provider.go
+++ b/pkg/vmprovider/providers/vsphere/contentlibrary/content_library_provider.go
@@ -196,10 +196,11 @@ func (cs *provider) RetrieveOvfEnvelopeFromLibraryItem(ctx context.Context, item
 		_ = downloadedFileContent.Close()
 	}()
 
+	// OVF file is validated during upload, err here can be internet error.
 	envelope, err := ovf.Unmarshal(downloadedFileContent)
 	if err != nil {
 		logger.Error(err, "error parsing the OVF envelope")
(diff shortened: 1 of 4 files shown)


dilyar85:
@zyiyi11 Can you apply this change to the v1a2 provider as well since the implementation is checked in now?  
Also, I think we can remove the corresponding check at https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/vmprovider/providers/vsphere/contentlibrary/content_library_provider.go#L322-L324 too.

pkg/vmprovider/providers/vsphere/contentlibrary/content_library_provider.go
```diff
@@ -196,10 +196,11 @@ func (cs *provider) RetrieveOvfEnvelopeFromLibraryItem(ctx context.Context, item
 		_ = downloadedFileContent.Close()
 	}()
 
+	// With Image Service validate OVF file, err here can be internet error.
```
dilyar85, line 199: Nit: I'm not sure if it's accurate to say that the "Image Service" is validating OVF files here. You could just mention that OVF files will be validated before pushing.

dilyar85 (approved):

PR #238, opened 2023-10-03
🌱 Reverse configSpec.firmware order of precedence to Image then Classes

sreyasn:

This change reverses of the order of precedence to populate the configSpec from the image first then the class. This is necessary since the vSphere UI is currently unable to create VM Classes with an empty firmware type and always defaults to 'efi', creating scenarios for VM boot failures when the image does not support the firmware. 

This reversal can be changed back once vSphere GUI supports creating VM Classes with an optional/empty firmware type as default for a user.

Fixes #
N/A

N/A

```
VM Image will take precedence over a VM Class to provide the configSpec.firmware type.
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/virtualmachine/configspec_test.go
+++ b/pkg/vmprovider/providers/vsphere2/virtualmachine/configspec_test.go
@@ -13,6 +13,7 @@ import (
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/lib"
+	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere2/constants"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere2/virtualmachine"
 	"github.com/vmware-tanzu/vm-operator/test/builder"
 )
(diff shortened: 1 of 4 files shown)


bryanv (approved):

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec_test.go
```diff
@@ -108,6 +111,37 @@ var _ = Describe("CreateConfigSpec", func() {
...
+			})
+		})
+
+		When("vm has an invalid firmware override annotation", func() {
+			BeforeEach(func() {
+				vm.Annotations[constants.FirmwareOverrideAnnotation] = "foo"
+			})
+
+			It("config spec doesn't have the invalid val", func() {
+				Expect(configSpec.Firmware).ToNot(Equal(vm.Annotations[constants.FirmwareOverrideAnnotation]))
```
zyiyi11, line 142: Just to confirm, if vmi doesn't have firmware specified, and annotation has invalid val. ConfigSpec firmware will be empty val?

zyiyi11 (approved):

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec_test.go
sreyasn: So it goes like,
If 
- annotation has invalid val or doesn't exist then
- vmi doesn't have firmware specified and 
- vmclass doesn't have firmware specified 

then configSpec will be empty when passed on to vpxd.

zyiyi11: so annotation takes precedence to image, image take precedence to vmclass?

sreyasn: annotation is an override to everything. yes.

PR #239, opened 2023-10-04
✨ Update VM backup to skip unchanged data and include instance ID

dilyar85:
**What does this PR do, and why is it needed?**
- Move `ExtraConfigToMap` and `MergeExtraConfig` util functions to "pkg/util/configspec.go"
- Persist cloud-init instance ID to keep it unchanged upon VM restore.
- Optimize backup process to skip reconfiguring the VM with unchanged data.
- Persist backup info in VM's EC after successfully updating the VM object to avoid additional calls from the controller.

**Are there any special notes for your reviewer**:
All the corresponding changes to v1a2 are introduced in #240 in an effort to reduce the size of this PR.

**Please add a release note if necessary**:

```release-note
Update VM backup to skip unchanged data and include instance ID.
```

**Testing Done:**
- Updated test code to cover new functionality
- Manually deployed the change to a testbed and verified expected VM backup behaviors

```console
# Instance ID stored in VM's ExtraConfig
$ govc vm.info -e "test-vm" | grep vmservice.virtualmachine.instanceid
    vmservice.virtualmachine.instanceid:     H4sIAAAAAAAA/0o1T00xNjIz0jVONUvTNTE1NtK1MLc0000yMk80MU61TDMzSwIAAAD//wEAAP//U5zXbyQAAAA=

# Verbose logs from VMOP reconciling a VM already backed up
I1004 19:04:42.948471       1 backup.go:51] vsphere "msg"="Skipping VM kube data backup as unchanged" "vmName"="sdiliyaer-test/test-vm"
I1004 19:04:42.949844       1 backup.go:65] vsphere "msg"="Skipping VM instance ID backup as it already exists" "vmName"="sdiliyaer-test/test-vm"
I1004 19:04:42.951217       1 backup.go:79] vsphere "msg"="Skipping VM bootstrap data backup as unchanged" "vmName"="sdiliyaer-test/test-vm"
I1004 19:04:42.957320       1 backup.go:93] vsphere "msg"="Skipping VM disk data backup as unchanged" "vmName"="sdiliyaer-test/test-vm"
```

PR #240, opened 2023-10-05
🌱 Add VM backup implementation for v1a2

dilyar85:
This patch adds the implementation code for VM backup in v1a2. It includes all the follow-up changes in #239 with some minor updates specific to v1a2.

```release-note
Add VM backup implementation for v1a2.
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/constants/constants.go
+++ b/pkg/vmprovider/providers/vsphere2/constants/constants.go
@@ -144,7 +144,7 @@ const (
 	// BackupVMDiskDataExtraConfigKey is the ExtraConfig key to the VM's disk info
 	// data in JSON, compressed using gzip and base64-encoded.
 	BackupVMDiskDataExtraConfigKey = "vmservice.virtualmachine.diskdata"
-	// BackupVMInstanceIDExtraConfigKey is the ExtraConfig key to the VM's
+	// BackupVMCloudInitInstanceIDExtraConfigKey is the ExtraConfig key to the VM's
 	// Cloud-Init instance ID, compressed using gzip and base64-encoded.
-	BackupVMInstanceIDExtraConfigKey = "vmservice.virtualmachine.instanceid"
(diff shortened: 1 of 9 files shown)


pkg/util/configspec.go
dilyar85, line 1: This file is temporarily changed to pass unit tests. It will be undone after rebasing with #239 in.

pkg/vmprovider/providers/vsphere2/constants/constants.go
```diff
@@ -134,4 +134,17 @@ const (
 	V1alpha2SubnetMask = "V1alpha2_SubnetMask"
 	// V1alpha2FormatNameservers is an alias for versioned templating function V1alpha2_FormatNameservers.
 	V1alpha2FormatNameservers = "V1alpha2_FormatNameservers"
+
+	// BackupVMKubeDataExtraConfigKey is the ExtraConfig key to the VirtualMachine
+	// resource's Kubernetes spec data, compressed using gzip and base64-encoded.
+	BackupVMKubeDataExtraConfigKey = "vmservice.virtualmachine.kubedata"
+	// BackupVMBootstrapDataExtraConfigKey is the ExtraConfig key to the VM's
+	// bootstrap data object, compressed using gzip and base64-encoded.
+	BackupVMBootstrapDataExtraConfigKey = "vmservice.virtualmachine.bootstrapdata"
```
bryanv, line 143: For v1a2 vApp customization there is also be a map of maps: see GetVirtualMachineBootstrap(). I'm pretty sure - once we get around to supporting it - the inlined CloudInit customization can also specify a map of maps.

I'm fine if we just what to worry about this case later.

bryanv (approved):
Just one comment but not necessarily a block if you're OK with some more later work.

pkg/vmprovider/providers/vsphere2/constants/constants.go
dilyar85: Yes, I have also left a todo comment to address this case. I will work on it once the v1a2 customization is ready (thought it would be easier to validate my change then).

PR #241, opened 2023-10-05
 ✨ Improve v1a2 VM Network validation checks

bryanv:
First pass at better checks of the various Network fields during create. The existing table driven tests have reached their limit and are just too cumbersome to keep adding cases to. And it is annoying that the setup and assertion messages are not grouped together. Start new table driven tests for these new Network tests, and we should revisit the rest all the existing tests later once we've moved fully to v1a2. And I want to change we do the tests so that it is easier to share assertions between create and update.

Correct the Gateway4 and Gateway6 docs: these should not have the network prefix, just the IP.

```release-note
NONE
```


sreyasn (approved):
LGTM

dilyar85 (approved):

PR #245, opened 2023-10-12
:seedling: Platform-aware tooling

akutz:

This patch makes all of the `hack/tools` tooling binaries platform aware by placing them in `hack/tools/bin/GOHOSTOS_GHOSTARCH`. The binaries are now also built targeting the `GOOS`/`GOARCH` of the host, not the values that may have been specified via environment variables. This ensures someone overriding `GOOS` or `GOARCH` to cross-compile VM Operator does not result in the creation of tooling binaries unable to run on the local system.

Additionally, the tools targets are no longer marked `PHONY`, nor are the tools [order-only pre-requisites](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) for other targets. Because most tooling is managed via Go modules, the `go.mod` file is a pre-requisite for building tools. This means that tools will be built once per platform and only rebuilt if `go.mod` changes.

Fixes `NA`

I ran `make tools` from the root of the project to ensure all tooling still builds successfully.

```release-note
Tooling binaries are built using the OS/Arch of the local system.
```


aruneshpa (approved):
LGTM.

PR #249, opened 2023-10-17
✨ Synthesize v1a1 Prereq condition from v1a2 conditions

bryanv:
There is various weirdness around how and what we set the v1a1 Prereq condition to, so this isn't quite a complete match but hopefully good enough.

For now just leave the conditions as-is when going from v1a1 to v1a2. We might later want to remove the Prereq - or other v1a1 only conditions - during that conversion.

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/virtualmachine_conversion.go
+++ b/api/v1alpha1/virtualmachine_conversion.go
@@ -541,6 +541,88 @@ func Convert_v1alpha1_VirtualMachineStatus_To_v1alpha2_VirtualMachineStatus(
 	return nil
 }
 
+func translate_v1alpha2_Conditions_To_v1alpha1_Conditions(conditions []Condition) []Condition {
+	var preReqCond, vmClassCond, vmImageCond, vmSetResourcePolicy, vmBootstrap *Condition
+
+	for i := range conditions {
+		c := &conditions[i]
+
+		switch c.Type {
+		case VirtualMachinePrereqReadyCondition:
+			preReqCond = c
+		case v1alpha2.VirtualMachineConditionClassReady:
+			vmClassCond = c
+		case v1alpha2.VirtualMachineConditionImageReady:
+			vmImageCond = c
+		case v1alpha2.VirtualMachineConditionVMSetResourcePolicyReady:
+			vmSetResourcePolicy = c
+		case v1alpha2.VirtualMachineConditionBootstrapReady:
+			vmBootstrap = c
+		}
+	}
+
+	// Try to replicate how the v1a1 provider would set the singular prereqs condition. The class is checked
+	// first, then the image. Note that the set resource policy and metadata (bootstrap) are not a part of
+	// the v1a1 prereqs, and are optional.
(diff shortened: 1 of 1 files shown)


api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -552,6 +634,7 @@ func Convert_v1alpha2_VirtualMachineStatus_To_v1alpha1_VirtualMachineStatus(
 	out.Phase = convert_v1alpha2_Conditions_To_v1alpha1_Phase(in.Conditions)
 	out.VmIp, out.NetworkInterfaces = convert_v1alpha2_NetworkStatus_To_v1alpha1_Network(in.Network)
 	out.LastRestartTime = in.LastRestartTime
+	out.Conditions = convert_v1alpha2_Conditions_To_v1alpha1_Conditions(out.Conditions)
```
dilyar85, line 637: This should be `in.Conditions`?

bryanv: The autoConvert above already converted the types. I'll rename this from `convert_` to like `filter_` or something else.

zyiyi11 (approved):
LGTM. And we will have a `convert_v1alpha1_Conditions_To_v1alpha2_Conditions` on another MR?

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -541,6 +541,88 @@ func Convert_v1alpha1_VirtualMachineStatus_To_v1alpha2_VirtualMachineStatus(
...
+func convert_v1alpha2_Conditions_To_v1alpha1_Conditions(conditions []Condition) []Condition {
+	var preReqCond, vmClassCond, vmImageCond, vmSetResourcePolicy, vmBootstrap *Condition
+
+	for i := range conditions {
+		c := &conditions[i]
+
+		switch c.Type {
+		case VirtualMachinePrereqReadyCondition:
+			preReqCond = c
+		case v1alpha2.VirtualMachineClassConditionReady:
```
sreyasn, line 553: Is this VirtualMachineConditionClassReady instead?.. 
https://github.com/vmware-tanzu/vm-operator/blob/3c428b154b5caea84f2da3ee4b1ff7512ee516f5/api/v1alpha2/virtualmachine_types.go#L18

api/v1alpha1/virtualmachine_conversion.go
sreyasn: From the above seem like the autoconvert retains all the v1a2 Conditions as is for the v1a1 CR to use. This convert_/filter_ method is to primarily synthesize the v1a1 preReq condition then, BV?.

bryanv: Correct

bryanv:
> LGTM. And we will have a `convert_v1alpha1_Conditions_To_v1alpha2_Conditions` on another MR?

The straight type conversion is already there. Unless something breaks I don't intend to do anything v1a1 -> v1a2: v1a1 doesn't have the required info, and we'll let the VM reconcile populate what conditions it can.

PR #253, opened 2023-10-18
✨ Finalize network v1a2 naming convention

sreyasn:
This change checks for v1a1 network named interfaces before creating or updating netIfs with the new naming convention.

Fixes # N/A

N/A

```
N/A
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/network/network.go
+++ b/pkg/vmprovider/providers/vsphere2/network/network.go
@@ -16,6 +16,7 @@ import (
 	"github.com/vmware/govmomi/vim25"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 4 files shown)


dilyar85 (approved):
LGTM.

pkg/vmprovider/providers/vsphere2/network/network.go
```diff
@@ -259,30 +260,42 @@ func createNetOPNetworkInterface(
...
+		Name:      NetOPCRName(vmCtx.VM.Name, networkName, interfaceSpec.Name, true),
+	}
 
-	netIf := &netopv1alpha1.NetworkInterface{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      NetOPCRName(vmCtx.VM.Name, networkName, interfaceSpec.Name, false),
-			Namespace: vmCtx.VM.Namespace,
-		},
+	// check if a networkIf object exists with the older (v1a1) naming convention
+	if err = client.Get(vmCtx, netIfKey, netIf); ctrlruntime.IgnoreNotFound(err) != nil {
```
bryanv, line 273: Move this on to its own line `err := client.Get(...)`, and remove the `err` declaration on L263 - I prefer that any vars in the if itself are scoped to just that.

Or with my later comment, maybe something like 
```
if err := client.Get(...); err != nil {
    if !apierrors.IsNotFound(err) {
        return err
    }

    name =  NetOPCRName(..., false)
}
```

pkg/vmprovider/providers/vsphere2/network/network.go
```diff
@@ -259,30 +260,42 @@ func createNetOPNetworkInterface(
...
+		return nil, err
 	}
 
-	_, err := controllerutil.CreateOrUpdate(vmCtx, client, netIf, func() error {
```
bryanv, line 272: Still always call the `CreateOrUpdate`. We want to make sure that this VM is the controlling owner of the NIC CR before using it.

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator_unit_test.go
```diff
@@ -260,6 +262,29 @@ func unitTestsValidateCreate() {
 			ctx.vm.Annotations[vmopv1.FirstBootDoneAnnotation] = updateSuffix
 		}
 
+		if args.extraLongNetworkIfName {
```
bryanv, line 265: I have a strong preference that we start to move away from this struct/table driven test because it is just getting too unwieldy and hard to follow. Later in this file there's a `Context("Network") { ...` where I added newer style tests for the Network stuff - let's move these test in there.

sreyasn: gotcha. Done

bryanv (approved):

PR #256, opened 2023-10-23
🌱 Enabled v1a1 VM mutation webhooks on v1a2 FSS enabled env

sreyasn:
The v1a1 VM mutating webhooks are still required in a v1a2 environment to preserve fields in v1a1 unknown to v1a2 CR.
The mutation webhooks also run serially, so they don't have the problem of racing that a validation webhook would have.

Fixes # N/A

N/A

```
N/A
```


dilyar85 (approved):
LGTM. Thanks for adding the comment for clarification.

PR #258, opened 2023-10-24
:seedling: Allow privileged users to modify annotations on a VirtualMachine object

aruneshpa:

Currently, we have webhooks to prevent users from modifying system used annotations.  This is done by validating if the user making the request is the kube-admin user.  However, other parts of the infrastructure (such as WCP service) are not considered privileged, which is required for backup/restore.

This changes expands the definition of a privileged account by including WCP service solution user.  The code expects this particular environment variable to be present in the pod's environment variable.

Testing Done:

1. Create a ServiceAccount
2. Create a ClusterRoleBinding with a roleRef that gives the ServiceAccount permissions to edit/update/delete VM resources
3. Invoke create VM with a reserved annotation
```
$ k apply -f test-yamls/vm.yaml --as system:serviceaccount:default:test-sa
Error from server (metadata.annotations.virtualmachine.vmoperator.vmware.com/first-boot-done: Forbidden: modifying this annotation is not allowed for non-admin users): error when creating "test-yamls/vm.yaml": admission webhook "default.validating.virtualmachine.v1alpha1.vmoperator.vmware.com" denied the request: metadata.annotations.virtualmachine.vmoperator.vmware.com/first-boot-done: Forbidden: modifying this annotation is not allowed for non-admin users
```
5. Modify the VM operator deployment to add the env var.
6. Try to create the VM with the annotation again and verify that it goes through:
```
$ k apply -f test-yamls/vm.yaml --as system:serviceaccount:default:test-sa
virtualmachine.vmoperator.vmware.com/parunesh-test-4 created
```

Fixes N/A

N/A

```release-note
Allow privileged users to modify annotations on a VirtualMachine object
```

### DIFF
--- a/config/wcp/vmoperator/manager_env_var_patch.yaml
+++ b/config/wcp/vmoperator/manager_env_var_patch.yaml
@@ -81,3 +81,9 @@
   value:
     name: FSS_WCP_VMSERVICE_BACKUPRESTORE
     value: "<FSS_WCP_VMSERVICE_BACKUPRESTORE_VALUE>"
+
+- op: add
+  path: /spec/template/spec/containers/0/env/-
+  value:
+    name: PRIVILEGED_USERS
+    value: "<COMMA_SEPARATED_LIST_OF_USERS>"
--- a/pkg/lib/env.go
+++ b/pkg/lib/env.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -28,6 +29,13 @@ const (
 	VMServiceBackupRestoreFSS     = "FSS_WCP_VMSERVICE_BACKUPRESTORE"
 	MaxCreateVMsOnProviderEnv     = "MAX_CREATE_VMS_ON_PROVIDER"
 	DefaultMaxCreateVMsOnProvider = 80
+
+	// PrivilegedUsersEnv is the key for the environment variable
+	// containing a set of privileged users. Currently, users pointed by
+	// this env variable (if VM Service Backup/Restore FSS is enabled),
+	// along with kube-admin and system users are allowed to make changes
+	// to certain restricted annotations on a VirtualMachine resource.
+	PrivilegedUsersEnv = "PRIVILEGED_USERS"
 
 	InstanceStoragePVPlacementFailedTTLEnv = "INSTANCE_STORAGE_PV_PLACEMENT_FAILED_TTL"
 	// DefaultInstanceStoragePVPlacementFailedTTL is the default wait time before declaring PV placement failed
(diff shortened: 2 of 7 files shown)


pkg/lib/env.go
```diff
@@ -27,6 +27,7 @@ const (
 	WindowsSysprepFSS             = "FSS_WCP_WINDOWS_SYSPREP"
 	VMServiceBackupRestoreFSS     = "FSS_WCP_VMSERVICE_BACKUPRESTORE"
 	MaxCreateVMsOnProviderEnv     = "MAX_CREATE_VMS_ON_PROVIDER"
+	WCPServiceUserEnv             = "WCPSVC_USER"
```
akutz, line 30: Please document either this constant or the function `GetWCPServiceSolutionUser` explaining what the value of the env var is and how it is used.

akutz (changes requested):
Thanks @aruneshpa. I'm leaving one in-line comment for now and then this one in general -- why not make this generic? Instead of calling it `WCP_USER` or something like that, why not `PRIVILEGED_USERS` as an env var that can be a comma-separated list of users/roles that fall into the privileged user status?

aruneshpa:
> Thanks @aruneshpa. I'm leaving one in-line comment for now and then this one in general -- why not make this generic? Instead of calling it `WCPSVC_USER` or something like that, why not `PRIVILEGED_USERS` as an env var that can be a comma-separated list of users/roles that fall into the privileged user status?

Hmm. @akutz , we did not have a use-case of another privileged user and I don't expect to change that frequently.  That being said, I like the idea of making this extensible.  Taking your suggestion up.

config/wcp/vmoperator/manager_env_var_patch.yaml
```diff
@@ -81,3 +81,9 @@
   value:
     name: FSS_WCP_VMSERVICE_BACKUPRESTORE
     value: "<FSS_WCP_VMSERVICE_BACKUPRESTORE_VALUE>"
+
+- op: add
+  path: /spec/template/spec/containers/0/env/-
+  value:
+    name: PRIVILEGED_USERS
+    value: "<WCPSVC_USER>"
```
akutz, line 89: `<COMMA_SEPARATED_LIST_OF_USERS>`

akutz (changes requested):
See the update to the placeholder env var value. Also, please add some unit tests against mulitple users in the value.

akutz (approved):
lgtm, thank you!

pkg/lib/env.go
```diff
@@ -29,6 +30,13 @@ const (
 	MaxCreateVMsOnProviderEnv     = "MAX_CREATE_VMS_ON_PROVIDER"
 	DefaultMaxCreateVMsOnProvider = 80
 
+	// PrivilegedUsersEnv is the key for the environment variable
+	// containing a set of privileged users. Currently, users pointed by
```
dilyar85, line 34: The word "currently" might cause confusion because the env var privileged users are included only if the backup & restore FSS are enabled (which is not atm). Could you make this clearer in the comment, or remove the FSS check from the `IsPrivilegedAccount` function?

webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator_unit_test.go
```diff
@@ -410,11 +428,13 @@ func unitTestsValidateCreate() {
 		ctx = newUnitTestContextForValidatingWebhook(false)
 		oldFaultDomainsFunc = lib.IsWcpFaultDomainsFSSEnabled
 		oldImageRegistryFunc = lib.IsWCPVMImageRegistryEnabled
+		oldVMServiceBackupRestoreFunc = lib.IsVMServiceBackupRestoreFSSEnabled
```
dilyar85, line 431: I think the recommended way is to use `os.Setenv` and `os.Unsetenv`, see [this](https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/vmprovider/providers/vsphere/vmprovider_vm_utils_test.go#L70-L74) for an example.

dilyar85 (approved):
Left some minor comments and overall LGTM!

webhooks/virtualmachine/v1alpha1/validation/virtualmachine_validator_unit_test.go
aruneshpa: I noticed, and considered that.  One issue with that approach is the fact that we don't "reset" the values of method.  As such, we run the risk of causing other tests to fail.  IOW, this test will not leave the value of the env var was it was before the test execution.

pkg/lib/env.go
aruneshpa: Modified the text to indicate that the privileged users are only allowed when the FSS is enabled.  I think we need to keep the FSS check around to ensure that net-new changes are guarded by an FSS.

PR #261, opened 2023-11-01
🌱 Adds a Ready condition to CVMI and VMI resources

srm09:
This patch introduces 2 different approaches to add a Ready condition to the ClusterVirtualMachineImage and VirtualMachineImage CRs. 

For ClusterVirtualMachineImage, we have added the Ready condition in addition the existing conditions, and calculated the Ready condition as a summary of the existing conditions on the CVMI.
For VirtualMachineImage, we have removed the existing conditions and used the appropriate reason when setting the Ready condition on the VMI.

Fixes n/a

Following tests have been done:
- Existing CVMI resource's Status with previous condition types are replaced with the `Ready` condition **only**.
- Existing VMI resource's Status with previous condition types are replaced with the `Ready` condition **only**.
- New CVMI resource's Status has the Ready condition **only** with the appropriate reason.
- New VMI resource's Status has the Ready condition **only** with the appropriate reason.

```release-note
🌱 Adds a Ready condition to CVMI and VMI resources
```

### DIFF
--- a/api/v1alpha1/virtualmachineimage_conversion.go
+++ b/api/v1alpha1/virtualmachineimage_conversion.go
@@ -4,10 +4,13 @@
 package v1alpha1
 
 import (
-	"github.com/vmware-tanzu/vm-operator/api/v1alpha2"
-	"github.com/vmware-tanzu/vm-operator/api/v1alpha2/common"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiconversion "k8s.io/apimachinery/pkg/conversion"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
+
+	"github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+	"github.com/vmware-tanzu/vm-operator/api/v1alpha2/common"
 )
 
 func Convert_v1alpha1_ContentProviderReference_To_common_LocalObjectRef(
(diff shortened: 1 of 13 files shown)


dilyar85:
I don't have a strong preference between these two options. It appears that both are commonly used in other K8s components (e.g. Pod uses a summarised Ready condition while CertManager uses a single Ready condition).
I also looked up the K8s [API Conventions doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) which mentions that "Conditions should be added to explicitly convey properties that users and components care about." Based on this, I am inclined to drop the other conditions since they are not explicitly used by users or other components. It will also be more consistent and maintainable if we support different image providers in the future.

pkg/conditions2/getter.go
```diff
@@ -20,7 +20,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+	vmop "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
```
bryanv, line 23: I prefer to keep the existing alias. Prior commits tried to standardize on `vmopv1` where we could so the v1a1 and v1a2 code has few differences.

bryanv (approved):

pkg/conditions2/getter.go
srm09: I kept stumbling across the vmopv1 alias and double checking the import to verify if it is alpha1 or alpha2. 
Was hoping to eventually get the aliases sorted out across the codebase without an all encompassing PR to replace all occurrences at once.

dilyar85 (approved):
LGTM!

controllers/contentlibrary/v1alpha2/clustercontentlibraryitem/clustercontentlibraryitem_controller.go
```diff
@@ -156,25 +156,43 @@ func (r *Reconciler) ReconcileNormal(ctx *context.ClusterContentLibraryItemConte
 			savedStatus = cvmi.Status.DeepCopy()
 		}()
 
+		// Resetting the conditions to recalculate the latest state.
+		cvmi.SetConditions([]metav1.Condition{})
```
bryanv, line 160: This will cause the image object to be updated on every reconcile because we'll always set a new (current) LastTransitionTime. The condition helper Set() would otherwise preserve that unless there is a state transition.

srm09: I had specifically done this to remove the older conditions which might be present in case of brownfield environments. Having those v1a1 conditions is alright I suppose, should not be a big issue. Going forward, we are planning to rally around the `Ready` condition only, so the presence of those conditions in some cases should not be a problem.

bryanv: I'd agreed the older conditions are probably harmless. If not we can use the version conversion to remove them. But we do need to preserve the LastTransitionTime if there wasn't a state transition.

bryanv (approved):
Thanks for taking care of this.

api/v1alpha1/virtualmachineimage_conversion.go
```diff
@@ -86,6 +89,95 @@ func convert_v1alpha1_VirtualMachineImage_OVFEnv_To_v1alpha2_VirtualMachineImage
...
+	}
+
+	return v1a1Conditions
+}
+
+func convert_v1alpha1_VirtualMachineImageStatusConditions_To_v1alpha2_VirtualMachineImageStatusConditions(
+	conditions []Condition) []metav1.Condition {
+	var readyCondition *metav1.Condition
+	for _, condition := range conditions {
+		if condition.Status == corev1.ConditionFalse {
```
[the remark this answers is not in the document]
srm09: Done

PR #263, opened 2023-11-03
✨ Version conversion support for v1a2 BootDiskCapacity

bryanv:
In v1a1 we added the VsphereVolume to allow the resize of the OVF disks. It was something done quickly, and really never ended up being used, and the interface isn't very user friendly for general purpose for "VM Service" VMs. In v1a2 we just have a single field for resizing the boot disk. As best as we're able to propagate the boot disk field across versions.

```release-note
The v1alpha1 VM Spec.Volumes.VsphereVolumes has been replaced in v1alpha2 VM Spec.Advanced.BootDiskCapacity
```


srm09 (approved):
LGTM

zyiyi11 (approved):

PR #264, opened 2023-11-06
🌱 Move VM backup/restore related constants to API module

dilyar85:
This PR moves the constants used for VM backup/restore into the API module to enforce a contract. These constants are utilized when registering a VM Service VM as part of the VM restore process.

The WCP service codebase will be updated to use these constants imported from the vm-operator/api module once this change is applied.

```release-note
Move VM backup/restore related constants to the API module to enforce a contract.
```

### DIFF
--- a/api/v1alpha1/virtualmachine_types.go
+++ b/api/v1alpha1/virtualmachine_types.go
@@ -112,6 +112,28 @@ const (
 	FirstBootDoneAnnotation = "virtualmachine." + GroupName + "/first-boot-done"
 )
 
+// VirtualMachine backup/restore related constants.
+const (
+	// ManagedByExtensionKey and ManagedByExtensionType represent the ManagedBy
+	// field on the VM. They are used to differentiate VM Service managed VMs
+	// from traditional vSphere VMs.
+	ManagedByExtensionKey  = "com.vmware.vcenter.wcp"
+	ManagedByExtensionType = "VirtualMachine"
+
+	// VMBackupKubeDataExtraConfigKey is the ExtraConfig key to persist the VM's
+	// Kubernetes resource spec data, compressed using gzip and base64-encoded.
+	VMBackupKubeDataExtraConfigKey = "vmservice.virtualmachine.kubedata"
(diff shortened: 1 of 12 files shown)


api/v1alpha1/virtualmachine_types.go
```diff
@@ -112,6 +112,27 @@ const (
 	FirstBootDoneAnnotation = "virtualmachine." + GroupName + "/first-boot-done"
 )
 
+// VirtualMachine backup/restore related constants.
+const (
+	// ManagedByExtensionKey and ManagedByExtensionType represent the ManagedBy field on the VM.
+	// Historically, this field was used to differentiate VM Service managed VMs from traditional ones.
```
zyiyi11, line 118: by adding `historically`, does it mean this field is no longer used to differentiate?

api/v1alpha1/virtualmachine_types.go
```diff
@@ -112,6 +112,27 @@ const (
...
+	ManagedByExtensionKey  = "com.vmware.vcenter.wcp"
+	ManagedByExtensionType = "VirtualMachine"
+
+	// BackupVMKubeDataExtraConfigKey is the ExtraConfig key to the VirtualMachine
```
zyiyi11, line 122: Let's keep it consistent by calling VM or VirtualMachine

zyiyi11 (approved):
minor suggestion on comments. LGTM

api/v1alpha1/virtualmachine_types.go
dilyar85: I am not sure if this is used in any other context besides backup/restore before. The comments are copied from the previous place.

dilyar85: Added an additional comment to clarify in the latest commit.

aruneshpa (approved):
LGTM.

api/v1alpha1/virtualmachine_types.go
dilyar85: Updated the comment of these fields for better clarity.

PR #268, opened 2023-11-08
✨ Make VM Spec BootDiskCapacity field a pointer

bryanv:
This field is a resource.Quantity which when not set still outputs as "0". This field is not likely commonly used, and outputting of a zero size is potentially confusing or misleading. Also, in the k8s API, resource.Quantity fields are always a pointer.

While here, remove this field defaulting in the version conversion fuzzer since that is now supported after f65ef245.

```release-note
NONE
```


srm09:
Outputting a 0 value for boot disk capacity is misleading, ran a quick experiment to confirm that unset Quantity types are outputted as 0. Makes sense.

srm09 (approved):

PR #269, opened 2023-11-09
:seedling:Add validation on Bootstrap vAppConfig Properties

zyiyi11:
This change adds validation on Bootstrap vAppConfig Properties.
1) `Key` is required field https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha2/common/types.go#L75
2) `Value` is optional and is `ValueOrSecretKeySelector`, In which `From` and `Value` is mutually exclusive. https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha2/common/types.go#L49

Fixes # N/A

```release-note
Add validation on Bootstrap vAppConfig Properties
```

### DIFF
--- a/webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
@@ -203,7 +203,7 @@ func (v validator) validateBootstrap(
 		hasCloudConfig := !equality.Semantic.DeepEqual(cloudInit.CloudConfig, cloudinit.CloudConfig{})
 		hasRawCloudConfig := !equality.Semantic.DeepEqual(cloudInit.RawCloudConfig, corev1.SecretKeySelector{})
 		if hasCloudConfig && hasRawCloudConfig {
-			allErrs = append(allErrs, field.Invalid(p, cloudInit,
+			allErrs = append(allErrs, field.Invalid(p, "cloudInit",
 				"cloudConfig and rawCloudConfig are mutually exclusive"))
 		}
 	}
@@ -229,7 +229,7 @@ func (v validator) validateBootstrap(
 			hasSysPrep := !equality.Semantic.DeepEqual(sysPrep.Sysprep, sysprep.Sysprep{})
 			hasRawSysPrep := !equality.Semantic.DeepEqual(sysPrep.RawSysprep, corev1.SecretKeySelector{})
 			if hasSysPrep && hasRawSysPrep {
-				allErrs = append(allErrs, field.Invalid(p, sysPrep,
+				allErrs = append(allErrs, field.Invalid(p, "sysPrep",
 					"sysprep and rawSysprep are mutually exclusive"))
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -249,6 +249,19 @@ func (v validator) validateBootstrap(
...
+
+		if properties := vAppConfig.Properties; len(properties) != 0 {
+			for _, property := range properties {
+				if key := property.Key; key == "" {
+					allErrs = append(allErrs, field.Invalid(p.Child("properties").Child("key"), key,
+						"key is a required field in vAppConfig Properties"))
+				}
+				if value := property.Value; value.From != nil && value.Value != nil {
+					allErrs = append(allErrs, field.Invalid(p.Child("properties").Child("value"), value,
+						"From and Value is mutually exclusive"))
```
srm09, line 261:
```suggestion
						"from and value is mutually exclusive"))
```
since that is what be set in the YAML.

srm09 (commented):
minor nit in the error message.

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -249,6 +249,19 @@ func (v validator) validateBootstrap(
 			allErrs = append(allErrs, field.TypeInvalid(p, vAppConfig,
 				"properties and rawProperties are mutually exclusive"))
 		}
+
+		if properties := vAppConfig.Properties; len(properties) != 0 {
```
bryanv, line 253: This outer `if` can be removed, and iterate over ` ... := range vAppConfig.Properties {` instead

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -249,6 +249,19 @@ func (v validator) validateBootstrap(
...
+			for _, property := range properties {
+				if key := property.Key; key == "" {
```
bryanv, line 255: Nit: I don't think this scope `key` var add any clarity since we know it is `""` in the if body.

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -249,6 +249,19 @@ func (v validator) validateBootstrap(
...
+					allErrs = append(allErrs, field.Invalid(p.Child("properties").Child("key"), key,
+						"key is a required field in vAppConfig Properties"))
+				}
+				if value := property.Value; value.From != nil && value.Value != nil {
+					allErrs = append(allErrs, field.Invalid(p.Child("properties").Child("value"), value,
```
bryanv, line 260: Does using the `value` as the value cause the type to be a part of the error message? (We should avoid serializing types into the error strings)

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator_unit_test.go
```diff
@@ -511,6 +511,45 @@ func unitTestsValidateCreate() {
...
+											LocalObjectReference: corev1.LocalObjectReference{Name: "secret-name"},
+											Key:                  "key",
+										},
+										Value: pointer.String("value"),
+									},
+								},
+							},
+						}
+					},
+					validate: doValidateWithMsg("From and Value is mutually exclusive"),
```
bryanv, line 533: Can we use the whole error message here & below for a stronger assertion?

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
zyiyi11: Yes it does. Makes sense. I am only putting `"value"` as part of error strings then.

bryanv (approved):

PR #271, opened 2023-11-09
✨ Make CloudInit and Sysprep raw and inlined fields pointers

bryanv:
This makes it easier to determine which method is being used, and in the case of CloudInit, not display the inlined config when it is otherwise not being used due to a required non-empty field.

```release-note
NONE
```


pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
```diff
@@ -26,7 +25,7 @@ func BootstrapSysPrep(
 	var data string
 	key := "unattend"
 
-	if equality.Semantic.DeepEqual(sysPrepSpec.Sysprep, sysprep.Sysprep{}) {
+	if !reflect.ValueOf(sysPrepSpec.RawSysprep).IsZero() {
```
akutz, line 28: Why is this not just checking is not nil?

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -226,9 +222,7 @@ func (v validator) validateBootstrap(
 		}
 
 		if lib.IsWindowsSysprepFSSEnabled() {
-			hasSysPrep := !equality.Semantic.DeepEqual(sysPrep.Sysprep, sysprep.Sysprep{})
-			hasRawSysPrep := !equality.Semantic.DeepEqual(sysPrep.RawSysprep, corev1.SecretKeySelector{})
-			if hasSysPrep && hasRawSysPrep {
+			if sysPrep.Sysprep != nil && !reflect.ValueOf(sysPrep.RawSysprep).IsZero() {
```
akutz, line 225: Isn't RawSysprep by address? Why the reflection?

akutz (commented):
I had two small questions. If they are resolved, it looks good to me.

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
akutz: I see what happened. I expected you to make `RawCloudConfig` and `RawSysprep` by address as well.

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
akutz: See previous comment.

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
bryanv: Sure. They're all pointers now.

sreyasn (approved):

PR #273, opened 2023-11-13
🌱 Remove v1a1 VM mutation webhook when v1a2 FSS is enabled

sreyasn:

The v1a1 VM mutation webhook can be safely removed now because the v1a2 VM mutation webhook does the needful.

Fixes # N/A

We initially had certain things that only the v1a1 VM mutation webhook did. Currently both the v1a1 and v1a2 VM mutation webhook do the same things. 

```
```


zyiyi11 (approved):
/LGTM

PR #275, opened 2023-11-14
🐛 v1a2 VM mutation webhook updates

bryanv:
Use the admission request original raw json byte array as the diff patch source: this is needed to generate the correct json diff patches for nested optional fields. This matches a change made to controller-runtime webhook server that we'll eventually switch to.

Sync over remaining differences in the v1a2 webhook and tests from v1a1

```release-note
NONE
```


sreyasn (approved):
Looks good. Thanks for this change, BV.

PR #276, opened 2023-11-21
🌱 Update deprecated pr labeler version

srvaroa:

Bump the version of the PR Labeler action, updating to the newer configuration schema. The size configuration was changed a few versions ago. I'm looking to remove support so I'm sending some PRs to the very few repos that still rely on it.

I updated the reference to the action to master. This actually uses the latest stable version so it should avoid any unintended breakage.

Thanks for using the action!

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -31,7 +31,7 @@ jobs:
     # - "config/crd/bases/.*"  # generated CRDs
     # - "docs/apis/v*.md"      # generated API documentation,
     #                          # ex. docs/apis/v1alpha1.md
-    - uses: srvaroa/labeler@master
+    - uses: srvaroa/labeler@v1.9.0
       with:
         config_path: .github/configs/labeler.yml
       env:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     # - "config/crd/bases/.*"  # generated CRDs
     # - "docs/apis/v*.md"      # generated API documentation,
     #                          # ex. docs/apis/v1alpha1.md
-    - uses: srvaroa/labeler@v0.9
+    - uses: srvaroa/labeler@master
```
akutz, line 34: We want to avoid sliding version targets. Please point to a specific version instead.

akutz:
@srvaroa By the way, thank you very much for such a fantastic GitHub action!

.github/workflows/pr-labeler.yml
srvaroa: Thanks for the review, fixed!

srvaroa:
Thanks @akutz ! I have pinned the version to the latest available.

akutz (approved):
Thank you!

PR #278, opened 2023-11-29
✨ Support v1a2 VirtualMachine full version conversion

bryanv:
The difference between v1a1 and v1a2 VMs is pretty large, and a lot of v1a2 fields are new so there is nothing corresponding in v1a1 to be able to down convert to v1a1, and then back to v1a2. This means that, if set, the new v1a2 fields would be lost on a version conversion round trip.

Use the CAPI trick to store the serialized object in the annotations, and then deser it restore the v1a2 fields. This does get a little messy when we have to try to potentially merge updates down in v1a1 with the restored v1a2 fields.

```release-note
NONE
```


akutz (approved):
Lgtm

PR #281, opened 2023-11-30
✨Move tkg vmware-system-* annotations to vmi/cvmi status

sreyasn:

One of the goals w/ this is to make Metadata watches more effective in v1a2 by not having big yamls in the ObjectMeta.

This change moves the
- tkg vmware-system-* annotations to a v1a2 status field
- adds conversions for this status field to convert to annotations for v1a1.
- v1a1 annotations when up-converted are copied over to the v1a2 status field and removed from v1a2 annotations if already present

Fixes # N/A

Testing TKC deployments and scaled up to check vmware-system annotations are fetched from the v1a1 virtualmachineimage.

```
N/A
```

### DIFF
--- a/api/v1alpha1/virtualmachineimage_conversion.go
+++ b/api/v1alpha1/virtualmachineimage_conversion.go
@@ -4,6 +4,8 @@
 package v1alpha1
 
 import (
+	"strings"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiconversion "k8s.io/apimachinery/pkg/conversion"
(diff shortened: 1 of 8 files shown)


akutz:
Let's make sure we also engage TKG internally to jump-start their move to v1a2.

api/v1alpha1/virtualmachineimage_conversion.go
```diff
@@ -308,6 +345,11 @@ func (src *VirtualMachineImage) ConvertTo(dstRaw conversion.Hub) error {
 		return err
 	}
 
+	if err := convert_v1alpha1_VirtualMachineImageAnnotations_To_v1alpha2_VirtualMachineImage_VMwareSystemProperties(
```
bryanv, line 348: We want to remove the annotations that we're moved to the status. One of the goals w/ this is to make MD-only watches more effective by not have the big TKG yaml blog in the ObjectMeta.

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -349,8 +349,8 @@ func (v validator) validateNetworkInterfaceSpec(
 		networkIfCRName = fmt.Sprintf("%s-%s", vmName, interfaceSpec.Name)
 	}
 
-	for _, msg := range validation.NameIsDNSLabel(networkIfCRName, false) {
-		allErrs = append(allErrs, field.Invalid(interfacePath.Child("name"), interfaceSpec.Name, msg))
+	for _, msg := range validation.NameIsDNSSubdomain(networkIfCRName, false) {
```
bryanv, line 352: This looks fine but I misspoke yesterday - let's do this as a separate MR since it isn't related to the primary change in the MR.

api/v1alpha1/virtualmachineimage_conversion.go
sreyasn: fixed it.. PTAL when you get a chance

bryanv (approved):

PR #285, opened 2023-12-05
✨ Set more useful interface name in v1a2 VM Status

bryanv:
We have a long ways to go before we actually support multiple interfaces, both in bootstrap and in reconcile. The exceedingly common case is just one interface so special case that to provide the interface name in the Status.

```release-note
NONE
```


sreyasn (approved):
LGTM

akutz (approved):

PR #295, opened 2023-12-08
⚠️ Add note that the TCPSocket probe method is deprecated

bryanv:
This probe method was added at the last minute of the first WCP release as a stopgap to address a limitation with the NSX-T LB at the time. Since the probe is done via VM Operator pod on one of the Supervisor CP VMs it requires network connectivity that does not exist in certain environments, and TKGs - the primary user of this probe method - will be switching to different method that's introduced in v1a2.

```release-note
The TCPSocket probe action is deprecated and will be removed in a later API version.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -76,6 +76,9 @@ issues:
   - linters:
     - gosec
     text: "G114: Use of net/http serve function that has no support for setting timeouts"
+  - linters:
+    - staticcheck
+    text: "^SA1019: .*TCPSocket is deprecated"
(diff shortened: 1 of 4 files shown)


akutz (approved):
Lgtm, but did you need/want to add the ignore for the static linter similar to https://github.com/vmware-tanzu/vm-operator/pull/300/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R72?

PR #298, opened 2023-12-11
WIP: Adds inline sysprep support for VM bootstrapping

srm09:
This patch adds support for inline sysprep which was introduced in the API in v1alpha2

n/a

For the secrets, we are currently decoding them and adding the data to the Data field under BootstrapArgs. There could be a situation wherein 2 or more secrets are storing data under the same key which will cause the secret values to be overidden.

```release-note
Adds inline sysprep support for VM bootstrapping
```


api/v1alpha2/sysprep/sysprep.go
```diff
@@ -71,7 +71,7 @@ type GUIUnattended struct {
 	// you may want to increase it. This number may be determined by the list of
 	// commands executed by the GuiRunOnce command.
 	//
-	// Please note this field only matters if AutoLogin is true.
+	// Please note this field only matters if AutoLogon is true.
```
bryanv, line 74: There is another `AutoLogin` on L61. Thanks!

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
```diff
@@ -79,3 +84,70 @@ func BootstrapSysPrep(
 
 	return configSpec, customSpec, nil
 }
+
+func ConvertTo(from *vmopv1_prep.Sysprep, bootstrapData BootstrapData) *vimTypes.CustomizationSysprep {
```
bryanv, line 88: Can this be made private?

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
```diff
@@ -10,18 +10,22 @@ import (
 	vimTypes "github.com/vmware/govmomi/vim25/types"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+	vmopv1_prep "github.com/vmware-tanzu/vm-operator/api/v1alpha2/sysprep"
```
bryanv, line 13: Nit: something like `vmopv1sysprep` is closer to the usually convention

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
```diff
@@ -79,3 +84,74 @@ func BootstrapSysPrep(
...
 	return configSpec, customSpec, nil
 }
+
+func convertTo(from *vmopv1_prep.Sysprep, bootstrapData BootstrapData) *vimTypes.CustomizationSysprep {
+	sysprepCustomization := &vimTypes.CustomizationSysprep{}
+
+	if from.GUIUnattended != nil {
+		sysprepCustomization.GuiUnattended = vimTypes.CustomizationGuiUnattended{
+			Password: &vimTypes.CustomizationPassword{
+				Value:     *bootstrapData.SysprepPassword,
```
bryanv, line 94: Will this be `nil` when `GUIUnattended.AutoLogon` false?

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
```diff
@@ -79,3 +84,74 @@ func BootstrapSysPrep(
...
+		CommandList: from.GUIRunOnce.Commands,
+	}
+
+	if from.Identification != nil {
+		sysprepCustomization.Identification = vimTypes.CustomizationIdentification{
+			JoinWorkgroup: from.Identification.JoinWorkgroup,
+			JoinDomain:    from.Identification.JoinDomain,
+			DomainAdmin:   from.Identification.DomainAdmin,
+			DomainAdminPassword: &vimTypes.CustomizationPassword{
+				Value:     *bootstrapData.SysprepDomainPassword,
```
bryanv, line 125: Will this be `nil` when `Identification.JoinDomain` is `""`?

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
srm09: Removed the pointer to avoid an accidental NPE, also moved the logic to add it only if the value is set in `bootstrapData`.

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap_sysprep.go
srm09: Removed the pointer to avoid an accidental NPE, also moved the logic to add it only if the value is set in `bootstrapData`.

bryanv (approved):
LGTM but the tests need to updated to not use prior common key selector structure.

PR #302, opened 2023-12-12
🐛 Fixes VMI/CVMI condition conversion

srm09:
When the v1a2 FSS is enabled, all VMI and CVMIs are missing the conditions when requesting a v1a1 object. This patch introduces a fix wherein when no false conditions are reported during the down conversion of a VMI (or CVMI), it sets all the child conditions as True.

Also, the latest transition time for True conditions was missing during the down conversion. To fix that, we set the transition time from the Ready condition to each of the child conditions since we do not have the individual times for each of those conditions anymore.

n/a

```release-note
Sets last transition time for VMI and CVMI True conditions
```

### DIFF
--- a/api/v1alpha1/virtualmachineimage_conversion.go
+++ b/api/v1alpha1/virtualmachineimage_conversion.go
@@ -93,6 +93,10 @@ func convert_v1alpha1_VirtualMachineImage_OVFEnv_To_v1alpha2_VirtualMachineImage
 
 func convert_v1alpha2_VirtualMachineImageStatusConditions_To_v1alpha1_VirtualMachineImageStatusConditions(
 	conditions []metav1.Condition) []Condition {
+	if len(conditions) == 0 {
+		return []Condition{}
+	}
+
 	var (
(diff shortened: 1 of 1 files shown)


api/v1alpha1/virtualmachineimage_conversion.go
```diff
@@ -93,6 +93,10 @@ func convert_v1alpha1_VirtualMachineImage_OVFEnv_To_v1alpha2_VirtualMachineImage
 
 func convert_v1alpha2_VirtualMachineImageStatusConditions_To_v1alpha1_VirtualMachineImageStatusConditions(
 	conditions []metav1.Condition) []Condition {
+	if conditions == nil || len(conditions) == 0 {
```
bryanv, line 96: This can be simplified to just `len(conditions) == 0` since that covers `nil` too. Unless something complains or really needs an empty slic, I'd return `nil` here instead.

bryanv (approved):

sreyasn (approved):

api/v1alpha1/virtualmachineimage_conversion.go
srm09: Done

PR #304, opened 2023-12-13
✨ Add VirtualMachine conversion tests

bryanv:
Start with a nearly fully populated v1a2 VM and ensure that the VM resulting from conversion is the same. We can keep adding these explicit conversion tests to ensure that fields are converted as we expect instead of always depending on the fuzzer.

Fix a few conversion issues around the Network and Advanced fields.

Always restore the full VM status from the annotations. Our controller should be the only one updating the Status, and an update from the conversion will cause our controllers to reconcile and reevaluate the Status.

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/conversion_test.go
+++ b/api/v1alpha1/conversion_test.go
@@ -85,24 +85,29 @@ func TestFuzzyConversion(t *testing.T) {
 }
 
 func overrideVirtualMachineFieldsFuncs(codecs runtimeserializer.CodecFactory) []interface{} {
-	// TODO: The changes from v1a1 to v1a2 is quite large so several parts of the input objects are
-	// 	     defaulted out until we start to marshall the object in the annotations for down conversions
-	// 	  	 and back.
 	return []interface{}{
 		func(vmSpec *v1alpha1.VirtualMachineSpec, c fuzz.Continue) {
 			c.Fuzz(vmSpec)
 
-			for i := range vmSpec.Volumes {
-				// Not present in v1a2.
-				vmSpec.Volumes[i].VsphereVolume = nil
+			var volumes []v1alpha1.VirtualMachineVolume
+			for _, vol := range vmSpec.Volumes {
+				// vSphere volumes are gone in v1a2 so skip those.
+				if vol.VsphereVolume == nil {
+					volumes = append(volumes, vol)
+				}
 			}
+			vmSpec.Volumes = volumes
 
-			if vmSpec.AdvancedOptions != nil {
-				if provOpts := vmSpec.AdvancedOptions.DefaultVolumeProvisioningOptions; provOpts != nil {
+			if opts := vmSpec.AdvancedOptions; opts != nil {
+				if provOpts := opts.DefaultVolumeProvisioningOptions; provOpts != nil {
 					if provOpts.ThinProvisioned != nil {
 						// Both cannot be set.
 						provOpts.EagerZeroed = nil
 					}
+				}
+
+				if opts.ChangeBlockTracking != nil && !*opts.ChangeBlockTracking {
+					opts.ChangeBlockTracking = nil
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


sreyasn (approved):

akutz:
> an update from the conversion will cause our controllers to reconcile and reevaluate the Status.

Why do we _want_ to eval the status? Should we not configure a filter that ignores updates to the status sub-resource?

api/v1alpha1/conversion_test.go
```diff
@@ -85,25 +85,30 @@ func TestFuzzyConversion(t *testing.T) {
...
-	// 	     defaulted out until we start to marshall the object in the annotations for down conversions
-	// 	  	 and back.
 	return []interface{}{
 		func(vmSpec *v1alpha1.VirtualMachineSpec, c fuzz.Continue) {
 			c.Fuzz(vmSpec)
 
-			for i := range vmSpec.Volumes {
-				// Not present in v1a2.
-				vmSpec.Volumes[i].VsphereVolume = nil
+			var volumes []v1alpha1.VirtualMachineVolume
```
akutz, line 92: Okay, I am glad I am not crazy. I use this pattern all of the time, but a Go linter [caught me the other day](https://github.com/vmware-tanzu/vm-operator/actions/runs/7199732125/job/19612086923#step:5:24) and told me I had to use `//nolint:prealloc`. I am still unsure why. The code in question was the line `var result []any` from the following snippet:

```golang
func (ccu *cloudConfigUsers) MarshalYAML() (any, error) {
	if len(ccu.users) == 0 {
		return nil, nil
	}
	var result []any //nolint:prealloc
	if ccu.defaultUser {
		result = append(result, "default")
	}
	for i := range ccu.users {
		result = append(result, ccu.users[i])
	}
	return result, nil
}
```

Do you happen to know why? I am not sure if it is because the slice's element was `any` or because the linter picked up on the append pattern...

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -441,7 +449,7 @@ func convert_v1alpha2_VirtualMachineAdvancedSpec_To_v1alpha1_VirtualMachineAdvan
 }
 
 func convert_v1alpha2_BootDiskCapacity_To_v1alpha1_VirtualMachineVolume(capacity *resource.Quantity) *VirtualMachineVolume {
-	if capacity == nil || capacity.IsZero() {
+	if capacity == nil {
```
akutz, line 452: Just curious why this change was made? Not opposed to it, just curious.

akutz (approved):
Lgtm, thanks!

bryanv:
> > an update from the conversion will cause our controllers to reconcile and reevaluate the Status.
> 
> Why do we _want_ to eval the status? Should we not configure a filter that ignores updates to the status sub-resource?

If the Spec changes, then yes. I don't think we're using the filter to skip the Status updated induced reconcile ATM.

api/v1alpha1/virtualmachine_conversion.go
bryanv: The check was causing fuzzer failures when the capacity was zero. I think with this change the conversion is closer to a truer conversion.

api/v1alpha1/conversion_test.go
bryanv: I'm not sure what pattern the linter is used to complain on these or not.

akutz:
> > > an update from the conversion will cause our controllers to reconcile and reevaluate the Status.
> > 
> > 
> > Why do we _want_ to eval the status? Should we not configure a filter that ignores updates to the status sub-resource?
> 
> If the Spec changes, then yes. I don't think we're using the filter to skip the Status updated induced reconcile ATM.

Sorry, I mean we never "evaluate" the status, or rather, we shouldn't. We construct it by evaluating and reconciling the spec. I think we just thought different things when we used the word "reevaluate."

PR #305, opened 2023-12-14
✨ Validate CloudConfig using schema

akutz:

This patch ensures the CloudConfig YAML produced by the inline CloudInit spec is validated against the official CloudConfig schema from Cloud-Init.

This patch also relocates the `cloudinit` utility package from `./pkg/vmprovider/providers/vsphere2/cloudinit` to `./pkg/util/cloudinit`. Since we do no anticipate future API revisions as a _huge_ departure from what has been established in v1alpha2, it is safe to use those types in utility code as they can be find/replaced with later schema versions, ex. `v1alpha3`, when that time comes. Other than the reference to a `github.com/vmware-tanzu/vm-operator/api/v1alpha2/cloudinit.CloudConfig`, the code in `./pkg/util/cloudinit` is completely unrelated to the vSphere provider, hence its relocation under the utility package.

Fixes `NONE`

I considered using the validation logic to also verify the raw cloud config, but I elected to not do this for the following reasons:

1. The validation is strict. If there are unknown fields, it will fail.
2. We should allow the users to use the raw cloud config to pass in data we may consider to be invalid, but may work for a given guest.
3. The schema is a moving target, and so we only update it with new VM Op releases.

In other words, the raw cloud config is not just a means to send in data that our version of the schema may consider to be invalid, but it's a way to ensure users have a way to bypass validation entirely if that is what they want to do.

```release-note
Validate the inline CloudConfig using the CloudConfig schema.
```


bryanv (approved):

akutz:
Merging as there were successful integration test runs already, and the last two have failed due to an unrelated image flake and the second a race condition.

PR #310, opened 2023-12-15
🐛 Restore previous LinuxPrep default customization behavior

bryanv:
We have E2E tests that assume this behavior so preserve that in v1a2 as well. The pointer API change in b54465 changed the behavior to exit bootstrap early. This is just a quick fix to get pipelines fixed. There is an existing story to both improve this default behavior - like by inspecting the image OS, and to keep addressing the prior lack of tests in this area.

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
+++ b/pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
@@ -57,7 +57,8 @@ func DoBootstrap(
 
 	bootstrap := vmCtx.VM.Spec.Bootstrap
 	if bootstrap == nil {
-		return nil
+		// For backwards compat we have to keep going and default to LinuxPrep.
+		bootstrap = &vmopv1.VirtualMachineBootstrapSpec{}
 	}
 
 	cloudInit := bootstrap.CloudInit


pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
```diff
@@ -57,7 +57,8 @@ func DoBootstrap(
 
 	bootstrap := vmCtx.VM.Spec.Bootstrap
 	if bootstrap == nil {
-		return nil
+		// For backwards compat we have to keep going and default to LinuxPrep.
+		bootstrap = &vmopv1.VirtualMachineBootstrapSpec{}
```
akutz, line 61: I assume it will be safe to augment this once we introduce knowledge of the image's reported OS and default to Sysprep if Windows is detected?

akutz (approved):

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
bryanv: Yes, as long as we want to continue with the implied GOSC (IDK if the v1a1 behavior was intentional but it was surprising to me when tracking down e2e failures some months back)

akutz: It was not intentional so much as a byproduct of being first and once other things were added, it just...defaulted.

PR #315, opened 2023-12-19
🐛 Add provider test for nil v1a2 VM Spec nil fields

bryanv:
Post b54465, several spec fields were changed to pointers so add test to cover if those are nil, and fix a few bugs.

```release-note
NONE
```


dilyar85 (approved):
LGTM!

PR #317, opened 2023-12-21
🌱 Rename inlined CloudInit default user field

bryanv:
Use DefaultUserEnabled as the field name for when the default user should still be created when other users are specified.

Update the VM with CloudInit example yaml

```release-note
NONE
```


dilyar85 (approved):
LGTM!

akutz (approved):

PR #324, opened 2023-12-26
🌱 Update v1a2 VM printer columns

bryanv:
Sync power state and age from v1a1

Use the Spec for the image and class names: for the time being, the corresponding status fields are updated until after the VM has been created on VC.

```release-note
NONE
```


akutz (approved):
Thank you for this!

PR #325, opened 2023-12-27
🐛 Do not always initialize v1a2 VM Spec.Advanced field

bryanv:
During VM creation, if the Spec.Advanced field was nil, we'd always set the field to its empty structure. Later when we would Patch the k8s VM at the end of reconcile, the field would get updated. This meant that every VM would have the Advanced field set even when not using any of the advance features.

Fix several NPE bugs the assignment was masking.

```release-note
NONE
```


srm09 (approved):

akutz (approved):
lgtm, thank you bv!

PR #329, opened 2024-01-02
✨ Do not use os.Getenv or pkg/lib for config

akutz:

This patch removes the usage of both os.Getenv and pkg/lib from throughout the codebase as a means to configure VM Operator. This method was fragile and resulted in a serial test design. The new package pkg/config is used to define the VM Operator configuration as well as store/retrieve the config from a Go context.

Fixes `NA`

* This is a _very_ large patch, because it has to be. The only way to remove the `pkg/lib` package was to actually remove it.
* @bryanv, you'll find there are a few places where I do `pkgconfig.SetContext(suite...`. I loathe this, and anywhere I do it is basically the same as setting an env var. It only happens in the webhook integration tests. I need to refactor those now that we have Ginkgo 2.0 so that we can start the webhook server with different configs. I plan to do that with the test framework PR that needs _this_ PR.

```release-note
Thread-safe testing with config from context.
```


webhooks/virtualmachinewebconsolerequest/v1alpha2/validation/webconsolerequest_validator_suite_test.go
```diff
@@ -8,17 +8,17 @@ import (
...
-	"github.com/vmware-tanzu/vm-operator/pkg/lib"
+	pkgconfig "github.com/vmware-tanzu/vm-operator/pkg/config"
 	"github.com/vmware-tanzu/vm-operator/test/builder"
 	"github.com/vmware-tanzu/vm-operator/webhooks/virtualmachinewebconsolerequest/v1alpha2/validation"
 )
 
 // suite is used for unit and integration testing this webhook.
-var suite = builder.NewTestSuiteForValidatingWebhookwithFSS(
+var suite = builder.NewTestSuiteForValidatingWebhookWithContext(
+	pkgconfig.WithConfig(pkgconfig.Config{Features: pkgconfig.FeatureStates{VMOpV1Alpha2: true}}),
```
dilyar85, line 18: I noticed that you used both this way and `pkgconfig.SetContext(pkgconfig.NewContextWithDefaultConfig(), func(in pkgconfig.Config)...)` to set up config across your changes. Is there any specific reason between these choices or is it just a matter of preference?

dilyar85 (approved):

controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_unit_test.go
```diff
@@ -146,12 +146,12 @@ func unitTestsReconcile() {
...
-			})
-
-			AfterEach(func() {
-				Expect(os.Unsetenv(lib.VMServiceV1Alpha2FSS)).To(Succeed())
+			JustBeforeEach(func() {
+				pkgconfig.SetContext(ctx, func(in pkgconfig.Config) (out pkgconfig.Config) {
+					out = in
+					out.Features.VMOpV1Alpha2 = true
+					return
+				})
```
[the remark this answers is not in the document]
akutz: The context is recreated each time, so it's not necessary.

webhooks/virtualmachinewebconsolerequest/v1alpha2/validation/webconsolerequest_validator_suite_test.go
akutz: Great question.

* `pkgconfig.NewContext` -- the config does not get defaults
* `pkgconfig.NewContextWithDefaultConfig` -- the config has defaults
* `pkgconfig.SetContext(...` -- I can provide a base config and then override a few things

controllers/virtualmachine/v1alpha1/virtualmachine_controller_unit_test.go
```diff
@@ -17,7 +17,9 @@ import (
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha1"
 
 	virtualmachine "github.com/vmware-tanzu/vm-operator/controllers/virtualmachine/v1alpha1"
+	pkgconfig "github.com/vmware-tanzu/vm-operator/pkg/config"
 	vmopContext "github.com/vmware-tanzu/vm-operator/pkg/context"
+
```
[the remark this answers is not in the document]
akutz: Yes, but given the linter didn't complain, I'll likely leave it to avoid having to issue more tests. If I have to retest, I'll include it. Thanks!

webhooks/virtualmachinewebconsolerequest/v1alpha2/validation/webconsolerequest_validator_suite_test.go
dilyar85: Thanks again for the detailed explanation!

pkg/config/context.go
```diff
@@ -0,0 +1,98 @@
...
+	if obj == nil {
+		panic("config is missing from context")
+	}
+	config, ok := obj.(*threadSafeConfig)
+	if !ok {
+		panic("config in context is invalid type")
+	}
+	config.Lock()
+	defer config.Unlock()
+	copyOfConfig := config.Config
```
bryanv, line 77: This copy shouldn't be needed, right?

bryanv (approved):

pkg/config/context.go
akutz: It prevents someone from accidentally hanging onto a reference to the `Config` object stored in the context.

pkg/config/context.go
```diff
@@ -0,0 +1,98 @@
...
+	return context.WithValue(
+		ctx,
+		configContextKeyValue,
+		&threadSafeConfig{Config: config})
+}
+
+// SetContext allows the Config in the provided context to be updated.
+// This function panics if the provided context does not contain a Config.
+// This function is thread-safe.
+func SetContext(ctx context.Context, setFn func(config *Config)) context.Context {
```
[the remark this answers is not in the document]
akutz: Sounds good. Updated with update fn.

PR #331, opened 2024-01-03
🌱 Add tests for VM Condition version conversion

bryanv:
There will likely be some upcoming changes on how the v1a1 PrereqReady condition is handled and filtered out but add some tests first.

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/virtualmachine_conversion.go
+++ b/api/v1alpha1/virtualmachine_conversion.go
@@ -668,7 +668,6 @@ func Convert_v1alpha1_VirtualMachineStatus_To_v1alpha2_VirtualMachineStatus(
 
 	out.PowerState = convert_v1alpha1_VirtualMachinePowerState_To_v1alpha2_VirtualMachinePowerState(in.PowerState)
 	out.Network = convert_v1alpha1_Network_To_v1alpha2_NetworkStatus(in.VmIp, in.NetworkInterfaces)
-	out.LastRestartTime = in.LastRestartTime
 
 	// WARNING: in.Phase requires manual conversion: does not exist in peer-type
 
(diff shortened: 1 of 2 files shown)


api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -668,7 +668,6 @@ func Convert_v1alpha1_VirtualMachineStatus_To_v1alpha2_VirtualMachineStatus(
 
 	out.PowerState = convert_v1alpha1_VirtualMachinePowerState_To_v1alpha2_VirtualMachinePowerState(in.PowerState)
 	out.Network = convert_v1alpha1_Network_To_v1alpha2_NetworkStatus(in.VmIp, in.NetworkInterfaces)
-	out.LastRestartTime = in.LastRestartTime
```
bryanv, line 671: The autogenerated conversion already handles this.

api/v1alpha1/virtualmachine_conversion_test.go
```diff
@@ -321,6 +322,196 @@ func TestVirtualMachineConversion(t *testing.T) {
...
+			Type:               nextver.VirtualMachineConditionImageReady,
+			Status:             metav1.ConditionTrue,
+			LastTransitionTime: metav1.NewTime(now.AddDate(2, 0, 0)),
+		}
+		vmSetRPCond := metav1.Condition{
+			Type:               nextver.VirtualMachineConditionVMSetResourcePolicyReady,
+			Status:             metav1.ConditionTrue,
+			LastTransitionTime: metav1.NewTime(now.AddDate(3, 0, 0)),
+		}
+		vmBSCond := metav1.Condition{
```
srm09, line 346: `vmBSCond` at a first glance seems to be a **BS** condition like `Foo`, but un fact is **BootstrapReady**. 🤣

srm09 (approved):
/lgtm

api/v1alpha1/virtualmachine_conversion_test.go
bryanv: It goes along nicely with the `bsArgs` that's used elsewhere. Just a reflection of how I was feeling at the time :)

PR #333, opened 2024-01-04
🐛 Fix VirtualMachineImage condition version conversion

bryanv:
When going from an originally v1a1 VMI to v1a2 VMI, the condition may not have a Reason set since it was optional in v1a1. However, in v1a2 that is now a required field so ensure that field is set. We use the Status field since we don't have anything better.

Add a bunch of missing condition conversion tests.

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/virtualmachineimage_conversion.go
+++ b/api/v1alpha1/virtualmachineimage_conversion.go
@@ -95,8 +95,9 @@ func convert_v1alpha1_VirtualMachineImage_OVFEnv_To_v1alpha2_VirtualMachineImage
 
 func convert_v1alpha2_VirtualMachineImageStatusConditions_To_v1alpha1_VirtualMachineImageStatusConditions(
 	conditions []metav1.Condition) []Condition {
+
 	if len(conditions) == 0 {
-		return []Condition{}
+		return nil
 	}
 
 	var (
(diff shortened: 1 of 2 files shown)


api/v1alpha1/virtualmachineimage_conversion.go
```diff
@@ -153,25 +151,26 @@ func convert_v1alpha2_VirtualMachineImageStatusConditions_To_v1alpha1_VirtualMac
...
 	}
-	if imageSyncedCondition != nil {
-		v1a1Conditions = append(v1a1Conditions, *imageSyncedCondition)
-	}
 	if imageProviderReadyCondition != nil {
 		v1a1Conditions = append(v1a1Conditions, *imageProviderReadyCondition)
 	}
+	if imageSyncedCondition != nil {
+		v1a1Conditions = append(v1a1Conditions, *imageSyncedCondition)
+	}
```
srm09, line 159: Question: Is this change to reflect the order in which these were set in the controller?

srm09 (approved):

api/v1alpha1/virtualmachineimage_conversion.go
bryanv: Yah. It's also the order that things are processed right above here.

PR #334, opened 2024-01-04
🐛 Fixes unset/empty write file content in cloudinit

srm09:
This patch fixes the validation logic to include the content in the CloudInit WriteFiles section to be empty. Since the API declares it as a required field, we set it to empty during the marshaling process to generate the YAML cloudconfig document.

n/a

n/a

```release-note
🐛 Fixes empty or unset write file content in cloudinit
```

### DIFF
--- a/pkg/util/cloudinit/cloudconfig.go
+++ b/pkg/util/cloudinit/cloudconfig.go
@@ -139,9 +139,11 @@ func MarshalYAML(
 			// a string from the Content field.
 			content := secret.WriteFiles[in.WriteFiles[i].Path]
 			if content == "" {
-				if err := json.Unmarshal(
-					in.WriteFiles[i].Content, &content); err != nil {
-
+				inContent := in.WriteFiles[i].Content
+				if len(inContent) == 0 {
+					inContent = []byte(`""`)
+				}
+				if err := json.Unmarshal(inContent, &content); err != nil {
 					return "", err
 				}
 			}
--- a/pkg/util/cloudinit/cloudconfig_test.go
+++ b/pkg/util/cloudinit/cloudconfig_test.go
@@ -152,6 +152,97 @@ var _ = Describe("CloudConfig MarshalYAML", func() {
 		})
 	})
 
+	When("CloudConfig has WriteFiles", func() {
+		BeforeEach(func() {
+			cloudConfig = vmopv1cloudinit.CloudConfig{
+				Users: []vmopv1cloudinit.User{
+					{
+						Name: "bob.wilson",
+						HashedPasswd: &common.SecretKeySelector{
+							Name: "my-bootstrap-data",
+							Key:  "cloud-init-user-bob.wilson-hashed_passwd",
+						},
+					},
+				},
+				WriteFiles: []vmopv1cloudinit.WriteFile{
+					{
+						Path:    "/hello",
+						Content: []byte(`"world"`),
+					},
+					{
+						Path:    "/hi",
+						Content: []byte(`"there"`),
+					},
+				},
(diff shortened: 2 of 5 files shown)


pkg/util/cloudinit/cloudconfig_test.go
```diff
@@ -152,6 +152,79 @@ var _ = Describe("CloudConfig MarshalYAML", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				Expect(data).To(Equal(cloudConfigWithEmptyFileContent))
+			})
+		})
+
+		When("Write File content is empty", func() {
+			BeforeEach(func() {
+				cloudConfig.WriteFiles = append(cloudConfig.WriteFiles, vmopv1cloudinit.WriteFile{
+					Path:        "/foo",
+					Content:     []byte(``),
```
akutz, line 213: Please have a test where the `Content` field is set to `nil`

akutz (approved):
lgtm, but for one test. Thank you!

pkg/util/cloudinit/validate/validate_test.go
```diff
@@ -97,6 +97,24 @@ var _ = Describe("Validate CloudConfigJSONRawMessage", func() {
...
+			cloudConfig.WriteFiles[0].Content = nil
+		})
+		It("Should not return any errors", func() {
+			Expect(errs).To(HaveLen(0))
+		})
+	})
+
+	When("The CloudConfig's first file content is empty", func() {
+		BeforeEach(func() {
+			cloudConfig.WriteFiles[0].Content = []byte(``)
```
akutz, line 112: Also please have a test where this field is `nil` and not `[]byte(``)`

sreyasn (approved):
LGTM. Thanks Sagar

pkg/util/cloudinit/validate/validate_test.go
srm09: https://github.com/vmware-tanzu/vm-operator/pull/334/files#diff-71149d5c5d525b6bff3a89a726ea2fd1bc114febde84de2b25bb4feb79012ba2R103

Are you thinking of something other than the one added?

akutz: Somehow I missed that, thanks!

pkg/util/cloudinit/cloudconfig.go
```diff
@@ -139,9 +139,11 @@ func MarshalYAML(
 			// a string from the Content field.
 			content := secret.WriteFiles[in.WriteFiles[i].Path]
 			if content == "" {
-				if err := json.Unmarshal(
-					in.WriteFiles[i].Content, &content); err != nil {
-
+				inContent := in.WriteFiles[i].Content
+				if len(inContent) == 0 {
```
bryanv, line 143: Not a big deal but I think `if inContent == nil` is a little clearer here (or otherwise won't we assign the same thing below?)

bryanv (approved):

pkg/util/cloudinit/cloudconfig.go
akutz: No.  `[]byte(`""`)` is not the same as `len(inContent) == 0`

akutz: In the above case, the `inContent` is literally the JSON as it would be written in a JSON doc. It's "raw JSON" as it were.

akutz: Therefore even `[]byte{}` would be invalid since it is not an "empty string" per the JSON rules.

PR #338, opened 2024-01-05
Use the v1a1 PrereqReady condition to populate v1a2 conditions

sreyasn:

If we have a v1a1 PrereqReady condition we can use that to fill in the similar v1a2 Conditions.
Also remove the PrereqReady condition from the v1a2 Conditions

Fixes # N/A

N/A

```
```

### DIFF
--- a/api/v1alpha1/virtualmachine_conversion.go
+++ b/api/v1alpha1/virtualmachine_conversion.go
@@ -676,7 +676,7 @@ func Convert_v1alpha1_VirtualMachineStatus_To_v1alpha2_VirtualMachineStatus(
 }
 
 func translate_v1alpha2_Conditions_To_v1alpha1_Conditions(conditions []Condition) []Condition {
-	var preReqCond, vmClassCond, vmImageCond, vmSetResourcePolicy, vmBootstrap *Condition
+	var preReqCond, vmClassCond, vmImageCond, vmSetResourcePolicyCond, vmBootstrapCond *Condition
 
 	for i := range conditions {
 		c := &conditions[i]
@@ -689,9 +689,9 @@ func translate_v1alpha2_Conditions_To_v1alpha1_Conditions(conditions []Condition
 		case v1alpha2.VirtualMachineConditionImageReady:
 			vmImageCond = c
 		case v1alpha2.VirtualMachineConditionVMSetResourcePolicyReady:
-			vmSetResourcePolicy = c
+			vmSetResourcePolicyCond = c
 		case v1alpha2.VirtualMachineConditionBootstrapReady:
-			vmBootstrap = c
+			vmBootstrapCond = c
 		}
 	}
 
@@ -700,8 +700,8 @@ func translate_v1alpha2_Conditions_To_v1alpha1_Conditions(conditions []Condition
 	// the v1a1 prereqs, and are optional.
 	if vmClassCond != nil && vmClassCond.Status == corev1.ConditionTrue &&
 		vmImageCond != nil && vmImageCond.Status == corev1.ConditionTrue &&
-		(vmSetResourcePolicy == nil || vmSetResourcePolicy.Status == corev1.ConditionTrue) &&
-		(vmBootstrap == nil || vmBootstrap.Status == corev1.ConditionTrue) {
+		(vmSetResourcePolicyCond == nil || vmSetResourcePolicyCond.Status == corev1.ConditionTrue) &&
+		(vmBootstrapCond == nil || vmBootstrapCond.Status == corev1.ConditionTrue) {
 
 		p := Condition{
 			Type:   VirtualMachinePrereqReadyCondition,
@@ -743,8 +743,8 @@ func translate_v1alpha2_Conditions_To_v1alpha1_Conditions(conditions []Condition
 		return append(conditions, p)
 	}
 
-	if vmSetResourcePolicy != nil && vmSetResourcePolicy.Status == corev1.ConditionFalse &&
-		vmBootstrap != nil && vmBootstrap.Status == corev1.ConditionFalse {
+	if vmSetResourcePolicyCond != nil && vmSetResourcePolicyCond.Status == corev1.ConditionFalse &&
+		vmBootstrapCond != nil && vmBootstrapCond.Status == corev1.ConditionFalse {
 
 		// These are not a part of the v1a1 Prereqs. If either is false, the v1a1 provider would not
 		// update the prereqs condition, but we don't set the condition to true either until all these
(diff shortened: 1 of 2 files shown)


api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -879,13 +879,149 @@ func restore_v1alpha2_VirtualMachineReadinessProbeSpec(
...
+	}
+
+	// If we didn't find the v1a1 PrereqReady condition then there is nothing to do.
+	if preReqCond == nil {
+		return dst.Status.Conditions
+	}
+
+	// If any of the v1a2 conditions were already present, remove v1a1 PrereqReady condition and move on.
+	if vmClassCond != nil || vmImageCond != nil || vmSetResourcePolicy != nil || vmBootstrap != nil {
+		return append(dst.Status.Conditions[:preReqCondIdx], dst.Status.Conditions[preReqCondIdx+1:]...)
```
bryanv, line 913: Can we move this to L910 so we don't have to also repeat it at the end of the function too?

api/v1alpha1/virtualmachine_conversion_test.go
```diff
@@ -541,6 +541,285 @@ func TestVirtualMachineConversion(t *testing.T) {
...
+								Status:             corev1.ConditionTrue,
+								LastTransitionTime: now,
+							},
+						},
+					},
+				}
+
+				hub := &nextver.VirtualMachine{}
+				g.Expect(spoke.ConvertTo(hub)).To(Succeed())
+				g.Expect(hub.Status.Conditions).To(HaveLen(4))
```
srm09, line 571: nit: Add an assertion that the PreReqReady condition does not get set anymore. That's one of the things we are addressing with this patch.

api/v1alpha1/virtualmachine_conversion_test.go
```diff
@@ -541,6 +541,285 @@ func TestVirtualMachineConversion(t *testing.T) {
...
+				c = hub.Status.Conditions[2]
+				g.Expect(c.Type).To(Equal(nextver.VirtualMachineConditionVMSetResourcePolicyReady))
+				g.Expect(c.Status).To(Equal(metav1.ConditionTrue))
+				g.Expect(c.Reason).To(Equal(string(metav1.ConditionTrue)))
+				g.Expect(c.LastTransitionTime).To(Equal(now))
+				c = hub.Status.Conditions[3]
+				g.Expect(c.Type).To(Equal(nextver.VirtualMachineConditionBootstrapReady))
+				g.Expect(c.Status).To(Equal(metav1.ConditionTrue))
+				g.Expect(c.Reason).To(Equal(string(metav1.ConditionTrue)))
+				g.Expect(c.LastTransitionTime).To(Equal(now))
```
srm09, line 592: nit suggestion: group each condition and assertions into a group and have a new line separating each group to make it readable.

srm09 (commented):
AFAIK this is correct, but wanted to confirm this.
With this patch, we are not changing the behavior of how these conditions are set, right? Since even if a single condition is marked as False, the controller still goes ahead with checking and setting the other conditions as well.

Otherwise looks good to me.

srm09 (approved):

api/v1alpha1/virtualmachine_conversion.go
sreyasn: done. double-check if it looks okay?.

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -892,13 +892,150 @@ func restore_v1alpha2_VirtualMachineReadinessProbeSpec(
 	}
 }
 
+func convert_v1alpha1_PreReqsReadyCondition_to_v1alpha2_Conditions(
+	dst *v1alpha2.VirtualMachine) []metav1.Condition {
+
+	var preReqCond, vmClassCond, vmImageCond, vmSetResourcePolicy, vmBootstrap *metav1.Condition
```
bryanv, line 898: Add `Cond` suffix for the later two vars (sorry, WIP had it this way)

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -892,13 +892,150 @@ func restore_v1alpha2_VirtualMachineReadinessProbeSpec(
...
+			})
+		}
+
+	} else if preReqCond.Status == metav1.ConditionFalse {
+		// Infer what v1a2 conditions need to be true/false from preReq reason.
+		// Order of evaluation of objects for v1a1 PreReq Condition:
+		// 1. VM Class then
+		// 2. VM Image
+
+		if preReqCond.Reason == VirtualMachineClassNotFoundReason {
```
bryanv, line 974: Not to likely to happend but I think could depending on what version upgrade from, we also include the `VirtualMachineClassBindingNotFoundReason` here? And use the PrereqReady reason as the new Reason too.

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -892,13 +892,150 @@ func restore_v1alpha2_VirtualMachineReadinessProbeSpec(
...
+				Reason:             string(metav1.ConditionFalse),
+			})
+			// v1a1 Image preReq hasn't been evaluated yet -- mark unknown
+			conditions = append(conditions, metav1.Condition{
+				Type:               v1alpha2.VirtualMachineConditionImageReady,
+				Status:             metav1.ConditionUnknown,
+				LastTransitionTime: preReqCond.LastTransitionTime,
+				Reason:             string(metav1.ConditionUnknown),
+			})
+		} else if preReqCond.Reason == VirtualMachineImageNotFoundReason {
```
bryanv, line 988: Similarly too can we include the Content* and VirtualMachineImageNotReadyReason reason?

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -892,13 +892,150 @@ func restore_v1alpha2_VirtualMachineReadinessProbeSpec(
...
+}
+
 // ConvertTo converts this VirtualMachine to the Hub version.
 func (src *VirtualMachine) ConvertTo(dstRaw conversion.Hub) error {
 	dst := dstRaw.(*v1alpha2.VirtualMachine)
 	if err := Convert_v1alpha1_VirtualMachine_To_v1alpha2_VirtualMachine(src, dst, nil); err != nil {
 		return err
 	}
 
+	newConditions := convert_v1alpha1_PreReqsReadyCondition_to_v1alpha2_Conditions(dst)
```
bryanv, line 1036: Can we move this to L1042 instead? If we're able to restore from the saved annotation, we know that the PreReq won't be there (and L1049 unconditionally restores that Status anyways)

api/v1alpha1/virtualmachine_conversion_test.go
```diff
@@ -704,6 +713,313 @@ func TestVirtualMachineConversion(t *testing.T) {
...
 		})
+
+		t.Run("Converts PrereqReady Condition", func(t *testing.T) {
+			t.Run("PrereqReady is True with no prior v1a2 conditions", func(t *testing.T) {
+				g := NewWithT(t)
+
+				now := metav1.Now()
+				spoke := v1alpha1.VirtualMachine{
+					Spec: v1alpha1.VirtualMachineSpec{
+						VmMetadata: &v1alpha1.VirtualMachineMetadata{
```
bryanv, line 724: Can we also have a test w/o the `VmMetadata` and `ReasourcePolicyName` set? I'd like to have something to ensure that we don't NPE.

api/v1alpha1/virtualmachine_conversion.go
sreyasn: Just to confirm -- Do we wanna set the PreReqReady reason as the new reason for the v1a2 Conditions too?. Notice I've used False/True which are used as v1a2 reasons for VMClass/Image not Ready otherwise.

api/v1alpha1/virtualmachine_conversion.go
sreyasn: Basically only do this `convert_v1alpha1_PreReqsReadyCondition` when there is nothing to restore from the conversion annotation?.

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -902,6 +1040,11 @@ func (src *VirtualMachine) ConvertTo(dstRaw conversion.Hub) error {
 	// Manually restore data.
 	restored := &v1alpha2.VirtualMachine{}
 	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
+		// If there is no v1a2 object to restore, this was originally a v1a1 VM
+		// Make sure the v1a1 PreReq conditions are translated over to v1a2 correctly
+		newConditions := convert_v1alpha1_PreReqsReadyCondition_to_v1alpha2_Conditions(dst)
```
bryanv, line 1045: Not a big deal but I'd prb just remove the temp var and assign it directly 'cause it might not actually be any new conditions.

bryanv (approved):

PR #343, opened 2024-01-09
🌱 Fix v1a2 cloud-init example

bryanv:
ee0250839 remove these top-level fields in favor of always be apart of the interfaces.

```release-note
NONE
```


akutz (approved):
Lgtm, thanks!

PR #347, opened 2024-01-10
Update NetworkInterface Name field pattern validation

sreyasn:

Since the networkInterface name is used in the CRD, we need to make sure it is DNSSubDomain compliant. It cannot include A-Z or underscore.

Fixes #
N/A

Earlier: 
```
root@422c3dc1f0a803888601c2f7bd08a66a [ ~ ]# kubectl apply -f vm.yaml
Error from server (spec.network.interfaces[0].name: Invalid value: "vm-v1a2-ETH0": is the resulting network interface name: a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')): error when creating "vm.yaml": admission webhook "default.validating.virtualmachine.v1alpha2.vmoperator.vmware.com" denied the request: spec.network.interfaces[0].name: Invalid value: "vm-v1a2-ETH0": is the resulting network interface name: a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

After the fix: 

```
 [ ~ ]# kubectl apply -f vm.yaml
The VirtualMachine "vm-v1a2" is invalid: spec.network.interfaces[0].name: Invalid value: "ETH0": spec.network.interfaces[0].name in body should match '^[a-z0-9]{2,}$'

[ ~ ]# kubectl apply -f vm.yaml
The VirtualMachine "vm-v1a2" is invalid: spec.network.interfaces[0].name: Invalid value: "e": spec.network.interfaces[0].name in body should match '^[a-z0-9]{2,}$'

yaml snippet:
  network:
    interfaces:
    - name: e0
    
root@4216498efaf1087de1e59cf3390b1430 [ ~ ]# kubectl apply -f vm.yaml
virtualmachine.vmoperator.vmware.com/vm-v1a2 created
    
```

```NONE
```

### DIFF
--- a/api/v1alpha2/virtualmachine_network_types.go
+++ b/api/v1alpha2/virtualmachine_network_types.go
@@ -32,7 +32,7 @@ type VirtualMachineNetworkInterfaceSpec struct {
 	// ensure the provided device name does not conflict with any other devices
 	// inside the guest, ex. dvd, cdrom, sda, etc.
 	//
-	// +kubebuilder:validation:Pattern=^\w\w+$
+	// +kubebuilder:validation:Pattern="^[a-z0-9]{2,}$"
 	Name string `json:"name"`
 
 	// Network is the name of the network resource to which this interface is
--- a/config/crd/bases/vmoperator.vmware.com_virtualmachines.yaml
+++ b/config/crd/bases/vmoperator.vmware.com_virtualmachines.yaml
@@ -1449,7 +1449,7 @@ spec:
                             is CloudInit. Please note it is up to the user to ensure
                             the provided device name does not conflict with any other
                             devices inside the guest, ex. dvd, cdrom, sda, etc."
-                          pattern: ^\w\w+$
+                          pattern: ^[a-z0-9]{2,}$
                           type: string
                         nameservers:
                           description: "Nameservers is a list of IP4 and/or IP6 addresses


srm09 (approved):
Did we have any validation on the existing v1a1 types? How would that be handled if there is an offending value set in the network name field?

sreyasn:
@srm09 - 
v1a1 did not have a network name that the user could mention. We used a combination of the vm-name + network name from the CRDs. They are always DNS Subdomain compliant

config/crd/bases/vmoperator.vmware.com_virtualmachines.yaml
```diff
@@ -1449,7 +1449,7 @@ spec:
                             is CloudInit. Please note it is up to the user to ensure
                             the provided device name does not conflict with any other
                             devices inside the guest, ex. dvd, cdrom, sda, etc."
-                          pattern: ^\w\w+$
+                          pattern: ^[a-z0-9][a-z0-9]+$
```
bryanv, line 1452: Isn't the second character class missing `-`?

sreyasn: I didn't think we needed to allow `-` for network names.  The earlier `\w` doesn't include `-` too, right?.

sreyasn: We've discussed this offline and decided to go without allowing `-`. We will make it more lenient if required later.

api/v1alpha2/virtualmachine_network_types.go
```diff
@@ -32,7 +32,7 @@ type VirtualMachineNetworkInterfaceSpec struct {
 	// ensure the provided device name does not conflict with any other devices
 	// inside the guest, ex. dvd, cdrom, sda, etc.
 	//
-	// +kubebuilder:validation:Pattern=^\w\w+$
+	// +kubebuilder:validation:Pattern=^[a-z0-9][a-z0-9]+$
```
akutz, line 35: I don't know why I used two `\w`. Can you please see if `^[a-z0-9]{1,}$` works as well?

sreyasn: Spoke to @akutz offline. We want atleast two characters so `^[a-z0-9][a-z0-9]+$` or `^[a-z0-9]{2,}$` should do. I will test this out.

PR #349, opened 2024-01-11
🌱 Use Network Hostname if set for Sysprep computer name

bryanv:
The intention with the hostName field is to let the user specify a name that's different than the k8s VM object name.

```release-note
NONE
```


sreyasn (approved):

akutz (approved):

srm09 (approved):
The initial thought process was to keep this as a separate argument to be able to later facilitate exposing it to the user, in the Sysprep VM customization  API directly. 

But we can re-introduce this change if we choose to expose the field in the API. Either way is fine.

PR #351, opened 2024-01-16
🌱  Update storage policy changes for Pod VM support on stretched supervisor

srm09:
To support Pod VM storage on stretched supervisor, a new CR `StoragePolicyQuota` has been introduced by vSphere CSI. This patch:
- Brings in the new [CRD types](https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/pkg/apis/cnsoperator/storagepolicy/v1alpha1/storagepolicyquota_types.go) in the external directory
- Updates the VMOp manifest generation logic to include the FSS for the support of Pod VM on stretched supervisor.
- Improves the VM storage policy validation logic to take into account the new CR in cases of stretched supervisor with multiple AZs
- During VM create, updates the VM storage policy ID fetch logic to leverage the new CR for stretched supervisor with multiple AZs

n/a

```release-note
🌱  Update storage policy changes for Pod VM support on stretched supervisor
```


pkg/vmprovider/providers/vsphere/storage/storageclass.go
```diff
@@ -20,17 +24,36 @@ func GetStoragePolicyID(
...
+	zones, err := topology.GetAvailabilityZones(vmCtx, client)
+	if err != nil {
 		return "", err
 	}
 
-	policyID, ok := sc.Parameters["storagePolicyID"]
-	if !ok {
-		return "", fmt.Errorf("StorageClass %s does not have 'storagePolicyID' parameter", storageClassName)
-	}
+	if pkgconfig.FromContext(vmCtx).Features.PodVMOnStretchedSupervisor && len(zones) > 1 {
```
bryanv, line 34: Here and elsewhere can we change things around to only grab the AZs when PodVMOnStretchedSupervisor is enabled?

test/builder/fake.go
```diff
@@ -11,10 +11,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	imgregv1a1 "github.com/vmware-tanzu/image-registry-operator-api/api/v1alpha1"
+
 	ncpv1alpha1 "github.com/vmware-tanzu/vm-operator/external/ncp/api/v1alpha1"
 	netopv1alpha1 "github.com/vmware-tanzu/vm-operator/external/net-operator/api/v1alpha1"
 	topologyv1 "github.com/vmware-tanzu/vm-operator/external/tanzu-topology/api/v1alpha1"
+	"github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/pkg/syncer/cnsoperator/apis"
```
bryanv, line 18: Can we import alias this to like `cnsapis` or similar?

test/builder/fake.go
```diff
@@ -11,10 +11,13 @@ import (
...
 	cnsv1alpha1 "github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/pkg/syncer/cnsoperator/apis/cnsnodevmattachment/v1alpha1"
+	storagev1alpha1 "github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/pkg/syncer/cnsoperator/apis/storagepolicy/v1alpha1"
```
bryanv, line 20: I think in most other places besides `test/...` this is import aliased as`cnsstoragev1`

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -349,28 +351,57 @@ func (v validator) validateStorageClass(ctx *context.WebhookRequestContext, vm *
...
+		}, storagePolicyQuota); err != nil {
+			return append(allErrs, field.Invalid(scPath, scName, fmt.Sprintf(storageClassNotFoundFmt, vm.Namespace)))
+		}
 
-	prefix := scName + storageResourceQuotaStrPattern
-	for _, resourceQuota := range resourceQuotas.Items {
-		for resourceName := range resourceQuota.Spec.Hard {
-			if strings.HasPrefix(resourceName.String(), prefix) {
-				return nil
+		// TODO(srm09): Is this validation necessary?
```
bryanv, line 368: I don't think so, at least not yet. This is just replacing how we enforce the storage NS association, and since the new thing is per-class specific and namespace-scoped that's implied.

bryanv (approved):
I think you're missing a `// +kubebuilder:rbac:...` directive for the new StoragePolicyQuota - config/rbac/role.yaml should be updated.

srm09:
> I think you're missing a // +kubebuilder:rbac:... directive for the new StoragePolicyQuota - config/rbac/role.yaml should be updated.

Ohh good catch, updated!

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -349,28 +352,48 @@ func (v validator) validateStorageClass(ctx *context.WebhookRequestContext, vm *
...
-	// TODO: This validation shouldn't be done in the webhook.
-
-	sc := &storagev1.StorageClass{}
-	if err := v.client.Get(ctx, client.ObjectKey{Name: scName}, sc); err != nil {
-		return append(allErrs, field.Invalid(scPath, scName, fmt.Sprintf(storageClassNotFoundFmt, vm.Namespace)))
+	var zones []topologyv1.AvailabilityZone
+	if pkgconfig.FromContext(ctx).Features.PodVMOnStretchedSupervisor {
+		azs, err := topology.GetAvailabilityZones(ctx, v.client)
+		if err != nil {
+			return append(allErrs, field.Invalid(scPath, scName, fmt.Sprintf(storageClassNotFoundFmt, vm.Namespace)))
```
divyenpatel, line 359: Can we return diff error. `storageClassNotFoundFmt` is not the right error when failed to get AZs.

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -349,28 +352,48 @@ func (v validator) validateStorageClass(ctx *context.WebhookRequestContext, vm *
...
-	if err := v.client.List(ctx, resourceQuotas, client.InNamespace(vm.Namespace)); err != nil {
-		return append(allErrs, field.Invalid(scPath, scName, err.Error()))
-	}
+	if pkgconfig.FromContext(ctx).Features.PodVMOnStretchedSupervisor && len(zones) > 1 {
+		storagePolicyQuota := &cnsstoragev1.StoragePolicyQuota{}
+		if err := v.client.Get(ctx, client.ObjectKey{
+			Namespace: vm.Namespace,
+			Name:      util.CNSStoragePolicyQuotaName(scName),
+		}, storagePolicyQuota); err != nil {
+			return append(allErrs, field.Invalid(scPath, scName, fmt.Sprintf(storageClassNotFoundFmt, vm.Namespace)))
```
divyenpatel, line 370: Can we say StoragePolicyQuota not found for the storageclass in the namespace?

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -349,28 +352,48 @@ func (v validator) validateStorageClass(ctx *context.WebhookRequestContext, vm *
...
 	scName := vm.Spec.StorageClass
 
-	// TODO: This validation shouldn't be done in the webhook.
-
-	sc := &storagev1.StorageClass{}
-	if err := v.client.Get(ctx, client.ObjectKey{Name: scName}, sc); err != nil {
-		return append(allErrs, field.Invalid(scPath, scName, fmt.Sprintf(storageClassNotFoundFmt, vm.Namespace)))
+	var zones []topologyv1.AvailabilityZone
+	if pkgconfig.FromContext(ctx).Features.PodVMOnStretchedSupervisor {
+		azs, err := topology.GetAvailabilityZones(ctx, v.client)
```
divyenpatel, line 357: Is it possible to cache and save AZs in the webhook once retrieved, otherwise every time this validation will invoke API to get AZs.

srm09: Its a simple API call to query the AZ list which is limited, and it will be called on every VM create which isn't a very frequent operation. Plus, in the entire VMOp codebase, we haven't cached the AZ list. Should not be an issue even at scale.

PR #353, opened 2024-01-17
✨ Disable DRS PowerOff for VMs with vGPU/Dynamic Direct Path IO devices

akutz:

It used to be the case that VI Admins could not manage the power state of VM Service managed VMs on vSphere via API or the HTML5 client (H5C). This presented a challenge to maintenance mode. If a VM could not be automatically evacuated from a host entering maintenance mode, the host could not successfully enter maintenance mode. In vSphere 7, VMs with vGPUs were not automatically vMotioned, and in vSphere 8+, VMs with PCI Passthru devices still are not vMotioned. There are also other CPU related settings that might pin a VM to a host, preventing DRS from vMotioning a VM to evacuate it off of a host entering maintenance mode.

Normally a VI Admin would power off a VM that blocks maintenance mode, but since VM Service managed VMs used to disallow VI Admins from doing so, a work-around was needed. VM Operator would apply a special flag in a VM's ExtraConfig to inform DRS to power off this VM if a host is entering maintenance mode. This flag was only applied for VMs with vGPUs or PCI Passthru devices because these were the only configurations allowed by VM Operator that would block evacuation-by-vMotion. And again, at the time this was implemented, VI Admins could not simply choose to power off these VM Service managed VMs as the VI Admins would do for any other VM. Since it is not possible to power on a VM on a host marked for maintenance mode, attempts to power the VM on by VM Operator will force DRS to attempt to find another compatible host for the VM.

Starting in vSphere 8, VMs with vGPUs *can* be automatically vMotioned, but VMs managed by VM Service are still powered off due to the special flag. Also starting in vSphere 8, VM Service managed VMs *can* be powered off by VI Admins in H5C. Beginning with vSphere 8.0U2, it is also possible to realize VM Service managed VMs using the *entire* virtual hardware and configuration surface of a vSphere VM, increasing the number of ways in which a VM Service managed VM might block a VM from being evacuated by vMotion.

Due to these factors, it is no longer desirable for DRS to automatically power off a VM Service managed VM. Instead, if a VM Service VM is blocking a host from entering maintenance mode, the VI Admin may treat it like any other VM and power off the VM using the web client or the vSphere API.

Fixes `NA`

`NA`

```release-note
DRS will no longer evacuate VMs with vGPUs and/or Dynamic Direct Path I/O devices from a host entering maintenance mode by powering off the VMs. Instead, VI Admins may configure a vSphere Cluster to indicate the level of stun time a VM with a vGPU is allowed to tolerate during a vMotion. Additionally, whether the VM cannot be evacuated due to a vGPU or any other reason, VI Admins may now power off the VM via the vSphere web client just like any other VM.
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/constants/constants.go
+++ b/pkg/vmprovider/providers/vsphere2/constants/constants.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2023 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2021-2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package constants
@@ -32,8 +32,12 @@ const (
 	// EnableDiskUUIDExtraConfigKey Enable UUID ExtraConfig key.
 	EnableDiskUUIDExtraConfigKey = "disk.enableUUID"
 
-	// MMPowerOffVMExtraConfigKey ExtraConfig key to enable DRS to powerOff VMs when underlying host enters into
-	// maintenance mode. This is to ensure the maintenance mode workflow is consistent for VMs with vGPU/DDPIO devices.
+	// MMPowerOffVMExtraConfigKey ExtraConfig key to enable DRS to powerOff VMs
+	// when the underlying host enters into maintenance mode. This is to ensure
+	// the maintenance mode workflow is consistent for VMs with vGPU/DDPIO
+	// devices.
+	// Deprecated: Admins can power off the VMs as they would VMs that are not
+	//             managed by VM Operator.
 	MMPowerOffVMExtraConfigKey = "maintenance.vm.evacuation.poweroff"
 
 	// NetPlanVersion points to the version used for Network config.
--- a/pkg/util/configspec.go
+++ b/pkg/util/configspec.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022-2023 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package util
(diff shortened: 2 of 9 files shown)


pkg/vmprovider/providers/vsphere2/constants/constants.go
```diff
@@ -32,8 +32,12 @@ const (
...
 	EnableDiskUUIDExtraConfigKey = "disk.enableUUID"
 
-	// MMPowerOffVMExtraConfigKey ExtraConfig key to enable DRS to powerOff VMs when underlying host enters into
-	// maintenance mode. This is to ensure the maintenance mode workflow is consistent for VMs with vGPU/DDPIO devices.
+	// MMPowerOffVMExtraConfigKey ExtraConfig key to enable DRS to powerOff VMs
+	// when the underlying host enters into maintenance mode. This is to ensure
+	// the maintenance mode workflow is consistent for VMs with vGPU/DDPIO
+	// devices.
+	// Deprecated: Admins can power off the VMs as they would VMs that are not
+	//             managed by VM Operator.
```
aruneshpa, line 40: Hmm.  They _can_.  But the virtual machine controller in VM operator will immediately try to power on the VM, right?  I think if we go by this route, we need to have admins leverage the `pause-reconcile` annotation.

akutz: Not necessary. Once a host has been marked for maintenance mode, a VM cannot be powered back onto that host. This is no different than DRS powering off the VM. We are just stating the admin will.

aruneshpa: Makes sense.  I was missing the fact that if a host is marked for MM, any attempt to power on the VM will place them on a different host automatically.  Thanks!

akutz: I added the following to the PR description:

> Since it is not possible to power on a VM on a host marked for maintenance mode, attempts to power the VM on by VM Operator will force DRS to attempt to find another compatible host for the VM.

bryanv (approved):

PR #357, opened 2024-01-19
🐛 Update bootstrap_cloudinit to use MergeExtraConfig

dilyar85:
This PR updates the `bootstrap_cloudinit.go` file to use the `MergeExtraConfig` function so that the existing EC also gets updated. This ensures that the VM is reconfigured with accurate data when updated (e.g. cloud-init `guestinfo.metadata` with a different instance ID).

```release-note
Update existing ExtraConfig keys if the values are changed.
```


aruneshpa (approved):

aruneshpa (approved):
LGTM.

PR #363, opened 2024-01-22
✨ Add DeviceName to network interfaces spec

bryanv:
When using cloud-init this allows the device name in the guest to be set to something different that what the name is in the k8s VM and used for the corresponding network interface CR name.

This decouples the two names if the user wants to. Note that since the netplan we create always matches by MAC address, continue to use the interface spec name as the netplan Ethernets key so it is easier to match up the VM Spec with the generated netPlan.

Fixes # NA

```release-note
DeviceName field in the interfaces spec can be used to specify the network device name when using cloud-init.
```

### DIFF
--- a/api/v1alpha1/virtualmachine_conversion_test.go
+++ b/api/v1alpha1/virtualmachine_conversion_test.go
@@ -82,7 +82,8 @@ func TestVirtualMachineConversion(t *testing.T) {
 					SearchDomains: []string{"foo.local", "bar.local"},
 					Interfaces: []nextver.VirtualMachineNetworkInterfaceSpec{
 						{
-							Name: "vds-interface",
+							Name:       "vds-interface",
+							DeviceName: "eth10",
 							Network: nextver_common.PartialObjectRef{
 								TypeMeta: metav1.TypeMeta{
 									Kind:       "Network",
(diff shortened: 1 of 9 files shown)


akutz (approved):
@bryanv ,

This looks good, but I do wonder -- should we stop using `Name` to rename devices at all if `DeviceName` is not specified and simply default to the scalar `ethN` is `DeviceName` is omitted? I suppose that complicates things if someone sets the `DeviceName` to `eth0` for the second NIC and omits the device name for the first NIC, which would get `eth0` by default. Thoughts on dropping `Name` from the source of the guest's device name?

akutz:
> v1a1 always renamed to eth# but that was buried in the implementation since the spec didn't have name/deviceName field.

I am aware, which is why I stated that as the example. Anyway, I am good with the PR. I just did not know if we wanted to fall back to the name field. However, given if we do not, there's an opportunity for overlap, this is fine.

Actually, come to think of it -- do you not need to validate the DeviceName is not set to the value of any other device's Name field?

bryanv:
> Actually, come to think of it -- do you not need to validate the DeviceName is not set to the value of any other device's Name field?
> 
> In fact, should we not have a validator that ensures `DeviceName` is unique for each element in the interfaces list? No two NICs should be able to have the same `DeviceName`, nor should they be able to have the same name as _another_ NIC's `Name` field, right?

Yes, I mentioned this when asking you to look at this. I wasn't sure if we wanted to (re)add the DeviceName back - it was in v1a2 but only in the top-level NetworkSpec and not in the InterfaceSpec so never really honored - and given the deadline if we just wanted to punt on this to v1a3.

PR #365, opened 2024-01-24
🌱 Include required storage class in StorageReady condition

bryanv:
If we require a storage class - which as of today will always be the case in WCP, include that as a part of the StorageReady condition check. There is little point in letting the VM creation flow get past this point since we will be able to create the VM.

There is some future that will make this optional for certain VMs so we're not going to enforce this in the webhook at the moment. There is also future UX improvement here: we don't allow the storage class to change - this will also be relaxed in the future - so the only way out of this is to delete the k8s VM resource and recreate it with the intended storage class. We could use the presence of this condition to allow it to be changed.

Remove API doc mentioning this field being defaulted when there is a sensible default, pushing this to v1a3+. The uncertainty about how and when the storage class will be optional for some VMs make me worried about trying to determine a default this moment, so keep the existing v1a1 behavior for v1a2.

Also apply #375 that was inadvertently done to the old v1a1 provider to the new v1a2 provider.

This is just an incremental minor improvement change - as noted above this requirement and flow will be changing with future work.

```release-note
NONE
```

### DIFF
--- a/api/v1alpha2/virtualmachine_types.go
+++ b/api/v1alpha2/virtualmachine_types.go
@@ -220,10 +220,6 @@ type VirtualMachineSpec struct {
 	// Please see https://kubernetes.io/docs/concepts/storage/storage-classes/
 	// for more information on Kubernetes storage classes.
 	//
-	// This field is optional in the cases where there exists a sensible
-	// default value, such as when there is a single StorageClass
-	// resource available in the same Namespace as the VM being deployed.
-	//
 	// +optional
 	StorageClass string `json:"storageClass,omitempty"`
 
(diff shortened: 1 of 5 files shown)


akutz:
Can we re-work this to make the storage class check happen in the controller and only if the VM does not already exist on vSphere?

bryanv:
> Can we re-work this to make the storage class check happen in the controller and only if the VM does not already exist on vSphere?

The controller itself doesn't really have that info avail and IMO not really the right place to do this. The code here is only called after we've determined the VM does not exist on VC and we're in the process of trying to create it.

zyiyi11 (commented):
We mentioned in v1a2 api that StorageClass can be omitted
```
	// This field is optional in the cases where there exists a sensible
	// default value, such as when there is a single StorageClass
	// resource available in the same Namespace as the VM being deployed.
```
I wonder how do we handle this case? ^^

bryanv:
> We mentioned in v1a2 api that StorageClass can be omitted
> 
> ```
> 	// This field is optional in the cases where there exists a sensible
> 	// default value, such as when there is a single StorageClass
> 	// resource available in the same Namespace as the VM being deployed.
> ```
> 
> I wonder how do we handle this case? ^^

We don't. There was some other defaulting docs that was removed in v1a2, and I think I just missed this one. I'd want to push this out v1a3+ at this point; I'll create another MR for that.

akutz (approved):

PR #368, opened 2024-01-24
🌱 Commonize setting of optional VM ready conditions

bryanv:
The VM SetResourcePool, Bootstrap, and Network are optional so only set their corresponding ready conditions when they're specified.  We were setting the Bootstrap ready condition when Bootstrap was nil. We we were not setting the other two when not specified or disabled.  Add some conditions.Delete() calls to make the not setting a condition explicit.

Add a bunch of expected condition assertions

Fixes #

I'm cool if we want to set these conditions to true when not specified instead. It's just that they should all behave the same.

```release-note
NONE
```


zyiyi11 (approved):
/LGTM

akutz (approved):
It amazes me the size of PRs when the actual change is two lines. Yay testing? :)

PR #371, opened 2024-01-26
🌱 Set HW Version in CreateConfigSpec()

bryanv:
The logic for determining the ConfigSpec version was spread out in two different functions and done separately after creating the ConfigSpec. The version can come from the class ConfigSpec or the class (passthru) devices, image, VM Spec MinHardwareVersion, or if the VM has PVCs at create time. Now have one function that is responsible for determining the version.

Prior logic was a little confusing - hopefully this will make the intention clearer

```release-note
NONE
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/virtualmachine/configspec.go
+++ b/pkg/vmprovider/providers/vsphere2/virtualmachine/configspec.go
@@ -4,12 +4,15 @@
 package virtualmachine
 
 import (
+	"fmt"
+
 	"github.com/vmware/govmomi/vim25/types"
 	"k8s.io/utils/pointer"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 	pkgconfig "github.com/vmware-tanzu/vm-operator/pkg/config"
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/util"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere2/constants"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmprovider/providers/vsphere2/instancestorage"
 )
@@ -28,7 +31,9 @@ func CreateConfigSpec(
 
 	configSpec := types.VirtualMachineConfigSpec{}
 
-	// If there is a class ConfigSpec, then that is our initial ConfigSpec.
+	// If there is a class ConfigSpec, then that is our initial ConfigSpec. Note that our caller
+	// will synthesize a class ConfigSpec from the class's Hardware.Devices if the class doesn't
+	// have a ConfigSpec.
 	if vmClassConfigSpec != nil {
 		configSpec = *vmClassConfigSpec
 	}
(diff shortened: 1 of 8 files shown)


pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
```diff
@@ -951,9 +936,7 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpecExtraConfig(
 		ecMap[k] = renderTemplateFn(k, v)
 	}
 
-	hasPassthroughDevices := len(util.SelectVirtualPCIPassthrough(util.DevicesFromConfigSpec(createArgs.ConfigSpec))) > 0
```
bryanv, line 954: This was slightly different than what we had on the config spec side - this would only check the type for passthrough, not the backing for vGPU/DDPIO. I don't think we intentionally made this different, and I think just checking the type is what was intended.

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec_test.go
```diff
@@ -142,6 +146,114 @@ var _ = Describe("CreateConfigSpec", func() {
...
 			})
 		})
+
+		When("VM Image hardware version is set", func() {
+			BeforeEach(func() {
+				vmImageStatus.HardwareVersion = pointer.Int32(101)
+			})
+
+			It("config spec has expected version", func() {
+				Expect(configSpec.Version).To(Equal("vmx-101"))
```
sreyasn, line 156: Nit: Add comment for Dummy VM having a PVC --  the image Version gets considered because of it.

sreyasn (approved):
LGTM!. Thank you BV.

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec.go
```diff
@@ -106,6 +114,48 @@ func CreateConfigSpec(
...
+		var minVerFromDevs int32
+		if util.HasVirtualPCIPassthroughDeviceChange(configSpec.DeviceChange) {
+			minVerFromDevs = max(imageVersion, constants.MinSupportedHWVersionForPCIPassthruDevices)
+		} else if hasPVC(vm) {
+			// This only catches volumes set at VM create time.
+			minVerFromDevs = max(imageVersion, constants.MinSupportedHWVersionForPVC)
+		}
+
+		// If both are zero, VC will use the cluster's default version.
+		return max(vmMinVersion, minVerFromDevs)
```
akutz, line 148: Should we not just be setting `configSpecVersion = minVerFromDevs` and letting it fall through outside the `if` block?

akutz (commented):
I had one question...

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec.go
bryanv: I've gone back & forth a bit on how to organize this function. I think this early return works out best b/c a `configSpec.Version == ""` is kind of the exceptional case nowadays, and to fall through here  we need another `configSpec.Version != ""` check to preserve the prior & desired behavior of when to update that field.

bryanv: Sure.

PR #372, opened 2024-01-26
🌱 Ensure Network field is set in network interfaces

bryanv:
While we still do the right thing if this is not set, by ensuring the network type is set v1a1 version conversion will show the correct networkType.

I'm just doing this on the create for now. Update will get sorted out later on multi-nic / the fly network changes work.

```release-note
NONE
```


dilyar85 (approved):
LGTM!

PR #373, opened 2024-01-26
🌱 Skip VM Service B/R workflow for TKG node VMs

dilyar85:
This PR skips persisting backup ExtraConfig keys in TKG node VMs when the `AutoVADPBackupRestore` feature is enabled.  
TKG node VMs will be backed up and restored differently (outside of VM-Operator reconciliation), hence do not require the backup ExtraConfig keys. These keys are only necessary for VM Service managed VMs.

N/A

The `HasTKGLabels` function has been moved to a new util file as it is used in multiple locations.

```release-note
Skip persisting backup ExtraConfig keys in TKG VMs when the AutoVADPBackupRestore feature is enabled.
```

**Testing Done**
- Deployed TKG VMs to an internal testbed without this patch (the VMs had backup ExtraConfig keys added):
```
$ govc vm.info -e "tkc-default-h4vct-rnppx" | grep "vmservice.virtualmachine" | awk '{print $1}'
vmservice.virtualmachine.resource.yaml:
vmservice.virtualmachine.additional.resources.yaml:
vmservice.virtualmachine.cloudinit.instanceid:
```
- Updated the internal testbed with this patch and redeployed TKG VMs (no longer had backup ExtraConfig keys):
```
$ govc vm.info -e "tkc-default-scqkx-zhwx9" | grep "vmservice.virtualmachine" | wc -l
0
```

### DIFF
--- a/pkg/util/kube/vm.go
+++ b/pkg/util/kube/vm.go
@@ -13,10 +13,11 @@ const (
 	CAPVClusterRoleLabelKey = "capv.vmware.com/cluster.role"
 )
 
-func HasTKGLabels(vmLabels map[string]string) bool {
-	_, ok := vmLabels[CAPWClusterRoleLabelKey]
-	if !ok {
-		_, ok = vmLabels[CAPVClusterRoleLabelKey]
-	}
-	return ok
+// HasCAPILabels returns true if the VM has a label indicating it was created by
+// Cluster API such as CAPW or CAPV.
+func HasCAPILabels(vmLabels map[string]string) bool {
+	_, hasCAPWLabel := vmLabels[CAPWClusterRoleLabelKey]
+	_, hasCAPVLabel := vmLabels[CAPVClusterRoleLabelKey]
+
+	return hasCAPWLabel || hasCAPVLabel
 }
--- a/pkg/util/kube/vm.go
+++ b/pkg/util/kube/vm.go
@@ -1,0 +1,22 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package kube
+
+const (
+	// CAPWClusterRoleLabelKey is the key for the label applied to a VM that was
+	// created by CAPW.
+	CAPWClusterRoleLabelKey = "capw.vmware.com/cluster.role" //nolint:gosec
+
+	// CAPVClusterRoleLabelKey is the key for the label applied to a VM that was
+	// created by CAPV.
+	CAPVClusterRoleLabelKey = "capv.vmware.com/cluster.role"
+)
+
+func HasTKGLabels(vmLabels map[string]string) bool {
+	_, ok := vmLabels[CAPWClusterRoleLabelKey]
+	if !ok {
+		_, ok = vmLabels[CAPVClusterRoleLabelKey]
+	}
+	return ok
+}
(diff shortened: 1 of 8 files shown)


pkg/vmprovider/providers/vsphere2/vmprovider_vm_test.go
```diff
@@ -56,7 +58,11 @@ func vmTests() {
 	)
 
 	BeforeEach(func() {
-		testConfig = builder.VCSimTestConfig{WithV1A2: true}
+		testConfig = builder.VCSimTestConfig{
+			WithV1A2:                  true,
+			WithContentLibrary:        true,
+			WithAutoVADPBackupRestore: true,
```
sreyasn, line 64: Nit: 
Can we create a separate subsection with a "WithAutoVADPBackupRestore: enabled" for a Basic VM/TKGS VM creation. This way we can test for `with and without FSS enabled` for the presence/absence of the extraConfigKey. We can then piggyback on existing tests after FSS is enabled on main.

sreyasn (approved):
Change looks good to me!. Thanks Sai

pkg/vmprovider/providers/vsphere2/vmprovider_vm_test.go
dilyar85: I think this should be included in the default basic VM tests because AutoVADPBackupRestore will be enabled with v1a2. Adding a separate block without the B/R FSS enabled in `vsphere2/vmprovider_vm_test.go` seems unnecessary and would create additional work when we remove the FSS in the future.

sreyasn: I guess it's okay cause we're close to enabling on main but I would generally prefer keeping tests for both scenarios around to have code coverage with/without FSS enabled.. v1a2 is enabled up top here because the vmprovider/vsphere2/ dir only handles v1a2 VMs. 

I get your point though. Resolving..

akutz:
Can you please adjust this PR's title / description? It was not until I read [this comment](https://github.com/vmware-tanzu/vm-operator/pull/373/files#diff-38f57caef14c92af0ed193fb12eef3df3bb811208d6a2f46f49652fb55807946R399) that I realized what this change did:

```golang
// Skip TKG VMs since they are backed up differently than VM Service VMs.
```

I suggest "Skip B/R workflow for TKG node VMs" and then explain the ExtraConfig key bit in the description.

akutz (approved):

pkg/util/kube/vm.go
```diff
@@ -0,0 +1,22 @@
...
+	// CAPWClusterRoleLabelKey is the key for the label applied to a VM that was
+	// created by CAPW.
+	CAPWClusterRoleLabelKey = "capw.vmware.com/cluster.role" //nolint:gosec
+
+	// CAPVClusterRoleLabelKey is the key for the label applied to a VM that was
+	// created by CAPV.
+	CAPVClusterRoleLabelKey = "capv.vmware.com/cluster.role"
+)
+
+func HasTKGLabels(vmLabels map[string]string) bool {
```
akutz, line 16: To be fair, this is really "HasCAPILabels", right?

dilyar85: Yeah I just moved it from the bootstrap.go file which had this function name. Thought we should keep it as is, in case TKG VMs have other identified labels added in the future.

pkg/util/kube/vm.go
```diff
@@ -0,0 +1,22 @@
...
+	_, ok := vmLabels[CAPWClusterRoleLabelKey]
```
aruneshpa, line 17: nit: I find this more readable:
```suggestion
_, hasCAPWLabel := vmLabels[CAPWClusterRoleLabelKey]
_, hasCAPILabel := vmLabels[CAPVClusterRoleLabelKey]

return hasCAPWLabel || hasCAPILabel
```
But I will leave it to you.

aruneshpa (approved):
LGTM with one minor suggestion.

pkg/util/kube/vm.go
dilyar85: Updated as suggested.

PR #374, opened 2024-01-30
 🐛 Disallow empty sysPrep bootstrap

bryanv:
If neither the raw or cooked sysprep data is provided the Windows GOSC Customize() call will fail, and since an empty text for raw sysprep also fails just disallow that configuration.

If wanting to do an empty Sysprep ends up being a common request - so the networking is configured - we can investigate more to see if there is some noop default we can fallback to.

```release-note
NONE
```


dilyar85 (approved):
LGTM!

sreyasn (approved):

PR #375, opened 2024-01-31
🌱 Simplify get storage policy ID from storage class name

bryanv:
This is a partial revert of 177578e. Regardless if the PodVMOnStretchedSupervisor FFS is enabled and if this setup is stretched we can always resolve the storage policy ID via the cluster scoped StorageClass. The storage namespace association has already be validated by this point.

While here don't export GetStoragePolicyID().

```release-note
NONE
```


akutz (approved):

divyenpatel (approved):
/approve

PR #382, opened 2024-02-03
🐛 Set Sysprep bootstrap secret data in createArgs & updateArgs

dilyar85:
This seems like a miss in https://github.com/vmware-tanzu/vm-operator/commit/a8c499d14118a0b9c687d850135fe1594121b7ce when we added the inline Sysprep bootstrap support.
If a Secret key is specified in the inline Sysprep bootstrap, it does get retrieved, but it's not currently being passed to the create/updateArgs. This causes the `CustomizationGuiUnattended` failed because of missing data. 

N/A.

N/A.

```release-note
Set Sysprep bootstrap secret data in createArgs & updateArgs to ensure successful GOSC when VM bootstrap spec contains inline Sysprep.
```


sreyasn (approved):

sreyasn:
Thank you @dilyar85

bryanv (approved):

PR #383, opened 2024-02-05
🌱 Preserve class ConfigSpec Annotation if set

bryanv:
If the VC VM already has an Annotation then leave it as-is instead of always reverting back to our default.

```release-note
If the VirtualMachineClass ConfigSpec has the Annotation field set, then a VM created from the class will have that Annotation.
```


dilyar85 (approved):
Thanks for fixing this!

sreyasn (approved):
Looks like we only need this change in update. LGTM!.

PR #384, opened 2024-02-06
🌱 Make manager the default container

aruneshpa:

This change makes `manager` the default container so that commands like `kubectl exec` and `kubectl logs` pick that container by default.

Also handle a few trivial things like:
- Changing the now unsupported `verbosity` with `-v` flag in kind load
- Removing the incorrect kustomize path since we now add HOSTOSARCH in the binary path

# Testing Done:
- Generated the manifests and verified that the annotation is present: 
```
$ grep -r 'kubectl.kubernetes.io/default-container: manager' artifacts/local-deployment.yaml artifacts/local-deployment.yaml:
kubectl.kubernetes.io/default-container: manager
```

- Verified that kubectl logs pick up the default container
```
$ k logs -n vmware-system-vmop vmware-system-vmop-controller-manager-6f8d6f8c78-wh5z8 | head Defaulted container "manager" out of: manager, kube-rbac-proxy
I0206 20:42:47.145487       1 main.go:53] entrypoint "msg"="Starting VM Operator controller" "buildnumber"="00000000" "buildtype"="dev" "commit"="42c578ddf9bd43d579475ddf23948f6a3fa7fc53" "version"="1.8.0-479-g42c578dd"
```

N/A

N/A

```release-note
Make "manager" the default container for kubectl log and exec
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -493,7 +493,7 @@ deploy-local-kind: docker-build load-kind ## Deploy controller in the kind clust
 
 .PHONY: load-kind
 load-kind: ## Load the image into the kind cluster
-	kind load docker-image $(IMG) --name $(KIND_CLUSTER_NAME) --loglevel debug
+	kind load docker-image $(IMG) --name $(KIND_CLUSTER_NAME) -v 10
 
 ## --------------------------------------
 ## Development - run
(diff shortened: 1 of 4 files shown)


zyiyi11 (approved):
/LGTM. Thanks for the change

PR #386, opened 2024-02-07
🌱 Update VM backup to avoid unnecessary reconfigures

dilyar85:
This PR updates the backup code to check `spec` (by `metadata.generation`), `labels`, and `annotations` field changes of VM CR to determine if the existing backup is up-to-date. Previously, it used `resourceVersion`, which could lead to unnecessary VM reconfigurations whenever the VM status was updated (e.g. VM readiness failed and its "lastTransitionTime" was updated in each no-op reconciliation).

N/A.

The instance-id backup is removed as part of this PR as the corresponding restore workflow doesn't require it anymore.

```release-note
Skip VM backup reconfiguration if there are no changes in the VM spec, labels, or annotations.
```

**Testing Done**:
- Added/updated backup test cases to cover no-op changes
- Deployed this patch to an internal testbed and verified that the VM backup reconfiguration was triggered only for spec, label, and annotation changes.


pkg/vmprovider/providers/vsphere2/virtualmachine/backup.go
```diff
@@ -141,24 +121,71 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
-			}
+	curBackup := ecMap[vmopv1.AdditionalResourcesYAMLExtraConfigKey]
+	backupVers, err := getBackupResourceVersions(curBackup)
+	if err != nil {
+		return "", err
+	}
+
+	isLatestBackup := true
+	for _, curRes := range resources {
+		if backupVers[string(curRes.GetUID())] != curRes.GetResourceVersion() {
```
aruneshpa, line 186: Should we also include labels and annotations here to re-create the bootstrap resources as close to the backup?  VM operator doesn't consume these, but it's possible that an automation in customers' env relies on the presence of these labels/annotations.

cc @akutz for this thoughts.

dilyar85: resourceVersion should be updated for any object changes, so we be able to capture label/annotation changes here by comparing the resourceVersion.

dilyar85: Here is an example diff of adding an annotation `foo:bar` to a Secret object:
```
$ diff --suppress-common-lines /tmp/secret-1.yaml /tmp/secret-2.yaml
6a7
>     foo: bar
24a26,34
>   - apiVersion: v1
>     fieldsType: FieldsV1
>     fieldsV1:
>       f:metadata:
>         f:annotations:
>           f:foo: {}
>     manager: kubectl-annotate
>     operation: Update
>     time: "2024-02-07T15:26:24Z"
27c37
<   resourceVersion: "860371"
---
>   resourceVersion: "1596751"
```

aruneshpa: One more question: do you see _any_ other reason the resource version of the object could be updated (other than changes to the data)?  I am trying to avoid the same scenarios as VM resources where status updates updated the RV.

aruneshpa (approved):
Tiny clarification question.  If the answer to that is no, mark that comment as resolved and merge it.

dilyar85:
> One more question: do you see any other reason the resource version of the object could be updated (other than changes to the data)? I am trying to avoid the same scenarios as VM resources where status updates updated the RV.

Today, we only store the bootstrap Secret/ConfigMap in the additional resource backup. The only possible changes could be to the data or metadata (specifically, annotations and labels, as the rest of the metadata is immutable). Therefore, the resourceVersion isn't likely to change frequently.

However, if in the future we store objects other than bootstrap in the additional resource required for backup/restore, we may need to consider checking for the generation in case the object has a spec/status field.

PR #388, opened 2024-02-08
🌱 Add validation for autoLogonCount and update description

sreyasn:
This change adds validations and updates logonCount description to say must be specified
when AutoLogon is true.

Fixes # N/A

https://learn.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/microsoft-windows-shell-setup-autologon-logoncount

```
Update AutoLogonCount description to say must be specified when AutoLogon is true

```

### DIFF
--- a/api/v1alpha2/sysprep/sysprep.go
+++ b/api/v1alpha2/sysprep/sysprep.go
@@ -67,7 +67,7 @@ type GUIUnattended struct {
 	// you may want to increase it. This number may be determined by the list of
 	// commands executed by the GuiRunOnce command.
 	//
-	// Please note this field only matters if AutoLogon is true.
+	// Please note this field must be specified with a non-zero positive integer if AutoLogon is true.
 	//
 	// +optional
 	AutoLogonCount int32 `json:"autoLogonCount,omitempty"`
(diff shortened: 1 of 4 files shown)


api/v1alpha2/sysprep/sysprep.go
```diff
@@ -67,7 +67,8 @@ type GUIUnattended struct {
 	// you may want to increase it. This number may be determined by the list of
 	// commands executed by the GuiRunOnce command.
 	//
-	// Please note this field only matters if AutoLogon is true.
+	// Please note this field only matters if AutoLogon is true. AutoLogonCount
+	// must be specified if AutoLogon is true.
```
bryanv, line 71: This seems a little redundant. Do you think this is still clear if we remove the existing `Please note this field only matters if AutoLogon is true`?

We can add a webhook validation for this?

sreyasn: yeah. how about ?
// Please note this must be specified if AutoLogon is true.

sreyasn: I'll add some validation for this. good call.

bryanv: > yeah. how about ? // Please note this field must be specified if AutoLogon is true.

I would make a mention that it must be positive since it isn't a pointer and specifying zero is no good.

bryanv (approved):

PR #389, opened 2024-02-09
🐛 Fix VirtualMachineWebConsoleRequest test race

bryanv:
Both the unit and integration tests shared the same fakeVMProvider variable so these could race. Usually I'd see ticket being "" because the other test called Reset() on the fake VM Provider.

Add assertions that the request response is the expected ticket instead of being not empty.

In v1a2 use the new CRD name to identify the v1a2 test suite

```release-note
NONE
```


sreyasn (approved):
Thank you, BV!

dilyar85 (approved):
Thanks for fixing this, BV!

PR #390, opened 2024-02-12
✨ Make the VM network interfaces immutable

bryanv:
In VDS we have difficulty in matching up the network interfaces in our VM spec with the VM's devices. Unlike NCP/NSX-T, VDS does not generate the MAC address - we defer to VC to generate it - and an overly restrictive limitation that on VDS, the EthCard ExternalID cannot be set on device create (but can later set it via an edit). So for now, try to constrain things a bit by making our network interfaces immutable so we can depend on the device order during create.

Limit network interfaces to 10 since that is the VC limit

This is just the first small step to make this better supported. In general, multiple interfaces just wouldn't have worked in either NSX-T or VDS so limit what we can make work for now until we have more time to solve this to support hot adding and removal of interfaces.

Much more work in this area is needed. One note is that this starts to relax the "only allow network changes when the VM is 'off'". Since we cannot really know the VM's actual power state at that time so we used the Spec. One could get around this by quickly updating the spec from Off -> On without it impacting the VMs actual power state. 

```release-note
The VM Spec Network.Interfaces is immutable in that the Name or Network cannot be changed, and interfaces cannot be added or removed.

There is a maximum of 10 network interfaces to reflect the VC limit.
```

### DIFF
--- a/api/v1alpha2/virtualmachine_network_types.go
+++ b/api/v1alpha2/virtualmachine_network_types.go
@@ -218,9 +218,13 @@ type VirtualMachineNetworkSpec struct {
 	// If the Interfaces field is empty and the Disabled field is false, then
 	// a default interface with the name eth0 will be created.
 	//
+	// The maximum number of network interface allowed is 10 because of the limit
+	// built into vSphere.
+	//
 	// +optional
 	// +listType=map
 	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
 	Interfaces []VirtualMachineNetworkInterfaceSpec `json:"interfaces,omitempty"`
 }
 
@@ -414,12 +418,21 @@ type VirtualMachineNetworkInterfaceIPStatus struct {
 // VirtualMachineNetworkInterfaceStatus describes the observed state of a
 // VM's network interface.
 type VirtualMachineNetworkInterfaceStatus struct {
-	// Name describes the unique name of this network interface, used to
-	// distinguish it from other network interfaces attached to this VM.
+	// Name describes the corresponding network interface with the same name
+	// in the VM's desired network interface list. If unset, then there is no
+	// corresponding entry for this interface.
 	//
-	// Please note this name is not related to the name of the device as it is
-	// surfaced inside of the guest.
-	Name string `json:"name"`
+	// Please note this name is not necessarily related to the name of the
+	// device as it is surfaced inside of the guest.
+	//
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// DeviceKey describes the unique hardware device key of this network
+	// interface.
+	//
+	// +optional
+	DeviceKey int32 `json:"deviceKey,omitempty"`
 
 	// IP describes the observed state of the interface's IP configuration.
 	//
@@ -471,8 +484,6 @@ type VirtualMachineNetworkStatus struct {
 	// Interfaces describes the status of the VM's network interfaces.
 	//
 	// +optional
-	// +listType=map
-	// +listMapKey=name
 	Interfaces []VirtualMachineNetworkInterfaceStatus `json:"interfaces,omitempty"`
 
 	// PrimaryIP4 describes the VM's primary IP4 address.
--- a/pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
+++ b/pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
@@ -642,29 +642,46 @@ func (s *Session) ensureCNSVolumes(vmCtx context.VirtualMachineContextA2) error 
 	return nil
 }
 
+func (s *Session) fixupMacAddresses(
+	vmCtx context.VirtualMachineContextA2,
+	resVM *res.VirtualMachine,
+	updateArgs *VMUpdateArgs) error {
+
+	missingMAC := false
+	for i := range updateArgs.NetworkResults.Results {
+		if updateArgs.NetworkResults.Results[i].MacAddress == "" {
+			missingMAC = true
+			break
+		}
+	}
+	if !missingMAC {
+		// Expected path in NSX-T since it always provides the MAC address.
+		return nil
+	}
+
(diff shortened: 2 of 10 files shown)


api/v1alpha2/virtualmachine_network_types.go
```diff
@@ -461,6 +461,7 @@ type VirtualMachineNetworkStatus struct {
 	// +optional
 	// +listType=map
 	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
```
zyiyi11, line 464: I thought this should be added to Spec instead of Status?

webhooks/virtualmachine/v1alpha2/validation/virtualmachine_validator.go
```diff
@@ -926,38 +924,78 @@ func (v validator) validateUpdatesWhenPoweredOn(ctx *context.WebhookRequestConte
...
+
+	oldNetwork := oldVM.Spec.Network
+	newNetwork := vm.Spec.Network
+
+	if oldNetwork == nil && newNetwork == nil {
+		return allErrs
+	}
+
+	// The mutation webhook will add interfaces.
+	if ctx.IsPrivilegedAccount {
```
akutz, line 974: The mutation webhook isn't running in the context of the privileged account. It would be the caller's account. It would only be a privileged account if it was the VM Op service account, cluster-admin, or in the priv account list. The mutation webhook for adding a default NIC only runs on Create, not update, so do we even need this check anyway?

pkg/vmprovider/providers/vsphere2/vmlifecycle/update_status.go
```diff
@@ -164,19 +160,36 @@ func getGuestNetworkStatus(
...
+				continue
+			}
+
+			var name string
+			if ethCardIdx < len(networkInterfaces) {
+				name = networkInterfaces[ethCardIdx].Name
+				ethCardIdx++
+			} else {
+				// Provide a default name that can otherwise hopefully help identify the interface.
+				name = fmt.Sprintf("nic-%d-%d", i, deviceKey)
```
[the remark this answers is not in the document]
bryanv: It is a little harder to use the same convention since we don't have an interface spec, so we don't have the interface name or the (optional) network name. We could default the interface name to ethX, but then we have to worry about collisions and whatnot. I'd prefer to see using the interface name for the name here in the status, and introduce the CR name as a object ref in v1a3+. 
I think I'll change this to make the device key num more clear and it have it be an invalid interface spec name, like `eth-deviceKey:12345` so we don't have to worry about a collision.

api/v1alpha2/virtualmachine_network_types.go
```diff
@@ -209,9 +209,13 @@ type VirtualMachineNetworkSpec struct {
 	// If the Interfaces field is empty and the Disabled field is false, then
 	// a default interface with the name eth0 will be created.
 	//
+	// The maximum number of network interface allowed is 10 because of the
```
akutz, line 212: Please change to "because of the limit built into vSphere." That way it covers both vC and ESX.

pkg/vmprovider/providers/vsphere2/vmlifecycle/update_status.go
```diff
@@ -164,19 +160,39 @@ func getGuestNetworkStatus(
...
+				continue
+			}
+
+			var name string
+			if networkInterfaceIdx < len(networkInterfaces) {
+				name = networkInterfaces[networkInterfaceIdx].Name
+				networkInterfaceIdx++
+			} else {
+				// Provide a default name that can otherwise hopefully help identify the interface.
+				name = fmt.Sprintf("eth-deviceKey:%d", deviceKey)
```
akutz, line 191: Why not just omit the name altogether? It should not be required in the status anyway.

akutz (commented):
A question about the name in the status...

pkg/vmprovider/providers/vsphere2/vmlifecycle/update_status.go
bryanv: The name is the map key so it is required: https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha2/virtualmachine_network_types.go#L466

If we want to change the API, then sure we can leave it blank but TBH I'm not a huge fan of that. If nothing else, using some fallback name based on the device ID here helps track changes to the IP config pushed down from VM Tools for the dev w/o having to use like the MAC.

akutz: If we want to include the device key, should we just introduce it as a field in the status?

pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
```diff
@@ -642,30 +642,47 @@ func (s *Session) ensureCNSVolumes(vmCtx context.VirtualMachineContextA2) error
...
+		return nil
+	}
+
+	networkDevices, err := resVM.GetNetworkDevices(vmCtx)
+	if err != nil {
+		return err
+	}
+
+	// Just zip these together until we can do interface identification.
+	for i := 0; i < min(len(networkDevices), len(updateArgs.NetworkResults.Results)); i++ {
```
aruneshpa, line 668: Do we need to call out the interface matching based on ordering in the API docs, or would that be too confusing (or TMI) for users?

aruneshpa (approved):
LGTM.

pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
bryanv: IMO too confusing right at the moment

pkg/vmprovider/providers/vsphere2/vmlifecycle/update_status.go
bryanv: Done - remove Name as the list key and added a device key field

akutz (approved):
lgtm, thanks! (barring the small changes we discussed)

PR #391, opened 2024-02-13
🌱 Make more v1a2 VM Status Network fields pointers

bryanv:
This helps declutter the output for the common VM configuration.

kubectl output can be cluttered with a lot of otherwise unset/empty structs.

This of course comes with the trade off that NPE can occur if we're not careful

```release-note
NONE
```


zyiyi11 (approved):

dilyar85 (approved):
LGTM

akutz (approved):

PR #392, opened 2024-02-13
:sparkles: Add default NetworkInterface when VPC network

zyiyi11:
This change adds default network interface `SubnetSet` when VPC Network in VM mutation webhook. 

Fixes N/A

We don't need to determine the network type/name because VPC will pick a namespace default(SubnetSet with the label `nsxoperator.vmware.com/default-subnetset-for: VirtualMachine`)

Testing done: when deployed a VM under VPC network without specifying network, default network interface is added.
```
  Network:
    Interfaces:
      Name:  eth0
      Network:
        API Version:  nsx.vmware.com/v1alpha1
        Kind:         SubnetSet
        Name:
```

```release-note
Add default NetworkInterface when VPC network
```

### DIFF
--- a/webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator.go
+++ b/webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2023 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2019-2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package mutation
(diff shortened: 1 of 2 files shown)


dilyar85 (approved):
LGTM

webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator.go
```diff
@@ -215,6 +217,9 @@ func AddDefaultNetworkInterface(ctx *context.WebhookRequestContext, client clien
 		if netName == "" {
 			netName = defaultNamedNetwork
 		}
+	case pkgconfig.NetworkProviderTypeVPC:
```
bryanv, line 220: Nit: can we move this above named to keep the "real" network types together.

bryanv (approved):

PR #396, opened 2024-02-20
🐛 Fix behavior of vm.spec.minHardwareVersion

akutz:

This patch fixes the behavior of the VM field `spec.minHardwareVersion` so that it ensures the VM's hardware stays above or at the specified, minimum hardware version. Previously this field only controlled the hardware version during deploy, even though the API was not documented this way. The intent was always for this field to ensure a VM's hardware version stays above or at the specified, minimum hardware version.

Fixes NA

* ~Do not merge until https://github.com/vmware/govmomi/pull/3365 is merged and this PR is updated to reference GoVmomi from its main repository.~
  * https://github.com/vmware/govmomi/pull/3365 has been merged.
* I confirmed internally that the VM Operator service account has the privilege required to upgrade a VM's hardware, `VirtualMachine.Config.UpgradeVirtualHardware`

```release-note
Fix VM spec.minHardwareVersion to match its documented behavior.
```


pkg/util/vsphere/vm/hardware_version_test.go
```diff
@@ -0,0 +1,225 @@
...
+				mgdObj.Self.Value = "does-not-exist"
+				initialHardwareVersion = "vmx-13"
+				expectedErr = errors.New("failed to retrieve properties ServerFaultCode: The object has already been deleted or has not been completely created")
+			})
+			Context("and the hardware version is cached in the managed object", func() {
+				BeforeEach(func() {
+					mgdObj.Summary.Config.HwVersion = "vmx-13"
+				})
+				Context("and fetchProperties is true", func() {
+					It("an upgrade op should return an error", func() {})
```
aruneshpa, line 116: Can we please move the validations in the respective `It` blocks?  While it might add a few additional lines of code, think that will help the readability massively.

For example, we could have 
```golang
			result, err := vmutil.ReconcileMinHardwareVersion(
				logr.NewContext(ctx, suite.GetLogger()),
				ctx.VCClient.Client,
				mgdObj,
				fetchProperties,
				minHardwareVersion)
```

in the `JustBeforeEach` block and we verify the `result` and the `err` here.

akutz: Actually, you couldn't have it as you wrote it as it would scope `result` and `err` to the `JustBeforeEach` block :)

As for the suggestion, I need to give it more thought. My initial take is that this is part and parcel of Ginkgo DSL, and that folks should embrace it. But like I said, I need to marinate on it. Either way I will not change the tests in this PR as they follow the same pattern from `power_state_test.go`, and I do not want to rewrite both right now. 

And besides, this is kind of the purpose of Ginkgo -- optimizing the tests.

akutz: I addressed this in the latest rebase.

sreyasn (approved):
Thanks for this change, @akutz

api/v1alpha2/virtualmachine_types.go
```diff
@@ -338,15 +338,19 @@ type VirtualMachineSpec struct {
 	// +optional
 	Reserved *VirtualMachineReservedSpec `json:"reserved,omitempty"`
 
-	// MinHardwareVersion specifies the desired minimum hardware version
-	// for this VM.
+	// +optional
```
bryanv, line 341: Did you intend to move these above the description comment? Typically we have them after it.

pkg/util/vsphere/vm/hardware_version.go
```diff
@@ -0,0 +1,81 @@
...
+	fetchProperties bool,
+	minHardwareVersion int32) (bool, error) {
+
+	var (
+		hwVersion int32
+		log       = logr.FromContextOrDiscard(ctx)
+		obj       = object.NewVirtualMachine(client, mo.Self)
+	)
+
+	log = log.WithValues(
```
bryanv, line 33: Can we early return when `minHardwareVersion == 0` b/c that will be the most common value and we know we have nothing to do?

Also, this seems like a lot of repeated logging in the steady state when the VM hardware version will change at most a few times in its lifetime.

api/v1alpha2/virtualmachine_types.go
akutz: It was intentional. I was going to slowly move the rest over time. That way generated GoDocs don't include the kubebuilder annotations. If you feel strongly, we can leave this alone and move them all at once later.

bryanv: OK. This is fine with me.

bryanv (approved):

PR #401, opened 2024-02-22
⚠️ Remove v1a1 code

bryanv:

This removes the bulk of the v1a1 code. Some v1a2 FFS checks are still there and will either be removed or kept as a guide for when we have to do this again. Similarly, there's also various FooA2() functions that I've kept as-is for now. Controllers and webhooks are also still under a v1a2 directory.

Fixes #

There is some more package rename and clean up - like the provider vsphere2 - to do later. 

This still needs to go through internal builds, etc so I'm not going to get merged super soon.

```release-note
NONE
```


akutz:
I'm concerned about the RBAC removals. For example, the RBAC for the old, v1a1 VM publish request. Won't our service account still require this? Or for that matter, the other v1a1 RBAC?

bryanv:
> I'm concerned about the RBAC removals. For example, the RBAC for the old, v1a1 VM publish request. Won't our service account still require this? Or for that matter, the other v1a1 RBAC?

The PublishRequest RBAC should be and looks to be unchanged in this MR. The VMOP types RBAC removed here are for v1a1 only types with corresponding code that has not been enabled for some time since their FFS predatess v1a2, so this MR should remove the remaining k8s client references to them. The Namespace RBAC shouldn't be needed anymore since that info is now obtained from the AZ CRs but we can add that back if wanted since it will be required for some upcoming work.

pkg/prober/prober_manager.go
```diff
@@ -60,24 +60,28 @@ type manager struct {
 	// if the time duration set in the AddAfter is not zero. vmReadinessProbeList can be used to avoid
 	// adding VMs to the readiness queue when this VM is already in the heap but not in the queue.
 	readinessMutex       sync.Mutex
-	vmReadinessProbeList map[string]*vmopv1.Probe
+	vmReadinessProbeList map[string]vmopv1.VirtualMachineReadinessProbeSpec
 }
 
 // NewManger initializes a prober manager.
-func NewManger(client client.Client, record vmoprecord.Recorder, vmProvider vmprovider.VirtualMachineProviderInterface) Manager {
+func NewManger(
```
sreyasn, line 67: not your change, typo method name: `Manager`

sreyasn (approved):
Change looks good to me.

PR #404, opened 2024-02-26
🌱 Assign whole BootstrapData struct in VM update reconcile path

bryanv:
Until we can clean up the larger issues around this code this will prevent another missed assignment like was fixed in aac0374c.

Fixes #

```release-note
NONE
```


akutz (approved):
Makes sense, thanks.

sreyasn (approved):

PR #408, opened 2024-02-28
✨ Use ginkgo in hack/test files

akutz:

This patch updates the hack/test files (unit and integration) to directly invoke ginkgo instead of go test. This change enables test summaries in GitHub actions, making it easier to determine what has failed.

Fixes `NA`

```release-note
NONE
```


Makefile
```diff
@@ -142,19 +142,29 @@ help: ## Display this help
...
 
 .PHONY: test-integration
-test-integration: $(ETCD) $(KUBE_APISERVER)
+test-integration: | $(GINKGO) $(ETCD) $(KUBE_APISERVER)
 test-integration: ## Run integration tests
-	KUBECONFIG=$(KUBECONFIG) hack/test-integration.sh $(INT_COV_FILE)
+	@rm -f $(INT_COV_FILE)
+	ENABLE_UNIT_TESTS=false \
+	ENABLE_INTEGRATION_TESTS=true \
+	KUBECONFIG=$(KUBECONFIG) \
```
bryanv, line 165: I think the `KUBECONFIG` here is a left over from long ago when these tests used to use kind. It shouldn't be needed, and the tests pass locally when I removed it if you want to remove it in this MR.

bryanv (approved):

PR #409, opened 2024-02-28
✨ Add GuestDeviceName to network interfaces spec

bryanv:

When using cloud-init this allows the device name in the guest to be set to something different that what the name is in the k8s VM and used for the corresponding network interface CR name.

This can be useful if desired device name in the guest does not conform to what k8s allows since Name field is used for part of the interface CR name.

Fixes #

```release-note
When using CloudInit for the bootstrap, the network device name in the guest can be specified with the GuestDeviceName field in the network interface spec. If omitted, the device will be renamed to the Name field of the network interface spec.
```

### DIFF
--- a/api/v1alpha2/virtualmachine_network_types.go
+++ b/api/v1alpha2/virtualmachine_network_types.go
@@ -27,10 +27,10 @@ type VirtualMachineNetworkInterfaceSpec struct {
 	// Name describes the unique name of this network interface, used to
 	// distinguish it from other network interfaces attached to this VM.
 	//
-	// This value is also used to rename the device inside the guest when the
-	// bootstrap provider is CloudInit. Please note it is up to the user to
-	// ensure the provided device name does not conflict with any other devices
-	// inside the guest, ex. dvd, cdrom, sda, etc.
+	// When the bootstrap provider is Cloud-Init and GuestDeviceName is not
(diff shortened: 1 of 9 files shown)


api/v1alpha2/virtualmachine_network_types.go
```diff
@@ -44,6 +44,15 @@ type VirtualMachineNetworkInterfaceSpec struct {
 	// +optional
 	Network common.PartialObjectRef `json:"network,omitempty"`
 
+	// GuestDeviceName is used to rename the device inside the guest when the
+	// bootstrap provider is Cloud-Init. Please note it is up to the user to
+	// ensure the provided device name does not conflict with any other devices
+	// inside the guest, ex. dvd, cdrom, sda, etc.
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=^\w\w+$
```
zyiyi11, line 53: Do we wanna match the pattern in Name? `Pattern="^[a-z0-9]{2,}$"` will exclude underscores.

zyiyi11 (approved):

api/v1alpha2/virtualmachine_network_types.go
bryanv: The validation for this field is more lenient since we don't use it for the k8s CR name for the network inteface.

PR #410, opened 2024-02-29
🌱 Only default to LinuxPrep GOSC on Linux VMs

bryanv:

In the old v1a1 code, we'd always fallback to LinuxPrep regardless if the guest was Linux or not, and we have tests that assume that behavior. Improve that here by only doing that on guests that would indicate to be some known Linux distro.

Fixes #

IDK how intentional the prior behavior was but since we have tests that assume this behavior and it is a LCD way of passing in the network config keep doing so. 

```release-note
If Bootstrap is omitted and the guest OS identifier is Linux, then the bootstrap will be defaulted to LinuxPrep.
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
+++ b/pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
@@ -5,6 +5,7 @@ package vmlifecycle
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/task"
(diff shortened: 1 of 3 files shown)


pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
```diff
@@ -274,3 +278,36 @@ func isCustomizationPendingError(err error) bool {
 	}
 	return false
 }
+
+// linuxGuestIDPrefixes is derived from the vimTypes.VirtualMachineGuestOsIdentifier values.
+var linuxGuestIDPrefixes = [...]string{
```
aruneshpa, line 283: Does it make sense to add this list and the `isLinuxGuest` methods to govmomi  and reference from there?

Govmomi already seems to have enums for all the guest OSes:

https://github.com/vmware/govmomi/blob/main/vim25/types/enum.go#L8316

aruneshpa (approved):
One question.  Otherwise looks good.

pkg/vmprovider/providers/vsphere2/vmlifecycle/bootstrap.go
bryanv: It would nice but we'd want to come along from the auto generated but it looks like that info isn't provided along with those enums.

You can tease the OSFamily from the GuestID out of the EnvironmentBrowser but the ergonomics to do at this point in the code is just not that friendly and just don't think it is worth it to do so for right now.

PR #414, opened 2024-03-05
🌱 Remove cache work-around for ConfigMaps/Secrets and centralize infra controllers

akutz:

This patch removes the special cache work-arounds put in place to avoid caching ConfigMap and Secret resources due to the amount of memory required to do so. It used to be these resources were not cached at all except for special namespaces. Now it is possible to use logic built into Controller-Runtime to achieve the same result.

This patch also refactors all of the infra controllers under the `controllers/infra` package.
    
Finally, the patch separates the infra ConfigMap and Secret controllers which once shared a single controller, even though this was anti-pattern.

Fixes `NA`

I added some negative tests to `controllers/infracluster` in order to assert the resources are only watched for the given namespaces.

```release-note
NONE
```


controllers/infra/secret/infra_secret_controller.go
aruneshpa, line 1: Suggestion: drop the "infra_" prefix from the file names since that's already present on the path?

aruneshpa (approved):

controllers/infra/secret/infra_secret_controller.go
akutz: I thought of that, but if the path isn't noticed, the short prefix indicates this isn't simply a secret controller. Does that make sense?

akutz: I do not feel strongly, and am happy to drop the prefix.

akutz: I will leave this alone for now, merge it, and we can come back to this later if it is an issue.

PR #422, opened 2024-03-07
🐛 Fix inability to get ConfigMap/Secret objs

akutz:

This patch addresses a bug created by https://github.com/vmware-tanzu/vm-operator/pull/414 and the move to the new, built-in cache logic in Controller-Runtime v0.17.x. In adopting this logic a mistake was made regarding how the logic corresponded to bypassing the cache for certain object types. This patch fixes the issue by once again disabling the global cache for ConfigMap and Secret resources, but this time using specific caches for the infra controllers.

A new test package, pkg/manager/test, has also been created to verify that the global cache is in fact bypassed for the disabled objects.

Fixes `NA`

```release-note
Fix error when getting ConfigMap and Secret resources
```

### DIFF
--- a/controllers/infra/configmap/infra_configmap.go
+++ b/controllers/infra/configmap/infra_configmap.go
@@ -12,7 +12,8 @@ import (
 )
 
 const (
-	// WcpClusterConfigMapNamespace is the namespace of the wcp-cluster-config ConfigMap.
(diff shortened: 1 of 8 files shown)


akutz:
FYI, I verified this on an internal testbed. The errors that used to occur are corrected.

bryanv (approved):

PR #423, opened 2024-03-07
🌱 Don't watch PVCs in non-InstanceStorage envs

bryanv:

For envs when InstanceStorage is not enabled (VDS) there is no point in watching PVCs since that watch is only needed for when we're waiting for the IS PVCs to get bound.

For envs when InstanceStorage in enabled (NSX-T) InstanceStorage volumes generally are not used used but we're still watching PVCs. This change may be superseded by a later change that only starts the Watch() when we first encounter a VM that has InstanceStorage volumes.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/volume/v1alpha2/volume_controller.go
+++ b/controllers/volume/v1alpha2/volume_controller.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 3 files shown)


aruneshpa (approved):

akutz:
I like it, but in the long run should we not add a dynamic watch on the type once we see a VM that uses a class with ISV?

akutz (approved):

bryanv:
> I like it, but in the long run should we not add a dynamic watch on the type once we see a VM that uses a class with ISV?

Yah the MR description mentions doing that next. This is just the immediate easiest thing to do atm.

bryanv:
Superseded by https://github.com/vmware-tanzu/vm-operator/pull/432

PR #424, opened 2024-03-07
✨ Add client Index on the CnsNodeVmAttachment Spec.NodeUUID

bryanv:

When find the CnsNodeVmAttachments for the given VM, we need to match up the attachments' Spec.NodeUUID with the VM's BiosUUID. Add an index on the NodeUUID field so that if the namespace has a lot of attachments the List() only returns the attachments for our VM.

Fixes #

This is just a WIP MR.

```release-note
NONE
```


zyiyi11 (approved):
You probably want to update release note to reflect this?

PR #428, opened 2024-03-12
🐛 Reduce VM props fetching in steady state

bryanv:

During VM reconcile, we go to VC to fetch the VM properties twice. Once to see if anything in the VM's HW config needed to be updated, and again to populate the VM Status.

In the common case - say the VM is already on or off - no changes to the VM config are done, so the properties fetched first are still entirely valid. Reuse those properties in that case.

CPU profiling showed that we'd spend a lot of time demarshalling the response, plus the latency of going to VC, this would slow down our VM reconcile since we have a set pool of workers.

There is still a lot of room for improvement here to be smarter about what properties are fetched based on the VM desired state.  I've got more work planned in this area for later feature work that will improve upon this.

Fixes #

Still lots of work needed in this area, especially to sanely support upcoming features. 

```release-note
NONE
```


akutz:
Thanks BV! I want to take a closer look at this.

pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
```diff
@@ -777,32 +782,206 @@ func (s *Session) attachClusterModule(
...
+		return
+	}
+	if opResult == vmutil.ReconcileMinHardwareVersionResultUpgraded {
+		refetchProps = true
+	}
+
+	return refetchProps, err
+}
+
+func (s *Session) suspendedVirtualMachine(
```
akutz, line 831: This is a very confusing function name. Can you please rename it to be something more descriptive? Or add a comment?

pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
```diff
@@ -777,32 +782,206 @@ func (s *Session) attachClusterModule(
 	return s.Client.ClusterModuleClient().AddMoRefToModule(vmCtx, moduleUUID, resVM.MoRef())
 }
 
+func (s *Session) poweredOffVirtualMachine(
```
akutz, line 785: This is a very confusing function name. Can you please rename it to be something more descriptive? Or add a comment?

akutz (approved):
lgtm, thanks BV!

PR #429, opened 2024-03-13
🐛 Fix "make generate" due to testlabels mod

akutz:

This patch fixes the "make generate" target, which was failing due to the introduction of the testlabels module. Apparently this was not caught by the PR testing, so we need to look into that. The controller-gen binary was not happy with the ./pkg/... path that included another module, i.e. ./pkg/constants/testlabels. Instead the command has been updated to specify a single path, the recursive pattern for the main module, which results in scanning more paths than before, but importantly excludes the testlabels module.

Fixes `NA`

```release-note
Fix bug in "make generate" target complaining about testlabels module
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,9 +292,7 @@ generate-manifests: ## Generate manifests e.g. CRD, RBAC etc.
 		output:webhook:dir=$(WEBHOOK_ROOT) \
 		webhook
 	$(CONTROLLER_GEN) \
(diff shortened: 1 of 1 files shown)


dilyar85 (approved):
LGTM

akutz:
Closing this PR. @bryanv did not see the same issue, and it turns out the problem was I had a `go.work` file that did not include the new testlabels package.

PR #430, opened 2024-03-13
:seedling: NSX LB annotation only for NSX-T Networking

zyiyi11:
As of today, VPC Networking only works with AVI LB. And AVI does not monitor any label or annotation for any behavior change in both L4 svc and Ingress. So we only return NSX LB under NSX-T Networking instead of !VDS.

Fixes N/A

```release-note
N/A
```


pkg/vmprovider/providers/vsphere2/network/network.go
```diff
@@ -557,9 +557,8 @@ func createVPCNetworkInterface(
 	}
 
 	switch interfaceSpec.Network.Kind {
-	case "SubnetSet":
-		vpcSubnetPort.Spec.SubnetSet = networkName
-	case "":
```
bryanv, line 562: Can you drop this part of the commit? Because of some other recent changes I think we can handle this better with a small change in the mutation webhook.

bryanv (changes requested):
Just one comment

PR #432, opened 2024-03-14
🌱 Defer PVC watch until first VM with instance storage

bryanv:

We only need to watch PVCs for instance storage since we need to wait for them to be bound to a specific host. Otherwise the watch is not needed at all. To reduce CPU and memory usage for a not always enabled or commonly used feature, defer the watch until the first instance storage VM is reconciled.

Also add a labelSelector with the existing instance storage label PVC watch so this watch only matches PVCs we care about.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/volume/v1alpha2/volume_controller.go
+++ b/controllers/volume/v1alpha2/volume_controller.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


controllers/volume/v1alpha2/volume_controller.go
```diff
@@ -274,6 +307,12 @@ func (r *Reconciler) reconcileInstanceStoragePVCs(ctx *context.VolumeContextA2)
 		return true, nil
 	}
 
+	err := r.startPVCWatch()
+	if err != nil {
```
dilyar85, line 311: Nit:
```suggestion
	if err := r.startPVCWatch(); err != nil {
```

dilyar85 (approved):
LGTM!

PR #435, opened 2024-03-18
🌱 Defer PVC watch in Volume controller until needed

bryanv:

We only need to be aware of PVCs in the Volume controller because of instance storage. This feature isn't enabled in all envs, and is not that commonly used in general, so defer the costs of the Watch() until we need to do so.

Improve getInstanceStoragePVCs() to not return PVCs that did not exist. Things still worked because these PVCs would fail the IsControlledBy() check but it is clearer to just not return these PVCs in the first place.

One goal is to offset some of the memory used by the CnsNodeVmAttachment field index added in 7a6e39d.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/volume/v1alpha2/volume_controller.go
+++ b/controllers/volume/v1alpha2/volume_controller.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -18,6 +19,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -31,6 +33,7 @@ import (
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 	pkgconfig "github.com/vmware-tanzu/vm-operator/pkg/config"
 	"github.com/vmware-tanzu/vm-operator/pkg/context"
+	pkgmgr "github.com/vmware-tanzu/vm-operator/pkg/manager"
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
 	"github.com/vmware-tanzu/vm-operator/pkg/util"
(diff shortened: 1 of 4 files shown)


controllers/volume/v1alpha2/volume_controller.go
```diff
@@ -95,15 +97,57 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
...
+		isPVCLabels := metav1.LabelSelector{
+			MatchLabels: map[string]string{constants.InstanceStorageLabelKey: "true"},
+		}
+		labelSelector, err := metav1.LabelSelectorAsSelector(&isPVCLabels)
+		if err != nil {
+			return err
+		}
+
+		var pvcWatchStarted bool
+		pvcWatchStartedLock := sync.Mutex{}
```
akutz, line 116: nit: These two lines could be rewritten as:

```golang
		var (
			pvcWatchStarted     bool
			pvcWatchStartedLock sync.Mutex
		)
```

I normally see mutexes declared that way.

controllers/volume/v1alpha2/volume_controller.go
```diff
@@ -95,15 +97,57 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
...
+
+		r.StartPVCWatchFunc = func() error {
```
akutz, line 118: nit: I think it would improve readability to make this into a stand-alone function that returns the watcher function, ex. 

```golang
func (r *Reconciler) getPVCWatcherFn(
	ctx *context.ControllerManagerContext,
	mgr manager.Manager) func() error {

	// ...
}
```

controllers/volume/v1alpha2/volume_controller.go
```diff
@@ -95,15 +97,57 @@ func AddToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) er
...
+			pvcWatchStartedLock.Lock()
+			defer pvcWatchStartedLock.Unlock()
+
+			if pvcWatchStarted {
```
akutz, line 122: I think it would make it more clear if `pvcWatchStarted` and `pvcWatchStartedLock` were fields declared on the reconciler. I really had to think about this to realize the current approach was not a bug.

controllers/volume/v1alpha2/volume_controller.go
```diff
@@ -474,10 +525,13 @@ func (r *Reconciler) getInstanceStoragePVCs(
 			Name:      vol.PersistentVolumeClaim.ClaimName,
 		}
 		pvc := &corev1.PersistentVolumeClaim{}
-		if err := r.Get(ctx, objKey, pvc); client.IgnoreNotFound(err) != nil {
-			errs = append(errs, err)
+		if err := r.PVCReader.Get(ctx, objKey, pvc); err != nil {
```
akutz, line 528: Should there be a nil check here in case `PVCReader` was not initialized?

pkg/manager/cache.go
```diff
@@ -7,11 +7,45 @@ import (
...
+	"k8s.io/apimachinery/pkg/labels"
 	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+// NewLabelSelectorCacheForObject creates a new cache that watches the specified
+// object type selected by the label selector in all namespaces . The cache is
+// added to the manager and starts alongside the other leader-election runnables.
+func NewLabelSelectorCacheForObject(
```
akutz, line 19: Can this PR please add a test to `./pkg/manager/test` for the new label selector cache, similar to the test there for the namespace cache?

akutz (commented):
Thanks for this change @bryanv! I left a few comments, thanks again!

pkg/manager/cache.go
akutz: nm: it appears we don't test the namespace cache after all. Deferring to the controller-runtime test sounds good.

controllers/volume/v1alpha2/volume_controller.go
bryanv: I went a little back & forth on how to do organize this. Ultimately ended up on the Reconcile w/ a function pointer because that makes unit testing this much easier since won't have a ControllerManagerContext or Manager and just the fake client.

akutz (approved):
Lgtm, thanks!

PR #437, opened 2024-03-20
📖 Update VirtualMachine CRD doc

dilyar85:
This patch updates the `imageName` field in the VM CRD to reflect the correct resolution order from VMI to CVMI. It also includes the user-friendly image resolution done by the mutation webhook for better clarity.  

```release-note
Update VirtualMachine API doc to correct the resolving order of the imageName field.
```

### DIFF
--- a/api/v1alpha2/virtualmachine_types.go
+++ b/api/v1alpha2/virtualmachine_types.go
@@ -189,24 +189,25 @@ type VirtualMachineSpec struct {
 	//
 	// This field may be used to specify the name of a VirtualMachineImage
 	// or ClusterVirtualMachineImage resource. The resolver first checks to see
-	// if there is a ClusterVirtualMachineImage with the specified name. If no
+	// if there is a VirtualMachineImage with the specified name. If no
 	// such resource exists, the resolver then checks to see if there is a
-	// VirtualMachineImage resource with the specified name in the same
+	// ClusterVirtualMachineImage resource with the specified name in the same
 	// Namespace as the VM being deployed.
 	//
-	// This field is optional in the cases where there exists a sensible
-	// default value, such as when there is a single VirtualMachineImage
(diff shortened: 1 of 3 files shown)


api/v1alpha2/virtualmachine_types.go
```diff
@@ -189,26 +189,23 @@ type VirtualMachineSpec struct {
...
 	// such resource exists, the resolver then checks to see if there is a
-	// VirtualMachineImage resource with the specified name in the same
+	// ClusterVirtualMachineImage resource with the specified name in the same
 	// Namespace as the VM being deployed.
 	//
-	// This field is optional in the cases where there exists a sensible
-	// default value, such as when there is a single VirtualMachineImage
-	// resource available in the same Namespace as the VM being deployed.
-	//
-	// +optional
```
akutz, line 201: Please leave `// +optional` for now since it will be in the future. We already enforce this in the webhook. FWIW, you should also always keep `// +optional` and `omitempty` in sync.

api/v1alpha2/virtualmachine_types.go
```diff
@@ -189,26 +189,23 @@ type VirtualMachineSpec struct {
...
 	ImageName string `json:"imageName,omitempty"`
 
 	// ClassName describes the name of the VirtualMachineClass resource used to
 	// deploy this VM.
-	//
-	// This field is optional in the cases where there exists a sensible
-	// default value, such as when there is a single VirtualMachineClass
-	// resource available in the same Namespace as the VM being deployed.
-	//
-	// +optional
```
akutz, line 211: Please leave `// +optional` for now since it will be in the future. We already enforce this in the webhook. FWIW, you should also always keep `// +optional` and `omitempty` in sync.

akutz (changes requested):
Thanks @dilyar85

akutz (approved):

PR #439, opened 2024-03-21
🐛 Better clear VM.Status.Network fields

bryanv:

6c4fbc4 introduced a few ways that we'd no longer update the VM.Status.Network if some of the moVM.Guest itself or its fields were nil.

Note that even on a powered off VM the moVM.Guest.Net.IPStacks is not empty, so as a side effect PrimaryIP{4,6} and Interfaces fields would get cleared.

Also note that we need to preserve the VM.Status.Network.Config field since that is currently only re-evaluated when the VM is powered on.

Fixes #

```release-note
NONE
```


akutz (approved):
Thank you!

PR #440, opened 2024-03-21
✨ Bump manager container CPU limit to 750m

bryanv:
With version conversion during certain v1a1 use cases that are Update() heavy with large VM objects this brings this brings performance back to prior values.

Some later work will split the webhook and controller deployments in to two so these limits can be finer tuned.

Fixes #

```release-note
manager container CPU limit raised from 250m to 750m
```


akutz (approved):
It would be nice to attach the two flame graphs to the PR so others can see the additional work added by the conversion webhooks.

bryanv:
Not leader & leader CPU profiles:

leader-profile
notleader-profile

PR #441, opened 2024-03-21
🌱 Handle when the Volume PVC ClaimName is changed

bryanv:

When the PVC for a Volume changes the prior CnsNodeVmAttachment needs to be deleted and a new one that refers to the new ClaimName needs to be created. since CNS won't reconcile an already attached volume.

Fixes #

All the more reason that I want to remove the CnsNodeVmAttachment workflow and do this in house.

```release-note
NONE
```

### DIFF
--- a/controllers/volume/v1alpha2/volume_controller_unit_test.go
+++ b/controllers/volume/v1alpha2/volume_controller_unit_test.go
@@ -706,6 +706,56 @@ func unitTestsReconcile() {
 			})
 		})
 
+		When("VM Spec.Volumes has CNS volume with existing CnsNodeVmAttachment for a different PVC", func() {
+
+			BeforeEach(func() {
+				vmVol = *vmVolumeWithPVC1
+				vm.Spec.Volumes = append(vm.Spec.Volumes, vmVol)
+
+				attachment := cnsAttachmentForVMVolume(vm, vmVol)
+				attachment.UID = "1d2b1552-8294-4a72-91c7-7cf52b2e0990"
+				attachment.Status.Attached = true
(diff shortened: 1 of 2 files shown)


controllers/volume/v1alpha2/volume_controller_unit_test.go
```diff
@@ -674,6 +674,51 @@ func unitTestsReconcile() {
...
+				By("change volume claim name", func() {
+					vmVol.PersistentVolumeClaim.ClaimName = "my-new-claim-name"
+					vm.Spec.Volumes = []vmopv1.VirtualMachineVolume{vmVol}
+
+					err := reconciler.ReconcileNormal(volCtx)
+					Expect(err).ToNot(HaveOccurred())
+				})
+
+				By("new CnsNodeVmAttachment is created with new claim name", func() {
+					attachment := getCNSAttachmentForVolumeName(vm, vmVol.Name)
```
dilyar85, line 711: Should we consider checking if the UID of this attachment has changed? This could help us verify that the attachment was actually recreated with a new claim name (to ensure that it would be reconciled by CNS).

dilyar85 (approved):
LGTM with a minor thought on the test.

controllers/volume/v1alpha2/volume_controller_unit_test.go
bryanv: I debated doing that but wasn't something we get for free with the fake client so I decided that Attached going from true to false along with assertVMVolStatusFromAttachmen() was good enough. I'll just fake up a UID on the attachment that is created in the `BeforeEach()`

PR #442, opened 2024-03-26
🐛 Fix local mkdocs in Python 3.11+

akutz:

In Python 3.11+, the pip command can no longer user the --user flag without encountering an error about an "externally-managed- environment", ex.

```shell
$ make docs-serve-python
pip3 install --user -r ./docs/requirements.txt
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try brew install
    xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-brew-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.
    
    If you wish to install a non-brew packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
make: *** [Makefile:561: docs-serve-python] Error 1
```

This change is documented at https://peps.python.org/pep-0668/. It may not be true for *all* environments, but it does happen on the managed macOS systems used by the Broadcom engineers working on VM Operator (at least if they are using Python via homebrew), so a work-around is needed.

The neatest solution is to use a Python virtual environment. Thus, now when the targets "docs-build-python" or
"docs-serve-python" are executed locally, a virtual environment is created at .venv/mkdocs.

Fixes `NA`

```release-note
The local docs targets now use a Python virtual environment at .venv/mkdocs.
```


dilyar85 (approved):
LGTM! I verified that the updated make targets worked for me locally using Python version 3.12.2.

PR #443, opened 2024-03-26
🌱 Remove Simple LB provider

bryanv:
This was added for local gce2e to act as a local LB when using kind. The local gce2e test env is gone now, and newer kind supports MetalLB which makes more sense to use.

Keep around the LB_PROVIDER envvar in case we need it in the future.

Fixes #

```release-note
NONE
```


akutz (approved):
Hell yes. Thank you BV.

PR #446, opened 2024-03-26
⚠️ Remove need for VM Image in Update path

bryanv:

Post create, I'd like us to never have to depend in the image resource existing since it may get deleted. Fix the two situations where the image was being used.

First the JSON_EXTRA_CONFIG contains RUN.container that uses a template to fill in the image name so vcsim creates a container under the covers. This was used for old local gce2e testing, and isn't used by us but some other teams have a desire to use it so keep it around. Now that is just set on create in the EC and we don't need to re-eval it later on.

Second remove the V1Alpha1Compatible condition check and always check the EC. This does change the behavior that if an image did not have the condition but had guestinfo.vmservice.defer-cloud-init=ready, we'll now always set it to enabled. I don't think this will really matter.

For now continue to set the FirstBootDoneAnnotation.

Fixes #

Could remove FirstBootDoneAnnotation too if we want to.

```release-note
NONE
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
+++ b/pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
@@ -6,6 +6,7 @@ package vsphere
 import (
 	goctx "context"
 	"fmt"
+	"maps"
 	"strings"
 	"sync"
 	"text/template"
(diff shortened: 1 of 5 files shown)


pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
```diff
@@ -942,23 +943,23 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpecExtraConfig(
...
+	if v, exists := ecMap[constants.ExtraConfigRunContainerKey]; exists {
+		// For local-vcsim the JSON_EXTRA_CONFIG is set and contains RUN.container so we use the
+		// template to fill in the image {{.Name }} so vcsim will create a container for the VM.
+		renderTemplateFn := func(name, text string) string {
+			t, err := template.New(name).Parse(text)
+			if err != nil {
+				return text
+			}
+			b := strings.Builder{}
+			if err := t.Execute(&b, createArgs.ImageStatus); err != nil {
```
bryanv, line 957: If we ever template more fields we should define an intermediate struct

pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
```diff
@@ -1106,56 +1114,12 @@ func (vs *vSphereVMProvider) vmUpdateGetArgs(
...
-		// when the vmMetadata is nil or when the transport requested is not CloudInit.
-		updateArgs.VirtualMachineImageV1Alpha1Compatible =
-			conditions.IsTrueFromConditions(vmImageStatus.Conditions, vmopv1.VirtualMachineImageV1Alpha1CompatibleCondition)
-	}
-
 	updateArgs.ConfigSpec = virtualmachine.CreateConfigSpec(
 		vmCtx,
 		vmClassConfigSpec,
 		&updateArgs.VMClass.Spec,
-		vmImageStatus,
```
bryanv, line 1149: The image status is used for the firmware and HW versions fallback. Neither is something we care about in the update path. This still kind of stinks and we need to keep hacking away at what's left over of the old session code.

pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
akutz: Maybe leave a TODO in the code instead of on the review?

pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
```diff
@@ -1087,6 +1088,13 @@ func (vs *vSphereVMProvider) vmUpdateGetArgs(
 	updateArgs.ResourcePolicy = resourcePolicy
 	updateArgs.BootstrapData = bsData
 
+	ecMap := maps.Clone(vs.globalExtraConfig)
+	maps.DeleteFunc(ecMap, func(k string, v string) bool {
```
akutz, line 1092: Cool, another neat, new Golang func!

pkg/vmprovider/providers/vsphere2/vmprovider_vm.go
akutz: If we add `spec.firmware` the way we have `spec.minHardwareVersion`, do we need to keep anything? Can we not just copy the values to those fields if they are empty?

bryanv: There is a `vmoperator.vmware.com/firmware` annotation. In the update path we don't use the those fields from the ConfigSpec created here, so what values don't really matter. Both `spec.minHardwareVersion` and the firmware annotation are what's used in the update path. I don't think a `spec.firmware` gains us anything specific for here, and IDK if that's used/cared enough about for a spec field. All this is still a confusing mix of pre and post CL DeployOVF w/ ConfigSpec work that needs to kept being chipped away at for resize work, and our sanity.

akutz (approved):

sreyasn (approved):
LGTM!

PR #447, opened 2024-04-01
✨ UpdateExtraConfig for powered on VMs, iterate list once

akutz:

This patch modifies the logic for updating ExtraConfig to support powered on VMs. This ensures powered on VMs that have vGPUs will have the flag in ExtraConfig removed that DRS queries to determine whether or not to power off the VM during vMotion. This was already handled in #353, but that logic is only executed before a VM is powered on, not for currently powered on VMs.

This patch also optimizes the behavior of the UpdateEC function to only iterate over the existing EC list once.

Fixes `NA`

```release-note
Update ExtraConfig for powered on VMs.
```

### DIFF
--- a/pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
+++ b/pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
@@ -300,59 +300,133 @@ func UpdateConfigSpecExtraConfig(
 	vmClassSpec *vmopv1.VirtualMachineClassSpec,
 	vm *vmopv1.VirtualMachine,
 	globalExtraConfig map[string]string,
-	imageV1Alpha1Compatible bool) {
+	imgV1A1Compat bool) {
 
-	extraConfig := maps.Clone(globalExtraConfig)
+	// Either initialize extraConfig to an empty map or from a copy of
+	// globalExtraConfig.
+	var extraConfig map[string]string
+	if globalExtraConfig == nil {
+		extraConfig = map[string]string{}
+	} else {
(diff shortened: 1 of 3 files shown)


dilyar85 (approved):
LGTM!

pkg/vmprovider/providers/vsphere2/session/session_vm_update.go
```diff
@@ -300,59 +300,111 @@ func UpdateConfigSpecExtraConfig(
...
 
-	virtualDevices := vmClassSpec.Hardware.Devices
-	if len(virtualDevices.VGPUDevices) > 0 || len(virtualDevices.DynamicDirectPathIODevices) > 0 ||
-		(classConfigSpec != nil && util.HasVirtualPCIPassthroughDeviceChange(classConfigSpec.DeviceChange)) {
-		setMMIOExtraConfig(vm, extraConfig)
+	// Ensure a VM with vGPUs or dynamic direct path I/O devices has the the
+	// correct flags in ExtraConfig for memory mapped I/O. Please see
+	// https://kb.vmware.com/s/article/2142307 for more information about these
+	// flags.
+	if hasvGPUOrDDPIODevices(classConfigSpec, vmClassSpec) {
```
bryanv, line 318: If we're doing this while the VM is on shouldn't it really be looking at the VM's ConfigInfo instead of the class?

akutz: Technically if the VM is deployed at all. I guess the current code path is an artifact of a single update post deploy.

bryanv (approved):

PR #448, opened 2024-04-01
✨ Remove VM Class as Config (DnD too) FSS

akutz:

This patch removes the feature states for the VM Class as Config and Day-and-Date features.

Fixes `NA`

```release-note
Remove the FSS_WCP_VM_CLASS_AS_CONFIG and FSS_WCP_VM_CLASS_AS_CONFIG_DAYNDATE feature states.
```


config/local/vmoperator/local_env_var_patch.yaml
```diff
@@ -17,10 +17,6 @@ spec:
           value: "false"
         - name: FSS_WCP_INSTANCE_STORAGE
           value: "true"
-        - name: FSS_WCP_VM_CLASS_AS_CONFIG
```
aruneshpa, line 20: I think these still need to be there for gce2e on older branches.

akutz: Only if we pull in the updated VM Op into those branches. Besides, I'd rather gce2e be updated to not depend on VM Op's deployment spec for the state of FSSs. It can break, and then it can be fixed to look at something else. Still, I'm not sure it will break at all since older branches should not be pointing to VM Op on main.

bryanv: It is used to determine if the FFS is enabled or not. I think the test will fail if the expected envar isn't there. It is quite brittle but not sure if we ever added a CM that contained the FFS states. Regardless gce2e ToT should be changed to either assume this one is enabled or default to enabled if not found.

bryanv (approved):

config/local/vmoperator/local_env_var_patch.yaml
akutz: > Regardless gce2e ToT should be changed to either assume this one is enabled or default to enabled if not found.

Yes, it should, but I do not think it needs to happen concurrently with upstream commits. It's just this patch won't be recommended from main until gce2e ToT is updated.

aruneshpa: Yep.  I think you all covered the reasoning pretty well.  And I agree that gce2e should handle this, but it doesn't.

The issue boils down to gce2e not being able to know if a missing FSS env var means FSS enabled and removed, or disabled.  This is specially tricky since we run gce2e from main for crossports to older branches (although that is a independent problem for another day).

PR #456, opened 2024-04-03
🌱 Add label to network interface CR for its VM name

bryanv:

Going forward for new VMs, once the network interfaces are mutable and we need to reconcile the network interface CRs we can use this to quickly list all the interfaces for the VM instead of having to list all in the namespace.

Fixes #

```release-note
NONE
```


akutz (approved):

dilyar85 (approved):
LGTM!

PR #458, opened 2024-04-04
🐛 Fix UpdateConfigSpecExtraConfig() when reconciling VM with vGPU/DDPIO

dilyar85:
This PR updates the `setMemoryMappedIOFlagsInExtraConfig()` function to skip if the VM is nil. This could be the case when it's called from `UpdateConfigSpecExtraConfig()` while reconciling a powered-on VM with a vGPU or DDPIO device.

Fixes #447.

```release-note
Fix UpdateConfigSpecExtraConfig() when reconciling a powered-on VM with vGPU or DDPIO device.
```


akutz (approved):
Thank you @dilyar85 !!

PR #463, opened 2024-04-09
🐛 Specify BIOS and instance UUIDs when creating VM

dougm:

When creating a new VirtualMachine:
- InstanceUUID is set to VM ObjectMeta.UID
- BiosUUID is included in VirtualMachineSpec
- cloud-init metadata instance ID is set to VM spec.biosUUID

Enable find by UUID preference in vcenter.GetVirtualMachine, as it is more stable than Inventory path.

### DIFF
--- a/api/v1alpha1/virtualmachine_conversion.go
+++ b/api/v1alpha1/virtualmachine_conversion.go
@@ -930,6 +930,12 @@ func restore_v1alpha3_VirtualMachineReadinessProbeSpec(
 	}
 }
 
+func restore_v1alpha3_VirtualMachineBiosUUID(
+	dst, src *vmopv1.VirtualMachine) {
+
+	dst.Spec.BiosUUID = src.Spec.BiosUUID
+}
+
 func convert_v1alpha1_PreReqsReadyCondition_to_v1alpha3_Conditions(
 	dst *vmopv1.VirtualMachine) []metav1.Condition {
 
@@ -1118,6 +1124,7 @@ func (src *VirtualMachine) ConvertTo(dstRaw ctrlconversion.Hub) error {
 	restore_v1alpha3_VirtualMachineBootstrapSpec(dst, restored)
 	restore_v1alpha3_VirtualMachineNetworkSpec(dst, restored)
 	restore_v1alpha3_VirtualMachineReadinessProbeSpec(dst, restored)
+	restore_v1alpha3_VirtualMachineBiosUUID(dst, restored)
 
 	// END RESTORE
 
--- a/api/v1alpha3/virtualmachine_bootstrap_types.go
+++ b/api/v1alpha3/virtualmachine_bootstrap_types.go
@@ -85,6 +85,12 @@ type VirtualMachineBootstrapSpec struct {
 // VirtualMachineBootstrapCloudInitSpec describes the CloudInit configuration
 // used to bootstrap the VM.
 type VirtualMachineBootstrapCloudInitSpec struct {
+	// InstanceID is the cloud-init metadata instance ID.
+	// If omitted, this field defaults to the VM's BiosUUID.
+	//
+	// +optional
+	InstanceID string `json:"instanceID,omitempty"`
+
 	// CloudConfig describes a subset of a Cloud-Init CloudConfig, used to
 	// bootstrap the VM.
 	//
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -436,6 +436,14 @@ type VirtualMachineSpec struct {
 	// behavior. In other words, please be careful when choosing to upgrade a
 	// VM to a newer hardware version.
 	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
+
+	// BiosUUID describes the desired BIOS UUID for a VM.
+	// If omitted, this field defaults to a random UUID.
+	// When the bootstrap provider is Cloud-Init, this value is
+	// used as the Cloud-Init instance ID.
+	// +optional
+	// +kubebuilder:validation:Format:=uuid4
+	BiosUUID string `json:"biosUUID,omitempty"`
 }
 
 // VirtualMachineReservedSpec describes a set of VM configuration options
--- a/pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
+++ b/pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
@@ -49,9 +49,12 @@ func BootStrapCloudInit(
 		sshPublicKeys = strings.Join(cloudInitSpec.SSHAuthorizedKeys, "\n")
 	}
 
-	uid := string(vmCtx.VM.UID)
-	// Use the existing instance ID if it is already set on the VM.
-	// This ensures cloud-init uses the same instance ID across B/R.
+	uid := cloudInitSpec.InstanceID
+	if uid == "" {
+		// InstanceID will be set to the VM bios uuid when creating new VMs.
+		// Use VM.UID for backward compatibility when InstanceID is not set.
+		uid = string(vmCtx.VM.UID)
+	}
 	if value, ok := vmCtx.VM.Annotations[vmopv1.InstanceIDAnnotation]; ok {
 		uid = value
 	}
--- a/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
+++ b/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -118,6 +119,7 @@ func (m mutator) Mutate(ctx *pkgctx.WebhookRequestContext) admission.Response {
 		SetCreatedAtAnnotations(ctx, modified)
 		AddDefaultNetworkInterface(ctx, m.client, modified)
 		SetDefaultPowerState(ctx, m.client, modified)
+		SetDefaultBiosUUID(ctx, m.client, modified)
 		if _, err := ResolveImageNameOnCreate(ctx, m.client, modified); err != nil {
 			return admission.Denied(err.Error())
 		}
@@ -309,6 +311,35 @@ func SetDefaultPowerState(
 		return true
 	}
 	return false
+}
+
+// SetDefaultBiosUUID sets a default bios uuid for a new VM.
+// If CloudInit is the Bootstrap method, CloudInit InstanceID is also set to BiosUUID.
+// Return true if a default bios uuid was set, otherwise false.
+func SetDefaultBiosUUID(
+	ctx *pkgctx.WebhookRequestContext,
+	client client.Client,
+	vm *vmopv1.VirtualMachine) bool {
+
+	var wasMutated bool
+
+	if vm.Spec.BiosUUID == "" {
+		// Default to a Random (Version 4) UUID.
+		// This is the same UUID flavor/version used by Kubernetes and preferred by vSphere.
+		vm.Spec.BiosUUID = uuid.New().String()
+		wasMutated = true
+	}
+
+	if vm.Spec.Bootstrap != nil {
+		if ci := vm.Spec.Bootstrap.CloudInit; ci != nil {
+			if ci.InstanceID == "" {
+				ci.InstanceID = vm.Spec.BiosUUID
+				wasMutated = true
+			}
+		}
+	}
+
+	return wasMutated
 }
 
 const (
--- a/webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
+++ b/webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
@@ -9,6 +9,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	"github.com/google/uuid"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -109,6 +110,55 @@ func intgTestsMutating() {
 					Expect(modified.Spec.Network.Interfaces[0].Name).To(Equal("eth0"))
 					Expect(modified.Spec.Network.Interfaces[0].Network.Kind).To(Equal("Network"))
 				})
+			})
+		})
+	})
+
+	Context("SetDefaultBiosUUID", func() {
+		When("Creating VirtualMachine", func() {
+			When("When VM BiosUUID is empty", func() {
+				BeforeEach(func() {
+					ctx.vm.Spec.BiosUUID = ""
+				})
+
+				It("Should set BiosUUID", func() {
+					Expect(ctx.Client.Create(ctx, ctx.vm)).To(Succeed())
+
+					vm := &vmopv1.VirtualMachine{}
+					Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(ctx.vm), vm)).To(Succeed())
+					Expect(vm.Spec.BiosUUID).ToNot(BeEmpty())
+				})
+			})
+			When("When VM BiosUUID is not empty", func() {
+				var id = uuid.New().String()
+
+				BeforeEach(func() {
+					ctx.vm.Spec.BiosUUID = id
+				})
+
+				It("Should not mutate BiosUUID", func() {
+					Expect(ctx.Client.Create(ctx, ctx.vm)).To(Succeed())
+					vm := &vmopv1.VirtualMachine{}
+					Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(ctx.vm), vm)).To(Succeed())
+					Expect(vm.Spec.BiosUUID).To(Equal(id))
+				})
+			})
+		})
+
+		When("Updating VirtualMachine", func() {
+			var id = uuid.New().String()
+
+			BeforeEach(func() {
+				ctx.vm.Spec.BiosUUID = id
+				Expect(ctx.Client.Create(ctx, ctx.vm)).To(Succeed())
+			})
+
+			It("Should not mutate BiosUUID", func() {
+				vm := &vmopv1.VirtualMachine{}
+				Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(ctx.vm), vm)).To(Succeed())
+				Expect(ctx.Client.Update(ctx, vm)).To(Succeed())
+				Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(ctx.vm), vm)).To(Succeed())
+				Expect(vm.Spec.BiosUUID).To(Equal(id))
 			})
 		})
 	})
--- a/webhooks/virtualmachine/mutation/virtualmachine_mutator_unit_test.go
+++ b/webhooks/virtualmachine/mutation/virtualmachine_mutator_unit_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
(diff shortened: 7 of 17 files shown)


akutz (commented):
Great start! Left some comments. Also, PR description mentions cloud-init, but I don't see that change in this PR?

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec.go
```diff
@@ -49,6 +49,8 @@ func CreateConfigSpec(
 		ExtensionKey: vmopv1.ManagedByExtensionKey,
 		Type:         vmopv1.ManagedByExtensionType,
 	}
+	configSpec.InstanceUuid = string(vmCtx.VM.UID)
```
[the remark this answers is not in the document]
dougm: Good catch, yes. Fixed and added test.

api/v1alpha2/virtualmachine_types.go
```diff
@@ -380,6 +380,10 @@ type VirtualMachineSpec struct {
 	// behavior. In other words, please be careful when choosing to upgrade a
 	// VM to a newer hardware version.
 	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
+
+	// BiosUUID describes the desired BIOS UUID of a VirtualMachine.
```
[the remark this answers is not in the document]
dougm: Thank you, done

webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator.go
```diff
@@ -295,6 +297,35 @@ func SetDefaultPowerState(
...
+
+	var wasMutated bool
+
+	if vm.Spec.BiosUUID == "" {
+		vm.Spec.BiosUUID = uuid.New().String()
+		wasMutated = true
+	}
+
+	if vm.Spec.Bootstrap != nil && vm.Spec.Bootstrap.CloudInit != nil {
+		if vm.Annotations == nil {
```
bryanv, line 316: Nit: we could move this inside the `if id != vm.Spec.BiosUUID` block since that's when we need to write to the map.

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec_test.go
```diff
@@ -6,11 +6,13 @@ package virtualmachine_test
 import (
 	goctx "context"
 
+	"github.com/google/uuid"
```
bryanv, line 9: Nit: sort this down below with the rest of the non dot imports (we should prb see if there's a linter for this).

api/v1alpha2/virtualmachine_types.go
```diff
@@ -380,6 +380,13 @@ type VirtualMachineSpec struct {
 	// behavior. In other words, please be careful when choosing to upgrade a
 	// VM to a newer hardware version.
 	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
+
+	// BiosUUID describes the desired BIOS UUID for a VM.
+	// If omitted, this field defaults to a random UUID.
+	// When the bootstrap provider is Cloud-Init, this value is
+	// used as the Cloud-Init instance ID.
+	// +optional
+	BiosUUID string `json:"biosUUID,omitempty"`
```
bryanv, line 389: Can you set this field in the big https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha1/virtualmachine_conversion_test.go#L61 conversion test to make sure that its preserved on conversions?

pkg/vmprovider/providers/vsphere2/vcenter/getvm.go
```diff
@@ -31,14 +31,11 @@ func GetVirtualMachine(
 		}
 	}
 
-	// For when we start to use the k8s VM.UID for the VC VM's InstanceUUID or UUID (aka BiosUUID):
-	/*
-		if instanceUUID := vmCtx.VM.UID; instanceUUID != "" {
-			if vm, err := findVMByUUID(vmCtx, vimClient, datacenter, string(instanceUUID), true); err == nil {
-				return vm, nil
-			}
+	if instanceUUID := vmCtx.VM.UID; instanceUUID != "" {
```
bryanv, line 34: Do we want to do this lookup first since it is support to be the most stable?

Should we only do this lookup when this UID matches the Spec value?

api/v1alpha2/virtualmachine_types.go
dougm: Thanks for pointing that out, done.

pkg/vmprovider/providers/vsphere2/vcenter/getvm.go
dougm: Yes, changed the order.

For new VMs:
- `InstanceUuid` is set to the VM's `ObjectMeta.UID`
- `Uuid` is set to VM's spec.biosUUID

For existing brownfield VMs, both InstanceUuid and Uuid were generated by VC. So I don't think we have a way to skip? But, we could add lookup for BiosUUID, as in:
```
findVMByUUID(vmCtx, vimClient, datacenter, vmCtx.VM.Spec.BiosUUID, false)
```
What do you think?

pkg/vmprovider/providers/vsphere2/virtualmachine/configspec_test.go
dougm: Done

webhooks/virtualmachine/v1alpha2/mutation/virtualmachine_mutator.go
dougm: Done

dougm (commented):
Thanks for the review @bryanv

dougm:
> Great start! Left some comments. Also, PR description mentions cloud-init, but I don't see that change in this PR?

@akutz sorry just realized I'd missed this comment. The PR propagates spec.biosUUID to cloud-init UID via `InstanceIDAnnotation`: https://github.com/vmware-tanzu/vm-operator/pull/463/files#diff-c8fd712c0c230b622f8d05fad6eb15f695c439d43330d4c863ca6316543f6c22R336

akutz:
Hey @dougm ,

You will need to add code to https://github.com/vmware-tanzu/vm-operator/blob/5b7945822399a295d3d4e5c792e1dae667cb710d/api/v1alpha1/virtualmachine_conversion.go#L1060-L1083 and  https://github.com/vmware-tanzu/vm-operator/blob/5b7945822399a295d3d4e5c792e1dae667cb710d/api/v1alpha2/virtualmachine_conversion.go#L13-L33 to restore the value of `spec.biosUUID` when converting from v1a1 or v1a2 back to v1a3.

dougm:
> to restore the value of `spec.biosUUID` when converting from v1a1 or v1a2 back to v1a3.
Done, thanks @akutz

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -1078,6 +1078,7 @@ func (src *VirtualMachine) ConvertTo(dstRaw conversion.Hub) error {
 	restore_v1alpha3_VirtualMachineReadinessProbeSpec(dst, restored)
 
 	dst.Status = restored.Status
+	dst.Spec.BiosUUID = restored.Spec.BiosUUID
```
bryanv, line 1081: To keep the same pattern can we have a `restore_v1alpha3_VirtualMachineBiosUUID()` function like we do above?

dougm: Done 👍

api/v1alpha3/virtualmachine_types.go
```diff
@@ -406,6 +406,13 @@ type VirtualMachineSpec struct {
 	// behavior. In other words, please be careful when choosing to upgrade a
 	// VM to a newer hardware version.
 	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
+
+	// BiosUUID describes the desired BIOS UUID for a VM.
+	// If omitted, this field defaults to a random UUID.
+	// When the bootstrap provider is Cloud-Init, this value is
+	// used as the Cloud-Init instance ID.
+	// +optional
+	BiosUUID string `json:"biosUUID,omitempty"`
```
akutz, line 415: This should also validate the format, ex.

```go
// +kubebuilder:validation:Format:=uuid4
```

See https://github.com/kubernetes/kubernetes/blob/ac8ac0fc170eec7cdf292d841f8d026a65ce1681/vendor/github.com/go-openapi/strfmt/format.go#L96-L169 for a list of not super well-known formats that can be used.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -310,6 +312,35 @@ func SetDefaultPowerState(
...
+// Return true if a default bios uuid was set, otherwise false.
+func SetDefaultBiosUUID(
+	ctx *context.WebhookRequestContext,
+	client client.Client,
+	vm *vmopv1.VirtualMachine) bool {
+
+	var wasMutated bool
+
+	if vm.Spec.BiosUUID == "" {
+		vm.Spec.BiosUUID = uuid.New().String()
```
akutz, line 326: Can we make this New4 or New5? Just something that makes the version of the UUID explicit. That way we can ensure the validation is consistent with the expected version.

api/v1alpha3/virtualmachine_types.go
dougm: Nice, will do.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
dougm: I don't see those in the [uuid package](https://pkg.go.dev/github.com/google/uuid), just New, New6, New7. Should I add a helper/util func in vm-operator instead?

webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
```diff
@@ -113,6 +114,55 @@ func intgTestsMutating() {
...
 	})
 
+	Context("SetDefaultBiosUUID", func() {
+		When("Creating VirtualMachine", func() {
+			When("When VM BiosUUID is empty", func() {
+				BeforeEach(func() {
+					ctx.vm.Spec.BiosUUID = ""
+				})
+
+				It("Should set BiosUUID to PoweredOn", func() {
```
sreyasn, line 124: typo: 'to PoweredOn'

webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
```diff
@@ -113,6 +114,55 @@ func intgTestsMutating() {
...
+			When("When VM BiosUUID is not empty", func() {
+				var id = uuid.New().String()
+
+				BeforeEach(func() {
+					ctx.vm.Spec.BiosUUID = id
+				})
+
+				It("Should not mutate BiosUUID", func() {
+					Expect(ctx.Client.Create(ctx, ctx.vm)).To(Succeed())
+					modified := &vmopv1.VirtualMachine{}
```
sreyasn, line 141: let's just call this 'vm' instead of 'modified' because it doesn't get mutated?

webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
```diff
@@ -113,6 +114,55 @@ func intgTestsMutating() {
...
+		When("Updating VirtualMachine", func() {
+			var id = uuid.New().String()
+
+			BeforeEach(func() {
+				ctx.vm.Spec.BiosUUID = id
+				Expect(ctx.Client.Create(ctx, ctx.vm)).To(Succeed())
+			})
+
+			It("Should not mutate BiosUUID", func() {
+				modified := &vmopv1.VirtualMachine{}
```
sreyasn, line 157: same nit as above

webhooks/virtualmachine/mutation/virtualmachine_mutator_unit_test.go
```diff
@@ -271,6 +272,59 @@ func unitTestsMutating() {
...
 	})
 
+	Describe("SetDefaultBiosUUID", func() {
+
+		Context("When VM BiosUUID is empty", func() {
+			BeforeEach(func() {
+				ctx.vm.Spec.BiosUUID = ""
+			})
+
+			It("Should set BiosUUID to PoweredOn", func() {
```
sreyasn, line 282: is 'to PoweredOn' a typo?

webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
dougm: Whoops, fixed. I'd copy-n-pasted `It` block from another test

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
akutz: Huh. Well, do we know what version of the UUID vSphere uses?

akutz: FWIW, the checks in kube are just regexes https://github.com/kubernetes/kubernetes/blob/ac8ac0fc170eec7cdf292d841f8d026a65ce1681/vendor/github.com/go-openapi/strfmt/default.go#L61-L79

akutz: According to @mikekim, VMware tends to prefer UUID4, actual "random" UUIDs. The Google UUID package lists the various types and what they are -- https://pkg.go.dev/github.com/google/uuid#New. It looks like `New` creates a UUID4, so we should be good with that.

dougm: Agreed, added comments indicating we're using random UUID4 here.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -962,6 +962,7 @@ func (v validator) validateImmutableFields(ctx *context.WebhookRequestContext, v
 	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.ImageName, oldVM.Spec.ImageName, specPath.Child("imageName"))...)
 	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.ClassName, oldVM.Spec.ClassName, specPath.Child("className"))...)
 	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.StorageClass, oldVM.Spec.StorageClass, specPath.Child("storageClass"))...)
+	allErrs = append(allErrs, validation.ValidateImmutableField(vm.Spec.BiosUUID, oldVM.Spec.BiosUUID, specPath.Child("biosUUID"))...)
```
dilyar85, line 965: A couple of questions about validating the `vm.spec.biosUUID` field:
- Will we add it for old VMs during the upgrade process? If so, this check would always fail the upgrade requests.
- Do we allow DevOps users to create a VM with a specified biosUUID? I don't see it mentioned in the API or enforced in the validation webhook, but I remember that being part of your design.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -406,6 +406,14 @@ type VirtualMachineSpec struct {
...
 	// VM to a newer hardware version.
 	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
+
+	// BiosUUID describes the desired BIOS UUID for a VM.
+	// If omitted, this field defaults to a random UUID.
+	// When the bootstrap provider is Cloud-Init, this value is
+	// used as the Cloud-Init instance ID.
+	// +optional
+	// +kubebuilder:validation:Format:=uuid4
+	BiosUUID string `json:"biosUUID,omitempty"`
```
dilyar85, line 416: I think this might break the B/R workflow for VMs using the cloud-init bootstrap provider. A new cloud-init instance ID is always required to ensure that the network configuration runs on the restored VM. If it's set from the `vm.spec.biosUUID`, the same value from the backup would be used in restored VMs and hence not trigger the network configuration.

dougm: > A new cloud-init instance ID is always required

Existing code comment says same ID should be used?
https://github.com/vmware-tanzu/vm-operator/blob/df36cc7c94996357e5042cc1ff0d0feb6a818f9f/pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go#L53-L54

But I've yet to read up on details of the workflow. Where else would `InstanceIDAnnotation` be set and to what value?

dilyar85: That comment about B/R is no longer applicable and should be removed. We explicitly delete this annotation during the restore to ensure a new cloud-init instance ID is set to the restored VM CR UID.  
The annotation is actually set during the upgrade process to the value of `iid-datasource-none`. This is to address an old bug when cloud-init was used sans userdata. See https://github.com/vmware-tanzu/vm-operator/pull/211 that initiated added this code in v1a1.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -310,6 +312,37 @@ func SetDefaultPowerState(
...
+	var wasMutated bool
+
+	if vm.Spec.BiosUUID == "" {
+		// Default to a Random (Version 4) UUID.
+		// This is the same UUID flavor/version used by Kubernetes and preferred by vSphere.
+		vm.Spec.BiosUUID = uuid.New().String()
+		wasMutated = true
+	}
+
+	if vm.Spec.Bootstrap != nil && vm.Spec.Bootstrap.CloudInit != nil {
```
aruneshpa, line 332: Would this also handle all the cases and be a bit more readable?
```go
if vm.Annotations == nil {
    vm.Annotations = map[string]string{}
}
                                                                                                                                                                                                                                                    	if vm.Spec.Bootstrap != nil && vm.Spec.Boot
if vm.Annotations[vmopv1.InstanceIDAnnotation] != vm.Spec.BiosUUID {
    vm.Annotations[vmopv1.InstanceIDAnnotation] = vm.Spec.BiosUUID
    wasMutated = true
}

```

dougm: I'd meant to only create the map if nil, but realized after your comment I missed checking for nil map above:
```go
id := vm.Annotations[vmopv1.InstanceIDAnnotation]
```
This code may also go away after we resolve the cloud-init.UID discussion

webhooks/virtualmachine/validation/virtualmachine_validator.go
dougm: Thanks @dilyar85 , both should be fixed now

api/v1alpha3/virtualmachine_types.go
dougm: Updated to use new `VirtualMachineBootstrapCloudInitSpec.InstanceID` field

pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
```diff
@@ -49,9 +49,10 @@ func BootStrapCloudInit(
 		sshPublicKeys = strings.Join(cloudInitSpec.SSHAuthorizedKeys, "\n")
 	}
 
-	uid := string(vmCtx.VM.UID)
-	// Use the existing instance ID if it is already set on the VM.
-	// This ensures cloud-init uses the same instance ID across B/R.
+	uid := cloudInitSpec.InstanceID
+	if uid == "" {
+		uid = string(vmCtx.VM.UID)
+	}
```
dilyar85, line 55: I suppose this is for backward compatibility to avoid updating an existing VM's cloud-init instance ID, right? It would be great if you could add a comment for clarity and future reference.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1154,3 +1159,18 @@ func (v validator) validateLabel(ctx *pkgctx.WebhookRequestContext, vm, oldVM *v
...
+	var allErrs field.ErrorList
+
+	if ctx.IsPrivilegedAccount {
+		return allErrs
+	}
+
+	if vm.Spec.BiosUUID != "" {
+		uuidPath := field.NewPath("spec", "biosUUID")
+		allErrs = append(allErrs, field.Forbidden(uuidPath, modifyLabelNotAllowedForNonAdmin))
+	}
```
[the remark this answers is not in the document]
dougm: `validateBiosUUID` is only called from `ValidateCreate`, not used in `ValidateUpdate`. I added a comment.

pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
dougm: Right, added comment.

dilyar85 (approved):
LGTM! Once this change is merged, we need to update the restore codebase accordingly to ensure a new cloud-init instance ID is used for creating a restored VM.

webhooks/virtualmachine/mutation/virtualmachine_mutator_unit_test.go
```diff
@@ -272,6 +273,47 @@ func unitTestsMutating() {
...
+				Expect(mutation.SetDefaultBiosUUID(&ctx.WebhookRequestContext, ctx.Client, ctx.vm)).To(BeFalse())
+				Expect(ctx.vm.Spec.BiosUUID).Should(Equal(oldVM.Spec.BiosUUID))
+			})
+		})
+
+		Context("When CloudInit is not empty", func() {
+			BeforeEach(func() {
+				ctx.vm.Spec.BiosUUID = uuid.New().String()
+				ctx.vm.Spec.Bootstrap = &vmopv1.VirtualMachineBootstrapSpec{
+					CloudInit: &vmopv1.VirtualMachineBootstrapCloudInitSpec{},
```
sreyasn, line 306: can we add another test where ctx.vm.Spec.BiosUUID and  vm.Spec.Bootstrap.CloudInit.InstanceId is preset?.

dougm: Added

api/v1alpha3/virtualmachine_bootstrap_types.go
```diff
@@ -85,6 +85,12 @@ type VirtualMachineBootstrapSpec struct {
 // VirtualMachineBootstrapCloudInitSpec describes the CloudInit configuration
 // used to bootstrap the VM.
 type VirtualMachineBootstrapCloudInitSpec struct {
+	// InstanceID is the cloud-init metadata instance ID.
+	// If omitted, this field defaults to the VM's BiosUUID.
+	//
+	// +optional
+	InstanceID string `json:"instanceID,omitempty"`
```
sreyasn, line 92: Can both devOps and Admin users specify this field ?

sreyasn (approved):
The change looks good to me!. Thanks @dougm.

api/v1alpha3/virtualmachine_bootstrap_types.go
dougm: Currently, yes, I don't think we called out a need for this field to be admin-only.

PR #466, opened 2024-04-11
🐛Use IsDNS1035Label for VirtualMachineService validations

sreyasn:

This change fixes the validation for VirtualMachineService names to
use IsDNS1035Label. This is the right one to use since we create k8s Service objects
with the same name and they need to comply with RFC 1035.

Service Port names are validated as DNS 1123 still.

Fixes #

N/A

```NONE

```


bryanv (approved):
Just put back the prior port name validation

PR #469, opened 2024-04-15
:sparkles: Implement admin override to pause VM reconciliation

zyiyi11:
This change will support admins to pause VM Service VM reconciliation by adding ExtraConfig Key [PauseVMExtraConfigKey](https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha3/virtualmachine_types.go#L155). After admin adds EC key, VM will have label [PausedVMLabelKey](https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha3/virtualmachine_types.go#L162) =`admin` and `VirtualMachineReconcileReady` condition set to false, showing reason being `VirtualMachineReconcilePaused`, indicates that VirtualMachine reconciliation is being paused by Admin.

This change also patches devops pause VM workflow. DevOps add [PauseAnnotation](https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha3/virtualmachine_types.go#L114) to VM,  it prevents a VM from being reconciled. To align this workflow with above admin Pause VM Workflow: VM will have label [PausedVMLabelKey](https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha3/virtualmachine_types.go#L162) = `devops` and `VirtualMachineReconcileReady` condition set to false, showing reason being `VirtualMachineReconcilePaused`, indicates that VirtualMachine reconciliation is being paused by DevOps.

Fixes N/A

This is the implementation after https://github.com/vmware-tanzu/vm-operator/pull/460

```release-note
Implement admin override to pause VM reconciliation
```

### DIFF
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -211,15 +211,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		VM:      vm,
 	}
 
-	// If the VM has a pause reconcile annotation, it is being restored on vCenter. Return here so our reconcile
-	// does not replace the VM being restored on the vCenter inventory.
-	//
-	// Do not requeue the reconcile here since removing the pause annotation will trigger a reconcile anyway.
-	if _, ok := vm.Annotations[vmopv1.PauseAnnotation]; ok {
-		vmCtx.Logger.Info("Skipping reconcile since Pause annotation is set on the VM")
-		return ctrl.Result{}, nil
-	}
-
 	patchHelper, err := patch.NewHelper(vm, r.Client)
 	if err != nil {
 		return ctrl.Result{}, errors.Wrapf(err, "failed to init patch helper for %s", vmCtx.String())
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/prober"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
+	vspherevm "github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/virtualmachine"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
 	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 )
@@ -265,12 +266,24 @@ func requeueDelay(ctx *pkgctx.VirtualMachineContext) time.Duration {
 func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineContext) (reterr error) {
 	ctx.Logger.Info("Reconciling VirtualMachine Deletion")
 
+	// If the VM reconciliation has been paused by the developer,
+	// skip deletion and return.
+	if _, exists := ctx.VM.Annotations[vmopv1.PauseAnnotation]; exists {
+		ctx.Logger.Info("VirtualMachine is not deleted because devops paused this VM")
+		return nil
+	}
+
 	if controllerutil.ContainsFinalizer(ctx.VM, finalizerName) {
 		defer func() {
 			r.Recorder.EmitEvent(ctx.VM, "Delete", reterr, false)
 		}()
 
 		if err := r.VMProvider.DeleteVirtualMachine(ctx, ctx.VM); err != nil {
+			// If VM can not be deleted due to reconciliation being paused, ignore that.
+			if errors.Is(err, vspherevm.ErrorVMPausedByAdmin()) {
+				ctx.Logger.Info("VM could not be deleted since it contains the pause reconcile ExtraConfig key")
+				return nil
+			}
 			ctx.Logger.Error(err, "Failed to delete VirtualMachine")
 			return err
 		}
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -1046,50 +1046,46 @@ func (s *Session) UpdateVirtualMachine(
 		return err
 	}
 
-	// If VM is being paused, only try to update status once.
-	if paused := updateVMPaused(vmCtx, moVM); paused {
-		vmCtx.Logger.Info("Pausing Updating VirtualMachine")
-		if updateErr := vmlifecycle.UpdateStatus(vmCtx, s.K8sClient, vcVM, moVM); updateErr != nil {
-			vmCtx.Logger.Error(updateErr, "Updating VM status failed when VirtualMachine is paused")
-		}
-		return nil
-	}
-
-	// Translate the VM's current power state into the VM Op power state value.
-	var existingPowerState vmopv1.VirtualMachinePowerState
-	switch moVM.Summary.Runtime.PowerState {
-	case vimtypes.VirtualMachinePowerStatePoweredOn:
-		existingPowerState = vmopv1.VirtualMachinePowerStateOn
-	case vimtypes.VirtualMachinePowerStatePoweredOff:
-		existingPowerState = vmopv1.VirtualMachinePowerStateOff
-	case vimtypes.VirtualMachinePowerStateSuspended:
-		existingPowerState = vmopv1.VirtualMachinePowerStateSuspended
-	}
-
 	var refetchProps bool
 	var err error
 
-	switch vmCtx.VM.Spec.PowerState {
-	case vmopv1.VirtualMachinePowerStateOff:
-		refetchProps, err = s.updateVMDesiredPowerStateOff(
-			vmCtx,
-			vcVM,
-			moVM,
-			existingPowerState)
+	// Only update VM's power state when VM is not paused.
+	if !isVMPaused(vmCtx, moVM) {
+		// Translate the VM's current power state into the VM Op power state value.
+		var existingPowerState vmopv1.VirtualMachinePowerState
+		switch moVM.Summary.Runtime.PowerState {
+		case vimtypes.VirtualMachinePowerStatePoweredOn:
+			existingPowerState = vmopv1.VirtualMachinePowerStateOn
+		case vimtypes.VirtualMachinePowerStatePoweredOff:
+			existingPowerState = vmopv1.VirtualMachinePowerStateOff
+		case vimtypes.VirtualMachinePowerStateSuspended:
+			existingPowerState = vmopv1.VirtualMachinePowerStateSuspended
+		}
 
-	case vmopv1.VirtualMachinePowerStateSuspended:
-		refetchProps, err = s.updateVMDesiredPowerStateSuspended(
-			vmCtx,
-			vcVM,
-			existingPowerState)
+		switch vmCtx.VM.Spec.PowerState {
+		case vmopv1.VirtualMachinePowerStateOff:
+			refetchProps, err = s.updateVMDesiredPowerStateOff(
+				vmCtx,
+				vcVM,
+				moVM,
+				existingPowerState)
 
-	case vmopv1.VirtualMachinePowerStateOn:
-		refetchProps, err = s.updateVMDesiredPowerStateOn(
-			vmCtx,
-			vcVM,
-			moVM,
-			getUpdateArgsFn,
-			existingPowerState)
+		case vmopv1.VirtualMachinePowerStateSuspended:
+			refetchProps, err = s.updateVMDesiredPowerStateSuspended(
+				vmCtx,
+				vcVM,
+				existingPowerState)
+
+		case vmopv1.VirtualMachinePowerStateOn:
+			refetchProps, err = s.updateVMDesiredPowerStateOn(
+				vmCtx,
+				vcVM,
+				moVM,
+				getUpdateArgsFn,
+				existingPowerState)
+		}
+	} else {
+		vmCtx.Logger.Info("VirtualMachine is paused. PowerState is not updated.")
 	}
 
 	if refetchProps {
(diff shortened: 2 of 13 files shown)


controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -211,15 +211,6 @@ func (r *Reconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Resu
 		VM:      vm,
 	}
 
-	// If the VM has a pause reconcile annotation, it is being restored on vCenter. Return here so our reconcile
-	// does not replace the VM being restored on the vCenter inventory.
-	//
-	// Do not requeue the reconcile here since removing the pause annotation will trigger a reconcile anyway.
-	if _, ok := vm.Annotations[vmopv1.PauseAnnotation]; ok {
```
bryanv, line 218: Previously this annotation also prevented ReconcileDelete() from being called. Is that something we intended to remove in this MR?

zyiyi11: Good point. I will check PauseAnnotation when calling ReconcileDelete. But in the case where admin adds PauseEC key and devops deletes the VM soon after, we can't prevent VM getting deleted, right?

sreyasn: Hmm. interesting. What is the expected workflow for the case you're describing above, @zyiyi11 ?. Who overrides who?.

From what I see in the implementation, the admin's control to pause reconcile is more like preventing any updates to the VM since it's via the ExtraConfig field set on the VM.  
- Tangential question: Talking about adding the extraConfig fields to the VM, how does the admin set this EC?. I guess this can happen via the VM Class or directly via the OVF?

Earlier the devops could outright pause all reconciles ( create/update/delete) since it was an annotation and the check was here in the controller.

dilyar85: > how does the admin set this EC?

@sreyasn VI Admins can set it in the VC UI or via govc as well.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1046,6 +1046,15 @@ func (s *Session) UpdateVirtualMachine(
...
 	}
 
+	// If VM is being paused, only try to update status once.
+	if paused := updateVMPaused(vmCtx, moVM); paused {
+		vmCtx.Logger.Info("Pausing Updating VirtualMachine")
+		if updateErr := vmlifecycle.UpdateStatus(vmCtx, s.K8sClient, vcVM, moVM); updateErr != nil {
+			vmCtx.Logger.Error(updateErr, "Updating VM status failed when VirtualMachine is paused")
+		}
+		return nil
+	}
```
dilyar85, line 1056: I would rename the `updateVMPaused` function to `isVMPaused` and simply return "admin/devops/both" or an empty string if the VM is not paused. Then, inside the `if` statement here, update the VM's labels accordingly.

Besides, I don't think the `UpdateStatus()` call is necessary here, as it goes through all VM's status fields. Instead, you could simply call `conditions.MarkFalse/MarkTrue` based on the returned value from the above function.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1099,3 +1108,38 @@ func (s *Session) UpdateVirtualMachine(
...
+
+	if _, ok := vm.Annotations[vmopv1.PauseAnnotation]; ok {
+		devopsPaused = true
+	}
+
+	// Source of truth is EC and Annotation
+	// No need to check previous label val
+	switch {
+	case adminPaused && devopsPaused:
+		vm.Labels[vmopv1.PausedVMLabelKey] = "both"
```
dilyar85, line 1136: Check if the vm.Labels is `nil` before writing to avoid panic.

controllers/virtualmachine/virtualmachine_controller.go
zyiyi11: On a second thought, I will add check on `PausedVMLabelKey` when triggering ReconcileDelete. So in this case, devops and admins can pause delete reconciles.

sreyasn: > But in the case where admin adds PauseEC key and devops deletes the VM soon after, we can't prevent VM getting deleted, right

@zyiyi11 , For this, the only way you could reliably prevent deletion from a devOps user when the EC was added by the Admin later is to check the EC directly in DeleteVirtualMachine(). Since the conditions VirtualMachineReconcileReady=False + PausedReason may not be updated by the time the devOps user tries the delete.

dilyar85: > do we allow 'Edit VM settings' to add Advanced Params/ExtraConfig via the VC UI for a VM Service VM?. hmm. never tried it via govc on a VM Service VM, that's probably possible.

@sreyasn Yep, it's doable in the UI once the VM is powered off. And there is also a warning prompt stating that the VM is managed by WCP. For govc, there is no restriction or warning prompt AFAIK.

pkg/providers/vsphere/session/session_vm_update.go
bryanv: I think we'd still want to keep the VM Status up to date. The "only try to update status once" comment doesn't make sense to me. 

I've got some larger changes planned to this area but I'd like to keep the single call to `UpdateStatus()` at the end of this function. Something like below would work and is still pretty readable. 

```
if !paused {
   switch vmCtx.VM.Spec.PowerState {
```

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1099,3 +1108,38 @@ func (s *Session) UpdateVirtualMachine(
...
+func updateVMPaused(
+	vmCtx context.VirtualMachineContext,
+	moVM *mo.VirtualMachine) bool {
+
+	vm := vmCtx.VM
+	var adminPaused, devopsPaused bool
+
+	for i := range moVM.Config.ExtraConfig {
+		if o := moVM.Config.ExtraConfig[i].GetOptionValue(); o != nil {
+			if o.Key == vmopv1.PauseVMExtraConfigKey && o.Value == constants.ExtraConfigTrue {
```
bryanv, line 1121: We can break whenever/if we find the key and `adminPaused` can be set depending on the value.

pkg/providers/vsphere/session/session_vm_update.go
zyiyi11: renamed the `updateVMPaused` function to `isVMPaused`. But I still decide to do label update inside the func. Returning bool for `isVMPaused` is more straightforward and can be used in later.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1099,3 +1104,44 @@ func (s *Session) UpdateVirtualMachine(
...
 }
+
+func isVMPaused(
+	vmCtx context.VirtualMachineContext,
+	moVM *mo.VirtualMachine) bool {
+
+	vm := vmCtx.VM
+	var adminPaused, devopsPaused bool
+
+	for i := range moVM.Config.ExtraConfig {
```
sreyasn, line 1115: can reuse isPausedByAdmin() here too?. and move isPausedByAdmin to a util directory

controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -234,9 +226,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
 
+	// When PausedVMLabelKey exists, VM is being paused from deletion.
 	if !vm.DeletionTimestamp.IsZero() {
-		err = r.ReconcileDelete(vmCtx)
-		return ctrl.Result{}, err
+		if _, exists := vm.Labels[vmopv1.PausedVMLabelKey]; !exists {
+			err = r.ReconcileDelete(vmCtx)
+			return ctrl.Result{}, err
+		}
+		vmCtx.Logger.Info("ReconcileDelete not called because VirtualMachine is currently paused")
```
bryanv, line 235: This is missing a return.

Can we move this check into `ReconcileDelete()` to keep Reconcile simpler and making this easier to unit test?

controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -280,6 +276,11 @@ func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineContext) (reterr
 		}()
 
 		if err := r.VMProvider.DeleteVirtualMachine(ctx, ctx.VM); err != nil {
+			// Ignore paused error to avoid endless reconcile during pause.
+			if err.Error() == constants.VMPausedByAdminError {
```
bryanv, line 280: How about using `errors.Is()` https://pkg.go.dev/errors#Is instead here?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1046,6 +1046,11 @@ func (s *Session) UpdateVirtualMachine(
 		return err
 	}
 
+	var paused bool
+	if paused = isVMPaused(vmCtx, moVM); paused {
```
bryanv, line 1050: Maybe something more like this is more readable?

```
if !isVMPaused() {
  var existingPowerState vmopv1.VirtualMachinePowerState
  ...
  switch ...
} else {
   vmCtx.Logger.Info()
```

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1099,3 +1104,34 @@ func (s *Session) UpdateVirtualMachine(
...
+func isVMPaused(
+	vmCtx pkgctx.VirtualMachineContext,
+	moVM *mo.VirtualMachine) bool {
+
+	vm := vmCtx.VM
+
+	adminPaused := vmutil.IsPausedByAdmin(moVM)
+
+	var devopsPaused bool
+	if _, ok := vm.Annotations[vmopv1.PauseAnnotation]; ok {
```
bryanv, line 1117: `_, devopsPaused := vm.Annotations[vmopv1.PauseAnnotation]` instead?

pkg/providers/vsphere/virtualmachine/delete.go
```diff
@@ -1,22 +1,37 @@
...
 	vcVM *object.VirtualMachine) error {
 
+	moVM := &mo.VirtualMachine{}
+	if err := vcVM.Properties(vmCtx, vcVM.Reference(), []string{"config.extraConfig"}, moVM); err != nil {
+		vmCtx.Logger.Error(err, "failed to get VM properties for DeleteVirtualMachine")
+		return err
+	}
+	// Throw an error to distinguish from successful deletion.
+	if paused := vmutil.IsPausedByAdmin(moVM); paused {
+		vmCtx.VM.Labels[vmopv1.PausedVMLabelKey] = "admin"
```
bryanv, line 32: Check for nil `VM.Labels` ?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1099,3 +1104,34 @@ func (s *Session) UpdateVirtualMachine(
...
+		devopsPaused = true
+	}
+
+	if vm.Labels == nil {
```
bryanv, line 1121: Maybe also check `adminPaused || devopsPaused` here since most of the the VMs won't be paused. 

Lots of ways to reduce a few duplicated bool checks, maybe like

```
if adminPaused || devopsPaused {
  if vm.labels == nil {
  }
  switch ... {	
  }
  return true
}
	
delete(vm.Labels, vmopv1.PausedVMLabelKey)
return false
```

pkg/providers/vsphere/virtualmachine/delete.go
```diff
@@ -1,22 +1,40 @@
...
 	vmutil "github.com/vmware-tanzu/vm-operator/pkg/util/vsphere/vm"
 )
 
 func DeleteVirtualMachine(
 	vmCtx pkgctx.VirtualMachineContext,
 	vcVM *object.VirtualMachine) error {
 
+	moVM := &mo.VirtualMachine{}
+	if err := vcVM.Properties(vmCtx, vcVM.Reference(), []string{"config.extraConfig"}, moVM); err != nil {
+		vmCtx.Logger.Error(err, "failed to get VM properties for DeleteVirtualMachine")
```
aruneshpa, line 27:
```suggestion
		vmCtx.Logger.Error(err, "failed to fetch config.extraConfig properties of VM for DeleteVirtualMachine")
```

Basically, some way to call out "what property were you trying to fetch".  Could be helpful in debugging if this call fails.

pkg/providers/vsphere/vmlifecycle/update_status_test.go
```diff
@@ -894,6 +894,74 @@ var _ = Describe("VSphere Bootstrap Status to VM Status Condition", func() {
...
+				},
+			}
+		})
+
+		JustBeforeEach(func() {
+			vmlifecycle.MarkReconciliationCondition(vm)
+		})
+
+		Context("PausedVMLabel is non-existent", func() {
+			It("sets condition true", func() {
```
aruneshpa, line 916: Here and below:
```suggestion
			It("sets VirtualMachineReconcileReady condition to true", func() {
```

pkg/util/vsphere/vm/vm.go
```diff
@@ -22,3 +25,18 @@ func ManagedObjectFromMoRef(moRef vimtypes.ManagedObjectReference) mo.VirtualMac
...
+
+func IsPausedByAdmin(moVM *mo.VirtualMachine) bool {
+	for i := range moVM.Config.ExtraConfig {
+		if o := moVM.Config.ExtraConfig[i].GetOptionValue(); o != nil {
+			// Break the loop when EC key is found.
+			if o.Key == vmopv1.PauseVMExtraConfigKey {
+				if o.Value == constants.ExtraConfigTrue {
+					return true
+				}
+				break
```
aruneshpa, line 37: nit: you could also do `return false` here as well.  And remove the comment on L32.

aruneshpa: It looks like the EC value is expected to be case sensitive?  IOW, we expect admins to specify "TRUE".  Anything else ("true, "True") won't work.  Is that documented somewhere in the API?  If not, we should call that out here: https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha3/virtualmachine_types.go#L154

controllers/virtualmachine/virtualmachine_controller.go
zyiyi11: Didn't return on purpose. So that when label is changed, it can be reconciled to update.
Moved to `ReconcileDelete()`, checking for PauseAnnotation instead of label. So that update on admin removing EC can be reconciled.
```
func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineContext) (reterr error) {
	ctx.Logger.Info("Reconciling VirtualMachine Deletion")

	// If the VM reconciliation has been paused by the developer,
	// skip deletion and return.
	if _, exists := ctx.VM.Annotations[vmopv1.PauseAnnotation]; exists {
		ctx.Logger.Info("VirtualMachine is not deleted because devops paused this VM")
		return nil
	}

	if controllerutil.ContainsFinalizer(ctx.VM, finalizerName) {
		defer func() {
			r.Recorder.EmitEvent(ctx.VM, "Delete", reterr, false)
		}()

		if err := r.VMProvider.DeleteVirtualMachine(ctx, ctx.VM); err != nil {
			// If VM can not be deleted due to reconciliation being paused, ignore that.
			if errors.Is(err, vspherevm.ErrorVMPausedByAdmin()) {
				ctx.Logger.Info("VM could not be deleted since it contains the pause reconcile ExtraConfig key")
				return nil
			}
			ctx.Logger.Error(err, "Failed to delete VirtualMachine")
			return err
		}

		controllerutil.RemoveFinalizer(ctx.VM, finalizerName)
		ctx.Logger.Info("Provider Completed deleting Virtual Machine", "time", time.Now().Format(time.RFC3339))
	}

	// BMV: Shouldn't these be in the ContainsFinalizer block?
	r.vmMetrics.DeleteMetrics(ctx)
	r.Prober.RemoveFromProberManager(ctx.VM)

	ctx.Logger.Info("Finished Reconciling VirtualMachine Deletion")
	return nil
}
```

aruneshpa (approved):
LGTM, with one comment.

PR #470, opened 2024-04-15
✨ Introduce vm.spec.image reference

akutz:

This patch introduces the field `spec.image` to the `VirtualMachine` API. This new field allows users to explicitly specify the reference to the `VirtualMachineImage` or `ClusterVirtualMachineImage` resource from which a VM is deployed. Additionally, the existing field `spec.imageName` is no longer mutated in place if the value of the field is an image's display name. Instead, the resolved image reference is mutated into `spec.image`.

This patch also removes `status.image` from the `VirtualMachine` API since it is now redundant.

Fixes `NA`

```release-note
Introduce VirtualMachine spec.image for explicitly specifying image used to deploy VM.
```


webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -310,24 +311,74 @@ func SetDefaultPowerState(
...
+			if !apierrors.IsNotFound(err) {
+				return false, err
+			}
+
+			if err := c.Get(
+				ctx,
+				client.ObjectKey{Name: imgName},
+				&vmopv1.ClusterVirtualMachineImage{}); err != nil {
+
+				return false, err
```
dilyar85, line 358: We might want to add some additional error messages here for better clarity. Otherwise, if a user specifies a non-existing "vmi-" image name, expecting it to be a namespace scope image, they would always see "failed to get ClusterVirtualMachineImage" returned from the current `err`.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -310,24 +311,74 @@ func SetDefaultPowerState(
...
 	c client.Client,
 	vm *vmopv1.VirtualMachine) (bool, error) {
-	// Return early if the VM image name is empty or already set to a vmi name.
-	imageName := vm.Spec.ImageName
-	if imageName == "" || strings.HasPrefix(imageName, "vmi-") {
+
+	// Return early if the image reference is used.
+	if vm.Spec.Image != nil {
+		return false, nil
+	}
```
dilyar85, line 329: What if `vm.Spec.Image` and `vm.Spec.imageName` are both set, but are pointing to different images? I don't see the validation webhook checking for this case. Should we check here in the mutation webhook and return an error in that case?

api/v1alpha3/virtualmachine_types.go
```diff
@@ -208,26 +208,50 @@ const (
...
 type VirtualMachineSpec struct {
+
+	// +optional
+
+	// Image describes the reference to the VirtualMachineImage or
+	// ClusterVirtualMachineImage resource used to deploy this VM.
+	//
+	// Please note, unlike the field spec.imageName, the value of
+	// spec.image.name MUST be a Kubernetes object name.
+	Image *VirtualMachineImageRef `json:"image,omitempty"`
```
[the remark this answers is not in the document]
akutz: I have been moving them as I touch fields. This prevents the kube-specific "annotation" from showing up in GoDocs.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
akutz: Fair. It used to be more general as well. Let me address that.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
akutz: I discussed that with @bryanv , and decided that `spec.image` is the authority. If it is set, `spec.imageName` is not used at all. Let me update the API docs to reflect that.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
akutz: I was thinking of a different location that used to have a more generic error. In this instance, I think this is probably okay:

> Otherwise, if a user specifies a non-existing "vmi-" image name, expecting it to be a namespace scope image

If it is non-existing, then they won't know what type to expect. If they expected it to be namespace scoped, they could have simply provided the actual kind in `spec.image`. 

I may add a message that indicates neither a namespace nor cluster scoped resource could be found, but still, this is probably fine as-is. Still, it's a good catch.

akutz: @dilyar85 I updated the error, PTAL.

akutz:
cc @fabriziopandini @chrischdi @sbueringer @sidharthsurana Please note this change since it removes `status.image` from the `VirtualMachine` API in v1a3. The field `spec.image` now contains the same information.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
dilyar85: Thanks for addressing this.

dilyar85 (approved):
LGTM!

pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -64,52 +64,60 @@ func GetVirtualMachineImageSpecAndStatus(
...
-		// Ensure the GVK for the object is synced back into the object since
-		// the object's APIVersion and Kind fields may be used later.
-		if err := kube.SyncGVKToObject(clusterVMImage, k8sClient.Scheme()); err != nil {
-			return nil, nil, nil, err
+	case "ClusterVirtualMachineImage":
+		key.Namespace = ""
+		var img vmopv1.ClusterVirtualMachineImage
+		if objErr = k8sClient.Get(vmCtx, key, &img); objErr == nil {
+			obj, spec, status = &img, &img.Spec, &img.Status
 		}
```
bryanv, line 96: Do we want a default case here to at least set `objErr` to something non-nil?

akutz: Handled in the resend.

akutz:
I was able to successfully deploy the following VMs to a testbed with this PR:

```yaml
kind: VirtualMachine
metadata:
  name: my-vm-v1a1
  namespace: my-namespace
spec:
  className: best-effort-small
  imageName: vmi-327c136fa21919948 # jammy-server-cloudimg-amd64
  vmMetadata:
    transport: CloudInit
---

apiVersion: vmoperator.vmware.com/v1alpha2
kind: VirtualMachine
metadata:
  name: my-vm-v1a2
  namespace: my-namespace
spec:
  bootstrap:
    cloudInit: {}
  className: best-effort-small
  imageName: jammy-server-cloudimg-amd64

---

kind: VirtualMachine
metadata:
  name: my-vm-v1a3
  namespace: my-namespace
spec:
  bootstrap:
    cloudInit: {}
  className: best-effort-small
  imageName: jammy-server-cloudimg-amd64
```

```shell
$ k get -A vm -owide
NAMESPACE      NAME         POWER-STATE   CLASS               IMAGE                   PRIMARY-IP4       AGE
my-namespace   my-vm-v1a1   PoweredOn     best-effort-small   vmi-327c136fa21919948   192.168.128.131   26m
my-namespace   my-vm-v1a2   PoweredOn     best-effort-small   vmi-327c136fa21919948   192.168.128.128   26m
my-namespace   my-vm-v1a3   PoweredOn     best-effort-small   vmi-327c136fa21919948   192.168.128.127   26m
```

bryanv (approved):

api/v1alpha3/virtualmachine_types.go
```diff
@@ -208,28 +208,53 @@ const (
...
-	// with the VM image resource name. If the display name resolves to multiple
-	// or no VM images, then the mutation webhook denies the request and outputs
-	// an error message accordingly.
+	// ClusterVirtualMachineImages), then a mutation webhook updates the
+	// spec.image field with the reference to the resolved VM image. If the
+	// display name resolves to multiple or no VM images, then the mutation
+	// webhook denies the request and returns an error.
 	//
-	// +optional
+	// Please note, this field is ignored when spec.image is non-empty.
```
sreyasn, line 257: I'm okay with this. A UX question I had.  

Why 
- the ignore 'ImageName' when 'spec.image' is non-empty strategy,  instead of 
- making 'ImageName' and 'spec.Image' mutually exclusive via a validation error?. 

Wanted to know your thoughts since I'm sure you've considered this and we could have an unused ImageName in spec now.

sreyasn (approved):
Change looks good to me otherwise.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -208,28 +208,53 @@ const (
...
 	// This field may also be used to specify the display name (vSphere name) of
 	// a VirtualMachineImage or ClusterVirtualMachineImage resource. If the
 	// display name unambiguously resolves to a distinct VM image (among all
 	// existing VirtualMachineImages in the VM's namespace and all existing
-	// ClusterVirtualMachineImages), then a mutation webhook updates this field
-	// with the VM image resource name. If the display name resolves to multiple
-	// or no VM images, then the mutation webhook denies the request and outputs
-	// an error message accordingly.
+	// ClusterVirtualMachineImages), then a mutation webhook updates the
+	// spec.image field with the reference to the resolved VM image. If the
```
ericvmw, line 253: Does this mean that clients should ignore this field after the spec is mutated? I can imagine the following 3 cases:

1. user specifies spec.image, and the match spec.imageName
2. user specifies spec.image, and a mis-match spec.imageName
3. user only specified this spec.imageName

For the first 2 cases, the spec.imageName is ignored, for the 3rd one, the web hook updates the spec.image based on the imageName. After the spec is mutated, case 1 and case 3 becomes the same, which is ok as it may not be necessary to track what is the original request. Case 2 can be a little confusing as it shows 2 mis-matching image in the spec (from UX perspective), it could be avoided with a check.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -208,28 +208,53 @@ const (
 	VirtualMachinePowerOpModeTrySoft VirtualMachinePowerOpMode = "TrySoft"
 )
 
+type VirtualMachineImageRef struct {
+	// Kind describes the type of image, either a namespace-scoped
+	// VirtualMachineImage or cluster-scoped ClusterVirtualMachineImage.
+	Kind string `json:"kind"`
+
+	// Name refers to the name of a VirtualMachineImage resource in the same
+	// namespace as this VM or a cluster-scoped ClusterVirtualMachineImage.
```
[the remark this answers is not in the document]
akutz: This was on purpose. I did not want to assume folks knew the cluster prefix meant it was cluster-scoped.

api/v1alpha3/virtualmachine_types.go
akutz: > making 'ImageName' and 'spec.Image' mutually exclusive via a validation error?.

We cannot do that without mutating `spec.imageName` to `empty` once the image is resolved. One purpose of this PR is to stop mutating the field in place.

api/v1alpha3/virtualmachine_types.go
akutz: cc @sreyasn since he asked a similar question up above

The reason we do not ensure `spec.image.name` and `spec.imageName` are the same on create is because the validation webhook runs after mutation, and so if `spec.imageName` is a display name, then `spec.image.name` will always be different, even if pointing to the same image.

The only way to do the above would be to do the validation during mutation.

akutz: I think I see a way to do this. Will update the PR.

api/v1alpha3/virtualmachine_types.go
akutz: See https://github.com/vmware-tanzu/vm-operator/pull/470#discussion_r1567949461 for other question.

api/v1alpha3/virtualmachine_types.go
akutz: Okay, this is addressed.

akutz:
Okay, it's finally working as intended:

```shell
# k -n my-namespace get vm -owide -w
NAME         POWER-STATE   CLASS               IMAGE                   PRIMARY-IP4   AGE
my-vm-v1a1   PoweredOn     best-effort-small   vmi-327c136fa21919948                 116s
my-vm-v1a2   PoweredOn     best-effort-small   vmi-327c136fa21919948                 115s
my-vm-v1a3   PoweredOn     best-effort-small   vmi-327c136fa21919948                 95s
my-vm-v1a2   PoweredOn     best-effort-small   vmi-327c136fa21919948   192.168.128.130   119s
my-vm-v1a1   PoweredOn     best-effort-small   vmi-327c136fa21919948   192.168.128.129   2m9s
my-vm-v1a3   PoweredOn     best-effort-small   vmi-327c136fa21919948   192.168.128.131   2m
```

Thanks for the reviews everyone!

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -310,62 +311,60 @@ func SetDefaultPowerState(
...
 		}
-		determinedImageName = cvmiList.Items[0].Name
-	default:
-		return false, errors.Errorf("multiple VM images exist for %q in cluster scope", imageName)
+		return true, nil
 	}
 
-	if determinedImageName == "" {
-		return false, errors.Errorf("no VM image exists for %q in namespace or cluster scope", imageName)
+	// Verify that the resolved image name matches the value in spec.image.
```
akutz, line 362: cc @sreyasn @ericvmw This is where we verify the image name matches the spec.

webhooks/virtualmachine/mutation/virtualmachine_mutator_unit_test.go
```diff
@@ -292,111 +336,220 @@ func unitTestsMutating() {
...
+				It("Should not mutate anything", func() {
+					Expect(mutatedErr).ToNot(HaveOccurred())
+					Expect(wasMutated).To(BeFalse())
+					Expect(ctx.vm.Spec.Image).ToNot(BeNil())
+					Expect(ctx.vm.Spec.Image.Kind).To(Equal(vmiKind))
+					Expect(ctx.vm.Spec.Image.Name).To(Equal(nsImg1ID))
+					Expect(ctx.vm.Spec.ImageName).To(BeEmpty())
+				})
+			})
+			When("spec.imageName points to a different image", func() {
```
akutz, line 541: cc @sreyasn @ericvmw this is where we test for the code at https://github.com/vmware-tanzu/vm-operator/pull/470/files#r1569229987

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
sreyasn: Nice. Thanks @akutz !!.

ericvmw: Thanks, but it could be that the kind matches and only the name doesn't match?

ericvmw (approved):

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
akutz: Both are verified.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -310,62 +311,60 @@ func SetDefaultPowerState(
...
+	if resolvedKind != vm.Spec.Image.Kind && resolvedName != vm.Spec.Image.Name {
```
akutz, line 363: @ericvmw -- see, both are verified. We are validating that the image that is resolved from spec.imageName *must* match the image specified by spec.image.

pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -64,52 +65,104 @@ func GetVirtualMachineImageSpecAndStatus(
...
+	}
 
-		obj, spec, status = clusterVMImage, &clusterVMImage.Spec, &clusterVMImage.Status
-	} else {
-		// Ensure the GVK for the object is synced back into the object since
-		// the object's APIVersion and Kind fields may be used later.
-		if err := kube.SyncGVKToObject(vmImage, k8sClient.Scheme()); err != nil {
-			return nil, nil, nil, err
+	if objErr != nil {
+		reason := "Unknown"
```
bryanv, line 145: I did it a while go as a quick hack until we do something better but there is a `errToConditionReasonAndMessage()` in this file that's unused elsewhere in similar situations.

akutz: I will look.

akutz: I am not going to use this right now as there are reasons not in the function that I would like to return.

PR #471, opened 2024-04-16
✨ Single test target / workflow

akutz:

This patch simplifies the Makefile test targets so there is now a single test target named "test". There is also a single test workflow in GitHub, reducing the duplicate tests that would run as part of the old unit/integration test division.

This patch also updates the code coverage workflow to warn if coverage falls below the minimum instead of failing the job.

Fixes `NA`

```release-note
Simplify testing with a single Makefile target named "test"
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,7 +165,7 @@ jobs:
     - name: Build Web Console Validator
       run: make web-console-validator-only
 
-  unit-test:
+  test:
     needs:
     - verify-go-modules
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


bryanv (approved):

akutz:
> > This patch also updates the code coverage workflow to warn if coverage falls below the minimum instead of failing the job.
> 
> Why are we not failing the job if coverage drops below a threshold?

Because adding a single test to v1a2 means the package is now covered, dropping out percentage from 70 to ~50. We had no tests in v1a2 before, but I added two for the two differences from v1a2 to v1a3.

PR #475, opened 2024-04-19
🌱 Update deprecated GitHub actions

akutz:
Per [the notice](https://github.com/vmware-tanzu/vm-operator/actions/runs/8754747978?pr=474) from a previous PR's GitHub actions, this patch updates several deprecated GitHub action versions:

* actions/setup-go -- v4 --> v5
* actions/download-artifact -- v3 --> v4
* actions/upload-artifact -- v3 --> v4
* actions/cache -- v3 --> v4

**What does this PR do, and why is it needed?**

**Which issue(s) is/are addressed by this PR?** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes `NA`

**Are there any special notes for your reviewer**:

**Please add a release note if necessary**:

```release-note
Updates GitHub actions cache, setup-go, upload-artifact, download-artifact
```

PR #476, opened 2024-04-19
🌱 Combine test and coverage actions

akutz:

This patch removes the distinct GitHub action for parsing the code coverage produced by the test action. Instead, the coverage logic is now handled at the end of the test action. Previously the coverage logic needed to be a distinct step since there were two, separate test jobs. However, now those are combined into one, meaning we can remove the need to upload the coverage artifact to be processed later.

Fixes `NA`

Depends on #475 

```release-note
NONE
```


dilyar85 (approved):
LGTM!

PR #480, opened 2024-04-22
✨ GuestInfo prober fetches required props only

akutz:

This patch updates the GuestInfo prober to only fetch the properties that are part of the specified probe definition. Previously the VM's entire config.extraConfig property path was retrieved, resulting in a very expensive operation, depending on the size of the data in that property. By focusing on specific properties, the number of bytes transferred is limited to the size of the property values required to evaluate the probe. This change is possible because vSphere supports fetching property paths with indexed values. For example, instead of fetching all of config.extraConfig, an array, just to access the element with the key "hello", it is possible to fetch the property path config.extraConfig["hello"].

Additionally, this patch forgoes the use of the GoVmomi VM helper method "Properties" for retrieving the values. GoVmomi uses Golang reflection to unmarshal the result of a property retrieval, and reflection is quite expensive. Instead, the GuestInfo prober inspects the results of the retrieval directly, without unmarshaling the data into a mo.VirtualMachine.

Fixes `NA`

Currently GoVmomi's vC Sim does not support indexed property path retrieval (cc @dougm), thus testing was performed internally against a real vSphere testbed using the idgit project. I also copied the container image for this PR to a testbed and deployed the following VM:

```yaml
kind: VirtualMachine
metadata:
  name: my-vm
  namespace: my-namespace
spec:
  bootstrap:
    cloudInit: {}
  className: best-effort-small
  imageName: jammy-server-cloudimg-amd64
  readinessProbe:
    periodSeconds: 10
    guestInfo:
    - key: metadata
      value: invalid
```

Once deployed, the VM has a `Ready` condition set to `False` because the probe for `metadata` does not equal the value `invalid`:

```yaml
  conditions:
  - lastTransitionTime: "2024-04-23T14:44:23Z"
    message: ""
    reason: NotReady
    status: "False"
    type: Ready
```

I then updated the VM as such:

```yaml
```yaml
kind: VirtualMachine
metadata:
  name: my-vm
  namespace: my-namespace
spec:
  bootstrap:
    cloudInit: {}
  className: best-effort-small
  imageName: jammy-server-cloudimg-amd64
  readinessProbe:
    periodSeconds: 10
    guestInfo:
    - key: metadata
      value: '.*'
```

Then the condition `Ready` transitioned to `True`:

```yaml
  conditions:
  - lastTransitionTime: "2024-04-23T14:47:08Z"
    message: ""
    reason: "True"
    status: "True"
    type: Ready
```

```release-note
GuestInfo prober performance improvement via direct and indexed property retrieval
```

### DIFF
--- a/pkg/providers/vm_provider_interface.go
+++ b/pkg/providers/vm_provider_interface.go
@@ -22,7 +22,7 @@ type VirtualMachineProviderInterface interface {
 	PublishVirtualMachine(ctx context.Context, vm *vmopv1.VirtualMachine,
 		vmPub *vmopv1.VirtualMachinePublishRequest, cl *imgregv1a1.ContentLibrary, actID string) (string, error)
 	GetVirtualMachineGuestHeartbeat(ctx context.Context, vm *vmopv1.VirtualMachine) (vmopv1.GuestHeartbeatStatus, error)
-	GetVirtualMachineGuestInfo(ctx context.Context, vm *vmopv1.VirtualMachine) (map[string]string, error)
+	GetVirtualMachineProperties(ctx context.Context, vm *vmopv1.VirtualMachine, propertyPaths []string) (map[string]any, error)
 	GetVirtualMachineWebMKSTicket(ctx context.Context, vm *vmopv1.VirtualMachine, pubKey string) (string, error)
 	GetVirtualMachineHardwareVersion(ctx context.Context, vm *vmopv1.VirtualMachine) (vimtypes.HardwareVersion, error)
 
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/property"
 	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 	apierrorsutil "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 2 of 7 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -223,6 +224,66 @@ func (vs *vSphereVMProvider) GetVirtualMachineGuestInfo(
...
+	}
+
+	propSet := []vimtypes.PropertySpec{{Type: "VirtualMachine"}}
+	if len(propertyPaths) == 0 {
+		propSet[0].All = vimtypes.NewBool(true)
+	} else {
+		propSet[0].PathSet = propertyPaths
+	}
+
+	rep, err := property.DefaultCollector(client.VimClient()).RetrieveProperties(
```
akutz, line 255: @dougm, do you know if it is safe to invoke `RetrieveProperties` from multiple threads in the same session with the `DefaultCollector`?

dougm: Yes it is safe

pkg/providers/vm_provider_interface.go
```diff
@@ -22,7 +22,8 @@ type VirtualMachineProviderInterface interface {
 	PublishVirtualMachine(ctx context.Context, vm *vmopv1.VirtualMachine,
 		vmPub *vmopv1.VirtualMachinePublishRequest, cl *imgregv1a1.ContentLibrary, actID string) (string, error)
 	GetVirtualMachineGuestHeartbeat(ctx context.Context, vm *vmopv1.VirtualMachine) (vmopv1.GuestHeartbeatStatus, error)
-	GetVirtualMachineGuestInfo(ctx context.Context, vm *vmopv1.VirtualMachine) (map[string]string, error)
+	// GetVirtualMachineGuestInfo(ctx context.Context, vm *vmopv1.VirtualMachine) (map[string]string, error)
```
aruneshpa, line 25: Why keep this around?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -223,6 +224,66 @@ func (vs *vSphereVMProvider) GetVirtualMachineGuestInfo(
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(rep.Returnval) == 0 {
+		return nil, fmt.Errorf("no properties")
+	}
+
+	result := map[string]any{}
```
aruneshpa, line 278: Isn't the value of a propSet always expected to be a string?  Or, are you accounting for properties where the key is present but the value is empty/unset?

aruneshpa (approved):
LGTM with tiny nits.

sreyasn (approved):
Thank you for this change! @akutz

pkg/providers/vsphere/vmprovider_vm.go
sreyasn: I think the rep.ReturnVal.PropSet[] is of type _DynamicProperty_ with _AnyType_ for Val.
 
```
// The DynamicProperty data object type represents a name-value pair.
type DynamicProperty struct {
	// Path to the property.
	Name string `xml:"name" json:"name"`
	// Value of the property.
	Val AnyType `xml:"val,typeattr" json:"val"`
}
```

akutz: @sreyasn is correct.

pkg/providers/vm_provider_interface.go
akutz: I was not necessarily going to do so. I just have not removed it yet. Or rather I forgot to remove it. I'll do so now. Thanks!

PR #483, opened 2024-04-24
🌱 Generate type for VM guest OS ID

akutz:
This patch generates a type for VirtualMachineGuestOSIdentifier that can be used in a future API change. The type has kubebuilder annotations that result in an API with validation for any field of this type.

Fixes `NA`

```release-note
NONE
```


dilyar85 (approved):
LGTM. Thanks for making this change!

Makefile
```diff
@@ -266,6 +266,7 @@ generate-go: ## Generate golang sources
 		paths=github.com/vmware-tanzu/vm-operator/api/... \
 		object:headerFile=./hack/boilerplate/boilerplate.generatego.txt
 	$(MAKE) -C ./pkg/util/cloudinit/schema $@
+	go generate ./api/v1alpha3
```
dilyar85, line 269: Should we put this before running the `controller-gen` command above?

akutz: Probably :) Thanks @dilyar85 !

PR #484, opened 2024-04-24
🌱 Make LinuxPrep.HWClockIsUTC nillable

akutz:

This patch makes the LinuxPrep.HWClockIsUTC a nillable field since it is internally. This was fine as optional, but not nillable for new VMs. But for importing existing VMs, we do not want to enable OR disable their clock UTC setting if unintended. cc @yanleizhao-vmware

Fixes `NA`

```release-note
Make VirtualMachine spec.bootstrap.linuxPrep.hardwareClockIsUTC nillable
```


api/v1alpha2/virtualmachine_conversion.go
```diff
@@ -58,6 +58,34 @@ func restore_v1alpha3_VirtualMachineImage(dst, src *vmopv1.VirtualMachine) {
 	dst.Spec.ImageName = src.Spec.ImageName
 }
 
+func restore_v1alpha3_VirtualMachineBootstrapLinuxPrep(dst, src *vmopv1.VirtualMachine) {
```
akutz, line 61: @bryanv Unlike the previous change, I think we should just bite the bullet and update the v1a2 type to be nillable as well. It is the same format on-the-wire, and it can only result in compile-time issues for folks using the types (unlike before where an NPE was possible). This is because it's a `bool` and not a `struct`, meaning what access method that used to work would not now fail at compile time?

akutz: Sgtm. I'll do that to avoid the potential for lossy data.

akutz: Okay, the PR is updatd.

bryanv (approved):

PR #495, opened 2024-05-02
 🌱 Store resolved Cloud-Init instance ID in spec

akutz:

This patch stores the Cloud-Init instance ID derived from either spec, the object UID, or the annotation, back into spec if the derived version is different than what is in spec. Additionally, the annotation that can overrride the instance ID is deleted once its value is stored in the spec.

This patch also updates some documentation to make the logic more clear, as well as provide historical context for why certain logic exists in the first place.

Fixes `NA`

```release-note
Store derived Cloud-Init instance ID back to spec.bootstrap.cloudInit.instanceID.
```

----
📚 Documentation preview 📚: https://vm-operator--495.org.readthedocs.build/en/495/

### DIFF
--- a/pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
+++ b/pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
@@ -49,17 +49,61 @@ func BootStrapCloudInit(
 		sshPublicKeys = strings.Join(cloudInitSpec.SSHAuthorizedKeys, "\n")
 	}
 
-	uid := cloudInitSpec.InstanceID
-	if uid == "" {
-		// InstanceID will be set to the VM bios uuid when creating new VMs.
-		// Use VM.UID for backward compatibility when InstanceID is not set.
-		uid = string(vmCtx.VM.UID)
+	// The Cloud-Init instance ID is from spec.bootstrap.cloudInit.instanceID.
+	iid := cloudInitSpec.InstanceID
+
+	if iid == "" {
+		// For new VM resources that use the Cloud-Init bootstrap provider, if
+		// the field spec.bootstrap.cloudInit.instanceID is empty when the VM
+		// resource is created, the field defaults to the value of
+		// spec.biosUUID.
+		//
+		// However, to maintain backwards compatibility with older VM resources
+		// that existed prior to the introduction of the field
+		// spec.bootstrap.cloudInit.instanceID, use the VM resource's object
+		// UID as the Cloud-Init instance ID if the value of
+		// spec.bootstrap.cloudInit.instanceID is empty.
+		iid = string(vmCtx.VM.UID)
 	}
-	if value, ok := vmCtx.VM.Annotations[vmopv1.InstanceIDAnnotation]; ok {
-		uid = value
+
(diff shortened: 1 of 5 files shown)


pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
```diff
@@ -49,17 +49,55 @@ func BootStrapCloudInit(
...
+		// Cloud-Init instance ID in case it needed to be something *other* than
+		// the Kubernetes object UID.
+		//
+		// If the instance ID from the annotation is non-empty *and* is
+		// different than the current instance ID value (whether it is from
+		// spec.bootstrap.cloudInit.instanceID or metadata.uid), go ahead and
+		// use the value from the annotation.
+		iid = v
+		delete(vmCtx.VM.Annotations, vmopv1.InstanceIDAnnotation)
+	}
```
dilyar85, line 90: The restore process actually does the opposite of what these comments describe. That is, if a backed-up VM has this annotation set, the restore service **deletes** this annotation before creating the VM in a K8s cluster. As you mentioned, a new instance ID (from the new object UID) ensures Cloud-Init runs all of the per-instance modules, including the network, so that the restored VM will get the correct new IP upon restore.

The actual use case of this annotation is in the upgrade process, which sets to `iid-datasource-none`. It addresses a previous bug when a VM is deployed with Cloud-Init without user data. See this PR that initially introduced this annotation: https://github.com/vmware-tanzu/vm-operator/pull/211 (I am sure you have the context since you proposed this fix. I am just posting it here for others' reference).

Since we have `cloudInitSpec.InstanceID` now, I think we could retire this annotation after updating our internal upgrade script to use this spec to set the `iid-datasource-none` to fix the original bug.

akutz: @dilyar85 So really it's my comments that are wrong, but the logic is fine? Let me update the comments. Thank you!

akutz: @dilyar85 Can you please review the new comments for accuracy? Thanks!

dilyar85 (approved):
LGTM. Thanks for the detailed comments with full context for future reference!

pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
```diff
@@ -49,17 +49,61 @@ func BootStrapCloudInit(
...
+		// spec.bootstrap.cloudInit.instanceID, use the VM resource's object
+		// UID as the Cloud-Init instance ID if the value of
+		// spec.bootstrap.cloudInit.instanceID is empty.
+		iid = string(vmCtx.VM.UID)
 	}
-	if value, ok := vmCtx.VM.Annotations[vmopv1.InstanceIDAnnotation]; ok {
-		uid = value
+
+	if v := vmCtx.VM.Annotations[vmopv1.InstanceIDAnnotation]; v != "" && v != iid {
+		// Before the introduction of spec.bootstrap.cloudInit.instanceID,
```
sreyasn, line 70: Thank you for the detailed comment @akutz

sreyasn (approved):

PR #497, opened 2024-05-02
📖 Invert kubebuilder annotations

akutz:

This patch inverts the kubebuilder annotations for the v1alpha3 schema. The value of this is that documentation generated from the GoDocs no longer contends with the kubebuilder comments. For example, this is true for the results of "make `generate-api-docs` as well as the API documentation hosted at https://pkg.go.dev. To validate the latter locally, install the `pkgsite` command with `go install golang.org/x/pkgsite/cmd/pkgsite@latest` and run `pkgsite .` from the root of this repository.

This patch also updates the tool used to generate docs from the CRDs, crd-ref-docs, to v0.0.12.

The following screenshot shows what the API types look like from https://pkg.go.dev. On the left is the example with the inverted comment order. On the right is the original order. The comments are much easier to read on the left.

image

Fixes `NA`

```release-note
Always place kubebuilder comments/annotations before comments on API types
```

----
📚 Documentation preview 📚: https://vm-operator--497.org.readthedocs.build/en/497/


dilyar85 (approved):
LGTM. Thanks for the diligence, Andrew!

PR #498, opened 2024-05-02
🐛 Honor v1a1 VMMetadata ConfigMapName in v1a3 using annotations

### LINKED ISSUE #486: RawCloudConfig should respect bootstrap data deployed from v1a1 ConfigMap
1. Deploy a `Secret` containing some cloud-init bootstrap user data:

```
$ cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: bootstrap-data
  namespace: sdiliyaer-test
type: Opaque
stringData:
  user-data: |
    #cloud-config
    ssh_pwauth: true
    users:
      - name: test-1
        plain_text_passwd: test-1
        lock_passwd: false
        sudo: ALL=(ALL) NOPASSWD:ALL
        shell: /bin/bash
EOF

secret/bootstrap-data created
```

2. Deploy a `ConfigMap` with the same name and namespace as the above Secret but containing different user-data:

```
$ cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: bootstrap-data
  namespace: sdiliyaer-test
data:
  user-data: |
    #cloud-config
    ssh_pwauth: true
    users:
      - name: test-2
        plain_text_passwd: test-2
        lock_passwd: false
        sudo: ALL=(ALL) NOPASSWD:ALL
        shell: /bin/bash
EOF

configmap/bootstrap-data created
```

3. Deploy a `v1a1 VM` using cloud-init and the above `ConfigMap` as its bootstrap-data:

```
$ cat <<EOF | kubectl apply -f -
kind: VirtualMachine
metadata:
  name: my-vm
  namespace: sdiliyaer-test
spec:
  imageName: ubuntu-impish-21.10-cloudimg
  className: best-effort-small
  vmMetadata:
    transport: CloudInit
    configMapName: bootstrap-data
EOF

virtualmachine.vmoperator.vmware.com/my-vm created
```

The VM is booted with the user-data from the `ConfigMap` (test-2 user) as specified in the v1a1 VM YAML spec.
However, it actually boots with the user-data from the `Secret` (test-1 user), as verified by the SSH user credentials below:

```console
$ kubectl get vm -A -o wide
NAMESPACE        NAME    POWER-STATE   CLASS               IMAGE                   PRIMARY-IP4     AGE
sdiliyaer-test   my-vm   PoweredOn     best-effort-small   vmi-d88f64306b7a6a392   192.168.128.9   5m13s

$ ssh test-1@192.168.128.9
test-1@192.168.128.9's password:
Welcome to Ubuntu Impish Indri (development branch) (GNU/Linux 5.13.0-14-generic x86_64)
...
test-1@my-vm:~$ exit
logout
Connection to 192.168.128.9 closed.

$ ssh test-2@192.168.128.9
test-2@192.168.128.9's password:
Permission denied, please try again.
```

As discussed internally with the team:
> In the up conversion lets add an annotation to indicate to use a CM during reconcile. And on down conversion use the annotation to fill in the right field in https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha1/virtualmachine_conversion.go#L209

> And mark ConfigMapName as deprecated in the v1a1 api

sreyasn:

Our initial implementation involved a configMap fallback when the secret for the VM was not present. This can cause issues when a secret with the same name is present in the same namespace. Therefore, this change populates an annotation during up conversion to handle cases where v1a3 needs to use a configMap instead of a secret. On down conversion back to v1a1, the ConfigMapName field is set based on existance of the same annotation.

Fixes #486 

Testing done on supervisor -- Deployed a VM in a namespace having both a secret and configMap of the same name. 
v1a1 VM referencing the configMap used the same to bootstrap instead of the secret. Also verified that the annotation exists on the v1a3 VM and not on the v1a1 VM on down conversion.

```
Honors v1a1 VMMetadata ConfigMapName field in v1a3 using annotations since v1a3 does not directly support bootstrapping via configMap transport resource
```


api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -207,19 +207,27 @@ func convert_v1alpha1_VmMetadata_To_v1alpha3_BootstrapSpec(
 }
 
 func convert_v1alpha3_BootstrapSpec_To_v1alpha1_VmMetadata(
-	in *vmopv1.VirtualMachineBootstrapSpec) *VirtualMachineMetadata {
+	btSpec *vmopv1.VirtualMachineBootstrapSpec, annotations map[string]string) *VirtualMachineMetadata {
```
bryanv, line 210: Nit: Typical convention is usually `in` & `out` (if present) for the var names in the conversion funcs

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -207,19 +207,27 @@ func convert_v1alpha1_VmMetadata_To_v1alpha3_BootstrapSpec(
...
+	if btSpec == nil || apiequality.Semantic.DeepEqual(*btSpec, vmopv1.VirtualMachineBootstrapSpec{}) {
 		return nil
 	}
 
-	// TODO: nextver only has a Secret bootstrap field so that's what we set in v1a1. If this was created
-	// as v1a1, we need to store the serialized object to know to set either the ConfigMap or Secret field.
+	// If this was created as v1a1, we can use the V1alpha1ConfigMapTransportAnnotation to know to set the ConfigMap field.
+	// nextVersions only support Secrets for bootstrap.
+	isV1a1MDConfigMapSet := false
+	if v, ok := annotations[vmopv1.V1alpha1ConfigMapTransportAnnotation]; ok && v == "true" {
```
bryanv, line 219: Maybe https://pkg.go.dev/strconv#ParseBool here instead (and just ignore the error). Or just have the presence of the key there be enough?

Thoughts are reducing the `isV1a1MDConfigMapSet` checks with like:

```
var secretName string
if cloudInit := in.CloudInit; cloudInit != nil {
  secretName =  cloudInit.RawCloudConfig.Name
...
}

if secretName != nil {
  // comment comment comment
  if _, ok := annotations[vmopv1.V1alpha1ConfigMapTransportAnnotation]; !ok {
     out.SecretName = secretName
  } else {
    out.ConfigMapName = secretName
  }
}

return out
```

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -1096,6 +1111,22 @@ func convert_v1alpha1_PreReqsReadyCondition_to_v1alpha3_Conditions(
...
+func Convert_v1alpha3_VirtualMachine_To_v1alpha1_VirtualMachine(in *vmopv1.VirtualMachine, out *VirtualMachine, s apiconversion.Scope) error {
+	if err := autoConvert_v1alpha3_VirtualMachine_To_v1alpha1_VirtualMachine(in, out, s); err != nil {
+		return err
+	}
+
+	// v1alpha3 Bootstrap to v1alpha1 VMMetadata conversion also requires a specific annotation lookup
+	out.Spec.VmMetadata = convert_v1alpha3_BootstrapSpec_To_v1alpha1_VmMetadata(in.Spec.Bootstrap, in.GetAnnotations())
+	// Remove the v1a1 configmap vmMetadata annotation if present during down-conversion back to v1a1. This is only
+	// necessary for later versions that don't support configMap based bootstrap.
+	if _, ok := out.Annotations[vmopv1.V1alpha1ConfigMapTransportAnnotation]; ok {
```
bryanv, line 1123: Can just always do the `delete()` since it noops when the key isn't there

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -1121,6 +1152,22 @@ func Convert_v1alpha1_VirtualMachine_To_v1alpha3_VirtualMachine(in *VirtualMachi
 		}
 	}
 
+	// For existing v1a1 VMs that have metadata transport resource type set to use
+	// a configMap, set an annotation during up-conversion to indicate that the
+	// provider needs to also use the configMap. The annotation will also be
+	// used to set the respective ConfigMapName field on down-conversion to v1alpha1
+	md := in.Spec.VmMetadata
+	if md != nil && !apiequality.Semantic.DeepEqual(*md, VirtualMachineMetadata{}) && md.ConfigMapName != "" {
```
bryanv, line 1160: I think we can get rid of the `DeepEqual()` because `ConfigMapName` wouldn't be empty

api/v1alpha1/virtualmachine_conversion.go
sreyasn: nice idea!

bryanv (approved):

PR #500, opened 2024-05-03
🌱 Make VM Spec ChangeBlockTracking nillable

bryanv:

When this field is unset, we should fallback to the VM Class ConfigSpec value if set.

Update logic in UpdateConfigSpecChangeBlockTracking() to prefer the VM Spec value if set.

Fixes #

Preferring fields set in the VM Spec is the resize behavior that we want for the upcoming resize work. This lets the
DevOps user override values set in the class.

```release-note
Make VirtualMachine spec.advanced.changeBlockTracking nillable
```


api/v1alpha2/virtualmachine_types.go
```diff
@@ -448,7 +448,7 @@ type VirtualMachineAdvancedSpec struct {
 	// VMware Data Recovery.
 	//
 	// +optional
-	ChangeBlockTracking bool `json:"changeBlockTracking,omitempty"`
+	ChangeBlockTracking *bool `json:"changeBlockTracking,omitempty"`
```
sreyasn, line 451: any implications to making v1a2 changes now?. Are we only trying to prevent additions/removals from older apis after we've made v1a3 the storage version.

sreyasn (approved):
Change looks good to me!

api/v1alpha2/virtualmachine_types.go
bryanv: It doesn't change the on the wire API but, yes, it will result in a compile error. @akutz and I talked about this for https://github.com/vmware-tanzu/vm-operator/pull/484 which is basically the same situation here - we'll take the compile error to avoid a lot of mess when trying to deal with conversion issues.

PR #501, opened 2024-05-06
:bug: Upgrade VMs for use with Instance ID Discovery

dougm:

For VMs created prior to v1alpha3, populate the new fields for use with discovery and B/R:
- spec.biosUUID
- cloudInit.instanceID

### DIFF
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -447,7 +447,7 @@ type VirtualMachineSpec struct {
 	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
 
 	// +optional
-	// +kubebuilder:validation:Format:=uuid4
+	// +kubebuilder:validation:Format:=uuid
 
 	// BiosUUID describes the desired BIOS UUID for a VM.
 	// If omitted, this field defaults to a random UUID.
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -33,6 +33,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/prober"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
 	vspherevm "github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/virtualmachine"
+	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/vmlifecycle"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
 	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 )
@@ -168,9 +169,9 @@ func upgradeSchema(ctx *pkgctx.VirtualMachineContext) {
 	if bs := ctx.VM.Spec.Bootstrap; bs != nil {
 		if ci := bs.CloudInit; ci != nil {
 			if ci.InstanceID == "" {
-				ci.InstanceID = string(ctx.VM.UID)
+				iid := vmlifecycle.BootStrapCloudInitInstanceID(*ctx, ci)
 				ctx.Logger.Info("Upgrade VirtualMachine spec",
-					"bootstrap.cloudInit.instanceID", ci.InstanceID)
+					"bootstrap.cloudInit.instanceID", iid)
 			}
 		}
 	}
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -156,6 +156,26 @@ func classToVMMapperFn(ctx *pkgctx.ControllerManagerContext, c client.Client, is
 	}
 }
 
+func upgradeSchema(ctx *pkgctx.VirtualMachineContext) {
+	// If empty, this VM was created before v1alpha3 added the spec.biosUUID field.
+	if ctx.VM.Spec.BiosUUID == "" && ctx.VM.Status.BiosUUID != "" {
+		ctx.VM.Spec.BiosUUID = ctx.VM.Status.BiosUUID
+		ctx.Logger.Info("Upgrade VirtualMachine spec",
+			"biosUUID", ctx.VM.Spec.BiosUUID)
+	}
+
+	// If empty, this VM was created before v1alpha3 added the instanceID field
+	if bs := ctx.VM.Spec.Bootstrap; bs != nil {
+		if ci := bs.CloudInit; ci != nil {
+			if ci.InstanceID == "" {
+				ci.InstanceID = string(ctx.VM.UID)
+				ctx.Logger.Info("Upgrade VirtualMachine spec",
+					"bootstrap.cloudInit.instanceID", ci.InstanceID)
+			}
+		}
+	}
+}
+
 func NewReconciler(
 	ctx context.Context,
 	client client.Client,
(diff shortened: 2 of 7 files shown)


controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -156,6 +156,24 @@ func classToVMMapperFn(ctx *pkgctx.ControllerManagerContext, c client.Client, is
...
+		ctx.VM.Spec.BiosUUID = ctx.VM.Status.BiosUUID
+		ctx.Logger.Info("Upgrade VirtualMachine biosUUID")
+	}
+
+	// If empty, this VM was created before v1alpha3 added the instanceID field
+	if bs := ctx.VM.Spec.Bootstrap; bs != nil {
+		if ci := bs.CloudInit; ci != nil {
+			if ci.InstanceID == "" {
+				ci.InstanceID = string(ctx.VM.UID)
+				ctx.Logger.Info("Upgrade VirtualMachine instanceID")
```
zyiyi11, line 171: Please add more details on logging. eg,`Update VirtualMachine cloudInit instanceID to use biosUUID`

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -962,10 +962,25 @@ func (v validator) validatePowerStateOnUpdate(
 	return allErrs
 }
 
+func cloudInit(vm *vmopv1.VirtualMachine) *vmopv1.VirtualMachineBootstrapCloudInitSpec {
```
zyiyi11, line 965: Maybe change this func to return bool, `IsBootstrapCloudInit(vm) bool`. so later it can check for `true` instead of nil.

controllers/virtualmachine/virtualmachine_controller.go
dougm: Added field path and value to both logs

webhooks/virtualmachine/validation/virtualmachine_validator.go
dougm: Done

api/v1alpha3/virtualmachine_types.go
```diff
@@ -447,7 +447,7 @@ type VirtualMachineSpec struct {
 	MinHardwareVersion int32 `json:"minHardwareVersion,omitempty"`
 
 	// +optional
-	// +kubebuilder:validation:Format:=uuid4
+	// +kubebuilder:validation:Format:=uuid
```
dougm, line 450: Wondering if we should just remove this validation and add our own. Unclear when this validation is actually applied, and does look correct for uuidv4. See: https://github.com/kubernetes/kubernetes/blob/master/vendor/k8s.io/kube-openapi/pkg/validation/strfmt/default.go#L64
See also: https://github.com/kubernetes/kubernetes/issues/97733

dilyar85: IIRC this validation is applied by the K8s API Server when a CR is created or updated.
I spot-checked a few VM instance IDs and noticed that they are not actually in the format of UUID v4?

dougm: > I spot-checked a few VM instance IDs and noticed that they are not actually in the format of UUID v4?

Can you expand on that?

I haven't seen the uuid4 validation fail when creating, only when updating with this MR.

The uuid package String() format used in vm-operator is always 8-4-4-4-12: 
```
xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```
See: https://github.com/google/uuid/blob/6e10cd1027e225e3ad7bfcc13c896abd165b02ef/uuid.go#L242

The regex for `uuid4` linked to above does not match, as it is 8-4-3-3-12

But the `uuid` format matches 8-4-4-4-12

And, kubernetes itself uses the same uuid package (8-4-4-4-12 format) for MetaData.UID:
https://github.com/kubernetes/kubernetes/blob/ac30b6a0e1572145964f320ceb9a18ed2c298e68/staging/src/k8s.io/apimachinery/pkg/util/uuid/uuid.go#L26

dilyar85: I believe the K8s regex used for `uuid4` is, in fact, correct. 

> UUID4Pattern = `(?i)^[0-9a-f]{8}-?[0-9a-f]{4}-?4[0-9a-f]{3}-?[89ab][0-9a-f]{3}-?[0-9a-f]{12}$`

This is in the format of 8-4-4-4-12, with the third section starting with `4` and the fourth starting with `8`, `9`, `a`, or `b`.

The validation failed during upgrade probably because the `spec.biosUUID` field was being set from the existing `status.biosUUID` field in the `upgradeSchema()` function. However, the default instance UUID used by vSphere VM isn't UUID v4 (e.g. `503e71cd-ff35-cd68-b222-29b587ac954d`). So this update from controller was rejected by the API Server due to the `+kubebuilder:validation:Format:=uuid4` in our CRD.

To ensure backwards compatibility, we cannot have uuid4 validation for existing VMs.

dougm: Ah, thank you, my eyes failed to parse the regex. Are you ok keeping the change to `uuid` validation?

> UUIDPattern = `(?i)^[0-9a-f]{8}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{12}$`

dilyar85: Np! I am fine with keeping it as `uuid` to ensure this doesn't break existing VMs with default UUIDs that are not v4. If we want to enforce UUID v4 for new VMs, we could leverage validating webhook to check this field during new VM creation.

dilyar85 (dismissed):

controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -341,6 +361,9 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineContext) (reterr
 	// Add this VM to prober manager if ReconcileNormal succeeds.
 	r.Prober.AddToProberManager(ctx.VM)
 
+	// Upgrade schema fields where needed
+	upgradeSchema(ctx)
```
dilyar85, line 365: I wonder if we should call this function before `CreateOrUpdateVirtualMachine()` so that the VM is reconciled with the correct upgraded spec?

controllers/virtualmachine/virtualmachine_controller.go
```diff
@@ -156,6 +156,26 @@ func classToVMMapperFn(ctx *pkgctx.ControllerManagerContext, c client.Client, is
...
+			"biosUUID", ctx.VM.Spec.BiosUUID)
+	}
+
+	// If empty, this VM was created before v1alpha3 added the instanceID field
+	if bs := ctx.VM.Spec.Bootstrap; bs != nil {
+		if ci := bs.CloudInit; ci != nil {
+			if ci.InstanceID == "" {
+				ci.InstanceID = string(ctx.VM.UID)
+				ctx.Logger.Info("Upgrade VirtualMachine spec",
+					"bootstrap.cloudInit.instanceID", ci.InstanceID)
```
dilyar85, line 173: This is missing the cloud-init instance ID annotation case. Should we replicate the following logic here?

https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go#L53-L103

dougm: Right, yes. Pushed a change so we can use the same logic in both places.

controllers/virtualmachine/virtualmachine_controller.go
dougm: Seems to work as expected either way, but moved to calling it before. I had originally done it after, since `Status.BiosUUID` is set in calls that happen during CreateOrUpdateVirtualMachine()

dilyar85: That field should already be set for existing VMs. As our internal upgrade script is executed before a new VMOP controller is applied and reconciles VMs, calling this function at the beginning also simulates the same process. Thanks for addressing this.

dilyar85 (approved):
LGTM!

PR #505, opened 2024-05-08
:sparkles: Support DHCP for VPC networking

zyiyi11:
This patch supports DHCP for VPC networking. VM Operator expect SubnetPort that uses DHCP to have `Status.NetworkInterfaceConfig.MACAddress` to be empty if Subnet/SubnetSet is configured for DHCP. This is the current SubnetPort status. https://github.com/vmware-tanzu/nsx-operator/issues/559 tracks the effort to return nil ipAddress for DHCP SubnetPort.
```
networkInterfaceConfig:
    ipAddresses:
    - gateway: 172.26.1.193
      ipAddress: /26
    logicalSwitchUUID: a22d4111-e7da-4396-94b7-ee80a1c25494
```

Fixes N/A

Manual testing done on a VPC testbed:
1. Create Subnet/SubnetSet that uses DHCP (same works with DHCP + CIDR, note that only Subnet can specify CIDR)
```
apiVersion: nsx.vmware.com/v1alpha1
kind: Subnet
metadata:
  name: test-subnet-dhcp
  namespace: yiyi-ns
spec:
  DHCPConfig:
    enableDHCP: true
---
apiVersion: nsx.vmware.com/v1alpha1
kind: Subnet
metadata:
  name: test-subnet-dhcp-cidr-20
  namespace: yiyi-ns
spec:
  ipv4SubnetSize: 64
  ipAddresses:
  - 172.26.128.0/24
  accessMode: Private
  DHCPConfig:
    enableDHCP: true
```
2. Deploy VM Service VM with above subnet using CloudInit bootstrap(note that images don't have dhclient installed, not gonna work)
3. VM successfully got IP address
```
  network:
    config:
      dns:
        hostName: test-vm-ubuntu-dhcp
      interfaces:
      - ip:
          dhcp:
            ip4:
              enabled: true
        name: eth0
    interfaces:
    - deviceKey: 4000
      ip:
        addresses:
        - address: 172.26.1.195
          state: preferred
        - address: fe80::250:56ff:fe9b:1569
          state: unknown
        macAddr: 00:50:56:9b:15:69
      name: eth0
    ipStacks:
    - dns:
        domainName: .
        hostName: test-vm-ubuntu-dhcp
        searchDomains:
        - .
      ipRoutes:
      - gateway:
          address: 172.26.1.193
          device: "0"
        networkAddress: 0.0.0.0/0
      - gateway:
          device: "0"
        networkAddress: 172.26.1.192/26
    primaryIP4: 172.26.1.195
```

```release-note
Support DHCP for VPC networking
```

### DIFF
--- a/pkg/providers/vsphere/network/network.go
+++ b/pkg/providers/vsphere/network/network.go
@@ -93,7 +93,7 @@ var (
 //     but eventually we should Watch() these resources. Note though, that in the very
 //     common case, the CR is reconciled before our poll timeout, so that does save us
 //     from bailing out of the Reconcile.
-//   - Both NCP and NetOP CR Status inform us of the backing and IPAM info. However, for
+//   - NCP, NetOP and VPC CR Status inform us of the backing and IPAM info. However, for
 //     our InterfaceSpec we allow for DHCP but neither NCP nor NetOP has a way for us to
 //     mark the CR to don't do IPAM or to check DHCP is even enabled on the network. So
 //     this burns an IP, and the user must know that DHCP is actually configured.
@@ -660,25 +660,22 @@ func vpcSubnetPortToResult(
 	}
 
 	ipConfigs := []NetworkInterfaceIPConfig{}
-	if macAddr := subnetPort.Status.NetworkInterfaceConfig.MACAddress; len(macAddr) == 0 {
-		// NSX Operator's way of saying DHCP.
-	} else {
-		// IPAddresses have CIDR format.
-		for _, ipAddr := range subnetPort.Status.NetworkInterfaceConfig.IPAddresses {
-			if ipAddr.IPAddress == "" {
-				continue
-			}
 
-			ip, _, _ := net.ParseCIDR(ipAddr.IPAddress)
-			isIPv4 := ip.To4() != nil
-			ipConfig := NetworkInterfaceIPConfig{
-				IPCIDR:  ipAddr.IPAddress,
-				IsIPv4:  isIPv4,
-				Gateway: ipAddr.Gateway,
-			}
-
-			ipConfigs = append(ipConfigs, ipConfig)
+	for _, ipAddr := range subnetPort.Status.NetworkInterfaceConfig.IPAddresses {
+		// An empty ipAddress is NSX Operator's way of saying DHCP.
+		if ipAddr.IPAddress == "" {
+			continue
 		}
+		// IPAddresses have CIDR format.
+		ip, _, _ := net.ParseCIDR(ipAddr.IPAddress)
+		isIPv4 := ip.To4() != nil
+		ipConfig := NetworkInterfaceIPConfig{
+			IPCIDR:  ipAddr.IPAddress,
+			IsIPv4:  isIPv4,
+			Gateway: ipAddr.Gateway,
+		}
+
+		ipConfigs = append(ipConfigs, ipConfig)
 	}
 
 	result := &NetworkInterfaceResult{
--- a/pkg/providers/vsphere/network/network.go
+++ b/pkg/providers/vsphere/network/network.go
@@ -660,22 +660,25 @@ func vpcSubnetPortToResult(
 	}
 
 	ipConfigs := []NetworkInterfaceIPConfig{}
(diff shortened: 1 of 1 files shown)


pkg/providers/vsphere/network/network.go
```diff
@@ -660,22 +660,25 @@ func vpcSubnetPortToResult(
 	}
 
 	ipConfigs := []NetworkInterfaceIPConfig{}
+	if macAddr := subnetPort.Status.NetworkInterfaceConfig.MACAddress; len(macAddr) == 0 {
```
bryanv, line 663: I don't think there is a need for the temp `macAddr` var here.

We could prb make this a little clearer by something like so we don't have an empty if block.

```
// An empty MAC address is NSX ...
if len(subnetPort.Status.NetworkInterfaceConfig.MACAddress) != 0 {
  // IPAddresses have CIDR format.
...
}
```

Maybe save off `subnetPort.Status.NetworkInterfaceConfig.MACAddress` to temp `macAddr` var so we don't have repeat it when filling out the `NetworkInterfaceResult` but that's not too big of a deal.

That all said, why use the the MAC address to check for DHCP? Could IPs be not empty with DHCP? No IPs is a clearer indication that will be using DHCP.

zyiyi11: Only because https://github.com/vmware-tanzu/nsx-operator/issues/559. There will  change `ipAddress: /26` to nil soon. And I don't have a preference on checking macAddress or ipAddress. If you feel strongly on checking ipAddress, then I will wait for above MR to be merged in nsx-operator first.

bryanv: Let's wait for that MR and use `IpAddresses` instead

pkg/providers/vsphere/network/network.go
```diff
@@ -660,22 +660,27 @@ func vpcSubnetPortToResult(
 	}
 
 	ipConfigs := []NetworkInterfaceIPConfig{}
+	ipAddresses := subnetPort.Status.NetworkInterfaceConfig.IPAddresses
+
+	// For now, we have an assumption that one subnetport only have one IP address.
+	// An empty ipAddress is NSX Operator's way of saying DHCP.
+	if len(ipAddresses) != 0 && len(ipAddresses[0].IPAddress) != 0 {
```
bryanv, line 667: I think the `len(ipAddresses[0].IPAddress) != 0` part of this can be removed because it will hit the `continue` in the loop

pkg/providers/vsphere/network/network.go
```diff
@@ -660,22 +660,27 @@ func vpcSubnetPortToResult(
 	}
 
 	ipConfigs := []NetworkInterfaceIPConfig{}
+	ipAddresses := subnetPort.Status.NetworkInterfaceConfig.IPAddresses
+
+	// For now, we have an assumption that one subnetport only have one IP address.
```
bryanv, line 665: This code though doesn't really have this assumption, right?

pkg/providers/vsphere/network/network.go
```diff
@@ -660,22 +660,27 @@ func vpcSubnetPortToResult(
...
+	if len(ipAddresses) != 0 {
+		// IPAddresses have CIDR format.
```
bryanv, line 667: This comment may make more sense near the `ParseCIDR()` call?

pkg/providers/vsphere/network/network.go
```diff
@@ -660,22 +660,27 @@ func vpcSubnetPortToResult(
 	}
 
 	ipConfigs := []NetworkInterfaceIPConfig{}
+	ipAddresses := subnetPort.Status.NetworkInterfaceConfig.IPAddresses
+
+	// For now, we have an assumption that one subnetport only have one IP address.
+	if len(ipAddresses) != 0 {
```
bryanv, line 666: This could be removed too since the only thing in this block is the `for` loop

bryanv (approved):
Just a few minor comments to clean up

pkg/providers/vsphere/network/network.go
zyiyi11: Nope.

PR #508, opened 2024-05-08
🌱 Remove "A2" suffix from dummy object function names

bryanv:

I used the A2 suffix during the v1a2 work where we had to deal with both v1a1 and v1a2 at the same time. That is behind us so clean up the names.

Move all the functions into their own file since they are not really util'ish.

Fixes #

```release-note
NONE
```


dilyar85 (approved):
LTGM! Thanks for cleaning these up, BV.

PR #510, opened 2024-05-09
✨ Support for classless VirtualMachine resources

akutz:

This patch introduces support managing VirtualMachine resources that were not originally deployed by VM Service, and therefore do not initially have an associated VirtualMachineClass.

Fixes `NA`

Depends on #516.

```release-note
Support for managing VirtualMachine resources sans valid value for spec.className as long as VM already exists in underlying platform.
```


pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -46,24 +49,81 @@ func errToConditionReasonAndMessage(err error) (string, string) {
...
-	return vmClass, nil
+	return obj, nil
+}
+
+func getVirtualMachineClassFromVM(
+	vmCtx pkgctx.VirtualMachineContext) (vmopv1.VirtualMachineClass, error) {
+
+	obj := vmopv1.VirtualMachineClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: vmCtx.VM.Spec.ClassName,
```
dilyar85, line 82: IIUC, this function will be called when the VM spec doesn't have a class name (i.e. `IsClasslessVM()` returns true). So I am not sure what the point of setting an empty name here is. Did you mean to set the namespace instead? If not, should we consider adding the namespace here as well?

dilyar85, line 82: IIUC, this function will be called when the VM spec doesn't have a class name (i.e. IsClasslessVM() returns true). So I am not sure what the point of setting an empty name here is. Did you mean to set the namespace instead? If not, should we consider adding the namespace here as well?

akutz: This is when it could have been "Imported" as well as empty. I can probably remove this, thanks.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -362,8 +363,21 @@ func (v validator) validateImageOnCreate(ctx *pkgctx.WebhookRequestContext, vm *
...
-	if vm.Spec.ClassName == "" {
-		allErrs = append(allErrs, field.Required(field.NewPath("spec", "className"), ""))
+	f := field.NewPath("spec", "className")
+
+	switch {
+	case vmopv1util.IsClasslessVM(*vm) &&
+		pkgcfg.FromContext(ctx).Features.VMImportNewNet:
+
+		// Restrict creating classless VM resources to privileged users.
+		if !ctx.IsPrivilegedAccount {
```
sreyasn, line 373: QQ: Is allowing admin users to create these 'sans-name' classes part of the import design?. Do we need it since vmop will synthesize one internally and not really 'create' the resource from the looks of this implementation.

akutz: It is part of the design as the Mobility service account is a priv user. They will be able to create VM resources with Instance / BIOS UUIDs that match existing VMs. Come to think of it, should we add some logic in the findVm code that insists a VM found by anything other than its inventory path is actually a VM Service VM? Whether that is insisting any VMs we find must have their ManagedBy set to "VM Service" or some other measure, how do we prevent DevOps users from setting `spec.biosUUID` to a value of an existing VM somewhere in vC, outside vCenter? Should we play it safe and make `spec.biosUUID` and the proposed `spec.instanceUUID` privileged fields?

cc @bryanv @aruneshpa

sreyasn: I thought `spec.biosUUID` was already going to be a privileged field ?. only that it was recently removed for revisiting -- https://github.com/vmware-tanzu/vm-operator/pull/494.

> how do we prevent DevOps users from setting spec.biosUUID to a value of an existing VM somewhere in vC, outside vCenter? make the proposed `spec.instanceUUID` privileged fields?
Agree with playing it safe. IIUC, the problem here could be that vmop fetches a VM with a devOps specified InstanceUUID and it's not the intended VM to be used.

akutz: > only that it was recently removed for revisiting -- https://github.com/vmware-tanzu/vm-operator/pull/494.

Yeah, it is still possible to do this. We used to do it for other fields as well. You have to do the validation in the mutation webhook.

akutz: Resolved conversation and reprotected `spec.biosUUID` with https://github.com/vmware-tanzu/vm-operator/pull/518.

pkg/providers/vsphere/vcenter/getvm.go
```diff
@@ -25,18 +25,28 @@ func GetVirtualMachine(
...
+	// Find by instance UUID.
+	if id := vmCtx.VM.UID; id != "" {
+		if vm, err := findVMByUUID(vmCtx, vimClient, datacenter, string(id), true); err == nil {
 			return vm, nil
 		}
 	}
 
-	if uniqueID := vmCtx.VM.Status.UniqueID; uniqueID != "" {
-		if vm, err := findVMByMoID(vmCtx, finder, uniqueID); err == nil {
+	// Find by BIOS UUID.
```
sreyasn, line 35: Do we need a comment for @dougm to remove this and use the spec.InstanceUUID here?

pkg/util/vmopv1/vm.go
```diff
@@ -138,3 +141,71 @@ func ResolveImageName(
...
+
+	// If the ConfigSpec contained a valid hardware version, then the version
+	// the VM will use is determined by comparing the ConfigSpec version and the
+	// one from the VM's spec.minHardwareVersion field and returning the largest
+	// of the two versions.
+	if configSpecVersion.IsValid() {
+		return max(vmMinVersion, configSpecVersion)
+	}
+
+	// A VM Class with an embedded ConfigSpec should have the version set, so
```
sreyasn, line 171: would probably make this more generic for future devs without the context of the embedded configSpec and a configSpec from HW devices
	// A VM Class with an embedded ConfigSpec should have the version set, so
	// this is a ConfigSpec we created from the HW devices in the class.

sreyasn:
```suggestion
	// A VM Class with an embedded ConfigSpec will in most cases have a version set, for brownfield 
	// VM Classes where the configSpec was synthesized from the HW devices .. blah blah.. 
```
maybe?

sreyasn (approved):

pkg/providers/vsphere/vcenter/getvm.go
akutz: I do not think so. Having a fallback to BIOS UUID is not going to hurt us, and is preferable to falling all the way back to the inventory path. vC returns an error if there are multiple results, so this is fine I think.

pkg/util/vmopv1/vm.go
akutz: It was an existing comment, but I can update it, sure. Thanks!

PR #512, opened 2024-05-10
🐛 Fix requeue delay if trying to create VMs

bryanv:

We limit the number of concurrent creates to avoid blocking all the reconcile workers from processing other VMs. We use a condition to determine if we need to force a queue via the timer. When the condition is not present, IsFalse returns false, but in that situation we want to requeue. Instead use IsTrue which returns false when the condition is not present.

Fixes #

```release-note
Fix requeue delay when retrying to create VM
```


dilyar85 (approved):

akutz (approved):
Lgtm, thank you!

PR #514, opened 2024-05-10
🌱 Add integration tests to cover VM create requeue delays

bryanv:

During the VM create and VM power on steps, we may need to requeue the VM for reconciliation internally. We do that via a RequeueAfter that controller runtime provides. Improve the integration test to cover these steps.

Parameterized the requeue delays values so they can be changed if needed if needed. When VM create is not blocking, and when we start to use the property.

Fixes #

```release-note
NONE
```


akutz (approved):
Lgtm, thank you!

sreyasn (approved):

PR #518, opened 2024-05-15
 🌱 Protect VirtualMachine spec.biosUUID

akutz:

This patch protects the VirtualMachine field spec.biosUUID from being set when a new VM is created unless by a privileged user. This logic resides in the VM's mutation webhook instead of the validation webhook, since the former also defaults the value, and the latter would have no way of knowing if the default value originated from the instigating user or the mutation webhook.

Fixes  `NA`

This effectively reverts https://github.com/vmware-tanzu/vm-operator/pull/494, but in a safe manner that will not break internal, e2e tests. This is also related to https://github.com/vmware-tanzu/vm-operator/pull/510#discussion_r1600731269.

Blocked by https://github.com/vmware-tanzu/vm-operator/pull/519. Will rebase this PR once #519 is merged.

```release-note
Protect spec.biosUUID so only privileged users may set it when creating VM resources.
```


dougm (approved):
Thanks @akutz , lgtm
I'd been meaning to revisit this after having to revert in edfaa873d / #494

PR #520, opened 2024-05-15
✨ Plumb through basic end to end resize

bryanv:

This is just the minimal plumbing needed to go through resize when the VM is in steady-state off. Lots of supporting plumbing is still needed like handling the last resize annotation when the VM Class does not exist, and building the full ConfigSpec that includes expected network devices, but this lets us work on the end to end resize.

This also copies the existing getUpdateArgsFn hack that we have as a holdover from the old Session days that we'll likely need to clean up as we have time.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -19,6 +19,7 @@ import (
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
 	"github.com/vmware-tanzu/vm-operator/pkg"
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/clustermodules"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/constants"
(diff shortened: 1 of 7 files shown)


pkg/providers/vsphere/vmprovider_vm_resize_test.go
```diff
@@ -0,0 +1,171 @@
...
+	Context("Resize VM", func() {
+
+		var (
+			vm         *vmopv1.VirtualMachine
+			vmClass    *vmopv1.VirtualMachineClass
+			configSpec vimtypes.VirtualMachineConfigSpec
+		)
+
+		BeforeEach(func() {
+			vmClass = builder.DummyVirtualMachineClass("initial-class")
```
sreyasn, line 107: Probably don't need this since 116 does it?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -44,6 +46,13 @@ type VMUpdateArgs struct {
 	NetworkResults network2.NetworkInterfaceResults
 }
 
+// VMResizeArgs contains the arguments needed to resize a VM on VC.
+type VMResizeArgs struct {
+	ConfigSpec vimtypes.VirtualMachineConfigSpec
+	VMClass    *vmopv1.VirtualMachineClass
```
sreyasn, line 52: Nit: Is this needed for later in the args?

pkg/providers/vsphere/vmprovider_vm_resize_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		Context("MemoryMB", func() {
+			BeforeEach(func() {
+				configSpec.MemoryMB = 1024
+			})
+
+			It("Resizes", func() {
+				cs := configSpec
+				cs.MemoryMB = 8192
+				vm.Spec.ClassName = createVMClass(cs).Name
```
sreyasn, line 160: Interesting. Didn't know we could do `createVMClass(cs).Name` in Golang. TIL

sreyasn (approved):
LGTM.

pkg/providers/vsphere/session/session_vm_update.go
bryanv: Yes we'll need it for annotation

PR #522, opened 2024-05-15
🐛 Fix verify-codegen.sh to get exit code correctly

dilyar85:

This PR fixes a bug in the verify-codegen.sh to get the actual exit code from the executed make commands. This ensures that the verify-codegen GitHub workflow job terminates with the correct exit code when the make generate commands fail.

The previous code would always capture the exit code as 0, e.g.:

```bash
if ! false; then echo "${?}"; else echo "fine"; fi;
```

would print out "0" even though the `false` command exits with `1`.

See https://github.com/vmware-tanzu/vm-operator/actions/runs/9098850495/job/25010046314?pr=519 where the `make generate` fails with an error message printed out, but the GitHub workflow was reported as successful. 

```release-note
Fix verify-codegen.sh used in GitHub CI workflow to exit correctly.
```

### DIFF
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -10,8 +10,8 @@ set -o pipefail
 # script is located.
 cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
-if ! make generate; then
-  exit_code="${?}"
+make generate || exit_code="${?}"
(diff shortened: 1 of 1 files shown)


akutz (approved):
lgtm, thanks @dilyar85 !

PR #523, opened 2024-05-16
🐛 Fix make generate

akutz:

This patch fixes an issue with the make generate target. Specifically, "go generate ./api/...". This used to work in Go 1.21, but it no longer works across modules in Go 1.22. This patch fixes the issue.

Fixes `NA`

```release-note
NONE
```


akutz:
> Worked for me locally when running `make generate` with this change. Thanks for fixing this!

No sir, thank _you_ for finding it in the first place!

PR #526, opened 2024-05-20
🐛 Check for valid image OS image labels keys and values

bryanv:

These values come from the CL Item which may not be valid k8s label values. Don't try to assign the label if the value is invalid since we won't be able to update the k8s image.

Fixes #

```release-note
The VirtualMachineImage OS ID, Version, and Type labels will only be add if the corresponding values are valid Kubernetes label values.
```


zyiyi11 (approved):
LGTM. Thanks BV.

PR #527, opened 2024-05-20
🐛 Update govmomi to consume https://github.com/vmware/govmomi/pull/3446

sreyasn:

This change relaxes the volume controller's requirement to have a supported hardware version (for govmomi) on updates to the VM with PVC volumes. Having such a restriction has caused issues where creation of VMs with cluster default versions (> supported) were allowed but updates with PVCs were prevented because govmomi doesn't recognize the version yet.

This can be
mitigated by making sure that
the hardware version is
- parseable (vmx-*),
- valid (non-zero) and
- greater than the supported version for PVC.

Fixes # N/A

N/A

```NONE
```


akutz (approved):
Thank you!

PR #528, opened 2024-05-21
:sparkles: VirtualMachineReplicaSet CRUD/Scale Up Scale Down support

aruneshpa:

This commit adds support for creating, deleting and scaling a VirtualMachineReplicaSet resource.  When a VirtualMachineReplicaSet is created, it is reconciled to create the desired number of virtual machine replicas specified in its spec.

Similarly, during scale out/scale in, the controller reconciles the replicas to match the desired state by either creating or deleting VirtualMachine objects.  Note that a new replica is only created when the controller detect that it requires a new replica (as opposed to a rollout that is triggered via modification of the template spec in a Deployment).

When the VirtualMachineReplicaSet resource is deleted, all of its owned virtual machine replicas are also cleaned up.  The controller relies on the Kubernetes garbage collection mechanism for this.

This change also adds support for deletion policies which can used used to determine how the controller picks a virtual machine to delete from the list of replicas that it controls during a scale in/down. The only supported policy for now is "Random" where all VMs that are not in deleting state get equal weight and the controller can pick any one of them randomly.  If a VM is already marked for deletion, that gets the highest deletion priority.

In several places, I have added a TODO to revisit since ReplicaSets offer an ability to signal when an application can be considered "available".  That is dependent on a "Ready" condition that is exposed by the replicas.  Since we don't have a consistent way to do that, I have removed the support for indicating availability of a VM replica set in the Status.  Consequently, the `MinReadySeconds` field from the spec is also removed since that is used to signal availability. These can be added in a followup.

### Testing Done:

1. Create a VirtualMachineReplicaSet

2. Get the resource to ensure all replicas ready
```
k get vmreplicaset -n parunesh-ns test-rs
NAME      REPLICAS   READY   AGE
test-rs   3          3       3m57s
```

3. Scale down the replica set
```
k scale vmreplicaset -n parunesh-ns test-rs --replicas=1
virtualmachinereplicaset.vmoperator.vmware.com/test-rs scaled
```

4. Ensure scale down has cleaned up replicas:
```
k get vmreplicaset -n parunesh-ns test-rs  -o wide
NAME      REPLICAS   READY   AGE
test-rs   1          1       5m15s

k get vm -n parunesh-ns
NAME            POWER-STATE   AGE
test-rs-zmdmm   PoweredOn     3m38s
```

5. Scale up the replica set
```
k scale vmreplicaset -n parunesh-ns test-rs --replicas=2
virtualmachinereplicaset.vmoperator.vmware.com/test-rs scaled
```

6. Ensure scaled up has created new replicas
```
k get vmreplicaset -n parunesh-ns test-rs  -o wide
NAME      REPLICAS   READY   AGE
test-rs   2          2       7m1s

k get vm -n parunesh-ns
NAME            POWER-STATE   AGE
test-rs-r9slq   PoweredOn     44s
test-rs-zmdmm   PoweredOn     4m57s
```

7. Delete the replica set and ensure all replicas are deleted.
```
k delete vmreplicaset -n parunesh-ns test-rs
virtualmachinereplicaset.vmoperator.vmware.com "test-rs" deleted

k get vm -n parunesh-ns
No resources found in parunesh-ns namespace.
```

None.

```release-note
Add support for VirtualMachineReplicaSet CRUD and Scale up/down
```

### DIFF
--- a/api/v1alpha3/virtualmachinereplicaset_types.go
+++ b/api/v1alpha3/virtualmachinereplicaset_types.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024-2024 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package v1alpha3
@@ -6,6 +6,45 @@ package v1alpha3
 import (
 	vmopv1common "github.com/vmware-tanzu/vm-operator/api/v1alpha3/common"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// VirtualMachineReplicaSetReplicaFailure is added in a replica set when one of
+	// its VMs fails to be created due to insufficient quota, limit ranges,
+	// security policy, host selection, or deleted due to the host being down or
+	// finalizers are failing.
+	VirtualMachineReplicaSetReplicaFailure = "ReplicaFailure"
+)
+
+const (
+	// VirtualMachinesCreatedCondition documents that the virtual machines controlled by
+	// the VirtualMachineReplicaSet are created. When this condition is false, it indicates that
+	// there was an error when creating the VirtualMachine object.
+	VirtualMachinesCreatedCondition = "VirtualMachinesCreated"
+
+	// VirtualMachinesReadyCondition reports an aggregate of current status of
+	// the virtual machines controlled by the VirtualMachineReplicaSet.
+	VirtualMachinesReadyCondition = "VirtualMachinesReady"
+
+	// VirtualMachineCreationFailedReason documents a VirtualMachineReplicaSet failing to
+	// generate a VirtualMachine object.
+	VirtualMachineCreationFailedReason = "VirtualMachineCreationFailed"
+
+	// ResizedCondition documents a VirtualMachineReplicaSet is resizing the set of controlled VirtualMachines.
+	ResizedCondition = "Resized"
+
+	// ScalingUpReason documents a VirtualMachineReplicaSet is increasing the number of replicas.
+	ScalingUpReason = "ScalingUp"
+
+	// ScalingDownReason documents a VirtualMachineReplicaSet is decreasing the number of replicas.
+	ScalingDownReason = "ScalingDown"
+)
+
+const (
+	// VirtualMachineReplicaSetNameLabel is the key of the label applied on all the
+	// replicas VirtualMachine objects that it owns.  The value of this label is the
+	// name of the VirtualMachineReplicaSet.
+	VirtualMachineReplicaSetNameLabel = "vmoperator.vmware.com/replicaset-name"
 )
 
 // VirtualMachineTemplateSpec describes the data needed to create a VirtualMachine
@@ -34,18 +73,19 @@ type VirtualMachineReplicaSetSpec struct {
 	Replicas *int32 `json:"replicas,omitempty"`
 
 	// +optional
+	// +kubebuilder:validation:Enum=Random
 	//
-	// Minimum number of seconds for which a newly created replica virtual
-	// machine should be ready for it to be considered available.
-	// Defaults to 0 (virtual machine will be considered available as soon as it
-	// is ready).
-	MinReadySeconds int32 `json:"minReadySeconds,omitempty"`
+	// DeletePolicy defines the policy used to identify nodes to delete when downscaling.
+	// Only supported deletion policy is "Random".
+	DeletePolicy string `json:"deletePolicy,omitempty"`
 
 	// +optional
 	//
 	// Selector is a label to query over virtual machines that should match the
 	// replica count. A virtual machine's label keys and values must match in order
 	// to be controlled by this VirtualMachineReplicaSet.
+	//
+	// It must match the VirtualMachine template's labels.
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
 	Selector *metav1.LabelSelector `json:"selector"`
 
@@ -75,18 +115,12 @@ type VirtualMachineReplicaSetStatus struct {
 	// ReadyReplicas is the number of ready replicas for this VirtualMachineReplicaSet. A
 	// virtual machine is considered ready when it's "Ready" condition is marked as
 	// true.
-	Readyreplicas int32 `json:"readyReplicas,omitempty"`
+	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
 
 	// +optional
 	//
-	// AvailableReplicas is the number of available replicas (ready for at
-	// least minReadySeconds) for this VirtualMachineReplicaSet.
-	AvailableReplicas int32 `json:"availableReplicas,omitempty"`
-
 	// ObservedGeneration reflects the generation of the most recently observed
 	// VirtualMachineReplicaSet.
-	//
-	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
 	// +optional
@@ -96,14 +130,22 @@ type VirtualMachineReplicaSetStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
+func (rs *VirtualMachineReplicaSet) GetConditions() []metav1.Condition {
+	return rs.Status.Conditions
+}
+
+func (rs *VirtualMachineReplicaSet) SetConditions(conditions []metav1.Condition) {
+	rs.Status.Conditions = conditions
+}
+
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:scope=Namespaced,shortName=vmreplicaset
+// +kubebuilder:resource:scope=Namespaced,shortName=vmrs;vmreplicaset
 // +kubebuilder:storageversion
 // +kubebuilder:subresource:status
 // +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
-// +kubebuilder:printcolumn:name="Replicas",type="integer",priority=1,JSONPath=".status.replicas"
-// +kubebuilder:printcolumn:name="Ready-Replicas",type="integer",priority=1,JSONPath=".status.readyReplicas"
-// +kubebuilder:printcolumn:name="Available-Replicas",type="integer",JSONPath=".status.availableReplicas"
+// +kubebuilder:printcolumn:name="Replicas",type="integer",JSONPath=".status.replicas",description="Total number of non-terminated virtual machines targeted by this VirtualMachineReplicaSet"
+// +kubebuilder:printcolumn:name="Ready",type="integer",JSONPath=".status.readyReplicas",description="Total number of ready virtual machines targeted by this VirtualMachineReplicaSet"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Time duration since creation of VirtualMachineReplicaSet"
 
 // VirtualMachineReplicaSet is the schema for the virtualmachinereplicasets API
 type VirtualMachineReplicaSet struct {
--- a/controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
+++ b/controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
@@ -1,4 +1,20 @@
-// Copyright (c) 2019-2024 VMware, Inc. All Rights Reserved.
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package virtualmachinereplicaset
@@ -6,11 +22,16 @@ package virtualmachinereplicaset
 import (
 	"context"
 	"fmt"
+	"maps"
 	"reflect"
 	"strings"
+	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	apierrorsutil "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -22,12 +43,27 @@ import (
 	"github.com/go-logr/logr"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
-
+	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/prober"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
+	"github.com/vmware-tanzu/vm-operator/pkg/util"
+)
+
+var (
+	// stateConfirmationTimeout is the amount of time allowed to wait for desired
+	// state. This is used to ensure a VirtualMachine reaches its desired state
+	// within a time period.
+	stateConfirmationTimeout = 10 * time.Second
+
+	// stateConfirmationInterval is the amount of time between polling
+	// attempts for the desired state.
+	stateConfirmationInterval = 100 * time.Millisecond
+
+	// replicaSetKind contains the schema.GroupVersionKind for the VirtualMachineReplicaSet type.
+	replicaSetKind = vmopv1.SchemeGroupVersion.WithKind("VirtualMachineReplicaSet")
 )
 
 const (
@@ -56,22 +92,21 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		Watches(&vmopv1.VirtualMachine{},
 			handler.EnqueueRequestsFromMapFunc(r.VMToReplicaSets(ctx)),
 		).
-		// TODO: Should we also watch the volumes so we get an event for changes
-		// to a volume mounted to a VM that is part of a ReplicaSet?
 		WithOptions(controller.Options{MaxConcurrentReconciles: ctx.MaxConcurrentReconciles}).
 		Complete(r)
-
 }
 
-// VMToReplicaSets is a mapper function to be used to enqueue requests for reconciliation for VirtualMachineSetReplicaSets that might adopt a VM.
-func (r *Reconciler) VMToReplicaSets(ctx *pkgctx.ControllerManagerContext) func(_ context.Context, o client.Object) []reconcile.Request {
+// VMToReplicaSets is a mapper function to be used to enqueue requests for
+// reconciliation for VirtualMachineSetReplicaSets that might adopt a VM.
+func (r *Reconciler) VMToReplicaSets(
+	ctx *pkgctx.ControllerManagerContext) func(_ context.Context, o client.Object) []reconcile.Request {
 
-	// If the replica VM already has a controller reference, return early. If
-	// not, enqueue a reconcile if the VM matches the selector specified in the
-	// ReplicaSet. Note that we don't support adopting a replica since that
-	// might involve relocating it across compute and storage entities -- which
-	// is not supported. Instead, we emit a Condition on the ReplicaSet resource
-	// indicating it has orphaned replicas.
+	// If the replica VM already has a controller reference, return early. If not,
+	// enqueue a reconcile if the VM matches the selector specified in the
+	// ReplicaSet.
+	// Note that we don't support adopting a replica since that might
+	// involve moving it across compute and storage entities -- which may not
+	// be possible.
 	return func(_ context.Context, o client.Object) []reconcile.Request {
 		result := []ctrl.Request{}
 		vm, ok := o.(*vmopv1.VirtualMachine)
@@ -79,21 +114,19 @@ func (r *Reconciler) VMToReplicaSets(ctx *pkgctx.ControllerManagerContext) func(
 			panic(fmt.Sprintf("Expected a VirtualMachine, but got a %T", o))
 		}
 
-		// If this VM already has a controller reference, no reconciles are
-		// queued.
-		for _, ref := range vm.ObjectMeta.GetOwnerReferences() {
-			if ref.Controller != nil && *ref.Controller {
-				return nil
-			}
+		// If this VM already has a controller reference, no reconcile requests
+		// are queued.
+		if metav1.GetControllerOfNoCopy(vm) != nil {
+			return nil
 		}
 
-		vmrsss, err := r.getReplicaSetsForVM(ctx, vm)
+		vmrss, err := r.getReplicaSetsForVM(ctx, vm)
 		if err != nil {
 			ctx.Logger.Error(err, "Failed getting VM ReplicaSets for VM")
 			return nil
 		}
 
-		for _, rs := range vmrsss {
+		for _, rs := range vmrss {
 			name := client.ObjectKey{Name: rs.Name, Namespace: rs.Namespace}
 			result = append(result, ctrl.Request{NamespacedName: name})
 		}
@@ -109,7 +142,8 @@ func (r *Reconciler) getReplicaSetsForVM(
 	vm *vmopv1.VirtualMachine) ([]*vmopv1.VirtualMachineReplicaSet, error) {
 
 	if len(vm.Labels) == 0 {
-		return nil, fmt.Errorf("vm %v has no labels, This is unexpected", client.ObjectKeyFromObject(vm))
+		// VMs managed by VirtualMachineReplicaSet are always expected to have labels set.
+		return nil, nil
 	}
 
 	rsList := &vmopv1.VirtualMachineReplicaSetList{}
@@ -118,9 +152,8 @@ func (r *Reconciler) getReplicaSetsForVM(
 	}
 
 	var rss []*vmopv1.VirtualMachineReplicaSet
-	for idx := range rsList.Items {
-		rs := &rsList.Items[idx]
-
+	for _, rs := range rsList.Items {
+		rs := rs
 		selector, err := metav1.LabelSelectorAsSelector(rs.Spec.Selector)
 		if err != nil {
 			continue
@@ -131,7 +164,7 @@ func (r *Reconciler) getReplicaSetsForVM(
 		}
 
 		if selector.Matches(labels.Set(vm.Labels)) {
-			rss = append(rss, rs)
+			rss = append(rss, &rs)
 		}
 	}
 
@@ -163,6 +196,7 @@ type Reconciler struct {
 }
 
 // +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinereplicasets,verbs=create;get;list;watch;update;patch;
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinereplicasets/status,verbs=get;update;patch
 
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
@@ -180,7 +214,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 
 	patchHelper, err := patch.NewHelper(rs, r.Client)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to init patch helper for %s: %w", rsCtx, err)
+		return ctrl.Result{}, fmt.Errorf("failed to init patch helper for %s: %w", rsCtx.String(), err)
 	}
 
 	defer func() {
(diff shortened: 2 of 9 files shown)


controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +225,461 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
 
-func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineReplicaSetContext) (reterr error) {
-	if controllerutil.ContainsFinalizer(ctx.ReplicaSet, finalizerName) {
-		defer func() {
-			r.Recorder.EmitEvent(ctx.ReplicaSet, "Delete", reterr, false)
-		}()
+// MustFormatValue returns the passed inputLabelValue if it meets the standards for a Kubernetes label value.
+// If the name is not a valid label value this function returns a hash which meets the requirements.
+// Reuse the hashing used by CAPI for MachineSets.
+func MustFormatValue(str string) string {
```
akutz, line 403: Given this function is a `Must` variant, I would rather see a `FormatValue` that returns `string, error` and then have `MustFormatValue` call that, and then using `panic` if `FormatValue` returns an error.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +225,461 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+// MustEqualValue returns true if the replica set name equals either the label
+// value or its hashed value.
+func MustEqualValue(rsName, labelValue string) bool {
+	return labelValue == MustFormatValue(rsName)
+}
 
-		// Handle deletion of replicas
+// getNewMachine creates a new Machine object. The name of the newly created resource is going
+// to be created by the API server, we set the generateName field.
+func (r *Reconciler) getNewMachine(rs *vmopv1.VirtualMachineReplicaSet) *vmopv1.VirtualMachine {
```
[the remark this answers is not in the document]
akutz: I imagine he is copying from CAPV a bit? Either way, I concur that `getNewVirtualMachine` is much clearer. Or `getNewReplica` even.

api/v1alpha3/virtualmachinereplicaset_types.go
```diff
@@ -34,18 +73,19 @@ type VirtualMachineReplicaSetSpec struct {
...
+	// Only supported deletion policy is "Random".
+	DeletePolicy string `json:"deletePolicy,omitempty"`
 
 	// +optional
 	//
 	// Selector is a label to query over virtual machines that should match the
 	// replica count. A virtual machine's label keys and values must match in order
 	// to be controlled by this VirtualMachineReplicaSet.
+
+	// It must match the pod template's labels.
```
akutz, line 88: `s/pod/VirtualMachine/g`

api/v1alpha3/virtualmachinereplicaset_types.go
```diff
@@ -8,6 +8,45 @@ import (
...
+	// VirtualMachinesCreatedCondition documents that the virtual machines controlled by
+	// the VirtualMachineReplicaSet are created. When this condition is false, it indicates that
+	// there was an error when creating the VirtualMachine object.
+	VirtualMachinesCreatedCondition = "VirtualMachinesCreated"
+
+	// VirtualMachinesReadyCondition reports an aggregate of current status of
+	// the virtual machines controlled by the VirtualMachineReplicaSet.
+	VirtualMachinesReadyCondition = "VirtualMachinesReady"
+
+	// VirtualMachineCreationFailedReason (Severity=Error) documents a VirtualMachineReplicaSet failing to
```
bryanv, line 29: Our conditions no longer have a `Severity`

api/v1alpha3/virtualmachinereplicaset_types.go
bryanv, line 88: pod -> vm?

akutz (commented):
First pass

api/v1alpha3/virtualmachinereplicaset_types.go
aruneshpa: Fixed.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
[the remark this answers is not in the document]
aruneshpa: That's it.  The machine and machineset references are because a lot of the logic here is a dupe of what CAPI does.  And fixed the name of the method.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +225,461 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+
+		// MustFormatValue is used here as the value of this label will be a
+		// hash if the VirtualMachineReplicaSet name is longer than max allowed
+		// label value length of 63 characters.
+		vm.Labels[vmopv1.VirtualMachineReplicaSetNameLabel] = MustFormatValue(ctx.ReplicaSet.Name)
+
+		if err := helper.Patch(ctx, vm); err != nil {
+			return ctrl.Result{}, errors.Wrapf(err, "failed to apply %s label to Machine %q",
+				vmopv1.VirtualMachineReplicaSetNameLabel, vm.Name)
+		}
```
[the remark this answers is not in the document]
aruneshpa: Hmm.  For now, let's keep it this way?  The adopt method is for adopting replicas.  This method then operates on _all_ replicas (adopted, and others) to ensure this label is applied.  If you don't feel strongly, let's keep it this way.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +225,461 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+			errs   []error
+		)
+
+		for i := 0; i < diff; i++ {
+			vm := r.getNewMachine(rs)
+			log := ctx.Logger.WithValues("vm", vm.Name)
+
+			log.Info(fmt.Sprintf("Creating VM %d of %d", i+1, diff))
+			if err := r.Client.Create(ctx, vm); err != nil {
+				log.Error(err, "Error while creating VirtualMachine")
```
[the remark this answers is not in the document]
aruneshpa: Hmm.  We are creating a VM using `GenerateName`.  So, it is _very_ unlikely that the VM exists -- it looks like it is about 7% chance of collision for every 1M requests (https://github.com/kubernetes/kubernetes/issues/115489).  So, this is mostly to guard against API server failures, in which case, I think we would want to know about this (for metrics/telemetry)?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: So, looks like the writer exposed by the FNV hasher never returns an error.  https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/hash/fnv/fnv.go;l=106

Still, I think the `Must` analogy and pattern applies here.  Moved to `util` package and followed your suggestion.  Thanks!

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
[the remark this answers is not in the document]
akutz: Why not use [`controllerutil.CreateOrPatch`](https://github.com/kubernetes-sigs/controller-runtime/blob/2e9781e9fc6054387cf0901c70db56f0b0a63083/pkg/controller/controllerutil/controllerutil.go#L306-L422)?

dilyar85 (approved):
LGTM. Thanks for addressing all the review comments.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -57,44 +78,43 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
...
 		vm, ok := o.(*vmopv1.VirtualMachine)
 		if !ok {
 			panic(fmt.Sprintf("Expected a VirtualMachine, but got a %T", o))
 		}
 
-		// If this VM already has a controller reference, no reconciles are
-		// queued.
+		// If this VM already has a controller reference, no reconcile requests
+		// are queued.
 		for _, ref := range vm.ObjectMeta.GetOwnerReferences() {
```
bryanv, line 105: `metav1.GetControllerOfNoCopy(vm) != nil` instead?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +223,441 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+			// Note: by setting the ownerRef on creation we signal to the
+			// VirtualMachine controller that this is not a stand-alone VM.
+			OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(rs, replicaSetKind)},
+			Namespace:       rs.Namespace,
+			Labels: map[string]string{
+				vmopv1.VirtualMachineReplicaSetNameLabel: util.MustFormatValue(rs.Name),
+			},
+			Annotations: rs.Spec.Template.Annotations,
+		},
+		TypeMeta: metav1.TypeMeta{
```
bryanv, line 420: Is this required?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +223,441 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		},
+		Spec: rs.Spec.Template.Spec,
+	}
 
-		controllerutil.RemoveFinalizer(ctx.ReplicaSet, finalizerName)
+	// Copy over the labels specified in replicaset's template to the
+	// VirtualMachine. Note that a direct assignment here would share the labels
+	// and consequently, any changes to the VM label will also impact the
+	// replicaset's template labels.
+	for k, v := range rs.Spec.Template.Labels {
```
bryanv, line 431: `maps.Copy()`?

Do we make sure that `rs.Spec.Template.Labels` does not contain `VirtualMachineReplicaSetNameLabel` ?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +223,441 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+
+		if len(errs) > 0 {
+			return apierrorsutil.NewAggregate(errs)
+		}
+
+		return r.waitForMachineCreation(ctx, vmList)
+	case diff > 0:
+		ctx.Logger.Info("ReplicaSet is scaling down",
+			"currentReplicas", len(vms),
+			"desiredReplicas", *(rs.Spec.Replicas),
```
bryanv, line 495: Need `()` here?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +223,441 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		if err != nil {
+			return err
+		}
+
+		var errs []error
+		vmsToDelete := getMachinesToDeletePrioritized(vms, diff, deletePriorityFunc)
+		for i, vm := range vmsToDelete {
+			log := ctx.Logger.WithValues("vm", vm.Name)
+			if vm.GetDeletionTimestamp().IsZero() {
+				log.Info(fmt.Sprintf("Deleting VM %d of %d", i+1, diff))
```
bryanv, line 510: Can we key and value the params here?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller_intg_test.go
```diff
@@ -0,0 +1,227 @@
...
+		EventuallyWithOffset(1, func() []string {
+			if rs := getVirtualMachineReplicaSet(ctx, objKey); rs != nil {
+				return rs.GetFinalizers()
+			}
+			return nil
+		}).Should(ContainElement(finalizerName), "waiting for VirtualMachineReplicaSet finalizer")
+	}
+
+	ensureReplicas := func(ctx *builder.IntegrationTestContext, labels map[string]string, desiredReplicas int) {
+		EventuallyWithOffset(1, func() bool {
```
bryanv, line 115: Using https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode pattern for these `Eventually`s can make them a lot clearer when the assertion fails

api/v1alpha3/virtualmachinereplicaset_types.go
```diff
@@ -8,6 +8,45 @@ import (
...
+	VirtualMachinesReadyCondition = "VirtualMachinesReady"
+
+	// VirtualMachineCreationFailedReason documents a VirtualMachineReplicaSet failing to
+	// generate a VirtualMachine object.
+	VirtualMachineCreationFailedReason = "VirtualMachineCreationFailed"
+
+	// ResizedCondition documents a VirtualMachineReplicaSet is resizing the set of controlled VirtualMachines.
+	ResizedCondition = "Resized"
+
+	// ScalingUpReason (Severity=Info) documents a VirtualMachineReplicaSet is increasing the number of replicas.
```
bryanv, line 36: Two left over serverities here

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: Thanks for this.

> Do we make sure that rs.Spec.Template.Labels does not contain VirtualMachineReplicaSetNameLabel

Hmm.  I think that should be okay?  If the template labels contain the replicaset name as a label, that will also be present as a selector on the repilcaset as well -- the webhook ensures that the `rs.spec.Selector.MatchLabels == rs.Spec.Template.Labels`.  I also don't see a validation like that in upstream k8s replicaset controller, or the machineset controller in cluster-api.

pkg/util/replicaset_label_test.go
```diff
@@ -0,0 +1,61 @@
...
+	Context("input is a valid label value", func() {
+		It("returns the input label", func() {
+			inputLabel := "valid-k8s-label-value"
+			ret, err := util.FormatValue(inputLabel)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ret).To(Equal(inputLabel))
+		})
+	})
+
+	Context("input is not a valid label", func() {
```
zyiyi11, line 28: nit: To keep wording consistent with below. `input is invalid label`, `does not panic, returns a hashed label value`

pkg/util/replicaset_label_test.go
```diff
@@ -0,0 +1,61 @@
...
+			Expect(actual).To(Equal(expected))
+		})
+	})
+})
+
+var _ = Describe("MustFormatValue", func() {
+	Context("input is invalid label", func() {
+		It("does not panic, returns a hashed label value", func() {
+			inputLabel := strings.Repeat("a", 100)
+			actual, err := util.FormatValue(inputLabel)
```
zyiyi11, line 49: I believe you meant to use `MustFormatValue` here

api/v1alpha3/virtualmachinereplicaset_types.go
```diff
@@ -96,14 +130,22 @@ type VirtualMachineReplicaSetStatus struct {
...
+	return rs.Status.Conditions
+}
+
+func (rs *VirtualMachineReplicaSet) SetConditions(conditions []metav1.Condition) {
+	rs.Status.Conditions = conditions
+}
+
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:scope=Namespaced,shortName=vmreplicaset
+// +kubebuilder:resource:scope=Namespaced,shortName=vmrs
```
akutz, line 142: Multiple short name aliases are allowed if you want to retain the original as well.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		return true
+	}
+
+	return false
 }
 
-func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineReplicaSetContext) error {
+// adoptOrphan sets the VirtualMachineReplicaSet as a controller OwnerReference
+// to the VirtualMachine.
+func (r *Reconciler) adoptOrphan(
```
akutz, line 258: `adoptOrphan`?

You are going to make the VM feel bad. I mean, what if its ESXi host was *just* lost in a disconnection accident? Too soon @aruneshpa, too soon...

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+	ctx *pkgctx.VirtualMachineReplicaSetContext,
+	rs *vmopv1.VirtualMachineReplicaSet,
+	vm *vmopv1.VirtualMachine) error {
+
+	_, err := controllerutil.CreateOrPatch(ctx, r.Client, vm, func() error {
+		return controllerutil.SetControllerReference(rs, vm, r.Scheme())
```
akutz, line 264: Do replica sets set ownership on _existing_ resources that have matching labels? The K8s replicaset controller _creates_ pods with the owner ref (https://github.com/kubernetes/kubernetes/blob/6d0aab2e38364d9d8e050b5f3a120a36ee588389/pkg/controller/replicaset/replica_set.go#L598), but I see no evidence it sets the owner ref on _adopted_ pods. Maybe I am missing something?

Ah, just found it. It's this bit in the replica set controller -- https://github.com/kubernetes/kubernetes/blob/6d0aab2e38364d9d8e050b5f3a120a36ee588389/pkg/controller/replicaset/replica_set.go#L756-L757. Which then leads to this in a PodControllerRefManager -- https://github.com/kubernetes/kubernetes/blob/6d0aab2e38364d9d8e050b5f3a120a36ee588389/pkg/controller/controller_ref_manager.go#L220-L234.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		ctx.ReplicaSet.Spec.Selector.MatchLabels = make(map[string]string)
+	}
+
+	if ctx.ReplicaSet.Spec.Template.Labels == nil {
+		ctx.ReplicaSet.Spec.Template.Labels = make(map[string]string)
+	}
+
+	selectorMap, err := metav1.LabelSelectorAsMap(ctx.ReplicaSet.Spec.Selector)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err,
```
akutz, line 293: We are not using `github.com/pkg/errors` anymore. Please change this to:

```go
fmt.Errorf("failed to convert VirtualMachineReplicaSet %q label selector to a map: %w", ctx.ReplicaSet.Name, err)
```

Please see https://github.com/vmware-tanzu/vm-operator/pull/538 for more information.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+
+	// Get all VirtualMachines linked to this VirtualMachineReplicaSet.
+	allVMs := &vmopv1.VirtualMachineList{}
+	if err := r.Client.List(ctx,
+		allVMs,
+		client.InNamespace(ctx.ReplicaSet.Namespace),
+		client.MatchingLabels(selectorMap),
+	); err != nil {
+		return ctrl.Result{},
+			fmt.Errorf("failed to list virtual machines matched by the VirtualMachineReplicaSet. selector: %q, err: %v",
```
akutz, line 306: Errors should be included in `fmt.Errorf` as `%w`. Please see https://github.com/vmware-tanzu/vm-operator/pull/538 and https://go.dev/blog/go1.13-errors for more information.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		res, err := controllerutil.CreateOrPatch(ctx, r.Client, vm, func() error {
+			// MustFormatValue is used here as the value of this label will be a
+			// hash if the VirtualMachineReplicaSet name is longer than max allowed
+			// label value length of 63 characters.
+			vm.Labels[vmopv1.VirtualMachineReplicaSetNameLabel] = util.MustFormatValue(ctx.ReplicaSet.Name)
+
+			return nil
+		})
+		if err != nil {
+			return ctrl.Result{}, errors.Wrapf(err, "failed to apply %s label to VirtualMachine %q",
```
akutz, line 358: We are not using `github.com/pkg/errors` anymore. Please change this to:

```go
fmt.Errorf("failed to apply %s label to VirtualMachine %q: %w", vmopv1.VirtualMachineReplicaSetNameLabel, vm.Name, err)
```

Please see https://github.com/vmware-tanzu/vm-operator/pull/538 for more information.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+	// TODO: Handle in place propagation of fields to machines before syncing replicas.
+
+	syncErr := r.syncReplicas(ctx, ctx.ReplicaSet, filteredVMs)
+
+	// Update the status of the VirtualMachineReplicaSet even in case of error
+	// since syncing might have resulted in replicas being added or removed.
+	r.updateStatus(ctx, ctx.ReplicaSet, filteredVMs)
+
+	if syncErr != nil {
+		return ctrl.Result{}, errors.Wrapf(syncErr, "failed to sync VirtualMachineReplicaSet replicas")
```
akutz, line 382: We are not using `github.com/pkg/errors` anymore. Please change this to:

```go
fmt.Errorf("failed to sync VirtualMachineReplicaSet replicas: %w", err)
```

Please see https://github.com/vmware-tanzu/vm-operator/pull/538 for more information.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+			// Note: by setting the ownerRef on creation we signal to the
+			// VirtualMachine controller that this is not a stand-alone VM.
+			OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(rs, replicaSetKind)},
+			Namespace:       rs.Namespace,
+			Labels: map[string]string{
+				vmopv1.VirtualMachineReplicaSetNameLabel: util.MustFormatValue(rs.Name),
+			},
+			Annotations: rs.Spec.Template.Annotations,
+		},
+		// TypeMeta: metav1.TypeMeta{
```
akutz, line 420: Is this supposed to be removed?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+// value or its hashed value.
+func MustEqualValue(rsName, labelValue string) bool {
+	return labelValue == util.MustFormatValue(rsName)
+}
 
-		// Handle deletion of replicas
+// getNewVirtualMachine creates a new VirtualMachine object. The name of the newly created resource is going
+// to be created by the API server, we set the generateName field.
+func (r *Reconciler) getNewVirtualMachine(rs *vmopv1.VirtualMachineReplicaSet) *vmopv1.VirtualMachine {
+	// gv := vmopv1.SchemeGroupVersion
```
akutz, line 407: Meant to remove this?

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		// },
+		Spec: rs.Spec.Template.Spec,
+	}
 
-		controllerutil.RemoveFinalizer(ctx.ReplicaSet, finalizerName)
+	vm.Labels = getLabelsFromVMReplicaSet(rs)
+
+	vm.Annotations = getAnnotationsFromReplicaSet(rs)
+
+	// TODO: Propagate the VirtualMachineDeploymentNameLabel from VirtualMachineReplicaSet to VirtualMachines if it exists.
```
akutz, line 431: Should this `TODO` be resolved in this PR? If not, let's file a GH issue and link to it here.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+
+	return vm
+}
+
+func getLabelsFromVMReplicaSet(rs *vmopv1.VirtualMachineReplicaSet) map[string]string {
+	labels := make(map[string]string)
```
akutz, line 437: This could be optimized as such `labels := make(map[string]string, len(rs.Spec.Template.Labels)` so that it is already sized according to the needs of the copy below.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+	maps.Copy(labels, rs.Spec.Template.Labels)
+
+	// Ensure that the ReplicaSetLabel is always present.
+	labels[vmopv1.VirtualMachineReplicaSetNameLabel] = util.MustFormatValue(rs.Name)
+
+	return labels
+}
+
+func getAnnotationsFromReplicaSet(rs *vmopv1.VirtualMachineReplicaSet) map[string]string {
+	annotations := make(map[string]string)
```
akutz, line 449: This could be optimized as such `annotations := make(map[string]string, len(rs.Spec.Template.Annotations)` so that it is already sized according to the needs of the copy below.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+}
+
+// syncReplicas scales VirtualMachine resources up or down.
+func (r *Reconciler) syncReplicas(
+	ctx *pkgctx.VirtualMachineReplicaSetContext,
+	rs *vmopv1.VirtualMachineReplicaSet,
+	vms []*vmopv1.VirtualMachine) error {
+
+	if rs.Spec.Replicas == nil {
+		return errors.Errorf("the Replicas field in Spec for VirtualMachineReplicaSet %v is nil, this should not be allowed", rs.Name)
```
akutz, line 464: We are not using `github.com/pkg/errors` anymore. Please change this to:

```go
fmt.Errorf("the Replicas field in Spec for VirtualMachineReplicaSet %q is nil, this should not be allowed: %w", rs.Name, err)
```

Please see https://github.com/vmware-tanzu/vm-operator/pull/538 for more information.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+				conditions.MarkFalse(
+					rs,
+					vmopv1.VirtualMachinesCreatedCondition,
+					vmopv1.VirtualMachineCreationFailedReason,
+					err.Error(),
+				)
+				continue
+			}
+
+			log.Info(fmt.Sprintf("Created VM %d of %d", i+1, diff))
```
akutz, line 499: Logging with `Sprintf` is anti-pattern. Please change this to:

```go
log.Info("Created VM", "index", i+1, "total", diff)
```

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		var (
+			vmList []*vmopv1.VirtualMachine
+			errs   []error
+		)
+
+		for i := 0; i < diff; i++ {
+			vm := r.getNewVirtualMachine(rs)
+			log := ctx.Logger.WithValues("vm", vm.Name)
+
+			log.Info(fmt.Sprintf("Creating VM %d of %d", i+1, diff))
```
akutz, line 485: Logging with `Sprintf` is anti-pattern. Please change this to:

```go
log.Info("Creating VM", "index", i+1, "total", diff)
```

akutz, line 485: Also, given that it is logged below as well as `Created VM`, how about we pick one of these statements and log it at a lower log level so we do not spam the logs? I usually prefer to log attempt statements at the higher log level and completions at lower log levels. Because generally, if we do not see an error, we know it completed. And if we are trying to gather timings, we can turn up logging.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+
+				if err := r.Client.Delete(ctx, vm); err != nil {
+					log.Error(err, "Unable to delete VM")
+					r.Recorder.Warnf(rs, "FailedDelete", "Failed to delete VM %q: %v", vm.Name, err)
+					errs = append(errs, err)
+					continue
+				}
+				r.Recorder.Eventf(rs, "SuccessfulDelete", "Deleted VM %q", vm.Name)
+			} else {
+				log.Info(fmt.Sprintf("Waiting for VM %d of %d to be deleted", i+1, diff))
```
akutz, line 537: Logging with `Sprintf` is anti-pattern. Please change this to:

```go
log.Info("Waiting for VM to be deleted", "index", i+1, "total", diff)
```

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+				}
+				return false, err
+			}
+
+			return true, nil
+		})
+
+		if pollErr != nil {
+			ctx.Logger.Error(pollErr, "Failed waiting for VirtualMachine to be created")
+			return errors.Wrap(pollErr, "failed waiting for VirtualMachine to be created")
```
akutz, line 586: We are not using `github.com/pkg/errors` anymore. Please change this to:

```go
fmt.Errorf("failed waiting for VirtualMachine to be created: %w", err)
```

Please see https://github.com/vmware-tanzu/vm-operator/pull/538 for more information.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller_intg_test.go
```diff
@@ -0,0 +1,230 @@
...
+					if err := ctx.Client.Get(ctx, rsKey, rs); err == nil && len(rs.Finalizers) > 0 {
+						rs.Finalizers = nil
+						_ = ctx.Client.Update(ctx, rs)
+					}
+				} else {
+					Expect(apierrors.IsNotFound(err)).To(BeTrue())
+				}
+			})
+
+			// Fake client doesn't support garbage collection, so we can't verify that all replicas have been cleaned up.
```
bryanv, line 150: The fake client is not used for integration tests. It is because env-test only runs apiserver and etcd and GC is done by kube-controller-manager

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
bryanv, line 264: I wouldn't think we want to possibility of the creating a VM when this is just suppose to adopt

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: No, removed it.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: Yep, same behavior with MachineSets adopting machines that match labels specified in the selector.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: I was aware of your change.  Converted a few errors from "errors" -> "fmt" but missed a bunch :(  Thanks for pointing this.  Updating.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: Yep.  Intended to remove it after my testing.  Cleaned up.

pkg/util/replicaset_label.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	hasher := fnv.New32a()
+	if _, err := hasher.Write([]byte(str)); err != nil {
+		// At time of writing the implementation of fnv's Write function can never return an error.
+		// If a future Go version changes the implementation, this code panics.
+		return "", err
+	}
+
+	return fmt.Sprintf("hash_%s_z", base64.URLEncoding.WithPadding(base64.NoPadding).EncodeToString(hasher.Sum(nil))), nil
```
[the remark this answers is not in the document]
aruneshpa: https://github.com/kubernetes-sigs/cluster-api/pull/7763 is the context.  Basically to ensure that the label values are always valid.  There are other ways around it (e.g., ensuring that the charset used to generate the random string is valid), but I chose the CAPI way.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: I agree that it is confusing.  However, I used this because:
- the list of VMs passed down to this method are VMs that have been retrieved from the API server (using the `List` API call).  So, they already exist?
- It is _slightly_ cleaner then using `mergeFrom` and `cilent.Patch`.

LMK what you think.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller_intg_test.go
aruneshpa: TIL.  Will update the comment to reflect this.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
bryanv: It could have been deleted between the List() and here (or even slightly before that if the ctr-runtime Client cache was stale). I don't think Patch() is too clumsy here but could always roll your own smaller wrapper, or just use Update() and if the RV is different just deal with the VM on a subsequent reconcile. 

Looks like there is another use of CreateOrPatch a little later on that looks problematic.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -202,38 +220,462 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+				ctx.Logger.Error(err, "VirtualMachineReplicaSet failed to adopt VirtualMachine", "vm", vm.Name)
+				r.Recorder.Warnf(ctx.ReplicaSet, "FailedAdopt", "Failed to adopt VirtualMachine %q: %v", vm.Name, err)
+
+				continue
+			}
+			ctx.Logger.Info("Adopted VirtualMachine", "vm", vm.Name)
+			r.Recorder.Eventf(ctx.ReplicaSet, "SuccessfulAdopt", "Adopted VirtualMachine %q", vm.Name)
+		}
+
+		filteredVMs = append(filteredVMs, vm)
```
bryanv, line 336: If the VM has a _different_ controller owner do we really want to add the label to it below?

bryanv: Oh I missed what `shouldExcludeVirtualMachine()` does. But maybe just inline that function so we don't have two calls to `metav1.GetControllerOf()` right after eachother?

aruneshpa: Fixed.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: A new PR.  And there's a separate effort tracking the entire VirtualMachineDeployment effort.  Will handle this as part of that.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: Reverted to using Patch.

akutz (approved):
lgtm, thank you!

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
aruneshpa: I am going to resolve this for now.  Totally fine with revisiting this if I missed something.

PR #533, opened 2024-05-23
🌱 Import net-operator API instead of vendoring it

akutz:

This PR contains two patches:

1. The first patch updates VM Op to import the net-operator API types from https://github.com/vmware-tanzu/net-operator-api instead of vendoring the types. The CRD for the net-operator `NetworkInterface` API is still vendored as it is required for testing.

2. The second patch removes the net-op API `Vmxnet3NetworkInterface` from the list of vendorered CRDs. This patch also removes the RBAC definition for this API as this API is not used anywhere by VM Operator.

Fixes `NA`

```release-note
Import net-operator-api and no longer vendor the types in `external/net-operator`.
```


.gitignore
```diff
@@ -1,3 +1,13 @@
+# Only vendor the net-op CRD for networkinterfaces.
```
bryanv, line 1: How about this instead so we don't have to update it as/if NetOP gains CRDs?

```
config/crd/external-crds/netoperator.vmware.com_*
!config/crd/external-crds/netoperator.vmware.com_networkinterfaces.yaml
```

bryanv (approved):

.gitignore
akutz: Good call!

zyiyi11 (approved):
Thanks Andrew!

PR #538, opened 2024-05-28
🌱 Replace github.com/pkg/errors with stdlib errors

akutz:

This patch replaces the use of `github.com/pkg/errors` library with the stdlib `errors` package. The latter gained the ability to embed wrapped errors and unwrap them in Go 1.13. The only difference between the two packages is that github.com/pkg/errors also includes the stack trace.

The `Wrap` (https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/errors.go#L181-L196) and `Wrapf` (https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/errors.go#L198-L213) functions from github.com/pkg/errors would include the stack trace for up to 32 callers (https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/stack.go#L163-L169). If the error was formatted with `+v`, then the error would emit the stack trace (https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/errors.go#L165-L179).

Given that VM Operator does not rely on this feature, it makes sense to deprecate the use of `github.com/pkg/errors` for the stdlib `errors` package instead. Please refer to https://pkg.go.dev/errors for more information on how to wrap, unwrap, and assert stdlib errors.

Fixes `NA`

```release-note
Replaces `github.com/pkg/errors` with stdlib `errors`
```

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -4,7 +4,7 @@
 package controllers
 
 import (
-	"github.com/pkg/errors"
+	"fmt"
 
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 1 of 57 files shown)


aruneshpa (approved):
Assuming that this is just a straight swap.  LGTM.

bryanv (approved):

PR #539, opened 2024-05-28
🌱  Only use gopkg.in/yaml.v3, remove k8s.io/utils, add pkg/util/ptr, and use net.ParseCIDR 

akutz:

This patch further simplifies VM Operator's dependency graph:

* This patch updates VM Operator to only use the package `gopkg.in/yaml.v3` directly. The package `gopkg.in/yaml.v2` is still used indirectly via `sigs.k8s.io/yaml`, but VM Operator will no longer import or use `gopkg.in/yaml.v2` directly.

* This patch also removes the direct dependency on `k8s.io/utils`. This module was used for the [`To`](https://pkg.go.dev/k8s.io/utils/ptr#To) function from `k8s.io/utils/ptr` and [`ParseIPNets`](https://pkg.go.dev/k8s.io/utils/net#ParseIPNets) function from `k8s.io/utils/net`.

  * The [`To`](https://pkg.go.dev/k8s.io/utils/ptr#To) function and `k8s.io/utils/ptr` package as a whole has been replaced with this repository's `./pkg/utils/ptr` package. This new package includes everything the external one did as well as a few new tricks.

  * The [`ParseIPNets`](https://pkg.go.dev/k8s.io/utils/net#ParseIPNets) function from `k8s.io/utils/net` is a little trickier. Looking at its [source code](https://github.com/kubernetes/utils/blob/fe8a2dddb1d00a41a4a4132ba33cb324aa7cd71c/net/ipnet.go#L28-L41), the `ParseIPNets` function is really just a wrapper for a legacy version of `net.ParseCIDR` called [`ParseCIDRSloppy`](https://github.com/kubernetes/utils/blob/fe8a2dddb1d00a41a4a4132ba33cb324aa7cd71c/net/parse.go#L29-L33):

    ```go
    // ParseCIDRSloppy is identical to Go's standard net.ParseCIDR, except that it allows
    // leading '0' characters on numbers.  Go used to allow this and then changed
    // the behavior in 1.17.  We're choosing to keep it for compat with potential
    // stored values.
    var ParseCIDRSloppy = forkednet.ParseCIDR
    ```

    Thus `ParseCIDRSloppy` allows an IP address like `001.001.001.001` whereas the stdlib Golang `net.ParseCIDR` would reject it. While it *is* a valid representation of `1.1.1.1` (try it with `ping` if you do not believe me!), it is not possible for the LB ranges VM Op validates with this function to have leading zeros. This is because Supervisor uses Golang's `net.ParseCIDR` to validate those IP4 networks, and any of them with leading zeros would be rejected.

Fixes `NA`

```release-note
Remove direct dependencies on gopkg.in/yaml.v2 and k8s.io/utils
```


pkg/util/ptr/ptr.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if a == nil {
+		// If a is nil, then we know b is nil as well.
+		return true
+	}
+
+	return *a == *b
+}
+
+// Overwrite copies src to dst as long as src is non-nil.
```
akutz, line 43: @bryanv FYI, I created this specifically to centralize the helper function you wrote in `./pkg/util/vmopv1/resize_devops.go`:

```go
func overridePtrField[T any](a *T, b **T) {
	if a != nil {
		*b = a
	}
}
```

I liked the helper and figured it made sense to live in this `ptr` package.

aruneshpa (approved):
LGTM!  
Thanks for making this change.

PR #540, opened 2024-05-29
🌱 Exclude ./api from coverage (but still test it)

akutz:

This patch excludes the `./api` module from coverage. This is because there is too much generated code in the API schemas that is not able to be tested due to the way the code is generated. For example, in `./api/v1alpha2/zz_generated.conversions.go`:

```go
func RegisterConversions(s *runtime.Scheme) error {
	if err := s.AddGeneratedConversionFunc((*ClusterVirtualMachineImage)(nil), (*v1alpha3.ClusterVirtualMachineImage)(nil), func(a, b interface{}, scope conversion.Scope) error {
		return Convert_v1alpha2_ClusterVirtualMachineImage_To_v1alpha3_ClusterVirtualMachineImage(a.(*ClusterVirtualMachineImage), b.(*v1alpha3.ClusterVirtualMachineImage), scope)
	}); err != nil {
		return err
	}
```

There are hundreds of these in the `v1alpha1`, `v1alpha2`, and `v1alpha3` schema packages, and for each one, there is no way to cover the logic where an error is returned due to the generated code.

Instead of having the API module cause the project's overall coverage metric dip into less-than-ideal numbers, this patch excludes the API module from coverage. Please note:

* The API module is *still* tested, and the `test` and `test-nocover` targets *will* fail if the API tests fail. 
* The API module is *still* reporting coverage, just not as part of the `cover.out` file that informs the PR results.

    The `test-api` target still uses `./hack/test.sh` which now defaults to using the `--cover` flag. However, `./hack/test.sh` only records the coverage results to a file if one is specified in the environment variable `COVERAGE_FILE`. Regardless if the results are emitted to a file, they are always available on `stdout`, including in GitHub action workflows ([example](https://github.com/vmware-tanzu/vm-operator/actions/runs/9287926215/job/25558137575?pr=540#step:5:126)).
 
The only difference is that the API module is no longer counted towards overall code coverage for the project.

Fixes `NA`

```release-note
Still test ./api, but exclude it from coverage due to all the generated code
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,9 +91,15 @@ IMAGE_TAG ?= latest
 IMG ?= ${IMAGE}:${IMAGE_TAG}
 
 # Code coverage files
-COVERAGE_FILE = cover.out
-INT_COV_FILE  = integration-cover.out
-FULL_COV_FILE = merged-cover.out
+COVERAGE_FILE := cover.out
+
+# Gather a set of root packages that have at least one file that matches
+# the pattern *_test.go as a child or descendent in that directory.
+# However, given this is not a cheap operation, only gather these packages if
+# the test-nocover target is one of the currenty active goals.
+ifeq (,$(filter-out test-nocover,$(MAKECMDGOALS)))
+COVERED_PKGS := $(shell find . -name '*_test.go' -not -path './api/*' -print | awk -F'/' '{print "./"$$2}' | sort -u)
+endif
 
 # Kind cluster name used in integration tests. Please note this name must
 # match the value in the Groovy CI script or else the CI process won't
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -142,16 +140,22 @@ help: ## Display this help
...
+test-api: | $(GINKGO)
+test-api: ## Run API tests
+	COVERAGE_FILE="" hack/test.sh ./api
+
 .PHONY: test-nocover
 test-nocover: | $(GINKGO)
+test-nocover: | test-api
 test-nocover: ## Run tests sans coverage
-	hack/test.sh
+	hack/test.sh ./controllers ./pkg ./test ./webhooks
```
dilyar85, line 152: It appears that no test cases are declared under the `./test` package. Consider removing it from this line?

dilyar85 (approved):
LGTM with a minor comment.

Makefile
akutz: Good catch!

akutz: Now the target tests all packages, minus `./api`, that has any test file in it directly or as a descendent.

PR #554, opened 2024-05-31
🐛 Fix issue reading infra config map tags

akutz:

This patch fixes an issue reading the infra config map due to the switch to sigs.k8s.io/yaml for marshaling/unmarshaling. It uses JSON tags instead of YAML tags. A test was also added to validate this.

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/controllers/infra/configmap/infra_configmap.go
+++ b/controllers/infra/configmap/infra_configmap.go
@@ -24,8 +24,8 @@ const (
 )
 
 type WcpClusterConfig struct {
-	VcPNID string `yaml:"vc_pnid"`
-	VcPort string `yaml:"vc_port"`
+	VcPNID string `json:"vc_pnid"`
+	VcPort string `json:"vc_port"`
 }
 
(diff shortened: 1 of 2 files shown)


controllers/infra/configmap/infra_configmap.go
```diff
@@ -49,6 +49,8 @@ func NewWcpClusterConfigMap(wcpClusterConfig WcpClusterConfig) (*corev1.ConfigMa
 		return nil, err
 	}
 
+	fmt.Println(string(bytes))
```
aruneshpa, line 52: Accidental, or you want this here for debugging?

aruneshpa (approved):
LGTM!
Thanks for addressing this so quickly.

controllers/infra/configmap/infra_configmap.go
akutz: On purpose for debugging, but did not mean to leave it. Thanks!

PR #555, opened 2024-06-03
🌱 Increase coverage for pkg/builder >95%

akutz:

This patch increases the code coverage for pkg/builder to above 95%!

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/pkg/builder/mutating_webhook_test.go
+++ b/pkg/builder/mutating_webhook_test.go
@@ -1,0 +1,200 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package builder_test
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/builder/mutating_webhook_test.go
```diff
@@ -0,0 +1,207 @@
...
+	})
+
+	AfterEach(func() {
+		err = nil
+		wh = nil
+		ctx = nil
+		gvk = getConfigMapSchemaGVK()
+		res = admission.Response{}
+		recorder = nil
+		scheme = nil
```
dilyar85, line 54: Nit: Some of these variable resets (e.g. ctx, scheme) are unnecessary here as they will always be initialized in the `BeforeEach`.
Also, could you help explain the purpose of `gvk = getConfigMapSchemaGVK()` here? I am a bit confused about why it uses the ConfigMap schema and why it's in the `AfterEach`...

dilyar85 (approved):
15% to >95%, amazing! 🚀

pkg/builder/mutating_webhook_test.go
akutz: Sure, some are not necessary, but I will likely not resend this to address that since it will require another run. Maybe in a follow-up. The `gvk` is used as part of building the `fakeMutator` and `fakeValidator` -- `gvk` is part of the response that can be set up in advance for testing.

akutz: I went ahead and addressed these, thank you!

PR #557, opened 2024-06-03
✨ Handle VM Class Resize annotations

bryanv:

This includes most of the infra around resize so that most of the remaining work is just being able to build the full resize Reconfigure ConfigSpec.

Improve the handling of user/devops fields by only updating the resize ConfigSpec if the current value - as from the ConfigInfo - is different.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm_resize_test.go
+++ b/pkg/providers/vsphere/vmprovider_vm_resize_test.go
@@ -20,6 +20,7 @@ import (
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere"
+	vmopv1util "github.com/vmware-tanzu/vm-operator/pkg/util/vmopv1"
 	"github.com/vmware-tanzu/vm-operator/test/builder"
 )
 
@@ -95,6 +96,22 @@ func vmResizeTests() {
 		return class
 	}
 
+	updateVMClass := func(class *vmopv1.VirtualMachineClass, cs vimtypes.VirtualMachineConfigSpec) {
+		class.Spec.ConfigSpec = encodedConfigSpec(cs)
+		class.Spec.Hardware.Cpus = int64(cs.NumCPUs)
+		class.Spec.Hardware.Memory = resource.MustParse(fmt.Sprintf("%dMi", cs.MemoryMB))
+		class.Generation++ // Fake client doesn't increment this.
+		ExpectWithOffset(1, ctx.Client.Update(ctx, class)).To(Succeed())
+	}
+
+	assertExpectedLastResizeAnnotation := func(vm *vmopv1.VirtualMachine, class *vmopv1.VirtualMachineClass) {
+		name, uid, generation, exists := vmopv1util.GetLastResizedAnnotation(*vm)
+		ExpectWithOffset(1, exists).To(BeTrue())
+		ExpectWithOffset(1, name).To(Equal(class.Name))
+		ExpectWithOffset(1, uid).To(BeEquivalentTo(class.UID))
+		ExpectWithOffset(1, generation).To(Equal(class.Generation))
+	}
+
 	Context("Resize VM", func() {
 
 		var (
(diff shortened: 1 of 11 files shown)


pkg/util/ptr/ptr.go
```diff
@@ -69,3 +69,32 @@ func Overwrite[T any](dst *T, src T) {
 		*dst = src
 	}
 }
+
+// OverwriteWithUser overwrites the *dst with an optional user value depending on
+// if the value needs to be updated based on the current value.
+func OverwriteWithUser[T comparable](dst **T, user, current *T) {
+	if dst == nil {
+		panic("dst is nil")
```
sreyasn, line 77: can it panic for classless VMs where the dst is not set?.

bryanv: For classless VMs the caller's ConfigSpec will be all default values. And `dst` always needs to point to a field in that  ConfigSpec.

pkg/providers/vsphere/vmprovider_vm_resize_test.go
```diff
@@ -156,25 +176,97 @@ func vmResizeTests() {
...
+
+					var o mo.VirtualMachine
+					Expect(vcVM.Properties(ctx, vcVM.Reference(), nil, &o)).To(Succeed())
+					Expect(o.Config.ChangeTrackingEnabled).To(HaveValue(BeTrue()))
+				})
+			})
+
+			Context("VM Classless VMs", func() {
+				BeforeEach(func() {
+					configSpec.ChangeTrackingEnabled = vimtypes.NewBool(false)
```
sreyasn, line 265: Nit: could we let the default val (false) just be picked up here instead of setting it explicitly?. thereby simulating the classlessVM scenario.

sreyasn (approved):

pkg/providers/vsphere/vmprovider_vm_resize_test.go
bryanv: It kind of doesn't matter. What makes the VM classless is when we set the `Spec.ClassName = ""`. The CS value here is used to create the VM so either we can test that it goes from `nil -> &true` or`&false -> &true`, and I kind of prefer the later.

sreyasn: > What makes the VM classless is when we set the Spec.ClassName = ""
agree.
> The CS value here is used to create the VM so either we can test that it goes from nil -> &true or&false -> &true, and I kind of prefer the later.
I wanted us to have testing as close to the actual implementation. As in, What would the overwrite do to a Classless VM's ConfigSpec ( which would only have default values). In testing, we could test by manually setting
`configSpec.ChangeTrackingEnabled = vimtypes.NewBool(false) ` but that's not what would happen actually with the happy path for classless. 

Anyways, it's just a Nit. I do see the benefit of having it set like that in BeforeEach().

bryanv: I'm still a little confused: for classless we won't have a class ConfigSpec so it will be default values. The BeforeEach is just used for creating the VM. This test is faking how real classless VMs would be "created" since they are really imported. But it is possible for an classless/imported VM to have this field - or any other - set.

sreyasn: I see what you mean. 
I was under the assumption that we were setting value (ChangeBlockTracking:false) for the configSpec argument in [OverwriteResizeConfigSpec() ](https://github.com/vmware-tanzu/vm-operator/blob/ef715a3111264925ab92ba06f66ac390430641cc/pkg/providers/vsphere/session/session_vm_update.go#L881)
```
if err := vmopv1util.OverwriteResizeConfigSpec(vmCtx, *vmCtx.VM, *moVM.Config, &configSpec); err != nil {
		return false, err
	}
```

But I see why you said it 'doesn't matter' now since the value we set is only used to create a vmclass() in the test which does really get used in the ClasslessVM scenario because we will not enter [this block](https://github.com/vmware-tanzu/vm-operator/blob/ef715a3111264925ab92ba06f66ac390430641cc/pkg/providers/vsphere/session/session_vm_update.go#L869) below and use this configSpec. 

`        if resizeArgs.VMClass != nil {`

PR #560, opened 2024-06-03
🌱 Increase coverage for session pkg from 33-->76%

akutz:

This patch increases the code coverage for the session package from 33% to 76%.

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -4,6 +4,7 @@
 package session
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"maps"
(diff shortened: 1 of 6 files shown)


dilyar85 (approved):
Thanks for adding these tests!

akutz:
> Thanks for adding these tests!

Thanks for the review! I am adding a little more coverage to this PR as I think if I do it will get us above 80% on the repo.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -584,7 +585,14 @@ func (s *Session) prePowerOnVMReconfigure(
...
 	defaultConfigSpec := &vimtypes.VirtualMachineConfigSpec{}
 	if !apiEquality.Semantic.DeepEqual(configSpec, defaultConfigSpec) {
-		vmCtx.Logger.Info("Pre PowerOn Reconfigure", "configSpec", configSpec)
+		var w bytes.Buffer
+		enc := vimtypes.NewJSONEncoder(&w)
+		if err := enc.Encode(configSpec); err != nil {
+			vmCtx.Logger.Error(err, "Failed to marshal ConfigSpec to JSON")
+			vmCtx.Logger.Info("Pre PowerOn Reconfigure", "configSpec", configSpec)
+		} else {
+			vmCtx.Logger.Info("Pre PowerOn Reconfigure", "configSpec", w.String())
```
akutz, line 594: @sreyasn / @bryanv This will be a helpful change as it will include the actual type information from the ConfigSpec as part of the logged reconfigure. We should do something similar for resize.

PR #563, opened 2024-06-05
✨ Add vm.spec.guestID API and its implementation

dilyar85:

This PR includes the following changes:
- A new `vm.spec.guestID` API in v1alpha3 and related conversion in v1alpha1 and v1alpha2.
    - Also add the missing instanceUUID conversion in v1a1
- Doc updates for this new API.
- Deploy a VM with `configSpec.guestID` if it's set in the VM Spec.
- Reconfigure a VM with the updated guestID (different from VM's ConfigInfo) before powering it on.
- Disallow updating `vm.spec.guestID` if a VM is powered on.
- Tests to cover all the above implementations. 

Depending on if a VM has been created, the VM's condition will be updated when an unsupported guestID is specified:
- If the VM has not yet been created, the `VirtualMachineCreated` condition will be marked as false with a message indicating an invalid guestID.
- If the VM has been created, the `GuestIDReconfigured` condition will be marked as false. It will be marked as true when a valid guestID is passed afterwards.

**Testing Done**:

- Valid Guest ID:
```console
# Deploy a VM with spec.guestID set:
$ cat << EOF | kubectl apply -f -
kind: VirtualMachine
metadata:
   name: test-vm-guest-id
   namespace: sdiliyaer-test
spec:
   className: best-effort-small
   imageName: ubuntu-impish-21.10-cloudimg
   guestID: otherGuest64
EOF
virtualmachine.vmoperator.vmware.com/test-vm-guest-id created

# VM has the expected configSpec.guestID set:
$ govc vm.info -json test-vm-guest-id | jq '.virtualMachines[].config.guestId'
"otherGuest64"

# Update the guestID when VM is powered on (should be denied):
$ cat << EOF | kubectl apply -f -
kind: VirtualMachine
metadata:
  name: test-vm-guest-id
  namespace: sdiliyaer-test
spec:
  className: best-effort-small
  imageName: ubuntu-impish-21.10-cloudimg
  guestID: ubuntu64Guest
EOF
...
error when patching "STDIN": admission webhook "default.validating.virtualmachine.v1alpha3.vmoperator.vmware.com" denied the request: spec.guestID: Forbidden: updates to this field is not allowed when VM power is on

# Update the guestID when VM is powered off (should be allowed):
$ cat << EOF | kubectl apply -f -
kind: VirtualMachine
metadata:
  name: test-vm-guest-id
  namespace: sdiliyaer-test
spec:
  className: best-effort-small
  imageName: ubuntu-impish-21.10-cloudimg
  guestID: ubuntu64Guest
  powerState: PoweredOff
EOF
virtualmachine.vmoperator.vmware.com/test-vm-guest-id configured

# Power on the VM and confirm VM has the updated guestID:
$ govc vm.info -json test-vm-guest-id | jq '.virtualMachines[].config.guestId'\
"ubuntu64Guest"

# Confirm the VM has the expected guestID reconfigured condition set:
$ kubectl get vm -n sdiliyaer-test test-vm-guest-id -o json | jq '.status.conditions[] | select(.type == "GuestIDReconfigured")'
{
  "lastTransitionTime": "2024-06-13T18:16:31Z",
  "message": "",
  "reason": "True",
  "status": "True",
  "type": "GuestIDReconfigured"
}
```

- Invalid Guest ID:

```console
# Deploy a new VM with an invalid guestID:
$ kubectl get vm -n sdiliyaer-test test-vm-guest-id-invalid-2 -o json | jq '.status.conditions[] | select(.type == "VirtualMachineCreated")'
{
  "lastTransitionTime": "2024-06-13T18:19:38Z",
  "message": "deploy error: An invalid argument \"configSpec.guestId\" was specified.",
  "reason": "Error",
  "status": "False",
  "type": "VirtualMachineCreated"
}

# Update an existing VM with an invalid guestID:
$ kubectl get vm -n sdiliyaer-test test-vm-guest-id-invalid-2 -o json | jq '.status.conditions[] | select(.type == "GuestIDReconfigured")'
{
  "lastTransitionTime": "2024-06-13T18:21:46Z",
  "message": "The specified guest ID value is not supported: otherGuestInvalid",
  "reason": "Invalid",
  "status": "False",
  "type": "GuestIDReconfigured"
}

# Update the VM with a valid guestID:
$ kubectl get vm -n sdiliyaer-test test-vm-guest-id-invalid-2 -o json | jq '.status.conditions[] | select(.type == "GuestIDReconfigured")'
{
  "lastTransitionTime": "2024-06-13T18:23:24Z",
  "message": "",
  "reason": "True",
  "status": "True",
  "type": "GuestIDReconfigured"
}
```

A separate change has been submitted internally to the vSphere Content Library service to avoid overwriting the VM's guestID from the OVF template.

```release-note
Introduce `vm.spec.guestID` API to specify a VM's guest operating system identifier.
```

----
📚 Documentation preview 📚: https://vm-operator--563.org.readthedocs.build/en/563/

### DIFF
--- a/api/v1alpha1/virtualmachine_conversion.go
+++ b/api/v1alpha1/virtualmachine_conversion.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2023 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2023-2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package v1alpha1
@@ -964,6 +964,10 @@ func restore_v1alpha3_VirtualMachineBootstrapCloudInitInstanceID(
 	dst.Spec.Bootstrap.CloudInit.InstanceID = iid
 }
 
+func restore_v1alpha3_VirtualMachineGuestID(dst, src *vmopv1.VirtualMachine) {
+	dst.Spec.GuestID = src.Spec.GuestID
+}
+
 func convert_v1alpha1_PreReqsReadyCondition_to_v1alpha3_Conditions(
 	dst *vmopv1.VirtualMachine) []metav1.Condition {
 
@@ -1184,6 +1188,8 @@ func (src *VirtualMachine) ConvertTo(dstRaw ctrlconversion.Hub) error {
 	restore_v1alpha3_VirtualMachineReadinessProbeSpec(dst, restored)
 	restore_v1alpha3_VirtualMachineBiosUUID(dst, restored)
 	restore_v1alpha3_VirtualMachineBootstrapCloudInitInstanceID(dst, restored)
+	restore_v1alpha3_VirtualMachineInstanceUUID(dst, restored)
+	restore_v1alpha3_VirtualMachineGuestID(dst, restored)
 
 	// END RESTORE
 
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -49,6 +49,10 @@ const (
 	// GuestBootstrapCondition exposes the status of guest bootstrap from within
 	// the guest OS, when available.
 	GuestBootstrapCondition = "GuestBootstrap"
+
+	// GuestIDCondition exposes the status of the guest ID configuration, when
+	// specified in the VirtualMachineSpec.
+	GuestIDCondition = "GuestID"
 )
 
 const (
@@ -484,6 +488,26 @@ type VirtualMachineSpec struct {
 	// When the bootstrap provider is Cloud-Init, this value is used as the
 	// default value for spec.bootstrap.cloudInit.instanceID if it is omitted.
 	BiosUUID string `json:"biosUUID,omitempty"`
+
+	// +optional
+
+	// GuestID describes the desired guest operating system identifier for a VM.
+	//
+	// The logic that determines the guest ID is as follows:
+	//
+	// 1. If this field is set, then its value is used.
+	// 2. Otherwise, if the VirtualMachineClass used to deploy the VM contains a
+	//    non-empty guest ID, then it is used.
+	// 3. Finally, if this field is still undetermined, and the VM is deployed
+	// from an OVF template that defines a guest ID, then that value is used.
+	//
+	// Please refer to https://bit.ly/4elnjP3 for a complete list of supported
+	// guest operating system identifiers.
+	//
+	// Please note that this field is immutable after the VM is powered on.
+	// To change the guest ID after the VM is powered on, the VM must be powered
+	// off and then powered on again with the updated guest ID spec.
+	GuestID string `json:"guestID,omitempty"`
 }
 
 // VirtualMachineReservedSpec describes a set of VM configuration options
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -50,9 +50,9 @@ const (
 	// the guest OS, when available.
 	GuestBootstrapCondition = "GuestBootstrap"
 
-	// GuestIDCondition exposes the status of the guest ID configuration, when
-	// specified in the VirtualMachineSpec.
-	GuestIDCondition = "GuestID"
+	// GuestIDReconfiguredCondition exposes the status of guest ID
+	// reconfiguration after a VM has been created, when available.
+	GuestIDReconfiguredCondition = "GuestIDReconfigured"
 )
 
 const (
--- a/docs/concepts/workloads/vm.md
+++ b/docs/concepts/workloads/vm.md
@@ -438,3 +438,20 @@ The value of `spec.biosUUID` is used to set the vSphere VM `config.uuid` field, 
 ``` bash
 govc vm.info -vm.uuid $(k get vm -o jsonpath='{.spec.biosUUID}' -n $ns $name)
 ```
+
+## Guest IDs
+
+The `spec.guestID` is an optional field that can be used to specify the guest operating system identifier when creating a new VM or powering on an existing VM.
+
+A complete list of supported guest IDs can be found [here](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U2/html/SDK/vsphere-ws/docs/ReferenceGuide/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html).
+
+The `govc` command can be used to query the currently supported guest IDs for a given vSphere environment:
+
+``` bash
+$ govc vm.option.info -cluster CLUSTER_NAME
+...
+almalinux_64Guest           AlmaLinux (64-bit)
+rockylinux_64Guest          Rocky Linux (64-bit)
+windows2022srvNext_64Guest  Microsoft Windows Server 2025 (64-bit)
+...
+```
--- a/pkg/providers/vsphere/virtualmachine/configspec.go
+++ b/pkg/providers/vsphere/virtualmachine/configspec.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022-2023 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package virtualmachine
@@ -113,6 +113,13 @@ func CreateConfigSpec(
 			lim := MemoryQuantityToMb(vmClassSpec.Policies.Resources.Limits.Memory)
 			configSpec.MemoryAllocation.Limit = &lim
 		}
+	}
+
+	// Initially set the guest ID in ConfigSpec to ensure VM is created with the expected guest ID.
+	// Afterwards, only update it if the VM spec guest ID differs from the VM's existing ConfigInfo.
+	// If the class also specifies a guest ID, it will be overridden by the VM spec guest ID.
+	if guestID := vmCtx.VM.Spec.GuestID; guestID != "" {
+		configSpec.GuestId = guestID
 	}
 
 	return configSpec
--- a/pkg/util/vsphere/vm/guest_id.go
+++ b/pkg/util/vsphere/vm/guest_id.go
@@ -17,15 +17,16 @@ import (
 // GuestIDProperty is the property name for the guest ID in the config spec.
 const GuestIDProperty = "configSpec.guestId"
 
-// UpdateVMGuestIDCondition updates the VM's GuestID condition to false if the
-// taskInfo contains an InvalidArgument fault with an invalid guest ID property.
-// Otherwise, it updates the condition to true.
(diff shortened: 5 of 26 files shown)


pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -115,6 +115,12 @@ func CreateConfigSpec(
 		}
 	}
 
+	// Initially set the guest ID in ConfigSpec to ensure VM is created with the expected guest ID.
+	// Afterwards, only update it if the VM spec guest ID differs from the existing ConfigInfo.
+	if guestID := string(vmCtx.VM.Spec.GuestID); guestID != "" {
```
bryanv, line 120: Do we ever want to honor an existing GuestID in the class ConfigSpec? If not, it should be cleared in `SanitizeVMClassConfigSpec()`

bryanv, line 120: Do we ever want to honor an existing GuestID in the class ConfigSpec? If not, it should be cleared in `SanitizeVMClassConfigSpec()`

dilyar85: I think we should continue to support GuestID in the ConfigSpec class as it is today, and if it's also specified in the VM spec, the latter should overwrite it. I will add a comment here and in the CRD API for better clarity.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -102,6 +102,10 @@ const (
 	// VirtualMachineReconcilePausedReason indicates that VirtualMachine
 	// reconciliation is being paused.
 	VirtualMachineReconcilePausedReason = "VirtualMachineReconcilePaused"
+
+	// VirtualMachineConditionPrePowerOnReconfigReady exposes the status of the
+	// VM's pre-power-on reconfiguration task.
+	VirtualMachineConditionPrePowerOnReconfigReady = "VirtualMachinePrePowerOnReconfigReady"
```
bryanv, line 108: I'm a little uneasy that this condition exposes an internal implementation detail. In a lot of cases we don't need that reconfigure b/c of the 7605 work, and ideally I want to get us to where the VM is created with its desired configuration in all cases. And with resize coming that is going to take the place of the current prepower on reconfigure, and that will have its own conditon.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1026,6 +1027,10 @@ func (v validator) validateUpdatesWhenPoweredOn(ctx *pkgctx.WebhookRequestContex
 		allErrs = append(allErrs, field.Forbidden(specPath.Child("bootstrap"), updatesNotAllowedWhenPowerOn))
 	}
 
+	if !equality.Semantic.DeepEqual(vm.Spec.GuestID, oldVM.Spec.GuestID) {
```
bryanv, line 1030: Can we get by with `!=` here instead?

api/v1alpha3/virtualmachine_types.go
```diff
@@ -484,6 +488,17 @@ type VirtualMachineSpec struct {
 	// When the bootstrap provider is Cloud-Init, this value is used as the
 	// default value for spec.bootstrap.cloudInit.instanceID if it is omitted.
 	BiosUUID string `json:"biosUUID,omitempty"`
+
+	// +optional
+
+	// GuestID describes the desired short guest operating system identifier for
+	// a VM. If omitted, the VM may still be created with a guest ID if it is
+	// defined in the OVF template used to deploy this VM.
+	// Please refer to https://bit.ly/3yJdL0e for a complete list of supported
```
dougm, line 497: Followed that link:
> Revision 14 March 2023  |  vSphere 8.0U1  |  vSphere vCenter SMS API 8.0U1

Maybe use [8.0U2](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U2/html/SDK/vsphere-ws/docs/ReferenceGuide/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html) instead? That doc also includes the since versions. Can also get the list supported by a live cluster using:
```console
% govc vm.option.info -cluster test-vpx-1717673857-26202-wcp.wcp-sanity-cluster
...
almalinux_64Guest           AlmaLinux (64-bit)
rockylinux_64Guest          Rocky Linux (64-bit)
windows2022srvNext_64Guest  Microsoft Windows Server 2025 (64-bit)
...
```
```

File: docs/concepts/workloads/vm.md [hunk 5]
```diff
@@ -438,3 +438,9 @@ The value of `spec.biosUUID` is used to set the vSphere VM `config.uuid` field,
 ``` bash
 govc vm.info -vm.uuid $(k get vm -o jsonpath='{.spec.biosUUID}' -n $ns $name)
 ```
+
+## Guest IDs
+
+The `spec.guestID` is an optional field that can be used to specify the guest operating system identifier when creating a new VM or when the existing VM is powered off.
+A complete list of supported guest IDs can be found [here](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/e1580753-930f-4c00-9105-77dd2e8913f8/98ce8e90-a51b-4e41-8dac-3e8fd5409f2b/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html).
```
dougm, line 445: Same here (U1 vs U2). Maybe the govc example could be included here too. Or even use it generate a list we could add to these docs?

webhooks/virtualmachine/validation/virtualmachine_validator.go
dougm: We also have a static list of valid guest IDs that could be used for validation: https://github.com/vmware/govmomi/blob/main/vim25/types/enum.go#L13439

dougm (commented):
lgtm, just some minor doc suggestions.

docs/concepts/workloads/vm.md
dilyar85: Updated the link and also added the govc command example.

api/v1alpha3/virtualmachine_types.go
dilyar85: Thanks! Updated with the newer version link.

webhooks/virtualmachine/validation/virtualmachine_validator.go
dilyar85: Fixed to use `!=`.
@dougm This field used to be validated by the above enum list in the API, but we switched to a string to avoid blocking VM Service VM creation/update in case a newer guestID is used, which hasn't yet been updated in the govmomi repo.

api/v1alpha3/virtualmachine_types.go
dilyar85: The purpose of this new condition is to inform users when they attempt to deploy a VM with an invalid guestID, as it's no longer verified by a generated list from the imported govmomi package. While implementing this change, I thought why make it specific to guestID only, hence the `PrePowerOnReconfigureReady` name above to expose any VM reconfigure failure (it still has config.guestID in the condition's reason if the task fails by that).

On the other hand, today VMOP already sends a K8s event if the `PrePowerOnReconfigure` task fails. This event contains specific invalid configs as well, such as guestID shown in the following example. So I think users are not entirely in the dark if they have specified an unsupported guestID:
```
Warning  CreateOrUpdateFailure  29s (x16 over 116s)  vmware-system-vmop/vmware-system-vmop-controller-manager-745fcd885b-v2xhm/virtualmachine-controller  reconfigure VM task failed: A specified parameter was not correct: configSpec.guestId
``` 
@akutz Any thoughts on dropping this condition and leveraging existing K8s events for this?

sreyasn: I'd prefer to not start exposing prePowerOnReconfigure as a condition too.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -484,6 +488,17 @@ type VirtualMachineSpec struct {
...
 	BiosUUID string `json:"biosUUID,omitempty"`
+
+	// +optional
+
+	// GuestID describes the desired short guest operating system identifier for
+	// a VM. If omitted, the VM may still be created with a guest ID if it is
+	// defined in the OVF template used to deploy this VM.
+	// Please refer to https://bit.ly/4elnjP3 for a complete list of supported
+	// guest operating system identifiers.
+	// This field overwrites the guest ID in the VM class and cannot be changed
```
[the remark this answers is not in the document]
dilyar85: Updated with the order of evaluation of this field added.

dougm (approved):
lgtm

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -1184,6 +1188,8 @@ func (src *VirtualMachine) ConvertTo(dstRaw ctrlconversion.Hub) error {
 	restore_v1alpha3_VirtualMachineReadinessProbeSpec(dst, restored)
 	restore_v1alpha3_VirtualMachineBiosUUID(dst, restored)
 	restore_v1alpha3_VirtualMachineBootstrapCloudInitInstanceID(dst, restored)
+	restore_v1alpha3_VirtualMachineInstanceUUID(dst, restored)
```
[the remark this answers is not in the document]
dilyar85: Done.

api/v1alpha3/virtualmachine_types.go
dilyar85: Updated with the Guest ID specific condition. Please refer to the PR's description for all possible success/failure cases.

pkg/util/vsphere/vm/guest_id.go
```diff
@@ -0,0 +1,54 @@
...
+
+// UpdateVMGuestIDCondition updates the VM's GuestID condition to false if the
+// taskInfo contains an InvalidArgument fault with an invalid guest ID property.
+// Otherwise, it updates the condition to true.
+func UpdateVMGuestIDCondition(
+	vmctx pkgctx.VirtualMachineContext,
+	configSpec vimtypes.VirtualMachineConfigSpec,
+	taskInfo *vimtypes.TaskInfo) {
+	// Return early if the current configSpec does not have a guest ID.
+	if configSpec.GuestId == "" {
```
bryanv, line 28: Shouldn't the condition be deleted here?

dilyar85: I was trying to avoid resetting the condition when the VM is reconfigured without the guestID set in the config spec. Thinking it through now, that should only happen when vm.spec.guestID is removed. So you're right, we should remove the condition here. I will update the implementation and add a test case for this. Thanks for catching this!

dilyar85: Also updated it to remove the condition when there is no error, instead of marking it as true, which will eventually be removed in a future VM reconfiguration anyways.

bryanv (approved):

PR #568, opened 2024-06-07
🐛 Fix VM controller VM Class reconcile request handler

### LINKED ISSUE #515: Do not requeue a VM to wait on IP if VM networking is disabled

The requeue logic related to a `VirtualMachine` resource and networking at https://github.com/vmware-tanzu/vm-operator/blob/015b5758d6af07d86be7c42b0d9159a687002d7e/controllers/virtualmachine/virtualmachine_controller.go#L278-L281 requeues an object for 10s if the VM is:

* powered on
* does not yet have an Ip4 or Ip6 address

This retry is unnecessarily aggressive for VMs with disabled networking.

The logic should check `spec.network.disabled` to determine if networking is disabled for the VM, ex.:

```go
		if !vm.Spec.Network.Disabled && (network == nil || (network.PrimaryIP4 == "" && network.PrimaryIP6 == "")) {
```

N/A

bryanv:

An empty VMClass.Spec.ControllerName name would not cause us to reconcile the VMs when its VM Class was changed.

Fix requeueDelay() to only fall into the IP retry if the VM is expected to have networking in the first place.

Fix several integration test issues:

Use a valid UUID for dummy InstanceUUID. The prior value along with the schema upgrade would fail due to field validation. Due to how the patch helpers breaks up the patches, this was causing an error retry instead of the expected requeueDelay() usage.

Remove a bogus Creating Phase test: the controller used to set the Phase in some cases, but Phase is gone and the controller doesn't set Conditions directly. This test wasn't really testing anything.

Fix a race when testing VM deletion so we actually assert the VM was deleted.

Move some assertions into the Eventually() block so errors are descriptive

Fixes #515 

```release-note
NONE
```

### DIFF
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -106,7 +106,11 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 // classToVMMapperFn returns a mapper function that can be used to queue reconcile request
 // for the VirtualMachines in response to an event on the VirtualMachineClass resource when
 // WCP_Namespaced_VM_Class FSS is enabled.
-func classToVMMapperFn(ctx *pkgctx.ControllerManagerContext, c client.Client, isDefaultVMClassController bool) func(_ context.Context, o client.Object) []reconcile.Request {
+func classToVMMapperFn(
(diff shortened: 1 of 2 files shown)


akutz (approved):
Lgtm, thank you!

PR #569, opened 2024-06-07
🌱 Update testing -- resize ExtraConfig

sreyasn:

Split into two separate tests because the output extraConfig does not maintain order.

Fixes #N/A

N/A

```NONE

```


bryanv (approved):

bryanv:
I've kind of had the thought of folding `MergeExtraConfig` and `ExtraConfigToMap` together, and in that one function have a map that is name to index for quick lookup.

akutz:
> I've kind of had the thought of folding `MergeExtraConfig` and `ExtraConfigToMap` together, and in that one function have a map that is name to index for quick lookup.

I have had the same thought, though I would keep  `ExtraConfigToMap` separate as it has value on its own and then possibly leverage it in `MergeExtraConfig`. 

Frankly, in `pkg/utils/vsphere`, I would:

```golang
func NewOptionValues(in []vimtypes.BaseOptionValue) OptionValues {
    if len(in) == 0 {
        return nil
    }

    out := make(OptionValues{}, len(in))

    for i := range in {
        switch tval := in[i].Value.(type) {
        case string:
            out[in[i].Key] = tval
        default:
            out[in[i].Key] = fmt.Sprintf("%v", tval)
        }
    }

    return out
}

type OptionValues map[string]string

func (ov OptionValues) ToList() []vimtypes.BaseOptionValue {}

func (ov OptionValues) Merge(in []vimtypes.BaseOptionValue) {}

func (ov OptionValues) MergeList(in OptionValues) {}
```

sreyasn:
Opened https://github.com/vmware-tanzu/vm-operator/issues/570 for the comment above, @akutz

PR #571, opened 2024-06-10
🌱 OptionValues helper

### LINKED ISSUE #570: Minor refactor of ExtraConfig methods -- pkg/utils/vsphere

https://github.com/vmware-tanzu/vm-operator/pull/569#issuecomment-2155584228
In pkg/utils/vsphere, I would:

```
func OptionValuesFromList(in []vimtypes.BaseOptionValue) OptionValues {
    if len(in) == 0 {
        return nil
    }

    out := make(OptionValues{}, len(in))

    for i := range in {
        switch tval := in[i].Value.(type) {
        case string:
            out[in[i].Key] = tval
        default:
            out[in[i].Key] = fmt.Sprintf("%v", tval)
        }
    }

    return out
}

type OptionValues map[string]string

func (ov OptionValues) ToList() []vimtypes.BaseOptionValue {}

func (ov OptionValues) Merge(in OptionValues) {}

func (ov OptionValues) MergeList(in []vimtypes.BaseOptionValue) {}
```
Something like the above. A very simple and efficient way to intake, manipulate, and reuse ExtraConfig or anything else that is a list of option values.

N/A

N/A

akutz:

This patch introduces the type OptionValues in "pkg/util". This new type simplifies working with []vimtypes.BaseOptionValue properties such as ExtraConfig.

@dougm , it occurs to me after writing this ... should this just be in GoVmomi?

Fixes #570

Please note there are _two_ commits for this PR:

1. The actual business logic.
2. A patch where the files affected from the previous patch have their `pkg/util` packages imported according to the correct rules as `pkgutil`. 

This enables reviewers to focus on the business logic patch separately, making it easier to see what actually changed.

```release-note
Introduce pkg/util.OptionValues for working with properties like ExtraConfig
```

### DIFF
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -326,8 +326,9 @@ func UpdateConfigSpecExtraConfig(
 		// with the class config spec extra config (ie) class config spec extra
 		// config keys takes precedence over the desired config spec extra
 		// config keys.
-		combinedExtraConfig := util.AppendNewExtraConfigValues(classConfigSpec.ExtraConfig, extraConfig)
-		extraConfig = util.ExtraConfigToMap(combinedExtraConfig)
+		extraConfig = util.OptionValues(classConfigSpec.ExtraConfig).
+			Append(util.OptionValuesFromMap(extraConfig)...).
+			StringMap()
 	}
 
 	// Note if the VM uses both LinuxPrep and vAppConfig. This is used in the
@@ -367,7 +368,8 @@ func UpdateConfigSpecExtraConfig(
 	// Update the ConfigSpec's ExtraConfig property with the results from
 	// above. Please note this *may* include keys with empty values. This
 	// indicates to vSphere that a key/value pair should be removed.
-	configSpec.ExtraConfig = util.MergeExtraConfig(config.ExtraConfig, extraConfig)
+	configSpec.ExtraConfig = util.OptionValues(config.ExtraConfig).
+		Diff(util.OptionValuesFromMap(extraConfig)...)
 }
 
 func isLinuxPrepAndVAppConfig(vm *vmopv1.VirtualMachine) bool {
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -28,7 +28,7 @@ import (
 	res "github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/resources"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/virtualmachine"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/vmlifecycle"
-	"github.com/vmware-tanzu/vm-operator/pkg/util"
+	pkgutil "github.com/vmware-tanzu/vm-operator/pkg/util"
 	"github.com/vmware-tanzu/vm-operator/pkg/util/resize"
 	vmopv1util "github.com/vmware-tanzu/vm-operator/pkg/util/vmopv1"
 	vmutil "github.com/vmware-tanzu/vm-operator/pkg/util/vsphere/vm"
@@ -326,8 +326,8 @@ func UpdateConfigSpecExtraConfig(
 		// with the class config spec extra config (ie) class config spec extra
 		// config keys takes precedence over the desired config spec extra
 		// config keys.
-		extraConfig = util.OptionValues(classConfigSpec.ExtraConfig).
-			Append(util.OptionValuesFromMap(extraConfig)...).
+		extraConfig = pkgutil.OptionValues(classConfigSpec.ExtraConfig).
+			Append(pkgutil.OptionValuesFromMap(extraConfig)...).
 			StringMap()
 	}
 
@@ -368,8 +368,8 @@ func UpdateConfigSpecExtraConfig(
 	// Update the ConfigSpec's ExtraConfig property with the results from
 	// above. Please note this *may* include keys with empty values. This
 	// indicates to vSphere that a key/value pair should be removed.
-	configSpec.ExtraConfig = util.OptionValues(config.ExtraConfig).
-		Diff(util.OptionValuesFromMap(extraConfig)...)
+	configSpec.ExtraConfig = pkgutil.OptionValues(config.ExtraConfig).
+		Diff(pkgutil.OptionValuesFromMap(extraConfig)...)
 }
 
 func isLinuxPrepAndVAppConfig(vm *vmopv1.VirtualMachine) bool {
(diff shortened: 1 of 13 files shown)


dougm:
> it occurs to me after writing this ... should this just be in GoVmomi?

I was thinking that too, along with some of the recent resize work that has gone in (which could be used in simulator at least). We could start a sheet/list of vmop helpers that would be generally useful if they were moved into govmomi. And we could use that to discuss which package(s) such helpers would land in govmomi.

pkg/util/option_values.go
```diff
@@ -0,0 +1,164 @@
...
+// AppendList combines the provided input with the current OptionValues data,
+// ignoring any keys from the input data that already exist in the existing
+// data.
+func (ov OptionValues) AppendList(in []vimtypes.BaseOptionValue) {
+	ov.mergeList(in, false)
+}
+
+// Merge combines the provided input with the current OptionValues data. Any
+// matching keys from the input data will overwrite keys in the existing data.
+func (ov OptionValues) Merge(in OptionValues) {
```
aruneshpa, line 115: I don't see this one being called anywhere as well.  If this is just for future proofing, I would vote for adding it when it's needed.  Sorry if I missed where it is referenced in the diff.

pkg/providers/vsphere/vmlifecycle/bootstrap_cloudinit.go
```diff
@@ -253,8 +253,12 @@ func GetCloudInitGuestInfoCustSpec(
 		extraConfig[constants.CloudInitGuestInfoUserdataEncoding] = "gzip+base64"
 	}
 
-	configSpec := &vimtypes.VirtualMachineConfigSpec{}
-	configSpec.ExtraConfig = util.MergeExtraConfig(config.ExtraConfig, extraConfig)
+	newExtraConfig := pkgutil.OptionValuesFromList(config.ExtraConfig)
+
+	configSpec := &vimtypes.VirtualMachineConfigSpec{
+		ExtraConfig: newExtraConfig.Diff(extraConfig).ToList(),
```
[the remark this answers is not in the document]
akutz: No, it is not merging as it does not impact the *existing* `OptionValues` object. It returns a difference of what is required to update the original based on the new values.

pkg/util/resize/configspec.go
```diff
@@ -265,8 +265,8 @@ func compareExtraConfig(
 		return
 	}
 
-	extraConfig := util.ExtraConfigToMap(cs.ExtraConfig)
-	outCS.ExtraConfig = util.MergeExtraConfig(ci.ExtraConfig, extraConfig)
+	extraConfig := pkgutil.OptionValuesFromList(ci.ExtraConfig)
+	outCS.ExtraConfig = extraConfig.DiffListToList(cs.ExtraConfig)
```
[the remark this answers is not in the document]
akutz: Nope, because of the ordering. Please note the comments on the functions. The `ToList` variants specifically maintain insertion order at the request of @bryanv to make ConfigSpec comparisons easier for resize.

pkg/util/option_values.go
akutz: The old function was called `MergeExtraConfig`, but it was improperly named. It should have been called `DiffExtraConfig`. I am keeping the `Merge` here in case someone _does_ need it in the future to ensure it is written the correct way.

pkg/util/option_values.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if len(out) == 0 {
+		return nil
+	}
+	return out
+}
+
+// DiffList returns the difference between the existing data and provided input.
+// Please note the output does not maintain the order of the input.
+func (ov OptionValues) DiffList(in []vimtypes.BaseOptionValue) OptionValues {
```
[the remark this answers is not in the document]
akutz: It is for consistency with the other functions, Append and Merge. Both of them have List options. This is to ensure folks that _want_ to use them have correctly written functions to use.

akutz:
@aruneshpa If you are good with our discussion, did you want to approve this?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -326,8 +326,7 @@ func UpdateConfigSpecExtraConfig(
 		// with the class config spec extra config (ie) class config spec extra
 		// config keys takes precedence over the desired config spec extra
 		// config keys.
-		combinedExtraConfig := util.AppendNewExtraConfigValues(classConfigSpec.ExtraConfig, extraConfig)
-		extraConfig = util.ExtraConfigToMap(combinedExtraConfig)
+		extraConfig.Append(pkgutil.OptionValuesFromList(classConfigSpec.ExtraConfig))
```
sreyasn, line 329: can we use AppendList() here ?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1033,8 +1033,10 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpecExtraConfig(
 		}
 	}
 
-	// The ConfigSpec's current ExtraConfig values (that came from the class) take precedence over what was set here.
-	createArgs.ConfigSpec.ExtraConfig = util.AppendNewExtraConfigValues(createArgs.ConfigSpec.ExtraConfig, ecMap)
+	// The ConfigSpec's current ExtraConfig values (that came from the class)
+	// take precedence over what was set here.
+	newExtraConfig := pkgutil.OptionValues(ecMap)
+	newExtraConfig.MergeList(createArgs.ConfigSpec.ExtraConfig)
```
sreyasn, line 1039: I see a behavior change from what happened with AppendNewExtraConfigValues(). The earlier method only added new fields from _ecMap_ that didn't exist in _createArgs.ConfigSpec.ExtraConfig_. 

Might need to do something like IIUC
```
createArgsEC := pkgutil.OptionValuesFromList(createArgs.ConfigSpec.ExtraConfig)
newEc := pkgutil.OptionValues(ecMap)
createArgsEc.Append(newEc)
```

pkg/util/resize/configspec.go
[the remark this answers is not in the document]
sreyasn: curious: Are we sure that by insertion ordering here, we also maintain the order that we get back from the configInfo when fetched from the VM?.

pkg/util/option_values.go
```diff
@@ -0,0 +1,164 @@
...
+// Please note the output does not maintain the order of the input.
+func (ov OptionValues) Diff(in OptionValues) OptionValues {
+	out := OptionValues{}
+	for k, newVal := range in {
+		if oldVal, ok := ov[k]; !ok || newVal != oldVal {
+			out[k] = newVal
+		}
+	}
+	if len(out) == 0 {
+		return nil
```
bryanv, line 56: Depending on how these calls are all chain, this does worry me a little that we can unexpectedly NPE, like if somebody would then call a func that ends up in `merge()` or `mergeList()`

pkg/util/resize/configspec.go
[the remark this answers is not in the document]
bryanv: There are some situations for writing simpler test assertions a deterministic ordering is helpful, but consider that ask secondary to whatever we want to have clearer non-test code.

pkg/util/option_values.go
akutz: How so? We allow `in` to be `nil` without error, and the functions that could NPE do not return anything, such as `Merge` and `MergeList`. These specifically do not return and cannot be used to chain calls.

bryanv: Let me rephrase a little bit - with the type alias you still need need to know the underlying type is nullable. I'm worried that something like below is too easy to write or end up with:

```
in := []vimtypes.BaseOptionValue{&vimtypes.OptionValue{Key: "foo", Value: "one"}}
more := []vimtypes.BaseOptionValue{&vimtypes.OptionValue{Key: "foo1", Value: "two"}}
x := pkgutil.OptionValuesFromList(in).DiffList(in)
....
x.MergeList(more)
```

Thinking back to a bug that was introduced with the switch to either the `slices` or `maps` package that an empty in param return nil.

sreyasn (approved):

bryanv (approved):

PR #572, opened 2024-06-10
🌱 Increase coverage for web-console-validation server from 51-->100%

dilyar85:

This patch increases the code coverage for the `pkg/webconsolevalidation` package from 51 to 100%.

```console
$ ./hack/test.sh ./pkg/webconsolevalidation
...

PASS
coverage: 100.0% of statements
composite coverage: 100.0% of statements
```

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/pkg/webconsolevalidation/server.go
+++ b/pkg/webconsolevalidation/server.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package webconsolevalidation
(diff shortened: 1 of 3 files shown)


akutz:
Thank you @dilyar85 !!

FWIW:

> `go test -coverprofile=/tmp/go-code-cover github.com/vmware-tanzu/vm-operator/pkg/webconsolevalidation`

The above can be simplified as:

`hack/test.sh ./pkg/webconsolevalidation`

:)

pkg/webconsolevalidation/server.go
```diff
@@ -20,19 +20,28 @@ import (
 // K8sClient is used to get the webconsolerequest resource from UUID and namespace.
 var K8sClient ctrlclient.Client
 
+// Function variables to allow for mocking in tests.
+var (
```
akutz, line 24: Might I ask why you do not just send this into the `InitServer` function as arguments?

akutz (changes requested):
Instead of package-scoped vars for testing that can complicate parallel testing, please send in the vars as arguments to `InitServer` unless there is some compelling reason to not do this.

pkg/webconsolevalidation/server.go
dilyar85: Updated as suggested and also removed the other pkg-scoped var `K8sClient` previously defined in this file.

akutz (approved):

PR #576, opened 2024-06-10
✨ Resize Console Preferences

sreyasn:

Support resizing a VM with the Virtual Machine ConfigSpec's console preferences setting.

Fixes #N/A

N/A

```
Support resizing VMs with console preferences settings
```

### DIFF
--- a/pkg/util/resize/configspec.go
+++ b/pkg/util/resize/configspec.go
@@ -34,6 +34,7 @@ func CreateResizeConfigSpec(
 	compareCPUPerfCounter(ci, cs, &outCS)
 	compareLatencySensitivity(ci, cs, &outCS)
 	compareExtraConfig(ci, cs, &outCS)
+	compareConsolePreferences(ci, cs, &outCS)
 
 	return outCS, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/util/resize/configspec.go
```diff
@@ -269,6 +270,40 @@ func compareExtraConfig(
...
+
+	if cs.ConsolePreferences == nil {
+		return
+	}
+
+	if ci.ConsolePreferences == nil {
+		// If configInfo console preferences is nil, there is nothing to compare.
+		// set desired values
+		outCS.ConsolePreferences = &vimtypes.VirtualMachineConsolePreferences{
+			PowerOnWhenOpened:        cs.ConsolePreferences.PowerOnWhenOpened,
```
bryanv, line 287: Shouldn't this also check that all field here are not-nil, right? Ideally we don't want just `outCS.ConsolePreferences  = &vimtypes.VirtualMachineConsolePreferences{}` because that is going to cause an unnecessary Reconfigure.

pkg/util/resize/configspec.go
```diff
@@ -269,6 +270,40 @@ func compareExtraConfig(
...
+			EnterFullScreenOnPowerOn: cs.ConsolePreferences.EnterFullScreenOnPowerOn,
+			CloseOnPowerOffOrSuspend: cs.ConsolePreferences.CloseOnPowerOffOrSuspend,
+		}
+
+		return
+	}
+
+	if ci.ConsolePreferences.PowerOnWhenOpened != cs.ConsolePreferences.PowerOnWhenOpened ||
```
bryanv, line 295: Is compare just the _pointers_ here what we want? I guess that's OK if we later see if after the `cmpPtr()`'s if fields`ConsolePreferences` are all nil and then just nil out `outCS.ConsolePreferences`?

sreyasn: Agree. going with what you suggested.

pkg/util/resize/configspec.go
sreyasn: yep.. that would be better. updated the comment to reflect the same logic

PR #577, opened 2024-06-11
🐛 EnsureAllDisksPointToController

akutz:

This patch adds a function for ensuring all disks in a ConfigSpec point to a valid controller.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -484,12 +484,25 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 func (vs *vSphereVMProvider) vmCreateDoPlacement(
 	vmCtx pkgctx.VirtualMachineContext,
 	vcClient *vcclient.Client,
-	createArgs *VMCreateArgs) error {
+	createArgs *VMCreateArgs) (retErr error) {
 
-	placementConfigSpec := virtualmachine.CreateConfigSpecForPlacement(
+	defer func() {
+		if retErr != nil {
+			conditions.MarkFalse(
+				vmCtx.VM,
+				vmopv1.VirtualMachineConditionPlacementReady,
+				"NotReady",
+				retErr.Error())
+		}
+	}()
+
+	placementConfigSpec, err := virtualmachine.CreateConfigSpecForPlacement(
 		vmCtx,
 		createArgs.ConfigSpec,
 		createArgs.StorageClassesToIDs)
+	if err != nil {
+		return err
+	}
 
 	result, err := placement.Placement(
 		vmCtx,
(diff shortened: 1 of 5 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -486,10 +486,13 @@ func (vs *vSphereVMProvider) vmCreateDoPlacement(
...
 	createArgs *VMCreateArgs) error {
 
-	placementConfigSpec := virtualmachine.CreateConfigSpecForPlacement(
+	placementConfigSpec, err := virtualmachine.CreateConfigSpecForPlacement(
 		vmCtx,
 		createArgs.ConfigSpec,
 		createArgs.StorageClassesToIDs)
+	if err != nil {
+		return err
+	}
```
dilyar85, line 495: Consider adding `conditions.MarkFalse(vmCtx.VM, vmopv1.VirtualMachineConditionPlacementReady, "NotReady", err.Error())` here?

pkg/util/ensure_disk_controller.go
```diff
@@ -0,0 +1,551 @@
...
+				// Please note that at this point it is not yet known if the
+				// controller key is a *valid* controller.
+				diskControllers.attach(controllerKey)
+			}
+		}
+
+		// Keep track of the largest device key used.
+		if d.Key < newDeviceKey {
+			newDeviceKey = d.Key
+		}
```
dilyar85, line 122: Looks like the code is doing the opposite of what the comment says?

akutz: Largest in the sense of largest *negative* number. I could say smallest to be technically accurate, but I thought this was more clear. Suggestions on how to phrase this?

pkg/providers/vsphere/vmprovider_vm.go
akutz: You are quite kind to not just state that I missed this :)

dougm (approved):
lgtm. Will add this to the list of helpers we should consider moving to govmomi. We could likely integrate with the [VirtualDeviceList](https://github.com/vmware/govmomi/blob/1db964ce5b84adcb1e069847455fa91224755703/object/virtual_device_list.go) related methods, used both client-side and server-side (vcsim)

pkg/util/ensure_disk_controller.go
dilyar85: Ah, I didn't realize that device keys in a ConfigSpec are negative numbers, so I was confused.
What you have updated looks good to me.

dilyar85 (approved):
LGTM!

sreyasn (approved):

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -484,12 +484,25 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 func (vs *vSphereVMProvider) vmCreateDoPlacement(
 	vmCtx pkgctx.VirtualMachineContext,
 	vcClient *vcclient.Client,
-	createArgs *VMCreateArgs) error {
+	createArgs *VMCreateArgs) (retErr error) {
+
+	defer func() {
```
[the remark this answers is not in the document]
sreyasn: Looking at the code. It does seem like we could have set the condition all along for this error. Since the result returned was not sufficient for IS.
Maybe @bryanv has a reasoning behind why it wasn't surfaced earlier?.

PR #579, opened 2024-06-12
🌱 Add the vCenter UUID annotation on the VM

aruneshpa:

This change adds the vCenter annotation on the VirtualMachine custom resource.  Among other things, this annotation can be used to determine whether a VM is being restored on the same vCenter where it was originally created.

```release-note
Add vCenter UUID annotation on the VM
```

### DIFF
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -139,7 +139,7 @@ const (
 
 	// VirtualMachineSameVMClassResizeAnnotation is an annotation that indicates the VM
 	// should be resized as the class it points to changes.
-	VirtualMachineSameVMClassResizeAnnotation = GroupName + "/same-vm-class-resize "
+	VirtualMachineSameVMClassResizeAnnotation = GroupName + "/same-vm-class-resize"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


dilyar85 (approved):

api/v1alpha3/virtualmachine_types.go
```diff
@@ -159,6 +159,12 @@ const (
 	PVCDiskDataExtraConfigKey = "vmservice.virtualmachine.pvc.disk.data"
 )
 
+const (
+	// VCUUIDAnnotation on a VirtualMachine contains the UUID of the
+	// VMware vCenter (VC) that is managing this virtual machine.
+	VCUUIDAnnotation = GroupName + "/vc-uuid"
```
akutz, line 165: Do we want to call out vCenter as the source of the ID? Why not keep the naming generic, such as:

```go
const (
	// InitialLocationID describes the ID of the location where a
	// VirtualMachine was initially deployed.
	InitialLocationID = GroupName + "/initial-location-id"
)
```

We do not plan to change this when the VM is restored on another platform, correct? We could instead introduce a second annotation:

```go
const (
	// CurrentLocationID describes the unique ID of the location where a
	// VirtualMachine is currently managed.
	CurrentLocationID = GroupName + "/current-location-id"
)
```

aruneshpa: That's a good point.  Updating.

aruneshpa: @akutz , Per our conversation, changed the text to "manager-id".  Let me know if this makes sense.

akutz (changes requested):
Lgtm, thanks @aruneshpa !

akutz (approved):
Oops, I selected "request changes" on accident. Approved :)

PR #583, opened 2024-06-14
🌱 Resize Brownfield VMs

bryanv:

VMs prior to the resize feature will not have the last-resize annotation that we will set on new VMs at create time. When the ClassName changes set the last-resized annotation so we'll evaluate a resize when the VM powerstate allows.

Add a quick for a braino and last second refactor that caused us to match all devices instead of just PCI passthru devices. I need to do some more clean up and refactoring of the tests to make this more natural to test.

Fixes #

```release-note
NONE
```

### DIFF
--- a/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
+++ b/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
@@ -134,6 +134,12 @@ func (m mutator) Mutate(ctx *pkgctx.WebhookRequestContext) admission.Response {
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
 
+		if ok, err := SetLastResizeAnnotation(ctx, modified, oldVM); err != nil {
+			return admission.Denied(err.Error())
+		} else if ok {
+			wasMutated = true
+		}
+
 		if ok, err := SetNextRestartTime(ctx, modified, oldVM); err != nil {
 			return admission.Denied(err.Error())
 		} else if ok {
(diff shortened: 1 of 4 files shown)


webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -450,3 +456,29 @@ func SetCreatedAtAnnotations(ctx context.Context, vm *vmopv1.VirtualMachine) {
...
+
+	if !pkgcfg.FromContext(ctx).Features.VMResize {
+		return false, nil
+	}
+
+	if vm.Spec.ClassName == oldVM.Spec.ClassName {
+		return false, nil
+	}
+
+	if _, _, _, ok := vmopv1util.GetLastResizedAnnotation(*vm); !ok {
```
bryanv, line 473: I know I mentioned some thoughts about removing the annotation if the class name changes back to what's saved in the annotation but I think in some situations that can result in some unexpected results so that's punted for now until I can think a bit more about it.

sreyasn:
Nit: can we rename commit and MR description to "Resize Brownfield VMs"

webhooks/virtualmachine/mutation/virtualmachine_mutator_unit_test.go
```diff
@@ -1099,4 +1100,44 @@ func unitTestsMutating() {
...
+				updated, err := mutation.SetLastResizeAnnotation(&ctx.WebhookRequestContext, ctx.vm, oldVM)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(updated).To(BeTrue())
+
+				className, _, _, exists := vmopv1util.GetLastResizedAnnotation(*ctx.vm)
+				Expect(exists).To(BeTrue())
+				Expect(className).To(Equal(oldVM.Spec.ClassName))
+			})
+		})
+	})
```
dilyar85, line 1142: Maybe worth adding another test case where the class name is changed, but the VM already has the last-resize annotation set (to validate that it doesn't overwrite the existing last-resize annotation value)?

dilyar85 (approved):

sreyasn (approved):

bryanv:
> Nit: can we rename commit and MR description to "Resize Brownfield VMs"

Sure

PR #592, opened 2024-06-21
✨Resize: Memory Reservation Locked to Max

sreyasn:

Support reconfiguring VMs with the Memory Reservation Locked to Max. Additionally, this change makes sure that the setting cannot be accidentally set to false when the desired config spec has PCI passthrough devices

Fixes # N/A

N/A

```
Support resize of Memory Reservation Locked to Max
```

### DIFF
--- a/pkg/util/resize/configspec.go
+++ b/pkg/util/resize/configspec.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/virtualmachine"
 	pkgutil "github.com/vmware-tanzu/vm-operator/pkg/util"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/ptr"
 )
 
 // CreateResizeConfigSpec takes the current VM state in the ConfigInfo and compares it to the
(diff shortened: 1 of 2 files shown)


pkg/util/resize/configspec.go
```diff
@@ -463,6 +466,23 @@ func compareVmxStatsCollectionEnabled(
...
 
+// compareMemoryReservationLockedToMax compares full memory reservation settings from config spec.
+func compareMemoryReservationLockedToMax(
+	ci vimtypes.VirtualMachineConfigInfo,
+	cs vimtypes.VirtualMachineConfigSpec,
+	outCS *vimtypes.VirtualMachineConfigSpec) {
+
+	memLockedMax := cs.MemoryReservationLockedToMax
+	if memLockedMax != nil && !*memLockedMax {
+		// If the desired configSpec has PCI pass-through devices, make sure memoryReservationLockedToMax is not false
```
bryanv, line 477: After the refactor of this function, maybe reword this a little - " .... memoryReservationLockedToMax must be true because ... " since at this point we know that locked is `&false`

bryanv (approved):

PR #594, opened 2024-06-24
✨ Add vm.spec.cdrom API and initial reconcile logic

dilyar85:

This PR introduces a new `vm.spec.cdrom` field and the initial reconciliation logic that adds/removes/updates CD-ROM devices, backed by an ISO type Virtual Machine image.

Fixes N/A.

This is a draft PR to receive early feedback on the implementation. The workflow has been verified in an internal testbed, and the following unit tests will be added to this PR:

- [ ] Add `cdrom_test.go` with full coverage for the new `cdrom.go` file.
- [ ] Add more tests in `session_vm_update_test.go` with the ISO FSS enabled. 

```release-note
Add vm.spec.cdrom API and initial reconcile logic to support VM deployment from ISO.
```


api/v1alpha3/virtualmachine_types.go
```diff
@@ -236,8 +236,44 @@ type VirtualMachineImageRef struct {
...
 }
 
+// VirtualMachineCdromSpec describes the desired state of a CD-ROM device.
+type VirtualMachineCdromSpec struct {
+	// +required
+
+	// ImageName is the Kubernetes object name of the VirtualMachineImage or
+	// ClusterVirtualMachineImage resource of type ISO that is used as the
+	// backing for the CD-ROM device.
+	ImageName string `json:"imageName"`
```
akutz, line 246: Should this be simply `image` and it refer to a `type CdromImageRef` where the name is required and kind is optional?

api/v1alpha3/virtualmachine_types.go
```diff
@@ -236,8 +236,44 @@ type VirtualMachineImageRef struct {
...
+	// device. It may be either a VirtualMachineImage (default value) or a
+	// ClusterVirtualMachineImage.
+	ImageKind string `json:"imageKind,omitempty"`
+
+	// +optional
+
+	// Disconnect indicates whether the CD-ROM device should be disconnected
+	// from the VM. If omitted or set to false, the CD-ROM device is added and
+	// connected to the VM when the VM is powered on.
+	Disconnect bool `json:"connect,omitempty"`
```
akutz, line 262: I assume the intent is that setting this to true also means the customer can expect the CD-ROM to be immediately disconnected for a running VM?

api/v1alpha3/virtualmachine_types.go
```diff
@@ -236,8 +236,44 @@ type VirtualMachineImageRef struct {
...
 type VirtualMachineSpec struct {
+	// +optional
+
+	// Cdrom describes the desired state of the VM's CD-ROM devices.
+	//
+	// For each CD-ROM device, an ISO-type VirtualMachineImage must be specified
+	// as the backing for the CD-ROM device.
+	//
+	// If multiple CD-ROM devices are specified, they will be attached with the
+	// exact order as they are listed in this field/
```
akutz, line 275: * There is a trailing `/` in the last sentence.
* Please consider rewriting the final sentence as `CD-ROM devices are attached in list-order.`

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,217 @@
...
+const (
+	vmiKind  = "VirtualMachineImage"
+	cvmiKind = "Cluster" + vmiKind
+
+	serverIDSuffix = "?serverId="
+)
+
+// UpdateCdromDeviceChanges updates the CD-ROM devices in the VM ConfigSpec to
+// reflect the desired state in the K8s VM Spec.
+func UpdateCdromDeviceChanges(
```
akutz, line 29: I wonder if this should be a variation on the `EnsureAllDisksHaveControllers` function, ex. `EnsureAllCDROMsHaveControllers`. It would only need to consider IDE and AHCI (SATA) controllers, so it would be much simpler...

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,217 @@
...
+	vmCtx pkgctx.VirtualMachineContext,
+	client ctrlclient.Client,
+	curDevices object.VirtualDeviceList) ([]vimtypes.BaseVirtualDeviceConfigSpec, error) {
+	if vmCtx.MoVM.Summary.Runtime.PowerState == vimtypes.VirtualMachinePowerStatePoweredOn {
```
akutz, line 33: This function should not check the power state. Or rather, I would like to see this function decomposed into one that does not care about the Kubernetes objects at all. For example:

```golang
// EnsureAllCDROMsHaveControllers ensures that all CD-ROM devices
// in the provided ConfigSpec point to a controller. If no controller
// exists, IDE and SATA controllers are added to the ConfigSpec as
// necessary for the CD-ROM devices.
//
// Please note the following table for the number of controllers of each type
// that are supported as well as how many devices (per controller) each
// supports:
//
// IDE
//   - controllers                                    2
//   - devices                                        2
//
// SATA
//   - controllers                                    4
//   - devices                                       30
func EnsureAllCDROMsHaveControllers(
     configSpec *vimtypes.ConfigSpec,
     existingDevices ...vimtypes.BaseVirtualDevice) error {

}

**Review by akutz — CHANGES_REQUESTED** [review]
Great work @dilyar85! Please see the requested changes.

File: pkg/util/vsphere/vm/cdrom.go [hunk 6]
```diff
@@ -0,0 +1,217 @@
...
+
+	if itemStatus.Type != imgregv1a1.ContentLibraryItemTypeIso {
+		return "", fmt.Errorf("expected ISO type image, got %s", itemStatus.Type)
+	}
+	if len(itemStatus.FileInfo) == 0 || itemStatus.FileInfo[0].StorageURI == "" {
+		return "", fmt.Errorf("no storage URI found in the content library item of image")
+	}
+
+	uri := itemStatus.FileInfo[0].StorageURI
+	if strings.Contains(uri, serverIDSuffix) {
```
dougm, line 150: With StorageURI currently being:
```
${datastore.info.url}/contentlib-${lib_id}/${item_id}/ubuntu-impish-21.10-cloudimg_${file_id}.ovf?serverId=${server_id}
```
We need to strip out `?serverId=${server_id}` here. And vCenter converts the datastore info.url to name:
```
[sharedVmfs-0] contentlib-${lib_id}/${item_id}/ubuntu-impish-21.10-cloudimg_${file_id}.ovf
```
Should we just change image-registry-operator's FileInfo[].StorageURI to to same format as above?
It would be easier to read and usable for file backing as-is. I can open an MR for the change.

Note, can find datastore.info.url with this property:
```console
% govc object.collect -s /*.wcp-sanity/datastore/sharedVmfs-0 info.url
ds:///vmfs/volumes/666d7a79-cb0d28b2-57c8-0645602e1b58/
```

dilyar85: Yep, I added a TODO comment to address this once it's supported in govmomi.

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,360 @@
...
+	var itemStatus imgregv1a1.ContentLibraryItemStatus
+	switch imageRef.Kind {
+	case vmiKind:
+		ns := vmCtx.VM.Namespace
+		var vmi vmopv1.VirtualMachineImage
+		if err := client.Get(vmCtx, ctrlclient.ObjectKey{Name: imageRef.Name, Namespace: ns}, &vmi); err != nil {
+			return "", err
+		}
+		var clitem imgregv1a1.ContentLibraryItem
+		if err := client.Get(vmCtx, ctrlclient.ObjectKey{Name: vmi.Spec.ProviderRef.Name, Namespace: ns}, &clitem); err != nil {
```
bryanv, line 297: The image types' `Spec.ProviderRef` could be nil

dougm (approved):
lgtm!

dilyar85:
Submitted https://github.com/vmware-tanzu/vm-operator/pull/611 to address the `ci/vulncheck-go` check failure.

config/crd/external-crds/imageregistry.vmware.com_clustercontentlibraryitems.yaml
```diff
@@ -205,6 +205,13 @@ spec:
                         in bytes on storage in vCenter.
                       pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                       x-kubernetes-int-or-string: true
+                    storageURI:
+                      description: |-
+                        StorageURI identifies the file on the storage backing. It is specific to
+                        the storage backing and available after the file is cached in vCenter.
+                        This URL is useful for creating a device that is backed by this file
```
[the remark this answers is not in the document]
dilyar85: Thanks for pointing this out. As this is from the image-registry-operator API, I will submit a change there to correct it.

config/crd/external-crds/imageregistry.vmware.com_contentlibraries.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     listKind: ContentLibraryList
     plural: contentlibraries
     shortNames:
-    - cl
```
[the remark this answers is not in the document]
dilyar85: I don't think so because it is already used for `cluster` in later K8s versions.

config/crd/external-crds/imageregistry.vmware.com_contentlibraryitems.yaml
```diff
@@ -205,6 +205,13 @@ spec:
                         in bytes on storage in vCenter.
                       pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                       x-kubernetes-int-or-string: true
+                    storageURI:
+                      description: |-
+                        StorageURI identifies the file on the storage backing. It is specific to
+                        the storage backing and available after the file is cached in vCenter.
+                        This URL is useful for creating a device that is backed by this file
```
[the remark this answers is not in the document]
dilyar85: Will address it in the image-registry-operator API

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,388 @@
...
+)
+
+// UpdateCdromDeviceChanges reconciles the desired CD-ROM devices specified in
+// VM.Spec.Cdrom with the current CD-ROM devices in the VM. It returns a list of
+// device changes required to update the CD-ROM devices.
+func UpdateCdromDeviceChanges(
+	vmCtx pkgctx.VirtualMachineContext,
+	client ctrlclient.Client,
+	curDevices object.VirtualDeviceList) ([]vimtypes.BaseVirtualDeviceConfigSpec, error) {
+	var (
```
akutz, line 35: Please leave a blank line between the function signature and the `var` statement.

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,388 @@
...
+}
+
+// UpdateConfigSpecCdromDeviceConnection updates the connection state of the
+// VM's existing CD-ROM devices to match what specifies in VM.Spec.Cdrom list.
+func UpdateConfigSpecCdromDeviceConnection(
+	vmCtx pkgctx.VirtualMachineContext,
+	client ctrlclient.Client,
+	config *vimtypes.VirtualMachineConfigInfo,
+	configSpec *vimtypes.VirtualMachineConfigSpec) error {
+	var (
```
akutz, line 109: Please leave a blank line between the function signature and the `var` statement.

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,388 @@
...
+
+	for _, specCdrom := range cdromSpec {
+		imageRef := specCdrom.Image
+		if imageRef == nil {
+			return fmt.Errorf("image ref is nil for CD-ROM: %q", specCdrom.Name)
+		}
+
+		cdrom, bFileName, err := getCdromAndBackingByImgRef(vmCtx, client, *imageRef, curDevices)
+		if err != nil {
+			vmCtx.Logger.Error(err, "failed to get CD-ROM device by image ref", "imageRef", *imageRef)
```
akutz, line 124: Please see my previous comments about returning a contextual error instead of logging the error and returning it.

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,388 @@
...
+// name from the given image reference.
+func getCdromAndBackingByImgRef(
+	vmCtx pkgctx.VirtualMachineContext,
+	client ctrlclient.Client,
+	imageRef vmopv1.VirtualMachineImageRef,
+	curDevices object.VirtualDeviceList) (vimtypes.BaseVirtualDevice, string, error) {
+
+	fileName, err := getIsoFilenameFromImageRef(vmCtx, client, imageRef)
+	if err != nil {
+		vmCtx.Logger.Error(err, "failed to get ISO file name from image ref", "imageRef", imageRef)
```
akutz, line 158: Please see my previous comments about returning a contextual error instead of logging the error and returning it.

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,388 @@
...
+// ensureAllCdromsHaveControllers ensures all CD-ROM device are assigned to an
+// available controller on the VM. First connect to one of the VM's two IDE
+// channels if one is free. If they are both in use, connect to the VM's SATA
+// controller if one is present. Otherwise, add a new AHCI (SATA) controller to
+// the VM and assign the CD-ROM to it.
+// It returns a list of device changes if a new controller is added.
+func ensureAllCdromsHaveControllers(
+	expectedCdromDevices map[string]vimtypes.BaseVirtualDevice,
+	curDevices object.VirtualDeviceList) []vimtypes.BaseVirtualDeviceConfigSpec {
+	var newControllerChanges []vimtypes.BaseVirtualDeviceConfigSpec
```
akutz, line 211: Please leave a blank line between the function signature and the `var` statement.

akutz (changes requested):

akutz (changes requested):
Thank you @dilyar85, this is really good, but I did have a few thoughts.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -242,8 +242,69 @@ type VirtualMachineImageRef struct {
...
+
+	// Connect describes the desired connection state of the CD-ROM device.
+	//
+	// When omitted or true, the CD-ROM device is added and connected to the
+	// VM. If the device already exists, it's updated to a connected state.
+	//
+	// When explicitly set to false, the CD-ROM device is added but remains
+	// disconnected from the VM. If the CD-ROM device already exists, it's
+	// updated to a disconnected state.
+	Connect bool `json:"connect"`
```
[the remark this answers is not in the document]
dilyar85: CC @aruneshpa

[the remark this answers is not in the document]
dilyar85: Changed to `Connected` for now and going to resolve this thread.

Will use the other thread for choosing between "Connected" or "Disconnected".

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,388 @@
...
+	// vCenter converts this to a different format during the CD-ROM creation:
+	// "[sharedVmfs-0] contentlib-${lib_id}/${item_id}/...{file_id}.iso"
+	//
+	// Therefore, we should return the storage URI in the vCenter format to
+	// ensure the existing CD-ROMs can be fetched by the backing file name.
+	//
+	// TODO(Sai): Update image-registry-operator to put storageURI in the
+	// vCenter format in content library item CR status and remove below lines.
+	uri := itemStatus.FileInfo[0].StorageURI
+	if idex := strings.Index(uri, serverIDSuffix); idex != -1 {
```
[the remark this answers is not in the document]
dilyar85: This is not needed anymore since we have the internal image-registry-operator updated with the expected format of URI stored in the content library item's CR status.  

Thanks for providing the example. I will keep that in mind for future similar changes.

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -0,0 +1,388 @@
...
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/util"
+)
+
+const (
+	vmiKind  = "VirtualMachineImage"
+	cvmiKind = "Cluster" + vmiKind
+
+	serverIDSuffix = "?serverId="
```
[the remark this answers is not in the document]
dilyar85: Deleted this constant as mentioned in the previous comment.

api/v1alpha3/virtualmachine_types.go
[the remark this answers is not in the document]
dilyar85: So this was named "Disconnected" initially and later changed to "Connected" after discussing with the team internally last week. The main concern was that "Disconnected" could be confusing in the code as it's negative and especially when used in a conditional statement (e.g. `if !cdrom.disconnected` to check if it's connected).

I don't have a strong preference between them but might still lean towards "disconnected" so that it makes more sense in the API. As API fields are more user-facing, I think we should prioritize choosing a more user-friendly term than an implementation detail inside the controller reconciliation process. 

@aruneshpa If you don't feel strongly about this, I can change this back to "disconnected" (or open to discussing any other name if you have other thoughts/concerns).

api/v1alpha3/virtualmachine_types.go
```diff
@@ -242,8 +242,70 @@ type VirtualMachineImageRef struct {
...
+type VirtualMachineCdromSpec struct {
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern="^[a-z0-9]{2,}$"
+
+	// Name describes the unique and immutable name of this CD-ROM device.
+	// If omitted or empty, the controller will generate it as "cdrom" + "num",
+	// where "num" is the device's index in the spec.cdrom list.
+	Name string `json:"name"`
+
+	// +kubebuilder:validation:Required
```
aruneshpa, line 255: Hmm.  Any reason for not relying on the absence of `+optional` tag? https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

aruneshpa (approved):
LGTM.  Thanks for the update.

api/v1alpha3/virtualmachine_types.go
dilyar85: Good point! Removed the unnecessary marker. While reading this doc, I also realized the CD-ROM's image field should not be a pointer since it's required and addressed that too in the latest commit.

Btw, I changed the `connect` field back to `disconnected` as discussed in [this thread](https://github.com/vmware-tanzu/vm-operator/pull/594#discussion_r1670657439) above. It isn't visible in the PR's file view as that comment is on an outdated line. Just wanted to make sure you're aware and aligned with this update.

api/v1alpha3/virtualmachine_types.go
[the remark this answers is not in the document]
dilyar85: Updated this field back to `Disconnected`: https://github.com/vmware-tanzu/vm-operator/pull/594/files#diff-3d050153d168a9ae5555513d460312ace08ab8ed5cc51a751fb7bd9375e4ba6cR264-R272

dilyar85:
To wrap up the above conversations about the "connection state" in the CD-ROM spec, the team has concluded to name that API field `connected` with a default true value. The PR has been updated to reflect this change.

akutz (approved):
Thank you @dilyar85 !!

PR #596, opened 2024-06-26
🌱 Handle more ExtraConfig fields during resize

bryanv:

Support MMPowerOffVMExtraConfigKey and V1Alpha1ExtraConfigKey EC fields during resize.

Rename the devops resize to just overwrite since we do set all of this when we don't have a VM Class. This is another step towards being able to replace the pre-power on reconfigure with resize.

Sanitize the VM Class ConfigSpec by removing the now deprecated MMPowerOff EC field.

Fixes #

```release-note
NONE
```


pkg/util/vmopv1/resize_overwrite.go
```diff
@@ -85,6 +89,44 @@ func overrideMMIOSize(
...
+		&vimtypes.OptionValue{Key: constants.MMPowerOffVMExtraConfigKey, Value: ""},
+	}
+}
+
+func updateV1Alpha1CompatibleEC(
+	vm vmopv1.VirtualMachine,
+	ci vimtypes.VirtualMachineConfigInfo,
+	_ *vimtypes.VirtualMachineConfigSpec) []vimtypes.BaseOptionValue {
+
+	bs := vm.Spec.Bootstrap
```
sreyasn, line 115: Nit: Add a comment about why this check is needed. vAppConfig + LinuxPrep ==  OvfEnv ?.

pkg/util/configspec.go
```diff
@@ -143,6 +145,10 @@ func SanitizeVMClassConfigSpec(
 	// Empty VmProfiles as storage profiles are disk specific
 	configSpec.VmProfile = []vimtypes.BaseVirtualMachineProfileSpec{}
 
+	configSpec.ExtraConfig = OptionValues(configSpec.ExtraConfig).Delete(
```
sreyasn, line 148: more like cautionary removal from the class configSpec here?. In case the admin decides to add it?

sreyasn (approved):

pkg/util/configspec.go
bryanv: Yes. It is deprecated and we'll always remove it but potentially saves a step.

PR #614, opened 2024-07-03
🐛  Ignore VM Class specified default guestID "otherGuest64"

sreyasn:

The VM Class specified guestID is not supported. Ignore if set.

The guestID is either specified explicity by
- the VM spec (or)
- falls back to the VM Image/OVF

Fixes # N/A

N/A

```
Ignore VM Class provided guestID
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/configspec.go
+++ b/pkg/providers/vsphere/virtualmachine/configspec.go
@@ -115,9 +115,8 @@ func CreateConfigSpec(
 		}
 	}
 
-	// Initially set the guest ID in ConfigSpec to ensure VM is created with the expected guest ID.
+	// If VM Spec guestID is specified, initially set the guest ID in ConfigSpec to ensure VM is created with the expected guest ID.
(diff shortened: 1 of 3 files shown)


dilyar85 (approved):
LGTM!

pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -115,9 +115,11 @@ func CreateConfigSpec(
 		}
 	}
 
-	// Initially set the guest ID in ConfigSpec to ensure VM is created with the expected guest ID.
+	// VM Class specified guestID is not supported. Ignore if set.
+	configSpec.GuestId = ""
```
bryanv, line 119: Does it make more sense to do this in `SanitizeVMClassConfigSpec()`?

sreyasn: you're right. It does get called before the CreateConfigSpec(). Will move it there. Thanks BV.

bryanv (approved):

PR #619, opened 2024-07-08
✨ Storage usage in VirtualMachine API status

akutz:

This patch updates the VirtualMachine API with the following changes:

* Elements in `status.volumes` now include "classic" (non-PVC) virtual disks, which can be distinguished with the `status.volumes[].type` field, set to either `Classic` or `Managed`.

* Elements in `status.volumes` now report the disk's limit/capacity as well as current, unique usage. For example, if the disk were part of a linked clone, only the data unique to the clone would have its size reported.

* The field `status.storage` now reports the VM's total committed, uncommitted, and unshared (i.e. unique) storage usage.

All storage information is reported as Kubernetes resource quantities.

The new code should have 100% coverage, ex:

image

image

image

image

image

image

image

Fixes `NA`

```release-note
Updates VirtualMachine API status to report total storage usage as well as usage per-volume (classic and managed).
```


api/v1alpha3/virtualmachine_storage_types.go
```diff
@@ -66,12 +66,48 @@ type InstanceVolumeClaimVolumeSource struct {
...
 // VirtualMachineVolumeStatus defines the observed state of a
 // VirtualMachineVolume instance.
 type VirtualMachineVolumeStatus struct {
 	// Name is the name of the attached volume.
 	Name string `json:"name"`
 
+	// +kubebuilder:default=Managed
+
+	// Type is the type of the attached volume.
+	Type VirtualMachineVolumeType `json:"type"`
```
akutz, line 94: @bryanv Do you know if we can safely make this non-optional? Since it has a default value I would think so, but I am not sure without checking, and we don't have any in-repo examples for setting up a test for an existing status already recorded without a type. 

This is all complicated of course by the fact that volume status info is now handled in two distinct locations.

akutz:
@bryanv I still need to figure out how to contend with the fact the volume controller replaces status.volumes with every reconcile. The new logic is thread-safe, but the existing volume logic sure is not. I have to address this, somehow, before we merge this PR.

api/v1alpha3/virtualmachine_storage_types.go
bryanv: I think we can be OK - on the initial up version conversion - when we don't have the the curver saved in the VM annotation - just assume everything is Managed since that is what everything was before. If we get it wrong, it will be fixed up on a subsequent reconcile.

api/v1alpha3/virtualmachine_storage_types.go
```diff
@@ -90,3 +126,40 @@ type VirtualMachineVolumeStatus struct {
...
+
+// VirtualMachineStorageStatus defines the observed state of a VirtualMachine's
+// storage.
+type VirtualMachineStorageStatus struct {
+
+	// +optional
+
+	// ChangeBlockTracking describes whether or not change block tracking is
+	// enabled for the VirtualMachine.
+	// ChangeBlockTracking *bool `json:"changeBlockTracking,omitempty"`
```
bryanv, line 138: Still TBD if we're moving this here, or just hadn't gotten around to writing the conversion code yet?

controllers/volume/volume_controller.go
```diff
@@ -69,11 +69,14 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 	r := NewReconciler(
 		ctx,
 		mgr.GetClient(),
+		mgr.GetAPIReader(),
 		ctrl.Log.WithName("controllers").WithName("volume"),
 		record.New(mgr.GetEventRecorderFor(controllerNameLong)),
 		ctx.VMProvider,
 	)
 
+	r.reader = mgr.GetAPIReader()
```
bryanv, line 78: This was already assigned in `NewReconciler()`

controllers/volume/volume_controller.go
```diff
@@ -892,10 +907,37 @@ func sanitizeCNSErrorMessage(msg string) string {
...
 }
+
+func updateVolumeStatusWithLimit(
+	ctx *pkgctx.VolumeContext,
+	c client.Reader,
+	status *vmopv1.VirtualMachineVolumeStatus) error {
+
+	var (
+		pvc    corev1.PersistentVolumeClaim
+		pvcKey = client.ObjectKey{Namespace: ctx.VM.Namespace, Name: status.Name}
```
bryanv, line 927: That `Name` isn't necessarily the PVC name - you want the `VirtualMachineVolume.PersistentVolumeClaim.ClaimName` value

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -678,3 +682,145 @@ func updateGuestNetworkStatus(vm *vmopv1.VirtualMachine, gi *vimtypes.GuestInfo)
...
+					Limit:    BytesToResourceGiB(di.CapacityInBytes),
+					Used:     BytesToResourceGiB(di.UniqueSize),
+				})
+		}
+	}
+
+	// Remove any status entries for classic disks that no longer exist in
+	// config.hardware.device.
+	i := 0
+	for _, vol := range vm.Status.Volumes {
```
bryanv, line 800: Nit - maybe https://pkg.go.dev/slices#DeleteFunc moderately easier to read?

controllers/volume/volume_controller.go
akutz: Doh, and you already told me as much. Sorry about that.

pkg/providers/vsphere/vmlifecycle/update_status.go
akutz: I forgot they added this, thank you.

akutz: Actually, I am going to keep this current pattern, and you will see why in a moment.

api/v1alpha3/virtualmachine_storage_types.go
akutz: I want to move it, but I'm hesitant to do it in *this* PR. Originally I had moved it, along with CBT, to the new status.storage. But then I realized that only really makes sense if we move spec.cbt and spec.volumes to spec.storage as well, and I did not think we should combine those changes in this PR.

controllers/volume/volume_controller.go
```diff
@@ -892,10 +923,61 @@ func sanitizeCNSErrorMessage(msg string) string {
...
 	}
 }
+
+func updateVolumeStatusWithLimit(
+	ctx *pkgctx.VolumeContext,
+	c client.Reader,
+	status *vmopv1.VirtualMachineVolumeStatus) error {
+
+	// Find the name of the PVC.
+	var pvcName string
```
bryanv, line 942: The only caller of this already has the `volume.PersistentVolumeClaim.ClaimName` handy just pass it in here?

controllers/volume/volume_controller.go
```diff
@@ -643,7 +657,12 @@ func (r *Reconciler) processAttachments(
 			// CNS noop reconciles of an attachment that is already attached so we must delete the existing
 			// attachment and create a new one.
 			if volume.PersistentVolumeClaim.ClaimName == attachment.Spec.VolumeName {
-				volumeStatus = append(volumeStatus, attachmentToVolumeStatus(volume.Name, attachment))
+				volumeStatus := attachmentToVolumeStatus(volume.Name, attachment)
+				volumeStatus.Used = existingManagedVolUsage[volume.Name]
+				if err := updateVolumeStatusWithLimit(ctx, r.reader, *volume.PersistentVolumeClaim, &volumeStatus); err != nil {
```
akutz, line 662: @bryanv I updated this as we discussed. I left the instance storage optimization since it's free with a loop no longer required.

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -799,6 +802,20 @@ func Convert_v1alpha3_VirtualMachineStatus_To_v1alpha1_VirtualMachineStatus(
 	out.LastRestartTime = in.LastRestartTime
 	out.Conditions = translate_v1alpha3_Conditions_To_v1alpha1_Conditions(out.Conditions)
 
+	out.Volumes = nil
+	for i := range in.Volumes {
+		if in.Volumes[i].Type != vmopv1.VirtualMachineStorageDiskTypeClassic {
+
+			// Only down-convert volume statuses if the volume is managed.
```
bryanv, line 809: Total nit - maybe change the check above to be affirmative `Type == Managed` to match this comment closer, and be clearer that this is PVCs only in case a third type comes along

bryanv (approved):

api/v1alpha1/virtualmachine_conversion.go
akutz: I specifically did this to handle the case where it's not yet set to `Managed` for some reason.

PR #620, opened 2024-07-09
:bug: Revert to returning early if the pause annotation is present on the VM

aruneshpa:

In the recent changes to support the break-glass admin overrides, we changed the behavior of the pause annotation for regular (non delete) reconciles so that we now _always_ add the finalizer regardless of the pause annotation.

This change fixes this by reverting the old behavior of returning early if the pause annotation is present.  This also means that now both reconcile methods (Create/Update and Delete) behave consistently when this annotation is present.

```release-note
Revert to returning early if the pause annotation is present on the VM
```

### DIFF
--- a/controllers/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine_controller.go
@@ -303,10 +303,9 @@ func requeueDelay(ctx *pkgctx.VirtualMachineContext) time.Duration {
 func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineContext) (reterr error) {
 	ctx.Logger.Info("Reconciling VirtualMachine Deletion")
 
-	// If the VM reconciliation has been paused by the developer,
-	// skip deletion and return.
+	// Return early if the VM reconciliation is paused.
 	if _, exists := ctx.VM.Annotations[vmopv1.PauseAnnotation]; exists {
-		ctx.Logger.Info("VirtualMachine is not deleted because devops paused this VM")
+		ctx.Logger.Info("Skipping deletion since VirtualMachine contains the pause annotation")
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


akutz:
Does this mean that an admin cannot use pause to prevent the deletion of the underlying vSphere VM?

aruneshpa:
> Does this mean that an admin cannot use pause to prevent the deletion of the underlying vSphere VM?

Hmm.  This change only impacts the behavior of the the pause _annotation_ by the devops persona.  If an admin pauses the VM by setting the ExtraConfig field, the deletion of vSphere VM will still be skipped.

https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/virtualmachine/delete.go#L32-L47

akutz:
> > Does this mean that an admin cannot use pause to prevent the deletion of the underlying vSphere VM?
> 
> Hmm. This change only impacts the behavior of the the pause _annotation_ by the devops persona. If an admin pauses the VM by setting the ExtraConfig field, the deletion of vSphere VM will still be skipped.
> 
> https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/virtualmachine/delete.go#L32-L47

Then I misunderstood and am fine with this. Thanks!

zyiyi11:
> Does this mean that an admin cannot use pause to prevent the deletion of the underlying vSphere VM?

IIUC, with this change:
- If admin paused the VM by using EC key, VM has the finalizer, during deletion, CR and vSphere VM not deleted. see: https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachine/virtualmachine_controller.go#L319

- If devops paused the VM by annotation, VM returns early in reconcile, it doesn't have the finalizer, during deletion, CR deleted and  vSphere VM remains.

- If both admin and devops paused this VM, VM returns early in reconcile, it doesn't have the finalizer. during deletion, CR deleted and  vSphere VM remains.

akutz (approved):

PR #622, opened 2024-07-10
🐛 Set VirtualDisk.UnitNumber when placing with multiple clusters

dougm:
UnitNumber is required for PlaceVmsXCluster w/ VpxdVmxGeneration fss enabled

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/configspec_test.go
+++ b/pkg/providers/vsphere/virtualmachine/configspec_test.go
@@ -445,8 +445,10 @@ var _ = Describe("CreateConfigSpecForPlacement", func() {
 			Expect(ok).To(BeTrue())
 			*/
 			dSpec2 := configSpec.DeviceChange[1].GetVirtualDeviceConfigSpec()
-			_, ok = dSpec2.Device.(*vimtypes.VirtualDisk)
+			disk, ok := dSpec2.Device.(*vimtypes.VirtualDisk)
 			Expect(ok).To(BeTrue())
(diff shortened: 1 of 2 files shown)


bryanv (approved):

pkg/providers/vsphere/virtualmachine/configspec_test.go
```diff
@@ -445,8 +445,9 @@ var _ = Describe("CreateConfigSpecForPlacement", func() {
 			Expect(ok).To(BeTrue())
 			*/
 			dSpec2 := configSpec.DeviceChange[1].GetVirtualDeviceConfigSpec()
-			_, ok = dSpec2.Device.(*vimtypes.VirtualDisk)
+			disk, ok := dSpec2.Device.(*vimtypes.VirtualDisk)
 			Expect(ok).To(BeTrue())
+			Expect(disk.UnitNumber).ToNot(BeNil())
```
dilyar85, line 450: Nit: Should we also check if it has the expected value of 0? I've seen errors when adding a CD-ROM device at a different unit number. Not sure if it's the same for a virtual disk but might be worth verifying it here.

dilyar85 (approved):
Thanks for fixing this!

pkg/providers/vsphere/virtualmachine/configspec_test.go
dougm: We can expect that here, added. In general, iirc, just need to make sure UnitNumber doesn't conflict with another device. And doesn't exceed the controller's max slots (e.g. SCSI is 16)

PR #626, opened 2024-07-15
🌱 StoragePolicyQuota controller

akutz:

This patch adds the controller that creates StoragePolicyUsage resources for StoragePolicyQuota resources applied to namespaces.

Fixes #

```release-note
StoragePolicyQuota controller creates StoragePolicyUsage resources for tracking VM usage
```

### DIFF
--- a/controllers/storagepolicyquota/storagepolicyquota_controller.go
+++ b/controllers/storagepolicyquota/storagepolicyquota_controller.go
@@ -1,0 +1,141 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package storagepolicyquota
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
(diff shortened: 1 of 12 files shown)


controllers/storagepolicyquota/storagepolicyquota_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	var obj spqv1.StoragePolicyQuota
+	if err := r.Get(ctx, req.NamespacedName, &obj); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	logger := ctrl.Log.WithName("StoragePolicyQuota").WithValues("name", req.NamespacedName)
+
+	// Ensure the GVK for the object is synced back into the object since
+	// the object's APIVersion and Kind fields may be used later.
+	kubeutil.MustSyncGVKToObject(&obj, r.Scheme())
```
bryanv, line 95: Is there something in this MR that needs this, or something that is coming later?

bryanv (approved):

controllers/storagepolicyquota/storagepolicyquota_controller.go
akutz: It's used to grab the API version from the object later. I can drop it and just grab it from the package instead if you prefer.

akutz: Ah, I remember now, it was for the SetControllerReference call at https://github.com/vmware-tanzu/vm-operator/pull/626/files#diff-933f182d85da229e61378c9b984cd5e434156ba585d6edc5c6258a3ce72b4504R123. It complained when the object lacked the GVK.

akutz: You can see where that function uses it at https://github.com/kubernetes-sigs/controller-runtime/blob/v0.18.4/pkg/controller/controllerutil/controllerutil.go#L81

abaruni (approved):

PR #627, opened 2024-07-15
✨ resize: Start to handle more device types

bryanv:

Start to flesh out handling resize of some of the easier device types. For these types, zip the expected and current devices of the same type together, and add, remove or edit as needed to update the VM configuration.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/resize/configspec_devices.go
+++ b/pkg/util/resize/configspec_devices.go
@@ -5,6 +5,7 @@ package resize
 
 import (
 	"reflect"
+	"slices"
 
 	"github.com/vmware/govmomi/object"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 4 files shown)


pkg/util/resize/configspec_devices.go
```diff
@@ -25,16 +26,633 @@ func compareHardwareDevices(
...
+			deviceChanges = append(deviceChanges, &vimtypes.VirtualDeviceConfigSpec{
+				Operation: vimtypes.VirtualDeviceConfigSpecOperationEdit,
+				Device:    editDev,
+			})
+		}
+	}
+
+	// Add new expected devices.
+	for idx := range expectedDevs[minLen:] {
+		deviceChanges = append(deviceChanges, &vimtypes.VirtualDeviceConfigSpec{
```
sreyasn, line 202: Prb wanna ignore defaultDevice 'Adds' too?. (or) do we wanna let it go to vpxd and let it drop

pkg/util/resize/configspec_devices.go
```diff
@@ -25,16 +26,633 @@ func compareHardwareDevices(
...
+
+	// For the expected and current devices of a type, zip the devices together to determine
+	// what, if any, edits need to be done. Then add and remove what is leftover.
+	minLen := min(len(expectedDevs), len(curDevs))
+	for idx := range minLen {
+		expectedDev, curDev := expectedDevs[idx], curDevs[idx]
+
+		editDev := matchDevFn(expectedDev, curDev)
+		if editDev != nil {
+			deviceChanges = append(deviceChanges, &vimtypes.VirtualDeviceConfigSpec{
```
sreyasn, line 193: IIRC, VMCI Device and Video Cards are default devices that can have settable properties on reconfigure -- edits are definitely possible. 
I'm unsure about an 'Edit' on a default device like PointingDevice though. need to verify.

sreyasn (commented):
first pass of the PR.

pkg/util/resize/configspec_devices.go
bryanv: The doc at least suggest the the host pointing dev field in the backing can be updated. However, using one of the allowed values from the VirtualPointingDeviceOptions it still fails. There might be something odd going with the field in general because on create valid values from the options appear to be ignore and always ends up as autodetect.

I might just comment out the PointingDevice checks for now since it won't be a very common thing for our usages.

pkg/util/resize/configspec_devices.go
bryanv: I'd prefer to let it go. First, if a default device somehow got removed and it is in the class's ConfigSpec we should try to re-add it. Second, and most common would be that the class's ConfigSpec is invalid in that it has multiple adds for a default device. I think we should still add it and let it fail, since things can already fail for a million other reasons.

sreyasn: makes sense. 
Just to put out there. 
AFAIK,  In the first case, the re-add will be ignored by vpxd in the 'reconfigure' call. So it won't make a difference and we can let vpxd make the call.  
The second case has potential for failure if invalid, true.

sreyasn (approved):

PR #631, opened 2024-07-18
:seedling: Introduce the ExtraConfig key to allow opt-out from auto register

aruneshpa:

Currently, we will unconditionally try to register a VM when it is restored using a backup/restore vendor, or a DR solution.  This isn't always ideal if the solution wants to prepare the infra, or customize the VM on vSphere before registering the VM with Supervisor.

This change introduces an ExtraConfig key that can be used to opt-out of the automatic registration workflow.

```release-note
Introduce the ExtraConfig key to allow opt-out from auto register
```

### DIFF
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -161,6 +161,22 @@ const (
 	// PVCDiskDataExtraConfigKey is the ExtraConfig key to persist the VM's
 	// PVC disk data in JSON, compressed using gzip and base64-encoded.
 	PVCDiskDataExtraConfigKey = "vmservice.virtualmachine.pvc.disk.data"
+
+	// DisableAutoRegistrationExtraConfigKey is the ExtraConfig key that can be
+	// set to "true" (case insensitive) on a virtual machine to opt-out of the
+	// automatic registration workflow.
+	//
+	// A "registration" refers to adopting a virtual machine so it is managed by
+	// VM operator on Supervisor. Typically, this involves creating a new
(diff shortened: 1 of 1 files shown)


api/v1alpha3/virtualmachine_types.go
```diff
@@ -161,6 +161,17 @@ const (
...
+	// DisableAutoRegistrationExtraConfigKey is the ExtraConfig key that can be
+	// set to "true" (case insensitive) on a virtual machine to opt-out of the
+	// automatic registration workflow.
+	//
+	// After a restore from a backup restore vendor, or a failover from a disaster
+	// recovery solution, vCenter automatically attempts to register the restored
+	// virtual machine with Supervisor. If the automatic registration is skipped by
+	// specifying this key, backup/restore and/or disaster recovery solutions are
+	// responsible to register the VM with Supervisor.
+	DisableAutoRegistrationExtraConfigKey = "vmservice.virtualmachine.disable-automatic-registration"
```
dilyar85, line 174: Not sure if there is any convention on the EC key name, but most of the existing ones use "." as a separator.

aruneshpa: The "." separator is used between things that provide a hierarchy (e.g., "vmservice", and "virtualmachine").  As far as the "disable-automatic-registration" is concerned, I could either use camel case of snake case.  I am seeing references for both in the existing EC keys.  Going to use camel case since that's what is more prevalent.

dilyar85 (approved):

bryanv (approved):

PR #632, opened 2024-07-18
✨ Reduce dependency graph for API module

akutz:

This patch updates the API module so its dependency graph is greatly reduced. The dependency on controller-runtime for things other than conversion was creating compile-time build issues for consumers of the API module.

The change encompasses three things:

1. Following the guide https://main.cluster-api.sigs.k8s.io/developer/providers/implementers-guide/create_api#registering-apis-in-the-scheme which demonstrates how to remove controller-runtime for registering API types with the scheme builder.

2. Moving the webhook registrations out of the API module. The module already did not have the validation/mutation hooks, but did, and still, has the conversion hooks since those *must* be in the API package. However, their hook registrations were moved out of the API module since the method of registration depended upon controller-runtime.

3. Removed the dependency on the testlabels module and just used the strings "api" and "fuzz" directly.

Fixes `NA`

```release-note
Reduce dependency graph for API module.
```


dougm (approved):
lgtm

dilyar85 (approved):
LGTM!

webhooks/conversion/v1alpha1/webhooks.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2023-2024 VMware, Inc. All Rights Reserved.
```
aruneshpa, line 1: I think this just needs to be 2024?  Same in other files.

aruneshpa (approved):
One nit related to copyright years.  LGTM otherwise.

webhooks/conversion/v1alpha1/webhooks.go
akutz: Ah, yeah, I moved them from the original location and just added the 2024. I only renamed them later. I'll update, thanks!

bryanv (approved):

PR #634, opened 2024-07-18
🐛 Only set vmPathName when using ISO image

dougm:
The storage policy datastore query is only needed for ISO images.

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm_test.go
+++ b/pkg/providers/vsphere/vmprovider_vm_test.go
@@ -332,6 +332,7 @@ func vmTests() {
 					propName        = "config.name"
 					propPowerState  = "runtime.powerState"
 					propExtraConfig = "config.extraConfig"
+					propPathName    = "config.files.vmPathName"
 				)
 				var (
 					err           error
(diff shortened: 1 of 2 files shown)


dilyar85 (approved):
LGTM!

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -382,6 +383,16 @@ func vmTests() {
...
 					})
 				})
+				When("getting "+propPathName, func() {
+					BeforeEach(func() {
+						propertyPaths = []string{propPathName}
+					})
+					It("should not be set", func() {
+						Expect(err).ToNot(HaveOccurred())
+						Expect(result).To(HaveLen(1))
+						Expect(result[propName]).To(BeNil()) // should only be set if cdrom is present
```
zyiyi11, line 393: Should we add a positive test where `cdrom` is present and `result[propName]` populated?

dougm: The positive test was already there, see: https://github.com/vmware-tanzu/vm-operator/pull/634/files#diff-407bbe7f375ec7c2885d2b2e0cb6c745e2c5a3ecd209b91a82905a296e92e079L2112

PR #636, opened 2024-07-19
:seedling: Rename the pause-reconcile annotation to paused

aruneshpa:

This change renames the "pause-reconcile" annotation used to pause the reconciliation of a VM to "paused".  This is consistent with other components using "paused" to indicate an entity being paused.

Additionally, introduce a helper package that can be used to check if an annotation is present (regardless of the value) in a readable way.

**Testing Done:**
1. Create a VM on v1alpha3 version with `paused` annotation set.
```
# k get --raw  https://127.0.0.1:6443/apis/vmoperator.vmware.com/v1alpha3/namespaces/parunesh-ns/virtualmachines/parunesh-vm | jq -r '.metadata.annotations | to_entries[] | select(.key == "vmoperator.vmware.com/paused")'
{
  "key": "vmoperator.vmware.com/paused",
  "value": "something"
}
```
2. Ensure that the VM doesn't have `pause-reconcile` annotation set
```
# k get --raw  https://127.0.0.1:6443/apis/vmoperator.vmware.com/v1alpha3/namespaces/parunesh-ns/virtualmachines/parunesh-vm | jq -r '.metadata.annotations | to_entries[] | select(.key == "vmoperator.vmware.com/pause-reconcile")'
#
```

3. Fetch the resource at v1alpha2 and v1alpha1 version and ensure that the `pause-reconcile` annotation is set.
```
# k get --raw  https://127.0.0.1:6443/apis/vmoperator.vmware.com/v1alpha2/namespaces/parunesh-ns/virtualmachines/parunesh-vm | jq -r '.metadata.annotations | to_entries[] | select(.key == "vmoperator.vmware.com/pause-reconcile")'
{
  "key": "vmoperator.vmware.com/pause-reconcile",
  "value": "something"
}
# k get --raw  https://127.0.0.1:6443/apis/vmoperator.vmware.com/v1alpha1/namespaces/parunesh-ns/virtualmachines/parunesh-vm | jq -r '.metadata.annotations | to_entries[] | select(.key == "vmoperator.vmware.com/pause-reconcile")'
{
  "key": "vmoperator.vmware.com/pause-reconcile",
  "value": "something"
}
```

```release-note
Rename the pause-reconcile annotation to paused
```

### DIFF
--- a/api/v1alpha1/virtualmachine_conversion.go
+++ b/api/v1alpha1/virtualmachine_conversion.go
@@ -1138,6 +1138,19 @@ func Convert_v1alpha3_VirtualMachine_To_v1alpha1_VirtualMachine(in *vmopv1.Virtu
 	// necessary for later versions that don't support configMap based bootstrap.
 	delete(out.Annotations, vmopv1.V1alpha1ConfigMapTransportAnnotation)
 
+	// Handle the renaming of pause annotation on down convert.
+	if val, ok := in.Annotations[vmopv1.PauseAnnotation]; ok {
+		if out.Annotations == nil {
+			out.Annotations = make(map[string]string)
+		}
+		out.Annotations[PauseAnnotation] = val
+
+		// Remove the pause annotation corresponding to the Hub.
+		// This would also remove the annotation if someone created a
+		// v1a1 VM with "paused" annotation.
+		delete(out.Annotations, vmopv1.PauseAnnotation)
+	}
+
 	return nil
 }
 
@@ -1180,6 +1193,20 @@ func Convert_v1alpha1_VirtualMachine_To_v1alpha3_VirtualMachine(in *VirtualMachi
 			annotations[vmopv1.V1alpha1ConfigMapTransportAnnotation] = "true"
 			out.GetObjectMeta().SetAnnotations(annotations)
 		}
+	}
+
+	// Handle the renaming of pause annotation on up convert.
+	if val, ok := in.Annotations[PauseAnnotation]; ok {
+		annotations := out.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[vmopv1.PauseAnnotation] = val
+
+		// Remove the pause annotation from v1alpha1.
+		// This would also remove the annotation if someone created a
+		// v1a1 VM with "pause-reconcile" annotation.
+		delete(annotations, PauseAnnotation)
 	}
 
 	return nil
--- a/pkg/util/annotations/helpers.go
+++ b/pkg/util/annotations/helpers.go
@@ -13,6 +13,10 @@ func HasForceEnableBackup(o metav1.Object) bool {
 	return hasAnnotation(o, vmopv1.ForceEnableBackupAnnotation)
 }
 
+func HasPaused(o metav1.Object) bool {
+	return hasAnnotation(o, vmopv1.PauseAnnotation)
+}
+
 func hasAnnotation(o metav1.Object, annotation string) bool {
 	annotations := o.GetAnnotations()
 	if annotations == nil {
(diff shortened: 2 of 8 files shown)


akutz:
I did realize a concern last night -- will this be confused with a VM being paused as in being in a suspended state?

aruneshpa:
> I did realize a concern last night -- will this be confused with a VM being paused as in being in a suspended state?

I think this should be fine since the suspended state is conveyed through spec and status.  It also helps that this is clearly called out in the API docs.

aruneshpa:
@bryanv and @akutz , the tricky thing I want to call out is, after this change, I am proactively removing the label that doesn't correspond to the destination schema version.  For example, if a v1a3 VM was created with *both* "pause-reconcile" and "paused" annotations, and this resource is fetched at v1a2, the conversion logic will drop the "paused" annotation from the resource (which the user had explicitly applied).  
This should be fine since these annotations are considered as reserved.  I would like your thoughts on this.

pkg/util/annotations/helpers.go
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package annotations
```
bryanv, line 4: How about moving these files under `pkg/util/vmopv1` instead? That's kind of mostly where we've been dumping util-y things that imports vmopv1

aruneshpa: Started making this change before I realized.  This *isn't* related to vmopv1 (even though it imports it).  This is a generic function which only operates on the object's metadata and checks whether an annotation (string) is present.  The schema is only imported for the strings :( 

I think a "pkg/util/vmopv1/annotations.HasPaused" package reads like it is specifically written for a schema (v1a3 in this case).  That's why I think it belongs in pkg/util subdir.  Let me know your thoughts.

bryanv: But it is using an annotation defined in our vmopv1 api. Outside of conversion everything just cares about the stoarage version.

bryanv:
> @bryanv and @akutz , the tricky thing I want to call out is, after this change, I am proactively removing the label that doesn't correspond to the destination schema version. For example, if a v1a3 VM was created with _both_ "pause-reconcile" and "paused" annotations, and this resource is fetched at v1a2, the conversion logic will drop the "paused" annotation from the resource (which the user had explicitly applied). This should be fine since these annotations are considered as reserved. I would like your thoughts on this.

I don't think that will be a big deal for this annotation. Note though that we do have the hub version stashed away in the annotation so if we really wanted/needed to we could restore it.

api/v1alpha1/virtualmachine_conversion.go
```diff
@@ -1138,6 +1138,20 @@ func Convert_v1alpha3_VirtualMachine_To_v1alpha1_VirtualMachine(in *vmopv1.Virtu
 	// necessary for later versions that don't support configMap based bootstrap.
 	delete(out.Annotations, vmopv1.V1alpha1ConfigMapTransportAnnotation)
 
+	// Handle the renaming of pause annotation on down convert.
+	if val, ok := in.Annotations[vmopv1.PauseAnnotation]; ok {
+		annotations := out.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
```
bryanv, line 1145: Aren't these missing a `SetAnnotation()` for this case?

bryanv (approved):

api/v1alpha1/virtualmachine_conversion.go
aruneshpa: Thanks for spotting this.  Fixed.

pkg/util/annotations/helpers.go
aruneshpa: Resolving this for now.  We can move the file later if needed.

dougm (approved):

dilyar85 (approved):
LGTM!

PR #637, opened 2024-07-19
:seedling: Introduce an annotation for unconditional opt into backup

aruneshpa:

Today, we have certain exclusion rules in place that govern whether or not we persist the configuration of VirtualMachine in the ExtraConfig for use by backup/restore or disaster recovery solutions.  However, there might be cases where an entity wants to explicitly opt into backup.

This change introduces an annotation that can be placed on the VirtualMachine resource that lets VM operator skip the exclusion checks when determining whether or not to store the VM's config in ExtraConfig.

```release-note
 Introduce an annotation for unconditional opt into backup
```


aruneshpa:
This is a WIP because I want to rebase it on the helpers added in the #636 MR.

dilyar85 (approved):
LGTM!

PR #639, opened 2024-07-24
🌱 resize: add condition to indicate if resize is available

bryanv:

If a resize is pending - likely because the VM is powered on and either the VM Spec.ClassName has been updated or the opt-in annotation is present and the class itself has changed - report that in a condition.

Fixes #

```release-note
NONE
```

### DIFF
--- a/api/v1alpha2/virtualmachine_types.go
+++ b/api/v1alpha2/virtualmachine_types.go
@@ -41,6 +41,10 @@ const (
 
 	// VirtualMachineConditionCreated indicates that the VM has been created.
 	VirtualMachineConditionCreated = "VirtualMachineCreated"
+
+	// VirtualMachineConfigurationSynced indicates that the VM's current configuration is synced to the
+	// current version of its VirtualMachineClass.
+	VirtualMachineConfigurationSynced = "VirtualMachineConfigurationSynced"
 )
 
 const (
--- a/pkg/providers/vsphere/vmlifecycle/update_status.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status.go
@@ -16,6 +16,7 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/govmomi/vmdk"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apierrorsutil "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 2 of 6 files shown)


pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -432,6 +436,43 @@ func MarkBootstrapCondition(
...
+		return
+	}
+
+	// Depending on what we did in the prior session update code for this VM, we might already
+	// have fetched the class but the way the code is structured today, it isn't very easy for
+	// us to pass that to here. So just refetch it again.
+	vmClass := vmopv1.VirtualMachineClass{}
+	err := k8sClient.Get(ctx, ctrlclient.ObjectKey{Name: className, Namespace: vm.Namespace}, &vmClass)
+	if err != nil {
+		if ctrlclient.IgnoreNotFound(err) == nil {
```
zyiyi11, line 461: For easier readability, suggest using `if apierrorsutil.IsNotFound(err) {`

pkg/providers/vsphere/vmprovider_vm_resize_test.go
```diff
@@ -222,6 +232,36 @@ func vmResizeTests() {
 			})
 		})
 
+		Context("Powered On VM", func() {
+			BeforeEach(func() {
+				configSpec.NumCPUs = 2
+				configSpec.MemoryMB = 1024
+			})
+
+			It("Resize Pending", func() {
```
zyiyi11, line 241: Can we also add a `Resize Succeed` test?

zyiyi11 (approved):
LGTM with some minor suggestions

pkg/providers/vsphere/vmlifecycle/update_status.go
bryanv: I was kind of aiming for one less import. The currently imported `apierrorsutil` isn't the package that has that method.

pkg/providers/vsphere/vmprovider_vm_resize_test.go
bryanv: This test was cribbed from the immediate prior "Powering On VM" resize test. For this test a powered on VM won't be resized - the primary purpose of this annotation is to indicate that a resize is avail and would be done the next time the VM is powered off.

sreyasn (approved):
Looks good to me!

api/v1alpha2/virtualmachine_types.go
```diff
@@ -41,6 +41,10 @@ const (
 
 	// VirtualMachineConditionCreated indicates that the VM has been created.
 	VirtualMachineConditionCreated = "VirtualMachineCreated"
+
+	// VirtualMachineResizeClassSynced indicates that the VM's current configuration is synced to the
+	// current version of its VirtualMachineClass.
+	VirtualMachineResizeClassSynced = "VirtualMachineResizeClassSynced"
```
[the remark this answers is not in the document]
bryanv: So it is there for any old v1a2 api users. For some recent'ish added Condition they were added them to v1a2 but not v1a1 so I did the same here.

PR #643, opened 2024-07-26
✨ Report storage use for VMs to StoragePolicyUsage resource

akutz:

This patch reports VM usage in a namespace to the namespace's StoragePolicyUsage resource.

Fixes `NA`

```release-note
Report storage usage for VMs per namespace.
```


pkg/util/vmopv1/vm.go
```diff
@@ -213,3 +216,123 @@ func IsClasslessVM(vm vmopv1.VirtualMachine) bool {
...
+
+			if !vm.DeletionTimestamp.IsZero() || vm.Status.Storage == nil {
+				// Ignore VMs being deleted or that are not reporting storage
+				// status.
+				continue
+			}
+
+			// Account for storage used by this VM.
+			if vm.Status.Storage.Unshared != nil {
+				totalUsed.Add(*vm.Status.Storage.Unshared)
```
akutz, line 276: @SandeepPissay / @xing-yang, question for y'all,

To calculate the total space that could be used by a VM (i.e. reserved), we should be summing Committed and Uncommitted from https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.vm.Summary.StorageSummary.html, right?

pkg/util/vmopv1/vm.go
```diff
@@ -213,3 +216,123 @@ func IsClasslessVM(vm vmopv1.VirtualMachine) bool {
...
+			if vm.Status.Storage.Committed != nil {
+				totalReserved.Add(*vm.Status.Storage.Committed)
+			}
+			if vm.Status.Storage.Uncommitted != nil {
+				totalReserved.Add(*vm.Status.Storage.Uncommitted)
+			}
+
+			// Subtract the PVC usage/limit from the total usage/reserved values
+			// as PVCs are counted separately.
+			for j := range vm.Status.Volumes {
```
akutz, line 289: @SandeepPissay / @xing-yang, just FYI -- 

We are subtracting space used/reserved by PVCs from the VM's total reserved space since PVCs are reported separately. This is what you would expect, correct?

pkg/util/vmopv1/vm.go
```diff
@@ -213,3 +216,123 @@ func IsClasslessVM(vm vmopv1.VirtualMachine) bool {
...
+				// status.
+				continue
+			}
+
+			// Account for storage used by this VM.
+			if vm.Status.Storage.Unshared != nil {
+				totalUsed.Add(*vm.Status.Storage.Unshared)
+			}
+
+			// Account for storage marked as reserved by this VM.
```
akutz, line 279: @SandeepPissay / @xing-yang, question for y'all,

To calculate the total space that could be used by a VM (i.e. reserved), we should be summing Committed and Uncommitted from https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.vm.Summary.StorageSummary.html, right?

SandeepPissay: The doc you pointed to says:

committed - Total storage space, in bytes, committed to this virtual machine across all datastores.
Essentially an aggregate of the property [committed](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.vm.StorageInfo.UsageOnDatastore.html#committed) across all datastores that this virtual machine is located on.
uncommitted - Additional storage space, in bytes, potentially used by this virtual machine on all datastores.
Essentially an aggregate of the property [uncommitted](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.vm.StorageInfo.UsageOnDatastore.html#uncommitted) across all datastores that this virtual machine is located on.


I think those properties will include any attached PVs as well. Also see my comment above on how we are using the provisioned capacity for storage quota accounting.

xing-yang: If we use "committed" and "uncommitted", could we be double counting?  I see that there is another value named "unshared". Should we use that instead?

unshared - Total storage space, in bytes, occupied by the virtual machine across all datastores, that is not shared with any other virtual machine.

pkg/util/vmopv1/vm.go
xing-yang: Yes.  Sandeep's explanation is right.

pkg/util/vmopv1/vm.go
akutz: > I think those properties will include any attached PVs as well. Also see my comment above on how we are using the provisioned capacity for storage quota accounting.

@SandeepPissay , yes, but as I noted elsewhere, we are subtracting these values.

akutz: > If we use "committed" and "uncommitted", could we be double counting? I see that there is another value named "unshared". Should we use that instead?

@xing-yang ,

I don't think so. Unshared is for used. Committed + uncommitted is potential (i.e. reserved), minus the PVCs.

pkg/util/vmopv1/vm.go
akutz: Ack. I asked Deepak the same internally and did not get a clear answer. I'll update the logic to only mark things reserved if the VM is not yet fully created.

akutz: @abaruni Based on @SandeepPissay's and @xing-yang's feedback, I will need to rely on the work you are doing for the OVF parsing to get the capacity for VMs not yet fully provisioned. I wonder what the most efficient way to do this is.

pkg/util/vmopv1/vm.go
xing-yang: Okay.

pkg/util/kube/cource/cource_context.go
```diff
@@ -0,0 +1,130 @@
...
+
+	leftMap := leftObj.(*sourceMap)
+	rightMap := rightObj.(*sourceMap)
+
+	leftMap.Lock()
+	rightMap.Lock()
+	defer leftMap.Unlock()
+	defer rightMap.Unlock()
+
+	for k, v := range rightMap.data {
```
bryanv, line 125: FWIW there is a `maps.Copy` - https://pkg.go.dev/maps#Copy

bryanv (approved):

PR #647, opened 2024-07-31
:sparkles: Update VirtualMachineImage and ClusterVirtualMachineImage resources to report disk information

abaruni:

This PR updates VirtualMachineImage and ClusterVirtualMachineImage resources to report disk information for the boot disk. The reported information is taken from the Disk Section of the OVF

Fixes #

```release-note
Update VirtualMachineImage and ClusterVirtualMachineImage API to report disk information
```


api/v1alpha1/virtualmachineimage_conversion.go
```diff
@@ -390,6 +391,14 @@ func (src *VirtualMachineImage) ConvertTo(dstRaw ctrlconversion.Hub) error {
 		return err
 	}
 
+	// Manually restore data.
+	restored := &vmopv1.VirtualMachineImage{}
+	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
+		return err
+	}
+
+	dst.Status.DiskInfo = restored.Status.DiskInfo
```
akutz, line 400: Why not just?

```golang
dst.Status = restored.Status
```

api/v1alpha1/virtualmachineimage_conversion.go
```diff
@@ -448,6 +458,14 @@ func (src *ClusterVirtualMachineImage) ConvertTo(dstRaw ctrlconversion.Hub) erro
 		return err
 	}
 
+	// Manually restore data.
+	restored := &vmopv1.ClusterVirtualMachineImage{}
+	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
+		return err
+	}
+
+	dst.Status.DiskInfo = restored.Status.DiskInfo
```
akutz, line 467: Why not just?

```golang
dst.Status = restored.Status
```

api/v1alpha3/virtualmachineimage_types.go
```diff
@@ -81,6 +82,40 @@ type VirtualMachineImageProductInfo struct {
 	FullVersion string `json:"fullVersion,omitempty"`
 }
 
+// VirtualMachineImageDiskInfo describes disk information for an image taken from DiskSection
+// of an OVF Envelope.
+type VirtualMachineImageDiskInfo struct {
```
akutz, line 87: Hey @ericvmw / @dilyar85,

I'm interested in your take on this as well -- should we be surfacing all of this detail? We definitely want the capacity and size, but I'm not sure we want _more_ than that? _Maybe_ the format? But the ID and file/parent ref information is really leaking implementation details IMO. What do y'all think?

api/v1alpha3/virtualmachineimage_types.go
```diff
@@ -81,6 +82,40 @@ type VirtualMachineImageProductInfo struct {
 	FullVersion string `json:"fullVersion,omitempty"`
 }
 
+// VirtualMachineImageDiskInfo describes disk information for an image taken from DiskSection
```
akutz, line 85: Suggest perhaps:

```golang
VirtualMachineImageDiskInfo describes information about any disks associated with
this image.
```

I'm not sure we need to leak where the information is derived.

akutz (changes requested):
Hey @abaruni,

This is great, thank you! I have some concerns about the amount of information we are leaking in the disk info section. Also, could you please update the PR description with a release note about adding disk info details to the VMI and CVMI APIs? Thanks!

api/v1alpha1/virtualmachineimage_conversion.go
abaruni: Wouldn't that overwrite what was done in `convert_v1alpha1_VirtualMachineImageSpec_To_v1alpha3_VirtualMachineImageStatus` on line 385?

api/v1alpha1/virtualmachineimage_conversion.go
abaruni: Wouldn't that overwrite what was done in `convert_v1alpha1_VirtualMachineImageSpec_To_v1alpha3_VirtualMachineImageStatus` on line 452?

api/v1alpha3/virtualmachineimage_types.go
abaruni: I was working with the requirement `should report the disk layout, including size, capacity, provisioning type, etc.`. The only one-to-one correspondence, and the ones that make sense in my mind, are capacity and size. But with the presence of the other explicitly named items and the presence of the `etc.`, I erred on the side of populating the full disk info as it appears in the envelope.

abaruni: Updated with suggestion

api/v1alpha3/virtualmachineimage_types.go
abaruni: Removed the fields that didn't make sense

api/v1alpha1/virtualmachineimage_conversion.go
bryanv: When the`Status.Type` was recently added, we decided on punting doing any version conversions for that field: TKG images have a huge plaintext yaml, so this will effectively double their size for down-version users. And CCI already has slowness issues with images. And with the {Cluster}VirtualMachineImages we - VMOP controller - are the only one that does updates on these types so the saved version in the Annotation won't ever be used.

api/v1alpha3/virtualmachineimage_types.go
```diff
@@ -200,6 +215,11 @@ type VirtualMachineImageStatus struct {
 
 	// +optional
 
+	// DiskInfo describes the observed disk information for this image.
```
bryanv, line 218: Should this make a mention that this is the just the boot disk?

For my education why are we only including the first disk, instead of having this be an array?

akutz: Except this will include any disk from the image, right?

akutz: I just realized this is not a slice :( When I saw the original fields included DiskId, I assumed the parent field was a slice. Talking with Adnan about it.

abaruni: There was some confusion and an assumption on my part. Will update this field to be a slice.

api/v1alpha1/virtualmachineimage_conversion.go
akutz: Yep. We decided to narrow this to a given field.

api/v1alpha3/virtualmachineimage_types.go
bryanv, line 218: Nit: maybe we want to make this description & field name plural?

pkg/providers/vsphere/contentlibrary/content_library_utils_test.go
```diff
@@ -149,6 +169,13 @@ var _ = Describe("UpdateVmiWithOvfEnvelope", func() {
 		Expect(image.Status.VMwareSystemProperties[0].Value).Should(Equal(versionVal))
 
 		Expect(conditions.Has(image, vmopv1.VirtualMachineImageV1Alpha1CompatibleCondition)).To(BeFalse())
+
+		Expect(len(image.Status.DiskInfo)).To(Equal(2))
```
bryanv, line 173: There is a `HaveLen` matcher that gives a more details when the length doesn't match

`Expect(image.Status.DiskInfo).To(HaveLen(2))`

bryanv (approved):

api/v1alpha3/virtualmachineimage_types.go
akutz: I think it is fine. DiskInfo == DiskInformation, and "information" can mean multiple disks IMO.

bryanv: I do like `Status.Disks` but totally cool with whatever y'all want to change it to or keep it as-is.

akutz (approved):
Lgtm, thanks @abaruni !

PR #649, opened 2024-07-31
Remove caBundle from conversion webhook defs

akutz:

This patch removes the caBundle field from the conversion webhook definitions. This field was set to the base-64 encoded newline, i.e. Cg==, to allow the creation of the webhooks since the field was previously marked as required. This was in error. The field was supposed to be marked optional, which was fixed in 2018. By then the pattern had already been established to ensure Cert-Manager could find the webhook def and replace the CABundle field with a valid value.

However, this pattern created havoc on updates, replacing certs when updating infra components with newline chars. A change in K8s 1.30 rc1 disallows this pattern, requiring clients to omit the field entirely.

This patch is in accordance with the new guidelines. Now that https://github.com/kubernetes/kubernetes/pull/124061 is merged, continuing to use "caBundle: Cg==" will cause the manifests to fail when used to update an existing CRD.

The K8s Slack thread
https://kubernetes.slack.com/archives/C0EG7JC6T/p1722441161968339 has more information.

Fixes #

```release-note

```


zyiyi11 (approved):
Thanks for the fix!

bryanv (approved):

PR #650, opened 2024-08-01
:seedling: Update vcsim tests under WCP_Workload_Domain_Isolation

zyiyi11:
This patch updates vcsim tests when WCP_Workload_Domain_Isolation enabled. 
- Updated `CreateWorkloadNamespace` to create namespaced Zone CR when FSS enabled.
- Test both FSS enabled/disabled for `zone_placement_test.go`.
- Test only FSS enabled for `folder_test.go`, `vmprovider_resourcepolicy_test.go` ect because they only cares `nsInfo.Namespace` during tests.

Fixes # N/A

```release-note
NONE
```

### DIFF
--- a/test/builder/vcsim_test_context.go
+++ b/test/builder/vcsim_test_context.go
@@ -118,10 +118,12 @@ type TestContextForVCSim struct {
 	RestClient     *rest.Client
 	Recorder       record.Recorder
 
-	// When WithFaultDomains is true:
 	ZoneCount       int
 	ClustersPerZone int
 	ZoneNames       []string
+
+	// withWorkloadIsolation stores VCSimTestConfig WithWorkloadIsolation value.
+	withWorkloadIsolation bool
 
 	// When WithContentLibrary is true:
 	ContentLibraryImageName string
@@ -195,7 +197,8 @@ func newTestContextForVCSim(
 	}
 
 	ctx.ClustersPerZone = clustersPerZone
-
+	// TODO: this can be removed once FSS_WCP_WORKLOAD_DOMAIN_ISOLATION enabled.
+	ctx.withWorkloadIsolation = config.WithWorkloadIsolation
 	return ctx
 }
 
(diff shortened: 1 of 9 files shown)


test/builder/vcsim_test_context.go
```diff
@@ -249,14 +252,31 @@ func (c *TestContextForVCSim) CreateWorkloadNamespace() WorkloadNamespaceInfo {
...
+		// When FSS_WCP_WORKLOAD_DOMAIN_ISOLATION is disabled, AvailabilityZone stores namespace info.
+		if !config.WithWorkloadIsolation {
+			az := &topologyv1.AvailabilityZone{}
+			Expect(c.Client.Get(c, client.ObjectKey{Name: azName}, az)).To(Succeed())
+			if az.Spec.Namespaces == nil {
+				az.Spec.Namespaces = map[string]topologyv1.NamespaceInfo{}
+			}
+			az.Spec.Namespaces[ns.Name] = nsInfo
+			Expect(c.Client.Update(c, az)).To(Succeed())
+			// When FSS_WCP_WORKLOAD_DOMAIN_ISOLATION is enabled, Namespaced Zone stores namespace info.
```
bryanv, line 264: It is probably clearer to move this comment after the `} else {`

test/builder/vcsim_test_context.go
```diff
@@ -218,7 +221,7 @@ func (c *TestContextForVCSim) AfterEach() {
 	c.UnitTestContext.AfterEach()
 }
 
-func (c *TestContextForVCSim) CreateWorkloadNamespace() WorkloadNamespaceInfo {
+func (c *TestContextForVCSim) CreateWorkloadNamespace(config VCSimTestConfig) WorkloadNamespaceInfo {
```
bryanv, line 224: Save the value in the `TestContextForVCSim` so this doesn't need to be updated.

zyiyi11: I am going with `save the value in the TestContextForVCSim`

test/builder/vcsim_test_context.go
```diff
@@ -144,6 +147,9 @@ type TestContextForVCSim struct {
 	datastore *object.Datastore
 
 	azCCRs map[string][]*object.ClusterComputeResource
+
+	// WithWorkloadIsolation stores VCSimTestConfig.WithWorkloadIsolation value.
```
bryanv, line 151: I don't think this needs to be public. Also group this near the existing `WithFaultDomains ...` comment (and maybe update that comment a bit).

bryanv (approved):

PR #654, opened 2024-08-06
✨ Perennial properties

akutz:

This patch does three things:

1. The VM object's namespace and name are stored in ExtraConfig as the keys `vmservice.namespace` and `vmservice.name`.

2. The VM's vSphere property `config.managedBy` is _always_ set to reflect the VM is managed by VM Service if it is not set to this value.

3. There is now support for *always* ensuring certain properties are set during each reconcile, regardless of the VM's power state or whether or not the VM is paused by the admin. For example, the aforementioned namespace/name ExtraConfig keys are included, as well as the VM's `config.managedBy` property.

Fixes `NA`

```release-note
Support perennial properties
```

### DIFF
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -4,7 +4,6 @@
 package session
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"maps"
@@ -21,6 +20,7 @@ import (
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
 	vmopv1common "github.com/vmware-tanzu/vm-operator/api/v1alpha3/common"
 	"github.com/vmware-tanzu/vm-operator/pkg"
+	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/clustermodules"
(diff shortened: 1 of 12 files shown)


pkg/util/vmopv1/resize_overwrite.go
```diff
@@ -53,13 +72,94 @@ func overwriteExtraConfig(
...
+
+	user := vimtypes.ManagedByInfo{
+		ExtensionKey: vmopv1.ManagedByExtensionKey,
+		Type:         vmopv1.ManagedByExtensionType,
+	}
+
+	overwrite(cs.ManagedBy, user, current)
+
+	var empty vimtypes.ManagedByInfo
+	if cs.ManagedBy != nil && *cs.ManagedBy == empty {
```
bryanv, line 117: From L105 `cs.ManagedBy` shouldn't be able to be `nil` at this point?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -798,27 +800,26 @@ func (s *Session) poweredOnVMReconfigure(
...
-		// this FSR internally to vSphere.
-		if configSpec.ChangeTrackingEnabled != nil {
-			if err := s.invokeFsrVirtualMachine(vmCtx, resVM); err != nil {
-				vmCtx.Logger.Error(err, "Failed to invoke FSR for CBT update")
-				return false, err
-			}
+	// Special case for CBT: in order for CBT change take effect for a powered on VM,
+	// a checkpoint save/restore is needed.  tracks the implementation of
+	// this FSR internally to vSphere.
+	if refetchProps && configSpec.ChangeTrackingEnabled != nil {
```
bryanv, line 819: Why does this need to check `refetchProps`?

akutz: Sorry, I know it's not super clear due to the variable name, but `refetchProps` is also `didReconfigureHappen`. Thus, we do not need to issue the FSR unless the reconfig actually occurred.

akutz: I updated the comments to reflect this.

bryanv (approved):

PR #659, opened 2024-08-08
🐛 Update CD-ROM API fields and related webhook code

dilyar85:

This PR updates CD-ROM API fields and docs for clarity and improves the VM webhook for ISO VM deployment.  
Specifically, this patch includes:
- Fix `AllowGuestControl` and `Connected` fields with boolean pointers.
- Update VM mutating webhook to not set the default CD-ROM name and re-populate `VirtualMachineImage` kind if previously set to default.
- Update VM validating webhook to require `guestID` if VM has a CD-ROM and handle duplicates across namespace and cluster scope images (since they will point to the same backing file name).

Fixes issues encountered in internal end-to-end tests for ISO VM deployment.

Tests have been updated accordingly with 100% code coverage of the changes.

```release-note
Update CD-ROM API fields and related webhook code to address gaps.
```

----
📚 Documentation preview 📚: https://vm-operator--659.org.readthedocs.build/en/659/


api/v1alpha3/virtualmachine_types.go
```diff
@@ -276,9 +276,10 @@ type VirtualMachineImageRef struct {
 type VirtualMachineCdromSpec struct {
 	// +kubebuilder:validation:Pattern="^[a-z0-9]{2,}$"
 
-	// Name describes the unique and immutable name of this CD-ROM device.
-	// If omitted or empty, the controller will generate it as "cdrom" + "num",
-	// where "num" is the device's index in the spec.cdrom list.
+	// Name consist of at least two lowercase letters or digits of this CD-ROM.
```
dougm, line 279: consist -> consists

api/v1alpha1/virtualmachine_conversion_test.go
```diff
@@ -224,8 +224,8 @@ func TestVirtualMachineConversion(t *testing.T) {
 							Name: "my-cdrom-image",
 							Kind: "VirtualMachineImage",
 						},
-						AllowGuestControl: true,
-						Connected:         true,
+						AllowGuestControl: &[]bool{true}[0],
```
dougm, line 227: Interesting, don't think I've seen that syntax before. You could also use `ptrOf(true)` here. That func is in both v1alpha[12] packages

pkg/providers/vsphere/session/session_vm_update_test.go
```diff
@@ -1553,8 +1553,8 @@ var _ = Describe("UpdateVirtualMachine", func() {
 									Name: vmiName,
 									Kind: vmiKind,
 								},
-								AllowGuestControl: true,
-								Connected:         true,
+								AllowGuestControl: &[]bool{true}[0],
```
dougm, line 1556: Looks like no `ptrOf` existing in this package, but you can also use `vimtypes.NewBool(true)`

pkg/util/vsphere/vm/cdrom.go
```diff
@@ -43,7 +43,7 @@ func UpdateCdromDeviceChanges(
 	for _, specCdrom := range vmCtx.VM.Spec.Cdrom {
 		imageRef := specCdrom.Image
 		// Sync the content library file if needed to connect the CD-ROM device.
-		syncFile := specCdrom.Connected
+		syncFile := *specCdrom.Connected
```
dougm, line 46: With these fields changed to `*bool`, is a nil check needed here? With something like:
```go
func boolVal(v *bool) bool {
	if v == nil {
		return false
	}
	return *v
}
...
syncFile := boolVal(specCdrom.Connected)
```

dougm (approved):
lgtm, just a few minor suggestions

api/v1alpha1/virtualmachine_conversion_test.go
dilyar85: I see both usages in the current code. Let me change it to calling that function since it's clearer and more readable.

pkg/providers/vsphere/session/session_vm_update_test.go
dilyar85: Looks like we have a similar function provided in `pkg/util/ptr`, will replace it with that.

pkg/util/vsphere/vm/cdrom.go
dilyar85: With the API defaulting this field to true, I don't think it could be nil? I also tried applying a VM YAML config with `connected: null` and found the field stored as the default true value in etcd. 

I agree it's always safe to double-check, and I'll use `pkg/util/ptr` to get its value safely.

api/v1alpha3/virtualmachine_types.go
dilyar85: Fixed.

api/v1alpha1/virtualmachine_conversion_test.go
dilyar85: Updated in the latest commit.

pkg/providers/vsphere/session/session_vm_update_test.go
dilyar85: Done.

pkg/util/vsphere/vm/cdrom.go
dilyar85: Updated with `ptr.Deref(specCdrom.Connected)`.

PR #662, opened 2024-08-08
✨ Implement Backup-version for Incremental Restore support

sreyasn:

This change
- Introduces a new annotation and an extraConfig Key to track the latest backup-version
- The backup version is a monotonically increasing number which uses the time.Now().Unix() method.
- Skip backup when backupVersion drift is detected (ie) backup version annotation > current backup extraConfig version. This indicates a restore being tried. backup will retry until the drift is fixed by a registration.
- Do backup when backupVersion annotation doesn't exist (or) when backupVersion and current backup extraconfig keys match (or) backup version annotation < extraConfig backup version.
- Update and persists a new version when a change is detected during each backup.
- Testing for the above.

Fixes # N/A

Manual testing: 

Backup version annotation populated on the VM resource:
> apiVersion: vmoperator.vmware.com/v1alpha3
> kind: VirtualMachine
> metadata:
>   annotations:
>     kubectl.kubernetes.io/last-applied-configuration: |
>       {"apiVersion":"vmoperator.vmware.com/v1alpha1","kind":"VirtualMachine","metadata":{"annotations":{},"name":"ubuntu-impish-new","namespace":"prod-vmsvc-reco-vds-3769"},"spec":{"className":"best-effort-small","imageName":"vmi-81f7a0f6aaee2bde6","networkInterfaces":[{"networkType":"vsphere-distributed"}],"powerState":"poweredOn","storageClass":"wcpglobal-storage-profile"}}
>     virtualmachine.vmoperator.vmware.com/first-boot-done: "true"
>     vmoperator.vmware.com/backup-version: "1723070407"
>     vmoperator.vmware.com/created-at-build-version: 1.8.6-309-g92143d52
>     vmoperator.vmware.com/created-at-schema-version: v1alpha3

```
Implement the backup-version extra config key and annotation workflows for Incremental restore
```

### DIFF
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -166,6 +166,15 @@ const (
 	// PVC disk data in JSON, compressed using gzip and base64-encoded.
 	PVCDiskDataExtraConfigKey = "vmservice.virtualmachine.pvc.disk.data"
 
+	// BackupVersionExtraConfigKey is the ExtraConfig key that indicates
+	// the version of the VM's last backup. It is a monotonically increasing counter and
+	// is only supposed to be used by IaaS control plane and vCenter for virtual machine registration
+	// post a restore operation.
+	//
+	// The BackupVersionExtraConfigKey on the vSphere VM and the VirtualMachineBackupVersionAnnotation
+	// on the VM resource in Supervisor indicate whether the backups are in sync.
+	BackupVersionExtraConfigKey = "vmservice.virtualmachine.backupVersion"
+
 	// DisableAutoRegistrationExtraConfigKey is the ExtraConfig key that can be
 	// set to "true" (case insensitive) on a virtual machine to opt-out of the
 	// automatic registration workflow.
@@ -191,6 +200,15 @@ const (
 	// vSphere.  There is no guarantee that the registration of the VM will be
 	// successful post a restore or failover operation.
 	ForceEnableBackupAnnotation = GroupName + "/force-enable-backup"
+
+	// VirtualMachineBackupVersionAnnotation is an annotation that indicates the VM's
+	// last backup version. It is a monotonically increasing counter and
+	// is only supposed to be used by IaaS control plane and vCenter for virtual machine registration
+	// post a restore operation.
+	//
+	// The VirtualMachineBackupVersionAnnotation on the VM resource in Supervisor and the BackupVersionExtraConfigKey on the vSphere VM
+	// indicate whether the backups are in sync.
+	VirtualMachineBackupVersionAnnotation = GroupName + "/backup-version"
 )
 
 const (
--- a/pkg/providers/vsphere/virtualmachine/backup.go
+++ b/pkg/providers/vsphere/virtualmachine/backup.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/vmware/govmomi/object"
@@ -19,6 +20,7 @@ import (
 	k8syaml "sigs.k8s.io/yaml"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	ctxop "github.com/vmware-tanzu/vm-operator/pkg/context/operation"
 	res "github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/resources"
@@ -31,6 +33,7 @@ type BackupVirtualMachineOptions struct {
 	VcVM                *object.VirtualMachine
 	DiskUUIDToPVC       map[string]corev1.PersistentVolumeClaim
 	AdditionalResources []client.Object
+	BackupVersion       string
 }
 
 // PVCDiskData contains the data of a disk attached to VM backed by a PVC.
@@ -55,22 +58,36 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
 		opts.VMCtx.Logger.Error(err, "failed to get VM properties for backup")
 		return err
 	}
-
 	curExCfg := pkgutil.OptionValues(moVM.Config.ExtraConfig)
 	var ecToUpdate pkgutil.OptionValues
 
-	vmYAML, err := getDesiredVMResourceYAMLForBackup(opts.VMCtx.VM, curExCfg)
-	if err != nil {
-		opts.VMCtx.Logger.Error(err, "failed to get VM resource yaml for backup")
-		return err
-	}
-	if vmYAML == "" {
-		opts.VMCtx.Logger.V(4).Info("Skipping VM resource yaml backup as unchanged")
-	} else {
-		ecToUpdate = append(ecToUpdate, &vimtypes.OptionValue{
-			Key:   vmopv1.VMResourceYAMLExtraConfigKey,
-			Value: vmYAML,
-		})
+	/*
+	 * When VMIncrementalRestore FSS is enabled, perform backups when
+	 * - no backup version annotation is present.
+	 * - the backup version annotation matches the current backup version in extraConfig.
+	 * - the backup version annotation is older than the extraConfig backup version key. This can only happen when reconfigure succeeds to add a new
+	 *   extraConfig backup version but the next reconcile failed to update the annotation on the VM resource.
+	 */
+	if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
+		if backupVersionAnnotation, ok := opts.VMCtx.VM.Annotations[vmopv1.VirtualMachineBackupVersionAnnotation]; ok && backupVersionAnnotation != "" {
+			if backupVersionEcVal, _ := curExCfg.GetString(vmopv1.BackupVersionExtraConfigKey); backupVersionEcVal != "" {
+				canBackup, err := canPerformBackup(opts, backupVersionEcVal, backupVersionAnnotation)
+				if err != nil {
+					return err
+				}
+
+				/* TODO: Improvements to fetch the backupVersion extraConfig again right before doing the backup reconfigure?.
+				 *  There is a minor chance for a race if a restore is triggered by a vendor between !canBackup check
+				 *  and the backup reconfigure call below. Chances are low since the vm reconcile only does a backup when it detects
+				 *  changes to the keys and a restore needs to be triggered around that exact time.
+				 */
+				// restore is detected if doBackup is false with no errors. wait for vm registration.
+				if !canBackup {
+					// TODO: add a condition to indicate backup is paused
+					return nil
+				}
+			}
+		}
 	}
 
 	additionalYAML, err := getDesiredAdditionalResourceYAMLForBackup(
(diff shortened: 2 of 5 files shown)


pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -53,22 +56,40 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
+	/*
+	 * When VMIncrementalRestore FSS is enabled, perform backups when
+	 * - the backup version annotation matches the current backup version in extraConfig.
+	 * - no backup version annotation is present.
+	 * - the backup version annotation is older than the extraConfig backup version key. This can only happen when reconfigure succeeds to add a new
+	 *   extraConfig backup version but the next reconcile failed to update the annotation on the VM resource.
+	 */
+	if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
+		if backupVersionAnnotation, ok := opts.VMCtx.VM.Annotations[vmopv1.VirtualMachineBackupVersionAnnotation]; ok && backupVersionAnnotation != "" {
+			currBackupVersionEcKey, _ := curExCfg.GetString(vmopv1.BackupVersionKey)
```
zyiyi11, line 71: Suggest use `backupVersionEcKey` instead of `currBackupVersionEcKey` since there is only one `backupVersionEcKey` and one `backupVersionAnnotation`. It can be confused with later `currVersion`.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -286,3 +369,11 @@ func getDesiredPVCDiskDataForBackup(
 
 	return diskDataBackup, nil
 }
+
+func mustSetLastBackupVersionAnnotation(vm *vmopv1.VirtualMachine, version string) {
```
zyiyi11, line 373: Perhaps `setLastBackupVersionAnnotation`? why `must`?

pkg/providers/vsphere/virtualmachine/backup.go
dilyar85: Also, the var should be `backupVersionEcVal` instead of Key?

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -53,22 +56,40 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
+				}
+
+				// if the storedVersion in the backup is older than the annotation, this is a restore.
+				if currVersion > storedVersion {
+					opts.VMCtx.Logger.V(4).Info("Skipping VM backup as a restore is detected , wait for VM registration")
+
+					return fmt.Errorf("backup version drift detected, annotation: %s, backup extraconfig: %s", backupVersionAnnotation, currBackupVersionEcKey)
+				}
+			}
+		}
```
dilyar85, line 92: Can we move this into a separate function, something like `isBackupNeeded()`, that returns a boolean and an error to indicate if VMOP should perform the backup process?

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -53,22 +56,40 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
+				if err != nil {
+					opts.VMCtx.Logger.Error(err, fmt.Sprintf("failed to parse backup version annotation to int64: %v", backupVersionAnnotation))
+					return err
+				}
+
+				// if the storedVersion in the backup is older than the annotation, this is a restore.
+				if currVersion > storedVersion {
+					opts.VMCtx.Logger.V(4).Info("Skipping VM backup as a restore is detected , wait for VM registration")
+
+					return fmt.Errorf("backup version drift detected, annotation: %s, backup extraconfig: %s", backupVersionAnnotation, currBackupVersionEcKey)
```
dilyar85, line 89: Do we need to return an error to force a reconciliation here? After VM registration, VMOP will receive reconciliation requests regardless, so we could just log with higher visibility and return.

pkg/providers/vsphere/virtualmachine/backup.go
sreyasn: I was split between doing this and keeping it as is too. Unlikely we'll reuse this anywhere but makes the code cleaner. moving into a separate function

pkg/providers/vsphere/virtualmachine/backup.go
sreyasn: thanks. I wanted your (or) @aruneshpa's opinion on this. I initially had an implementation that returned nil but changed it to force reconciliation. I'll change it back.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -115,12 +177,48 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
 	}
 
 	return nil
 }
 
+// isBackupNeeded checks if a backup can be performed. The method returns true when
+// the backup version annotation matches (or) is older (lesser) than the backup version in extraConfig.
+// It returns false when the backup version annotation is newer (greater) than the backup extraConfig version which indicates a vendor
+// triggered restore in progress.
+func isBackupNeeded(opts BackupVirtualMachineOptions, backupVersionEcVal, backupVersionAnnotation string) (bool, error) {
```
aruneshpa, line 203: I would recommend renaming this to something like `shouldPerformBackup` `canPerformBackup`, or `canBackup` because even if a backup is needed (because the config has changed on Supervisor), this method can return false.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -53,22 +56,29 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
+	 * When VMIncrementalRestore FSS is enabled, perform backups when
+	 * - no backup version annotation is present.
+	 * - the backup version annotation matches the current backup version in extraConfig.
+	 * - the backup version annotation is older than the extraConfig backup version key. This can only happen when reconfigure succeeds to add a new
+	 *   extraConfig backup version but the next reconcile failed to update the annotation on the VM resource.
+	 */
+	if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
+		if backupVersionAnnotation, ok := opts.VMCtx.VM.Annotations[vmopv1.VirtualMachineBackupVersionAnnotation]; ok && backupVersionAnnotation != "" {
+			backupVersionEcVal, _ := curExCfg.GetString(vmopv1.BackupVersionKey)
+			doBackup, err := isBackupNeeded(opts, backupVersionEcVal, backupVersionAnnotation)
```
aruneshpa, line 74: doBackup -> canBackup

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -53,22 +56,40 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
-		})
+	/*
+	 * When VMIncrementalRestore FSS is enabled, perform backups when
+	 * - the backup version annotation matches the current backup version in extraConfig.
+	 * - no backup version annotation is present.
+	 * - the backup version annotation is older than the extraConfig backup version key. This can only happen when reconfigure succeeds to add a new
+	 *   extraConfig backup version but the next reconcile failed to update the annotation on the VM resource.
+	 */
+	if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
+		if backupVersionAnnotation, ok := opts.VMCtx.VM.Annotations[vmopv1.VirtualMachineBackupVersionAnnotation]; ok && backupVersionAnnotation != "" {
```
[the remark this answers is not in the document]
dilyar85: @aruneshpa Reading from a `nil` annotations/map should be fine according to the [Go's Map definition](https://go.dev/ref/spec#Map_types:~:text=A%20nil%20map%20is%20equivalent%20to%20an%20empty%20map%20except%20that%20no%20elements%20may%20be%20added.):

> A nil map is equivalent to an empty map except that no elements may be added.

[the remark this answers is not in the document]
sreyasn: The check here shouldn't NPE because of the way golang treats nil maps.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -166,6 +166,10 @@ const (
 	// PVC disk data in JSON, compressed using gzip and base64-encoded.
 	PVCDiskDataExtraConfigKey = "vmservice.virtualmachine.pvc.disk.data"
 
+	// BackupVersionKey is the the ExtraConfig key that indicates
+	// the version of the VM's last backup.
+	BackupVersionKey = "vmservice.virtualmachine.backup.version"
```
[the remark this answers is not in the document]
sreyasn: thanks. made some additions to the existing comment. PTAL

pkg/providers/vsphere/virtualmachine/backup.go
sreyasn: > Should you also check for the case where the EC doesn't exist (the second return value of the GetString method)?

@aruneshpa , Do we care if the EC is doesn't exist vs ""?. Either ways at the moment the implementation decides to do a backup and fix it.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -55,22 +58,30 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
-			Value: vmYAML,
-		})
+	/*
+	 * When VMIncrementalRestore FSS is enabled, perform backups when
+	 * - no backup version annotation is present.
+	 * - the backup version annotation matches the current backup version in extraConfig.
+	 * - the backup version annotation is older than the extraConfig backup version key. This can only happen when reconfigure succeeds to add a new
+	 *   extraConfig backup version but the next reconcile failed to update the annotation on the VM resource.
+	 */
+	if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
```
aruneshpa, line 71: // I am worried that this might race with us persisting the backup a few lines below.  Think what happens if the VM is restored after this check succeeds (isBackupNeeded returns true).  In that case, the backup stored in the VM will be replaced a few lines below before registration has had a chance to run.  IMHO, this check should be as close to us trying to persist the backup as possible.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -55,22 +58,30 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
+		if backupVersionAnnotation, ok := opts.VMCtx.VM.Annotations[vmopv1.VirtualMachineBackupVersionAnnotation]; ok && backupVersionAnnotation != "" {
+			if backupVersionEcVal, _ := curExCfg.GetString(vmopv1.BackupVersionExtraConfigKey); backupVersionEcVal != "" {
+				canBackup, err := canPerformBackup(opts, backupVersionEcVal, backupVersionAnnotation)
+				if err != nil {
+					return err
+				}
+
+				// restore is detected if doBackup is false with no errors. wait for vm registration.
+				if !canBackup {
+					return nil
```
aruneshpa, line 81: Add a TODO here to track emitting an event/Condition.?
Add a log message for now.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -106,6 +117,58 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
 		})
 	}
 
+	if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
+		curBackup, _ := curExCfg.GetString(vmopv1.VMResourceYAMLExtraConfigKey)
+		vmBackupInSync, err := isVMBackupUpToDate(opts.VMCtx.VM, curBackup)
```
aruneshpa, line 122: Hmm.  Why not modify the `getDesiredVMResourceYAMLForBackup` method to set the Annotation if the FSS is enabled?  That way, you don't have to call `isVMBackupUpToDate` and duplicate the marshaling bit.

sreyasn: Setting the annotation and the backup version ecKey need to happen together and will happen on based on a few factors including the FSS 
factors (ie) 
`if (!ok || backupVersionAnnotation == "") || len(ecToUpdate) != 0 || !vmBackupInSync {`
and the FSS enabled.

getDesiredVMResourceYAMLForBackup will be taking on too much doing all this. We can clean up the _else_ block FSS is enabled and clean this up better.

pkg/providers/vsphere/virtualmachine/backup.go
sreyasn: makes sense. 
- We'll also be prepping a backup if we don't check early on too in the method. So we might need to check early on to prevent extra work being done.
- Also to reduce chances for the race as you mentioned, the properties need to be fetched again from the moVM right before trying Reconfigure() for backup. The extraConfig properties values fetched earlier in the method is what gets used until the end. So unless we fetch again, this race will still happen.

dilyar85: I don't think vendors treat VM Service VMs differently, and it's kind of impractical to have all vendors adopt this approach. Technically, VI Admins can pause the VM reconciliation before performing restore but it won't be an automated backup & restore workflow :)

sreyasn: Arunesh and I spoke about this offline. I've added to a todo to handle this minor case where a race could happen. resolving this for now

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -55,22 +58,36 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
-			Value: vmYAML,
-		})
+	/*
+	 * When VMIncrementalRestore FSS is enabled, perform backups when
+	 * - no backup version annotation is present.
+	 * - the backup version annotation matches the current backup version in extraConfig.
+	 * - the backup version annotation is older than the extraConfig backup version key. This can only happen when reconfigure succeeds to add a new
+	 *   extraConfig backup version but the next reconcile failed to update the annotation on the VM resource.
+	 */
+	if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
```
dilyar85, line 71: Nit: can store this in a var since it's checked multiple times below.

pkg/providers/vsphere/virtualmachine/backup.go
sreyasn: added a TODO for condition and a log message is printed out from canPerformBackup() for this case.

pkg/providers/vsphere/virtualmachine/backup.go
sreyasn: resolving. feel free to reopen if you feel strongly about this. we can talk about it offline.

dilyar85 (approved):

pkg/providers/vsphere/virtualmachine/backup.go
sreyasn: It's easier to grep for these Features.VMIncrementalRestore checks when someone goes about removing them for post-FSS enablement. Let's keep it as is.

aruneshpa (approved):
LGTM.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -55,22 +58,36 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
+		if backupVersionAnnotation, ok := opts.VMCtx.VM.Annotations[vmopv1.VirtualMachineBackupVersionAnnotation]; ok && backupVersionAnnotation != "" {
+			if backupVersionEcVal, _ := curExCfg.GetString(vmopv1.BackupVersionExtraConfigKey); backupVersionEcVal != "" {
+				canBackup, err := canPerformBackup(opts, backupVersionEcVal, backupVersionAnnotation)
+				if err != nil {
+					return err
```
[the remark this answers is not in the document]
sreyasn: both have logs line printed from canPerformBackup(). Didn't wan't to add another one here.

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -55,22 +58,36 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
...
+				}
+
+				/* TODO: Improvements to fetch the backupVersion extraConfig again right before doing the backup reconfigure?.
+				 *  There is a minor chance for a race if a restore is triggered by a vendor between !canBackup check
+				 *  and the backup reconfigure call below. Chances are low since the vm reconcile only does a backup when it detects
+				 *  changes to the keys and a restore needs to be triggered around that exact time.
+				 */
+				// restore is detected if doBackup is false with no errors. wait for vm registration.
+				if !canBackup {
+					// TODO: add a condition to indicate backup is paused
```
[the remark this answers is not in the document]
sreyasn: both have logs line printed from canPerformBackup(). Didn't wan't to add another one here.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -166,6 +166,15 @@ const (
...
 	PVCDiskDataExtraConfigKey = "vmservice.virtualmachine.pvc.disk.data"
 
+	// BackupVersionExtraConfigKey is the ExtraConfig key that indicates
+	// the version of the VM's last backup. It is a monotonically increasing counter and
+	// is only supposed to be used by IaaS control plane and vCenter for virtual machine registration
+	// post a restore operation.
+	//
+	// The BackupVersionExtraConfigKey on the vSphere VM and the VirtualMachineBackupVersionAnnotation
+	// on the VM resource in Supervisor indicate whether the backups are in sync.
+	BackupVersionExtraConfigKey = "vmservice.virtualmachine.backup.version"
```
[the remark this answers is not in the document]
sreyasn: sure.

PR #663, opened 2024-08-09
🌱 Handle zones that are being deleted under WCP_Workload_Domain_Isolation

zyiyi11:
This patch handles Zones that are being deleted. We don't want to allow new VMs to be created in zones that are being deleted.
- If VM deployment specifies a zone that is being deleted by SSO user, validation webhook will reject the creation.
- If VM deployment specifies a zone that is being deleted by admin, validation webhook will allow the creation.
- If VM deployment specifies a valid zone at create time, zone got deletionTimestamp later, still place it in the zone.
- For placement, filter out the to be deleted zones so we don't have it as a candidate.

Fixes # N/A

Testing done:
1. Bind 2 workload zones to supervisor, `zone-2` and `zone-3`.
2. Deploy VM with specified zone name places VM on expected zones.
3. Deploy 5 VMs with unspecified zone name, 4 landed on `zone-2` and 1 landed on `zone-3`
4. After `zone-2` deletion initiated
- Made sure `zone-2` has deletionTimestamp but not deleted yet with workloads still on.
- VM deployment with specified `topology.kubernetes.io/zone: zone-2` is rejected.
```
Error from server (metadata.labels[topology.kubernetes.io/zone]: Invalid value: "zone-2": cannot use zone that is being deleted): error when creating "vm1.yaml": admission webhook "default.validating.virtualmachine.v1alpha3.vmoperator.vmware.com" denied the request: metadata.labels[topology.kubernetes.io/zone]: Invalid value: "zone-2": cannot use zone that is being deleted
```
- Deploy 5 VMs with unspecified zone name, all allocated to the remaining `zone-3` as expected.

```release-note
Zone Placement will ignore Zones that are being deleted under WCP_Workload_Domain_Isolation
```

### DIFF
--- a/pkg/providers/vsphere/placement/zone_placement.go
+++ b/pkg/providers/vsphere/placement/zone_placement.go
@@ -112,6 +112,10 @@ func getPlacementCandidates(
 			zones = append(zones, zone)
 		}
 		for _, zone := range zones {
+			// Filter out the zone that is to be deleted, so we don't have it as a candidate when doing placement.
+			if zonePlacement && !zone.DeletionTimestamp.IsZero() {
+				continue
+			}
 			rpMoIDs := zone.Spec.ManagedVMs.PoolMoIDs
 			if childRPName != "" {
 				childRPMoIDs := lookupChildRPs(vmCtx, vcClient, rpMoIDs, zone.Name, childRPName)
--- a/pkg/topology/availability_zones_test.go
+++ b/pkg/topology/availability_zones_test.go
@@ -169,6 +169,24 @@ var _ = Describe("Availability Zones and Zones", func() {
 		ExpectWithOffset(1, apierrors.IsNotFound(err)).To(BeTrue())
 	}
 
+	assertGetNamespaceFolderAndRPMoIDZoneToDelete := func() {
+		obj := &topologyv1.Zone{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:  "ns-1",
+				Name:       "zone-to-delete",
+				Finalizers: []string{"test"},
+			},
+		}
+		err := client.Create(ctx, obj)
+		ExpectWithOffset(1, err).ToNot(HaveOccurred())
+		_, err = topology.GetZone(ctx, client, "zone-to-delete", "ns-1")
+		ExpectWithOffset(1, err).ToNot(HaveOccurred())
+		err = client.Delete(ctx, obj)
+		ExpectWithOffset(1, err).ToNot(HaveOccurred())
+		_, _, err = topology.GetNamespaceFolderAndRPMoID(ctx, client, "zone-to-delete", "ns-1")
+		ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	}
+
 	assertGetNamespaceFolderAndRPMoIDSuccessForAZ := func(azName string) {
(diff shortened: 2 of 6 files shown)


pkg/topology/availability_zones.go
```diff
@@ -69,12 +70,14 @@ func GetNamespaceFolderAndRPMoID(
 		if err != nil {
 			return "", "", err
 		}
+		if !zone.DeletionTimestamp.IsZero() {
```
bryanv, line 73: Won't this too prevent the creation of VMs that we want to allow?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1132,9 +1133,14 @@ func (v validator) validateAvailabilityZone(ctx *pkgctx.WebhookRequestContext, v
...
-			if _, err := topology.GetZone(ctx.Context, v.client, zoneLabelVal, vm.Namespace); err != nil {
+			zone, err := topology.GetZone(ctx.Context, v.client, zoneLabelVal, vm.Namespace)
+			if err != nil {
 				return append(allErrs, field.Invalid(zoneLabelPath, zoneLabelVal, err.Error()))
 			}
+			// prevent new VirtualMachines created on marked zone by SSO users.
+			if !ctx.IsPrivilegedAccount && !zone.DeletionTimestamp.IsZero() {
+				return append(allErrs, field.Invalid(zoneLabelPath, zoneLabelVal, invalidZone))
+			}
 		}
```
bryanv, line 1144: Should this be `} else {` here?

pkg/topology/availability_zones_test.go
```diff
@@ -169,6 +169,22 @@ var _ = Describe("Availability Zones and Zones", func() {
...
 
+	assertGetNamespaceFolderAndRPMoIDZoneToDelete := func() {
+		obj := &topologyv1.Zone{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:  "ns-1",
+				Name:       "zone-to-delete",
+				Finalizers: []string{"test"},
+			},
+		}
+		Expect(client.Create(ctx, obj)).To(Succeed())
```
bryanv, line 180: MIssing WithOffset in a few places here

pkg/providers/vsphere/placement/zone_placement.go
```diff
@@ -112,6 +112,10 @@ func getPlacementCandidates(
 			zones = append(zones, zone)
 		}
 		for _, zone := range zones {
+			// Filter out the zone that is to be deleted, so we don't have it as a candidate.
```
bryanv, line 115: I think this ends up being too strict for the `!zonePlacement` case above. I think draining zones just need to be ignored when we are doing zone placement.

zyiyi11: makes sense. Is it for the scenario where we have only one draining zone as candidate, we don't want to block and saying no candidate available?

bryanv: No. If the VM specified the zone at create time then it passed the webhook and we should try to create it. For when we're otherwise doing zone placement draining zones should not be included in the candidates. I don't want to go down the path of inferring SSO user or not via the OwnerRefs but we might be forced to do so later.

bryanv (approved):

PR #665, opened 2024-08-12
🐛 ExtraConfig keys as OptionVals when getting props / prober

akutz:

This patch updates the GuestInfo prober to correctly treat ExtraConfig properties retrieved by their key index as OptionValue types instead of thinking their value is returned directly.

Additionally, tests that use vC Sim have been added for the provider's GetVMProperties function that validate the new assumption about retrieving indexed ExtraConfig keys.

Fixes `NA`

```release-note
Fix bug (ish) with GuestInfo prober that incorrectly treated indexed keys as strings instead of OptionValues
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm_test.go
+++ b/pkg/providers/vsphere/vmprovider_vm_test.go
@@ -332,10 +332,12 @@ func vmTests() {
 
 			Context("GetVirtualMachineProperties", func() {
 				const (
-					propName        = "config.name"
-					propPowerState  = "runtime.powerState"
-					propExtraConfig = "config.extraConfig"
-					propPathName    = "config.files.vmPathName"
+					propName              = "config.name"
+					propPowerState        = "runtime.powerState"
+					propExtraConfig       = "config.extraConfig"
+					propPathName          = "config.files.vmPathName"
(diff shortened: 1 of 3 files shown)


pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -332,10 +332,12 @@ func vmTests() {
...
-					propName        = "config.name"
-					propPowerState  = "runtime.powerState"
-					propExtraConfig = "config.extraConfig"
-					propPathName    = "config.files.vmPathName"
+					propName              = "config.name"
+					propPowerState        = "runtime.powerState"
+					propExtraConfig       = "config.extraConfig"
+					propPathName          = "config.files.vmPathName"
+					propExtraConfigKeyKey = "vmservice.example"
+					propExtraConfigKey    = `config.extraConfig["` + propExtraConfigKeyKey + `"]`
```
dougm, line 340: Note that govmomi has `mo.Field` type now, which is mainly used on vcsim side, but can simplify client-side too:
```go
field := mo.Field{Path: propExtraConfig, Key: propExtraConfigKeyKey}
prop := field.String() // `config.extraConfig["vmservice.example"]`
```

dougm (approved):
lgtm

PR #668, opened 2024-08-15
🌱 Avoid duplicate VM ReconcileNormal events

bryanv:

Only emit a default event if ReconcileNormal fails and it was neither a create nor update op.

While here, fix a long standing bug if that we should update the probe manager status even if CreateOrUpdate fails. If the VM Spec.ReadinessProbe has been changed we should update what's being probed even if we reconcile fails for some reason.

Fixes #

```release-note
The VM CreateOrUpdateFailure event is renamed to ReconcileNormalFailure and only emitted on neither create or update operation
```


akutz (approved):
Thanks BV!

PR #673, opened 2024-08-20
✨ Bring your own (encryption) key (BYOK)

akutz:

This patch adds support for bringing your own encryption key used to encrypt/recrypt VMs.

Fixes `NA`

This PR will be rebased once #670 and #672 are merged.

```release-note
Support bringing your own (encryption) key (BYOK) to encrypt VM workloads
```

----
📚 Documentation preview 📚: https://vm-operator--673.org.readthedocs.build/en/673/


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1084,6 +1093,55 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpec(
...
+	if err := vs.k8sClient.Get(vmCtx, objKey, &obj); err != nil {
+		return err
+	}
+
+	providerID, keyID := obj.Spec.KeyProvider, obj.Spec.KeyID
+	if providerID == "" || keyID == "" {
+		return nil
+	}
+
+	isEncrypted := vmCtx.MoVM.Summary.Runtime.CryptoState != ""
```
bryanv, line 1120: I think this result in a Reconfigure() on every reconcile

api/v1alpha3/virtualmachine_types.go
```diff
@@ -420,6 +420,27 @@ type VirtualMachineSpec struct {
 
 	// +optional
 
+	// EncryptionClassName describes the name of the EncryptionClass resource
+	// used to encrypt this VM.
+	//
+	// Please note, this field is not required to encrypt the VM. If the
+	// underlying platform implements the vSphere Native Key Provider (NKP) then
```
rtsanev-github, line 427: if a default Key Provider is defined in the underlying platform...

It does not necessarily has to be native key provider, it can be standard key provider but it must be set as default in vCenter.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -420,6 +420,27 @@ type VirtualMachineSpec struct {
...
+	// the VM may still be fully or partially encrypted depending on the
+	// specified storage and VM classes.
+	//
+	// If an NKP is present and an encryption storage class is selected, the
```
rtsanev-github, line 431: NKP > Default Key Provider 

here and below

api/v1alpha3/virtualmachine_types.go
```diff
@@ -420,6 +420,27 @@ type VirtualMachineSpec struct {
...
+	// VM's home files and non-PVC disks will be encrypted.
```
rtsanev-github, line 432: Maybe its worth noting somewhere that in case there is a default key provider in vCenter AND EncryptionClassName (ECN) is specified, then ECN takes precedence.

test/builder/vcsim_test_context.go
```diff
@@ -648,18 +724,56 @@ func (c *TestContextForVCSim) setupK8sConfig(config VCSimTestConfig) {
...
 			},
 			Parameters: map[string]string{
 				"storagePolicyID": c.StorageProfileID,
 			},
-		}
-		Expect(c.Client.Create(c, storageClass)).To(Succeed())
+		})).To(Succeed())
+
+		c.EncryptedStorageClassName = "vm-encryption-policy"
+		c.EncryptedStorageProfileID = "4d5f673c-536f-11e6-beb8-9e71128cae77" // from vcsim
```
dougm, line 739: Use the exported const `pbm/simulator.DefaultEncryptionProfileID` instead ?
We also could/should add one for ID used in `StorageProfileID` above.

dougm (approved):
lgtm, just a few minor comments/suggestions

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1190,85 +1200,120 @@ func (s *Session) UpdateVirtualMachine(
...
 				existingPowerState)
 		}
 	} else {
 		vmCtx.Logger.Info("VirtualMachine is paused. PowerState is not updated.")
-		refetchProps, err = defaultReconfigure(vmCtx, vcVM, *vmCtx.MoVM.Config)
+		refetchProps, updateErr = defaultReconfigure(vmCtx, s.K8sClient, vcVM)
+	}
+
+	if updateErr != nil {
+		updateErr = fmt.Errorf("updating state failed w %w", updateErr)
```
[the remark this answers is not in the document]
akutz: Ha, yeah, you and @bryanv literally had the exact same thought process :)

test/builder/vcsim_test_context.go
akutz: Yep. It did not exist when I wrote this. I'll switch to that :)

controllers/storageclass/storageclass_controller.go
```diff
@@ -0,0 +1,157 @@
...
+
+	ctx = pkgcfg.JoinContext(ctx, r.Context)
+	ctx = ctxsc.JoinContext(ctx, r.Context)
+
+	var obj storagev1.StorageClass
+	if err := r.Get(ctx, req.NamespacedName, &obj); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	logger := ctrl.Log.WithName("StorageClass").WithValues("name", req.Name)
```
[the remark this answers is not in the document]
akutz: Yes, because here I'd rather not use reflection every time an object is reconciled. If I need to rebase this again, I'll add the reflected name to the reconciler struct and re-use it here.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -357,6 +361,65 @@ type VirtualMachineCdromSpec struct {
...
+	// selected, the VM's home files and non-PVC disks will be encrypted.
+	//
+	// If there is a default key provider and a and a VM Class with a virtual,
+	// trusted platform module (vTPM) is selected, the VM's home files will be
+	// encrypted.
+	//
+	// If the underlying vSphere platform does not have a default key provider,
+	// then this field is required when specifying an encryption storage class
+	// and/or a VM Class with a vTPM.
+	ClassName string `json:"className,omitempty"`
```
aruneshpa, line 386: What are your thoughts on calling this `encryptionClassName`?  There's no stuttering since the parent field is called `spec.crypto`.   IMHO, `spec.cryto.encryptionClassName` is more accurate than `spec.crypto.className`.

Not to mention we have another field called `className` which might be confusing.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -357,6 +361,65 @@ type VirtualMachineCdromSpec struct {
...
+	// +optional
+
+	// ClassName describes the name of the EncryptionClass resource
+	// used to encrypt this VM.
+	//
+	// Please note, this field is not required to encrypt the VM. If the
+	// underlying platform has a default key provider, the VM may still be fully
+	// or partially encrypted depending on the specified storage and VM classes.
+	//
+	// If there is a default key provider and an encryption storage class is
```
aruneshpa, line 376: Hmm.  I think this has to be more nuanced than that.

I think by "VM's home files", you mean files in VM_HOME directory.  But that also include PVC disks (FCDs can be created in either the designated fcd/ directory, or VM_HOME).

api/v1alpha3/virtualmachine_types.go
akutz: Let me think about it. I am leaning towards agreeing with you.

api/v1alpha3/virtualmachine_types.go
akutz: No, by VM home files I mean any files in the VM directory that are *not* disks.

api/v1alpha3/virtualmachine_types.go
akutz: Done, thanks @aruneshpa !

api/v1alpha3/virtualmachine_types.go
```diff
@@ -357,6 +361,65 @@ type VirtualMachineCdromSpec struct {
...
+	// selected, the VM's configuration and non-PVC disks will be encrypted.
+	//
+	// If there is a default key provider and a and a VM Class with a virtual,
```
dilyar85, line 379: Nit: extra "and a" here.

controllers/storageclass/storageclass_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	return &Reconciler{
+		Context:    ctx,
+		Client:     client,
+		Logger:     logger,
+		Recorder:   recorder,
+		vmProvider: vmProvider,
+	}
+}
+
+// Reconciler reconciles a VirtualMachineClass object.
```
dilyar85, line 63: "VirtualMachineClass": should this be "StorageClass" based on `controlledType = &storagev1.StorageClass{}`  above?

controllers/storageclass/storageclass_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	logger := ctrl.Log.WithName("StorageClass").WithValues("name", req.Name)
+	ctx = logr.NewContext(ctx, logger)
+
+	if !obj.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{}, r.ReconcileNormal(ctx, &obj)
```
dilyar85, line 92: Nit: move the `logger/ctx` update after the "if" statement since they aren't used until then?

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -167,6 +178,54 @@ func classToVMMapperFn(
...
+					reconcile.Request{
+						NamespacedName: client.ObjectKey{
+							Namespace: vm.Namespace,
+							Name:      vm.Name,
+						},
+					})
+			}
+		}
+
+		logger.Info(
```
dilyar85, line 221: Also use `logger.V(4)` here to avoid log spam (since this will be called regardless of any requests exist)?

docs/concepts/workloads/vm.md
```diff
@@ -467,3 +467,118 @@ rockylinux_64Guest          Rocky Linux (64-bit)
...
+```yaml
+status:
+  conditions:
+  - type: VirtualMachineEncryptionSynced
+    status: False
+    reason: InvalidState
+    message: Must be powered off to encrypt VM.
+```
+
+If the VM requested encryption and it was successful, then then condition will be:
```
dilyar85, line 565: extra "then"

docs/concepts/workloads/vm.md
```diff
@@ -467,3 +467,118 @@ rockylinux_64Guest          Rocky Linux (64-bit)
...
+
+The following fields may be used to determine the encryption status of a VM:
+
+| Name | Description |
+|------|-------------|
+| `status.crypto.encrypted` | The observed state of the VM's encryption. May be `config`, `disks`, or both. |
+| `status.crypto.providerID` | The provider ID used to encrypt the VM. |
+| `status.crypto.keyID` | The key ID used to encrypt the VM. |
+
+For example, the following is an example of the status of a VM encrypted with an encrypted storage class:
```
dilyar85, line 540: I noticed you have a mixed use of "encryption storage class" and "encrypted storage class" here and also in the code. Should we stick to one term ("encryption storage class") to keep it consistent with the API field?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -472,6 +476,42 @@ func (v validator) validateStorageClass(ctx *pkgctx.WebhookRequestContext, vm *v
...
+	)
+
+	if encClassName == "" {
+		return nil
+	}
+
+	if !pkgcfg.FromContext(ctx).Features.BringYourOwnEncryptionKey {
+
+		allErrs = append(allErrs, field.Invalid(
+			path, encClassName, "feature not enabled"))
```
dilyar85, line 496: This would allow users to specify `spec.crypto` fields (other than `encryptionClassName`) when the FSS is not enabled. Consider returning early if the `spec.crypto` is empty as well?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -472,6 +476,42 @@ func (v validator) validateStorageClass(ctx *pkgctx.WebhookRequestContext, vm *v
 	return append(allErrs, field.Invalid(scPath, scName, fmt.Sprintf(storageClassNotAssignedFmt, vm.Namespace)))
 }
 
+func (v validator) validateCrypto(
+	ctx *pkgctx.WebhookRequestContext,
+	vm *vmopv1.VirtualMachine) field.ErrorList {
+
+	var (
+		allErrs      field.ErrorList
+		path         = field.NewPath("spec", "crypto", "className")
```
dilyar85, line 485: Should it be `encryptionClassName` at the end? This is not captured in test since you hard-coded the field path with this incorrect name.

dilyar85 (commented):
Dropping the first round of review for the API, controller and webhook code. Will post another review for the rest later.

pkg/util/kube/storage.go
```diff
@@ -0,0 +1,154 @@
...
+	k8sClient ctrlclient.Client,
+	storageClassName string) (bool, error) {
+
+	var storageClass storagev1.StorageClass
+	if err := k8sClient.Get(
+		ctx,
+		ctrlclient.ObjectKey{Name: storageClassName},
+		&storageClass); err != nil {
+
+		if apierrors.IsNotFound(err) {
```
aruneshpa, line 126: Here and below, simplify with:

`return false, client.IgnoreNotFound(err)`

webhooks/virtualmachine/validation/virtualmachine_validator.go
akutz: Yes, because `crypto.useDefaultKeyProvider` is currently the default behavior for today. But yes, I should disallow setting that field to a non-nil value as well.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -357,6 +361,65 @@ type VirtualMachineCdromSpec struct {
...
+	//
+	// - If a VirtualMachine is encrypted with a provider other than the default
+	//   key provider, the VM will be rekeyed using the default key provider.
+	//
+	// When an explicit EncryptionClass is not provided and this value is false:
+	//
+	// - Deploying a VirtualMachine with an encryption storage policy or vTPM
+	//   will fail.
+	//
+	// - If a VirtualMachine is encrypted with a provider other than the default
```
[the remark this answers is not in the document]
akutz: Hmm, fair. However, I wrote the API docs from the perspective of the reconciler, not the webhooks (create vs update). Let me think on this.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -357,6 +361,65 @@ type VirtualMachineCdromSpec struct {
...
+	//   key provider, the VM will be not be rekeyed.
+	//
+	//   Please note, this could result in a VirtualMachine that cannot be
```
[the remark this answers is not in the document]
akutz: Because if the underlying key or provider is removed, and the VM is not rekeyed to use the default provider, then you cannot power on the VM since it is encrypted and there is no key to decrypt it.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -694,6 +762,35 @@ type VirtualMachineAdvancedSpec struct {
 	ChangeBlockTracking *bool `json:"changeBlockTracking,omitempty"`
 }
 
+type VirtualMachineEncryptionType string
+
+const (
+	VirtualMachineEncryptionTypeConfig VirtualMachineEncryptionType = "Config"
+	VirtualMachineEncryptionTypeDisks  VirtualMachineEncryptionType = "Disks"
```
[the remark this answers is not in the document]
akutz: Hmmm. True. I will likely update the status to indicate that `Disks` does not include PVCs.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -694,6 +762,35 @@ type VirtualMachineAdvancedSpec struct {
 	ChangeBlockTracking *bool `json:"changeBlockTracking,omitempty"`
 }
 
+type VirtualMachineEncryptionType string
+
+const (
+	VirtualMachineEncryptionTypeConfig VirtualMachineEncryptionType = "Config"
```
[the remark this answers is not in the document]
akutz: It does. But the documentation explains this (see `vm.md`). VMFiles is not descriptive either unless you already know.

pkg/util/kube/storage.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return policyID, nil
+}
+
+// MarkEncryptedStorageClass records the provided StorageClass as encrypted.
+func MarkEncryptedStorageClass(
+	ctx context.Context,
+	k8sClient ctrlclient.Client,
+	storageClass storagev1.StorageClass,
+	encrypted bool) error {
```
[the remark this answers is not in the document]
akutz: Because nothing stops someone from modifying the underlying Storage Policy to no longer support encryption. This function is called by the StorageClass controller that ensures we always represent the current state of the underlying policy.

pkg/util/kube/storage.go
```diff
@@ -0,0 +1,154 @@
...
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+
+		//
+		// The ConfigMap was not found.
+		//
+
+		if !encrypted {
+			// If the goal is to mark the StorageClass ss not encrypted, then we
```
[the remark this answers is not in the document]
akutz: I'm keeping it, and ack.

pkg/util/kube/storage.go
```diff
@@ -0,0 +1,154 @@
...
+			// do not need to actually create the underlying ConfigMap if it does
+			// not exist.
+			return nil
+		}
+
+		// The ConfigMap does not already exist and the goal is to mark the
+		// StorageClass as encrypted, so go ahead and create the ConfigMap and
+		// return early.
+		obj.OwnerReferences = []metav1.OwnerReference{ownerRef}
```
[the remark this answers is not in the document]
akutz: I am using OwnerRefs because Sandeep will not allow me to annotate the StorageClasses and he did not like me placing a finalizer on them either when I was directly modifying the ConfigMap with the SC names. Neither the GoDocs nor the Kube docs mention any limit.

pkg/util/kube/storage.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil
+	case !encrypted && !storageClassIsOwner:
+		// The StorageClass should be marked as not encrypted, which means it
+		// should not be in the ConfigMap. Since the StorageClass is not
+		// currently in the ConfigMap, we can return return early.
+		return nil
+	}
+
+	// Create the patch used to update the ConfigMap.
+	objPatch := ctrlclient.StrategicMergeFrom(obj.DeepCopyObject().(ctrlclient.Object))
```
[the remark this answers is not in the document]
akutz: I wrote CreateAndPatch, so I am aware of what it does. The model I am using here is more efficient since I do not need to ever update *or* create the object if `encryption` is false.

docs/concepts/workloads/vm.md
akutz: No, the extra then is a typo for `the` :)

dilyar85 (approved):
LGTM with some minor comments left above. Thanks @akutz for the clear docs and comments on this feature!

controllers/storageclass/storageclass_controller.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (r *Reconciler) ReconcileNormal(
```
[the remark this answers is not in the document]
akutz: That is correct. Any errors will be reported via controller-runtime. I do not believe in happy path logging except where it provides value. Here it does nothing but clutter logs.

pkg/util/kube/storage.go
```diff
@@ -0,0 +1,154 @@
...
+	encrypted bool) error {
+
+	var (
+		obj = corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: pkgcfg.FromContext(ctx).PodNamespace,
+				Name:      internal.EncryptedStorageClassNamesConfigMapName,
+			},
+		}
+		ownerRef = internal.GetOwnerRefForStorageClass(storageClass)
```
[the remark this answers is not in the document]
akutz: Then you did not keep reading :) It is used for testing, hence the `internal` package. That way I can use it between prod and test without exposing it to beyond these two packages.

api/v1alpha3/virtualmachine_types.go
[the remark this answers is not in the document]
akutz: I updated the GoDocs/OpenAPI docs as well to explain this further.

api/v1alpha3/virtualmachine_types.go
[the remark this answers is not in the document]
akutz: I updated the comment to make it more clear.

pkg/vmconfig/vmconfig_reconciler.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/vmware/govmomi/vim25/mo"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
+)
+
+// Reconciler is a type that modifies a VirtualMachine's configuration before
+// and/or after creating or updating the underlying, platform resource.
+type Reconciler interface {
```
aruneshpa, line 19: I know you are trying to introduce a pattern for this.  But reading through this code and the diff in the PR, my opinion is that we introduce this pattern when we have a _need_ for it.  Right now, it seems a bit too complicated for something that might not even be used by other components (if we never get to breaking the current pre/post power on reconfigure thing in this mode).

If you don't mind, I would recommend shelving all this, keeping this change specific to reconciling the StorageClasses and the VMs that are referencing them.

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -0,0 +1,690 @@
...
+
+	byokv1 "github.com/vmware-tanzu/vm-operator/external/byok/api/v1alpha1"
+	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/paused"
+	"github.com/vmware-tanzu/vm-operator/pkg/vmconfig/crypto/internal"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
+)
+
+type ptrCfgSpec = *vimtypes.VirtualMachineConfigSpec
```
aruneshpa, line 28: Hmm.  Is this a pattern?  What does this buy us compared to just using `*vimtypes.VirtualMachineConfigSpec`?

pkg/vmconfig/crypto/crypto_reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+	case ReasonInvalidChanges:
+		return "InvalidChanges"
+	case ReasonReconfigureError:
+		return "ReconfigureError"
+	}
+	return ""
+}
+
+func (r Reason) String() string {
+	return bitmask.String(r)
```
aruneshpa, line 95: Andrew, I would rather have us just use the reasons explicitly, as opposed to define a type for them, and then use bitmasks to figure out the reasons that apply.  This will be hard to maintain and I don't see many usecases for this pattern. 

My recommendation would be to shelve the bitmask manipulation entirely and references the reasons explicitly where needed.

akutz: The reasons *are* used explicitly, but there can be many different kinds and combinations. So I disagree there is not a use case for this in the future. There is literally a use case for it in this PR. The documentation in `vm.md` clearly outlines the reasons to the user. I appreciate the feedback, but I am going to keep this as-is.

pkg/vmconfig/crypto/crypto_reconciler_pre.go
akutz: It is used elsewhere in our codebase, yes. The types are exactly the same, and it enables more concise code. Even intellisense shows what it actually is.

pkg/vmconfig/vmconfig_reconciler.go
akutz: I appreciate the feedback, but I am going to keep this. It is part of my vision going forward, and this change highlights *that* the pattern works and *how* the pattern works. I initially did this change without the pattern, and it ended up looking very similar, just without the wrapper. All of VM Op should start adopting this pattern in order to make our reconciler less fragile than it is and so difficult to change.

api/v1alpha3/virtualmachine_types.go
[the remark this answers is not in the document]
akutz: So we cannot avoid this situation. If the underlying key provider is gone, then a user has to be able to set `spec.crypto.encryptionClassName` to empty to bypass the failure state of being unable to find the `EncryptionClass`. When they set it to empty, the VM is either rekeyed using the default provider or not, based on the value of `spec.crypto.useDefaultKeyProvider`. This is aligned with the existing behavior/options of VCD.

pkg/vmconfig/crypto/crypto_reconciler.go
akutz: FWIW, here is [a test case](https://github.com/vmware-tanzu/vm-operator/pull/673/files#diff-916fea486b6a687a518cffddb850c48ae19a10f9f42864ac3e3507126a6cc233R324-R341) that shows when the reason can be both `InvalidState` _and_ `InvalidChanges`

akutz: All that said, this _is_ something I am willing to revisit. I will give it more thought, but I am not going to let it block this PR. After it merges, I'll sit down and consider how many possible combinations there can be in practical terms. If it is just `InvalidState` and `InvalidChanges`, then I can have a distinct reason for that. Let me look.

akutz:
@aruneshpa I have addressed all of your feedback on the PR. While I did not accept it all, and stated my reasons, unless you have a strong objection, I am going to merge this PR so it does not continue to grow. We can revisit your concerns post-merge if that is alright?

aruneshpa:
> @aruneshpa I have addressed all of your feedback on the PR. While I did not accept it all, and stated my reasons, unless you have a strong objection, I am going to merge this PR so it does not continue to grow. We can revisit your concerns post-merge if that is alright?

Sounds good to me.  Thanks!

PR #677, opened 2024-08-21
🌱 Use PartialObjectMetadata to unmarshal VM for backup needed check

bryanv:

For VMs with like a bunch of volumes we don't need to decode the whole VM since we're only comparing fields in the metadata.

Use maps.Equal instead of DeepEqual to compare the label and annotation string maps.

Fixes #

```release-note
NONE
```


dilyar85 (approved):
TIL about `metav1.PartialObjectMetadata{}`. Thanks BV for optimizing this!

PR #678, opened 2024-08-22
🐛 Start infra capability controller for non-leader pods too

bryanv:

We want capabilities updates to be reflected in the non-leader pod as well so the webhooks are aware of the current enabled capabilities.

Remove the VMProvider from the infra capability controller. It was not used, and we don't want it on the non-leader pods due to VC session limits (and it would need the other infra controllers to be non-leader as well to pick up cred rotations)

Fixes #

```release-note
NONE
```


zyiyi11 (approved):
LGTM. 
just one nit: typo on title `non-loader`

PR #682, opened 2024-08-23
🐛 Fix panic when logging ConfigSpec w nil iface pointer

akutz:

This patch fixes a panic that occurs when the vimtypes JSON encoder is used to log a ConfigSpec that includes a field of type interface set to a pointer w a nil value. This patch:

* Prevents the issue discovered in the `GetOVFVAppConfigForConfigSpec` to `GetMergedvAppConfigSpec` code path
* Adds support for `SafeConfigSpecToString` as a function that can be used to prevent this from ever happening again.

Fixes `NA`

```release-note
Do not panic when logging ConfigSpec that includes field of type interface set to pointer with nil value
```

### DIFF
--- a/pkg/util/configspec.go
+++ b/pkg/util/configspec.go
@@ -6,6 +6,8 @@ package util
 import (
 	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
 	"reflect"
 
 	"github.com/vmware/govmomi/vim25"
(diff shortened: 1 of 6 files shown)


pkg/util/configspec.go
```diff
@@ -190,3 +191,35 @@ func EnsureMinHardwareVersionInConfigSpec(
 	}
 	configSpec.Version = configSpecHwVersion.String()
 }
+
+// SafeConfigSpecToString returns the string-ified version of the provided
+// ConfigSpec, first trying to use the special JSON encoder, then defaulting to
+// the normal JSON encoder.
+func SafeConfigSpecToString(
```
akutz, line 198: @dougm We should probably just add `String() string` to the ConfigSpec type in GoVmomi and implement this function there.

pkg/util/configspec.go
```diff
@@ -190,3 +191,35 @@ func EnsureMinHardwareVersionInConfigSpec(
...
+	configSpec *vimtypes.VirtualMachineConfigSpec) (s string) {
+
+	if configSpec == nil {
+		return ""
```
dougm, line 202: May not matter since this is just used for logging, but normal json.Marshal would return `"null"` in this case.

dougm (approved):
lgtm

pkg/util/configspec.go
akutz: Yeah, I was thinking the same after I posted the above.

akutz: And lets make docs clear on the func it's not meant for JSON encoding, but rather when wanting to print the ConfigSpec as a string.

pkg/util/configspec.go
akutz: Yep. Fair, I'll update it.

pkg/util/configspec.go
```diff
@@ -190,3 +192,35 @@ func EnsureMinHardwareVersionInConfigSpec(
...
+func SafeConfigSpecToString(
+	configSpec *vimtypes.VirtualMachineConfigSpec) (s string) {
+
+	if configSpec == nil {
+		return "null"
+	}
+
+	marshalWithNormalJSONEncoder := func() {
+		if data, err := json.Marshal(*configSpec); err != nil {
+			s = fmt.Sprintf("%v", *configSpec)
```
akutz, line 208: @dougm Since you already reviewed this, I just wanted to highlight that I also made this change post-your review. This way, no matter what, there's *something* returned.

aruneshpa (approved):
LGTM.

PR #685, opened 2024-08-23
🌱 Introduce condition to indicate the status of latest backup

sreyasn:

This change indicates the status of the latest backup as a condition on the VM resource to inform the latest backup.

Fixes # N/A

Any suggestions for naming the condition welcome

```
Add condition to indicate the status of latest backup
```

### DIFF
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -154,6 +154,20 @@ const (
 	ManagedByExtensionType = "VirtualMachine"
 )
 
+const (
+	// VirtualMachineBackupUpToDateCondition exposes the status of the latest VirtualMachine Backup, when available.
+	VirtualMachineBackupUpToDateCondition = "VirtualMachineBackupUpToDate"
+
+	// VirtualMachineBackupPaused documents that VirtualMachine backup is paused.
+	// This can happen after the backing virtual machine is restored by a backup/restore vendor, or a failover operation
+	// by a data protection solution. In either of these operations, VM operator does not persist backup information and
+	// waits for the virtual machine to be (re)-registered with VM Service.
+	VirtualMachineBackupPausedReason = "VirtualMachineBackupPaused"
+
+	// VirtualMachineBackupFailed documents that the VirtualMachine backup failed due to an error.
+	VirtualMachineBackupFailedReason = "VirtualMachineBackupFailed"
+)
+
 // VirtualMachine backup/restore related constants.
 const (
 	// VMResourceYAMLExtraConfigKey is the ExtraConfig key to persist VM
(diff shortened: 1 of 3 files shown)


pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -198,8 +210,11 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) error {
 
 		// Set the VirtualMachine's backup version annotation once reconfigure succeeds.
 		if pkgcfg.FromContext(opts.VMCtx).Features.VMIncrementalRestore {
-			setLastBackupVersionAnnotation(opts.VMCtx.VM, opts.BackupVersion)
-			// TODO: add a condition to indicate backup has resumed
+			setBackupVersionAnnotation(opts.VMCtx.VM, opts.BackupVersion)
+			conditions.MarkTrue(
```
bryanv, line 214: Is the expectation of this true condition to report the time of the last backup? Because the last transition time will only be updated when there is an actual transition.

sreyasn: Thanks for the tip, BV. I'm updating the condition's message with the latest version now. This should also force a lastTransition time update.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -154,6 +154,14 @@ const (
 	ManagedByExtensionType = "VirtualMachine"
 )
 
+const (
+	// VirtualMachineBackupCondition exposes the status of the latest VirtualMachine Backup, when available.
+	VirtualMachineBackupLatestCondition = "VirtualMachineBackupLatest"
```
aruneshpa, line 159:
```suggestion
	VirtualMachineBackupUpToDateCondition = "VirtualMachineBackupUpToDate"
```

Kind of like: https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/api/v1beta1/condition_consts.go#L52


And then, the reasons should be distinct as well.  Backup not up-to-date because a restore happened is different from an error comparing the backup version.
Same as how CAPI has different reasons for the uptodate condition:
https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/internal/controllers/controller.go#L416

```
	// RollingUpdateInProgressReason (Severity=Warning) documents a KubeadmControlPlane object executing a
	// rolling upgrade for aligning the machines spec to the desired state.
	RollingUpdateInProgressReason = "RollingUpdateInProgress"
```

aruneshpa (changes requested):
LGTM, with one comment.

api/v1alpha3/virtualmachine_types.go
sreyasn: Thanks. Updated and added a new FailedReason for err != nil . PTAL

dougm (approved):
lgtm

aruneshpa (approved):

PR #688, opened 2024-08-27
🐛 Handle disk backing changes when backing up PVCDiskData

dougm:
Vendor's partial restore (e.g. single disk) method may create a new, non-FCD disk. The restored disk has new UUID and lives the the VM's directory, not the fcd/ directory. Given that FCD disk names are unique, we consult the backup data to compare using `FileName` before `Uuid`.


dilyar85 (approved):
LGTM. Thanks Doug!

PR #689, opened 2024-08-29
🐛 Have EnsureDisksHaveControllers() alloc unit numbers

bryanv:

Some VC placement changes now require the disks to have unit number assigned so update EnsureDisksHaveControllers()  to do that.

Fixes #

```release-note
NONE
```


zyiyi11 (approved):
LGTM

akutz (changes requested):
I'm not fan of the introduction of the `panic`. We should simply indicate we cannot reconcile the VM. Please remove this and return `-1` that is handled as an error up the call stack, or provide clear evidence as to why it is needed.

akutz (approved):
@bryanv explained we only call the `Must` variant directly after adding a controller, so there should *always* be a unit number, otherwise it's a bug in the code.

PR #693, opened 2024-09-03
:bug: Do not reconcile volumes if VM has pause annotation set

aruneshpa:

Even if a VM has pause annotation set, we end up reconciling the volumes of a VM.  This results in volumes being detached from paused VMs.  This change fixes that behavior by returning early if the paused annotation is found on the VM.

```release-note
Do not reconcile volumes if VM has pause annotation set
```

### DIFF
--- a/controllers/virtualmachine/volume/volume_controller_intg_test.go
+++ b/controllers/virtualmachine/volume/volume_controller_intg_test.go
@@ -316,6 +316,43 @@ func intgTestsReconcile() {
 			Expect(err == nil || apierrors.IsNotFound(err)).To(BeTrue())
 		})
 
+		When("VM has a pause annotation", func() {
+			BeforeEach(func() {
+				vm.Annotations = make(map[string]string)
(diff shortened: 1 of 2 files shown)


controllers/virtualmachine/volume/volume_controller_intg_test.go
```diff
@@ -316,6 +316,43 @@ func intgTestsReconcile() {
...
+					vm.Status.BiosUUID = dummyBiosUUID
+					Expect(ctx.Client.Status().Update(ctx, vm)).To(Succeed())
+				})
+
+				By("Add CNS volume to Spec.Volumes", func() {
+					vm.Spec.Volumes = append(vm.Spec.Volumes, vmVolume1)
+					Expect(ctx.Client.Update(ctx, vm)).To(Succeed())
+				})
+
+				By("CnsNodeVmAttachment should be created for volume", func() {
```
dilyar85, line 346: should "not" be created?

dilyar85 (approved):
LGTM!

dougm (approved):
lgtm

PR #698, opened 2024-09-05
🌱 Drop managedFields and last-applied-config annotation from backup

dilyar85:

Persisting last-applied-config annotation and managed fields could doubles the size of the strings we are handling during backup.  Drop those when recording backup.

Fixes N/A.

- Updated unit tests to reflect this change and also verified backup data in an internal testbed:

```console
# Deployed a VM with last-applied annotation and managed fields:
$ kubectl get vm -n sdiliyaer-test test-vm -o yaml --show-managed-fields
kind: VirtualMachine
metadata:
  annotations:
      {"apiVersion":"vmoperator.vmware.com/v1alpha1","kind":"VirtualMachine","metadata":{"name":"test-vm","namespace":"sdiliyaer-test"},"spec":{"className":"best-effort-small","imageName":"ubuntu-impish-21.10-cloudimg","powerState":"poweredOn","storageClass":"wcpglobal-storage-profile","vmMetadata":{"secretName":"test-bootstrap-data","transport":"CloudInit"}}}
  managedFields:
  - apiVersion: vmoperator.vmware.com/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
...

# Confirmed neither last-applied annotation nor managed fields in the VM YAML backup:
$ govc vm.info -e 'test-vm' | grep vmservice.virtualmachine.resource.yaml | awk '{print $2}' | base64 -d | gunzip | yq eval '.metadata | keys' -
- annotations
- creationTimestamp
- finalizers
- generation
- labels
- name
- namespace
- resourceVersion
- uid

$ govc vm.info -e 'test-vm' | grep vmservice.virtualmachine.resource.yaml | awk '{print $2}' | base64 -d | gunzip | yq eval '.metadata.annotations | keys' -
- virtualmachine.vmoperator.vmware.com/first-boot-done
- vmoperator.vmware.com/created-at-build-version
- vmoperator.vmware.com/created-at-schema-version
- vmoperator.vmware.com/manager-id
```

- Verified the same result for additional bootstrap resources and incremental backups as well.

```release-note
Drop managedFields and last-applied-config from backup.
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/backup.go
+++ b/pkg/providers/vsphere/virtualmachine/backup.go
@@ -169,6 +169,7 @@ func BackupVirtualMachine(opts BackupVirtualMachineOptions) (result error) {
 			copyVM := opts.VMCtx.VM.DeepCopy()
 			setBackupVersionAnnotation(copyVM, opts.BackupVersion)
 			// Backup the updated VM's YAML with encoding and compression.
+			trimBackupFields(copyVM)
 			encodedVMYaml, err := getEncodedVMYaml(copyVM)
 			if err != nil {
 				opts.VMCtx.Logger.Error(err, "failed to get VM resource yaml for backup")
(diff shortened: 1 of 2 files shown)


dougm (approved):
lgtm

pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -271,12 +272,26 @@ func getDesiredVMResourceYAMLForBackup(
 		return "", err
 	}
 
+	// Use a copy to avoid modifying the VM object at the end of reconciliation.
+	copyVM := vm.DeepCopy()
```
aruneshpa, line 276: Line [168](https://github.com/vmware-tanzu/vm-operator/pull/698/files#diff-5b4870681742b58d08e67601c0496f92bade48b9cb23fe808895e318abb13ccfR169) in this file is also creating a deep copy of the VM.  Can we please dedupe that?

aruneshpa (approved):

pkg/providers/vsphere/virtualmachine/backup.go
dilyar85: They aren’t duplicated because that line will only be executed when the incremental FSS is enabled, and this function is called when the FSS is disabled. I thought about getting a deep copy before the FSS check, but it's not always needed when the FSS is enabled (e.g. when any of the `!ok || backupVersionAnnotation == "" || len(ecToUpdate) != 0 || !vmBackupInSync` returns false at line 161).

PR #700, opened 2024-09-07
Allow imageless VM creation when Incremental Restore FSS is enabled

aruneshpa:

Currently, we allow imageless VM creation when importing a VM is supported.  However, after a VM is restored by a VADP based vendor or a DR solution, the destination vCenter and Supervisor might not be the same as the primary.  As such, any references to images might be stale.

This change modifies the existing code to allow imageless VMs if the VM operator supports restoring into existing VMs.

```release-note
Allow imageless VM creation when Incremental Restore FSS is enabled
```


dilyar85 (approved):
LGTM!

bryanv (approved):

PR #701, opened 2024-09-08
🌱 Add image type label in VMI and CVMI objects

dilyar85:

This PR adds image type info as labels to VMI & CVMI objects so that they can be filtered by label selector from the client side. It also updates the `SyncStatusToLabels` function to remove any labels if the values from image status is empty or invalid.

Fixes N/A.

- Added tests to maintain 100% coverage:

```console
ok  	github.com/vmware-tanzu/vm-operator/pkg/util/image	0.481s	coverage: 100.0% of statements
```
 
- Deployed the change to an internal testbed and verified expected labels in both VMI and CVMI objects:

```console
$ kubectl get cvmi -o json | jq '.items[] | {"image.vmoperator.vmware.com/type": .metadata.labels["image.vmoperator.vmware.com/type"], ".status.type": .status.type}'
{
  "image.vmoperator.vmware.com/type": "ISO",
  ".status.type": "ISO"
}
{
  "image.vmoperator.vmware.com/type": "OVF",
  ".status.type": "OVF"
}

$ kubectl get vmi -n sdiliyaer-test -o json | jq '.items[] | {"image.vmoperator.vmware.com/type": .metadata.labels["image.vmoperator.vmware.com/type"], ".status.type": .status.type}'
{
  "image.vmoperator.vmware.com/type": "OVF",
  ".status.type": "OVF"
}
{
  "image.vmoperator.vmware.com/type": "ISO",
  ".status.type": "ISO"
}
```

```release-note
Add image type label in VMI and CVMI objects.
```

### DIFF
--- a/controllers/contentlibrary/utils/utils.go
+++ b/controllers/contentlibrary/utils/utils.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 VMware, Inc. All Rights Reserved.
+// Copyright (c) 2022-2024 VMware, Inc. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package utils
@@ -13,7 +13,7 @@ import (
 
 	imgregv1a1 "github.com/vmware-tanzu/image-registry-operator-api/api/v1alpha1"
 
-	"github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
 )
 
 // GetImageFieldNameFromItem returns the Image field name in format of "vmi-<uuid>"
@@ -65,7 +65,7 @@ func AddContentLibraryRefToAnnotation(to client.Object, ref *imgregv1a1.NameAndK
 	if annotations == nil {
 		annotations = map[string]string{}
 	}
-	annotations[v1alpha2.VMIContentLibRefAnnotation] = string(data)
+	annotations[vmopv1.VMIContentLibRefAnnotation] = string(data)
 	to.SetAnnotations(annotations)
 
 	return nil
--- a/pkg/util/image/status_to_label.go
+++ b/pkg/util/image/status_to_label.go
@@ -4,54 +4,62 @@
 package image
 
 import (
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
 )
 
-// SyncStatusToLabels copies the image's capabilities and OS information from
-// the image status to well-known labels on the image in order to make it
+// SyncStatusToLabels copies the image's type, capabilities and OS information
+// from the image status to well-known labels on the image in order to make it
 // easier for clients to search for images based on capabilities or OS details.
+// If the image status does not have the info or is invalid, remove the label.
 func SyncStatusToLabels(
 	obj metav1.Object,
 	status vmopv1.VirtualMachineImageStatus) {
 
-	var (
-		osID      = status.OSInfo.ID
-		osType    = status.OSInfo.Type
-		osVersion = status.OSInfo.Version
-		caps      = status.Capabilities
-	)
-
-	if osID == "" && osType == "" && osVersion == "" && len(caps) == 0 {
-		return
-	}
-
 	labels := obj.GetLabels()
 	if labels == nil {
 		labels = map[string]string{}
+		obj.SetLabels(labels)
 	}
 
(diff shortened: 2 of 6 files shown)


controllers/contentlibrary/utils/utils.go
```diff
@@ -65,8 +65,18 @@ func AddContentLibraryRefToAnnotation(to client.Object, ref *imgregv1a1.NameAndK
...
 	return nil
 }
+
+// AddImageTypeToLabel adds the image type label to the given VMI or CVMI object.
+func AddImageTypeToLabel(to client.Object, imageType string) {
+	labels := to.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels[vmopv1.VirtualMachineImageTypeLabel] = imageType
```
bryanv, line 80: Check that the label value is valid - there's other examples in this package.

dilyar85: Thanks! Addressed in https://github.com/vmware-tanzu/vm-operator/compare/83759cd66b4c75f0ab73da4260a9cb00ff0cda7d..719de5bea73dc46dae720b777067d51bfb2afe5b

controllers/contentlibrary/utils/utils.go
```diff
@@ -65,8 +66,24 @@ func AddContentLibraryRefToAnnotation(to client.Object, ref *imgregv1a1.NameAndK
...
+func AddImageTypeToLabel(to client.Object, imageType string) error {
+	if errStrs := validation.IsValidLabelValue(imageType); errStrs != nil {
+		return fmt.Errorf("invalid image type label value %q: %v", imageType, errStrs)
+	}
+
+	labels := to.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels[vmopv1.VirtualMachineImageTypeLabel] = imageType
```
dougm, line 86: Seems like this could added to`image.SyncStatusToLabels`, rather than adding AddImageTypeToLabel? https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/util/image/status_to_label.go#L16

controllers/contentlibrary/utils/utils.go
```diff
@@ -65,8 +66,24 @@ func AddContentLibraryRefToAnnotation(to client.Object, ref *imgregv1a1.NameAndK
...
 	to.SetAnnotations(annotations)
 
 	return nil
 }
+
+// AddImageTypeToLabel adds the image type label to the given object if the
+// label value is valid.
+func AddImageTypeToLabel(to client.Object, imageType string) error {
+	if errStrs := validation.IsValidLabelValue(imageType); errStrs != nil {
+		return fmt.Errorf("invalid image type label value %q: %v", imageType, errStrs)
```
bryanv, line 79: We really cannot return an error here. If the value isn't valid there is no point of endlessly retrying, and it will prevent the k8s object from being updated.

Either just return, set the value of an empty string, or delete the label if it exists in this case.

dilyar85: Addressed in the latest commit.

dougm (approved):
lgtm

pkg/util/image/status_to_label.go
```diff
@@ -10,21 +10,22 @@ import (
...
+		imgType   = status.Type
 		osID      = status.OSInfo.ID
 		osType    = status.OSInfo.Type
 		osVersion = status.OSInfo.Version
 		caps      = status.Capabilities
 	)
 
-	if osID == "" && osType == "" && osVersion == "" && len(caps) == 0 {
+	if imgType == "" && osID == "" && osType == "" && osVersion == "" && len(caps) == 0 {
 		return
```
bryanv, line 29: Should we remove the annotations in this case? Or set their values to ""?

dilyar85: Updated to remove.

pkg/util/image/status_to_label.go
```diff
@@ -4,54 +4,66 @@
...
-			}
+	// Delete any stale capability labels.
+	for k := range labels {
+		if _, ok := capKeys[k]; !ok && strings.HasPrefix(k, kp) {
+			delete(labels, k)
 		}
 	}
 
-	obj.SetLabels(labels)
+	if !reflect.DeepEqual(labels, obj.GetLabels()) {
```
bryanv, line 64: Why do this instead updating the objects labels in-place (which will already happen when object's labels != nil)

dilyar85: Good point. Updated to invoke `SetLabels` if the labels == nil and removed this block at the end. See: https://github.com/vmware-tanzu/vm-operator/compare/0155a2f9769e4842838101f19c2dac8ce6a8a85e..196d3999562481d1893eb4996acca22fadcebcde

PR #704, opened 2024-09-10
🐛 Use cached Get() of PVCs in Volume controller

bryanv:

The uncached Get() of every VM's PVCs on each reconcile causes an unacceptable perf regression.

The memory limit bump fallout from this change will be addressed later.

For now, keep the deferred instance-storage specific cache Watch(). It only matches IS PVCs via the label selector its cost is negligible but can be removed in a later commit.

Fixes #

```release-note
NONE
```


aruneshpa (approved):
LGTM.

sreyasn (approved):

PR #707, opened 2024-09-11
✨ Strip managed fields for cached objects

akutz:

This patch adds support for stripping managed fields for objects entering the cache. The controller-runtime [documentation for this transformer](https://github.com/kubernetes-sigs/controller-runtime/blob/38546806f2faf5973e3321a7bd5bb3afdbb5767d/pkg/cache/cache.go#L350-L362) states:

> If you are not explicitly accessing managedFields from your code, setting this as `DefaultTransform` on the cache can lead to significant reduction in memory usage.

as well as the PR, https://github.com/kubernetes-sigs/controller-runtime/pull/2791, which states:

> There are no known issues when using this as a `DefaultTransform` unless there is explicit code to access objects `ManagedFields` and can lead to a significant reduction in memory usage.

Given that VM Operator never directly accesses the managed fields for any object, there is no reason we should not be using this transformer.

Fixes `NA`

I would appreciate @sbueringer taking a look to make sure this change was implemented as intended. Thanks!

```release-note
Reduce memory usage by stripping managed fields from the cache
```

### DIFF
--- a/pkg/manager/cache.go
+++ b/pkg/manager/cache.go
@@ -35,9 +35,10 @@ func NewLabelSelectorCacheForObject(
 
 	cache, err := ctrlcache.New(mgr.GetConfig(),
 		ctrlcache.Options{
-			Scheme:     mgr.GetScheme(),
-			Mapper:     mgr.GetRESTMapper(),
-			SyncPeriod: resync,
+			Scheme:           mgr.GetScheme(),
+			Mapper:           mgr.GetRESTMapper(),
+			DefaultTransform: ctrlcache.TransformStripManagedFields(),
+			SyncPeriod:       resync,
 			ByObject: map[ctrlclient.Object]ctrlcache.ByObject{
 				object: {
 					Label: selector,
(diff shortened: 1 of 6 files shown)


dilyar85 (approved):
Thanks @akutz!

alvaroaleman:
Please do not ping me on projects I have zero involvement in.

akutz:
@bryanv / team,

Just to be on the safe side, I reached out to the author of the transformer on the PR that introduced it at https://github.com/kubernetes-sigs/controller-runtime/pull/2791#issuecomment-2343691274 to double-check whether this change will have any potential downsides.

(Please do not CC the author's name here as he does not want alerts on projects for which he is not involved)

pkg/manager/cache.go
```diff
@@ -70,12 +71,14 @@ func NewNamespacedCacheForObject(
...
-			Mapper:     mgr.GetRESTMapper(),
-			SyncPeriod: resync,
+			Scheme:           mgr.GetScheme(),
+			Mapper:           mgr.GetRESTMapper(),
+			DefaultTransform: ctrlcache.TransformStripManagedFields(),
+			SyncPeriod:       resync,
 			ByObject: map[ctrlclient.Object]ctrlcache.ByObject{
 				object: {
 					Namespaces: GetNamespaceCacheConfigs(namespaces...),
+					Transform:  ctrlcache.TransformStripManagedFields(),
```
sbueringer, line 81: This one is not required. The cache configs below Namespaces inherits the DefaultTransform from above.

sbueringer:
In general I would be curious how much memory it saves you, if you'll ever get around to measure the difference in scale tests :)

pkg/manager/cache.go
sbueringer: @akutz Just fyi, edited ^^ I was looking through the defaulting logic of cache options which is slightly complicated and also testing it.

As far as I can tell DefaultTransform is always used to default Transform on lower levels. Doesn't matter if:
* object is a cluster-wide object
* namespaces is nil
* regular case of a namespaced object that should be watched in >= 1 namespaces

akutz:
> In general I would be curious how much memory it saves you, if you'll ever get around to measure the difference in scale tests :)

Thanks @sbueringer,

We are definitely encountering a lot of memory overhead right now due to recently including PVCs in the cache. Given scale testing uses kubectl and that includes managed fields, we are hoping stripping them will result in a good chunk of memory being saved.

bryanv (approved):

PR #709, opened 2024-09-11
🐛  Fix NPE in checking VM paused by admin

dilyar85:

This PR fixes an potential NPE when the `mo.VirtualMachine` has a nil `Config *types.VirtualMachineConfigInfo` when checking if the VM is paused by VI Admin.

Fixes N/A.

This issue was reported from an internal test (3434725). 

```release-note
Fix NPE in checking if VM is paused by admin with nil config.
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/extraconfig.go
+++ b/pkg/providers/vsphere/virtualmachine/extraconfig.go
@@ -13,12 +13,17 @@ import (
 )
 
 func IsPausedByAdmin(moVM mo.VirtualMachine) bool {
-	for i := range moVM.Config.ExtraConfig {
-		if o := moVM.Config.ExtraConfig[i].GetOptionValue(); o != nil {
+	if moVM.Config == nil {
+		return false
+	}
+
+	for _, ec := range moVM.Config.ExtraConfig {
+		if o := ec.GetOptionValue(); o != nil {
 			if o.Key == vmopv1.PauseVMExtraConfigKey {
 				if value, ok := o.Value.(string); ok {
 					return strings.ToUpper(value) == constants.ExtraConfigTrue
 				}
+				return false
 			}
 		}
 	}
--- a/pkg/providers/vsphere/virtualmachine/extraconfig_test.go
+++ b/pkg/providers/vsphere/virtualmachine/extraconfig_test.go
@@ -36,6 +36,16 @@ func extraConfigTests() {
 			}
 		})
 
+		It("should return false when VM's Config is nil", func() {
+			mgdObj.Config = nil
+			Expect(virtualmachine.IsPausedByAdmin(mgdObj)).To(BeFalse())
+		})
+
+		It("should return false when VM's Config.ExtraConfig is nil", func() {
+			mgdObj.Config.ExtraConfig = nil
+			Expect(virtualmachine.IsPausedByAdmin(mgdObj)).To(BeFalse())
+		})
+
 		It("should return false when PauseVMExtraConfigKey is not set", func() {
 			Expect(virtualmachine.IsPausedByAdmin(mgdObj)).To(BeFalse())
 		})


zyiyi11 (approved):

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -13,11 +13,13 @@ import (
...
-			if o.Key == vmopv1.PauseVMExtraConfigKey {
-				if value, ok := o.Value.(string); ok {
-					return strings.ToUpper(value) == constants.ExtraConfigTrue
+	if c := moVM.Config; c != nil {
+		for i := range c.ExtraConfig {
+			if o := c.ExtraConfig[i].GetOptionValue(); o != nil {
+				if o.Key == vmopv1.PauseVMExtraConfigKey {
+					if value, ok := o.Value.(string); ok {
+						return strings.ToUpper(value) == constants.ExtraConfigTrue
+					}
```
bryanv, line 22: Why not have a `return false` here? We shouldn't have dup keys.

dilyar85: Good point! Didn't noticed it as I was just adding the nil check, updated.

bryanv:
I didn't just fix this when the bug came in because is this VM really paused or not? We don't know. I'm not sure if assuming not is the correct action here. And then based on that outcome BackupVirtualMachine(), ReconcileMinHardwareVersion(), and maybe elsewhere have similar usage that may need to be fixed/handled.

And this is very easy to miss we should have some top-level tests for moVM.Config = nil to try to catch regressions.

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -13,12 +13,17 @@ import (
 )
 
 func IsPausedByAdmin(moVM mo.VirtualMachine) bool {
-	for i := range moVM.Config.ExtraConfig {
-		if o := moVM.Config.ExtraConfig[i].GetOptionValue(); o != nil {
+	if moVM.Config == nil {
+		return false
+	}
+
+	for _, ec := range moVM.Config.ExtraConfig {
```
dougm, line 20: Another option, this loop could be replace with:
```go
ecList := object.OptionValueList(vmMO.Config.ExtraConfig)
return ecList.IsTrue(constants.ExtraConfigTrue)
```

aruneshpa:
> I didn't just fix this when the bug came in because is this VM really paused or not? We don't know. I'm not sure if assuming not is the correct action here. And then based on that outcome BackupVirtualMachine(), ReconcileMinHardwareVersion(), and maybe elsewhere have similar usage that may need to be fixed/handled.

Is the nil configInfo in this bug expected?  IOW, did we expect to receive a nil ConfigInfo, or was that due to other errors (e.g., not being able to fetch properties)?

If there was no error in fetching the properties of the VM, and the VM indeed has an empty ConfigInfo, then it is not paused by definition.  If this is due to some error, then we need to modify this method to return an error.

bryanv:
> Is the nil configInfo in this bug expected? IOW, did we expect to receive a nil ConfigInfo, or was that due to other errors (e.g., not being able to fetch properties)?

See https://github.com/vmware/govmomi/blob/3450f77a0db6e2b54bc632017ae95f8b2758ce27/object/virtual_machine.go#L428-L436. I don't much if any operations against the VM that we'd do in this path would succeed in this situation.

aruneshpa:
> > Is the nil configInfo in this bug expected? IOW, did we expect to receive a nil ConfigInfo, or was that due to other errors (e.g., not being able to fetch properties)?
> 
> See https://github.com/vmware/govmomi/blob/3450f77a0db6e2b54bc632017ae95f8b2758ce27/object/virtual_machine.go#L428-L436. I don't much if any operations against the VM that we'd do in this path would succeed in this situation.

Got it.  So yeah, it has to be an error in this case.  Besides, I don't thikn we can ever land with an empty/nil ConfigInfo in a non-error scenario.  [ConfigInfo](https://dp-downloads.broadcom.com/api-content/apis/API_VMA_001/8.0U2/html/vim.vm.ConfigInfo.html) contains the "name" of the VM -- which I would never expect to be empty unless we hit an error.

akutz:
This may be a non-issue with #673, as it refactors this code anyway.

PR #710, opened 2024-09-11
🐛 Don't fail if nameservers are empty in network CM

bryanv:

For some env this will be empty. Nameservers can be set via the VM Spec.Network

Fixes #

```release-note
No nameservers in the vmoperator-network-config ConfigMap does not prevent customization
```

### DIFF
--- a/pkg/providers/vsphere/config/config_test.go
+++ b/pkg/providers/vsphere/config/config_test.go
@@ -8,7 +8,9 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/config/config_test.go
```diff
@@ -96,6 +98,65 @@ func configTests() {
...
+		})
+
+		It("returns error if <worker_dns> is there", func() {
+			cm.Data = map[string]string{
+				config.NameserversKey: "<worker_dns>",
+			}
+			Expect(ctx.Client.Update(ctx, cm)).To(Succeed())
+
+			_, _, err := config.GetDNSInformationFromConfigMap(ctx, ctx.Client)
+			Expect(err).To(HaveOccurred())
```
mayankbh, line 144: nit: should this check the error returned?

mayankbh (approved):

sreyasn (approved):

PR #714, opened 2024-09-16
:sparkles: Define new unified storage quota management plugin webhook

abaruni:

Defines a new webhook server to act as the plug-in endpoint for the unified storage quota management admission webhook

Fixes #

```release-note
Support the unified storage quota extension service
```

### DIFF
--- a/webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
+++ b/webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
@@ -1,0 +1,347 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package validation
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+
+	v1 "k8s.io/api/admission/v1"
+	"k8s.io/api/admission/v1beta1"
+	storagev1 "k8s.io/api/storage/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
+	"github.com/vmware-tanzu/vm-operator/pkg/builder"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/record"
+)
+
+const (
+	webhookName            = "vmservice.cns.vsphere.vmware.com"
+	webhookPath            = "/getrequestedcapacityforvirtualmachine"
+	scParamStoragePolicyID = "storagePolicyID"
+)
+
+// RequestedCapacity represents response body returned by this webhook to the SPQ webhook.
+type RequestedCapacity struct {
+	// Capacity represents the size to be reserved for the given resource
+	Capacity resource.Quantity `json:"capacity"`
+
+	// StorageClassName represents the StorageClass associated with the given resource
+	StorageClassName string `json:"storageClassName"`
+
+	// StoragePolicyID represents the StoragePolicyId associated with the given resource
+	StoragePolicyID string `json:"storagePolicyId"`
+
+	// Reason indicates the cause for returning capacity as 0
+	Reason string `json:"reason,omitempty"`
+}
+
+type CapacityResponse struct {
+	RequestedCapacity
+	admission.Response
+}
+
+type RequestedCapacityHandler struct {
+	*pkgctx.WebhookContext
+	admission.Decoder
+
+	Client    client.Client
+	Converter runtime.UnstructuredConverter
+}
+
+// AddToManager adds the webhook to the provided manager.
+func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr ctrlmgr.Manager) error {
+	webhookNameLong := fmt.Sprintf("%s/%s/%s", ctx.Namespace, ctx.Name, webhookName)
+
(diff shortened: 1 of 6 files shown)


akutz (changes requested):
Some notes, but otherwise looks great. Thanks @abaruni !

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
```diff
@@ -0,0 +1,332 @@
...
+		// Subtract the PVC usage/limit from the total used value as PVCs
+		// are counted separately.
+		for _, volume := range vm.Status.Volumes {
+			if volume.Type == vmopv1.VirtualMachineStorageDiskTypeManaged {
+				if volume.Used != nil {
+					capacity.Sub(*volume.Used)
+				}
+			}
+		}
+	}
```
dilyar85, line 201: Would this ever be the case for VM creation requests here? IOW, isn't the status field always empty during resource creation and only populated by controllers afterward?

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
```diff
@@ -0,0 +1,332 @@
...
+	var capacity *resource.Quantity
+
+	if vm.Status.Storage == nil || !conditions.IsTrue(vm, vmopv1.VirtualMachineConditionCreated) {
+		if vm.Spec.Image == nil {
+			return CapacityResponse{Response: webhook.Errored(http.StatusBadRequest, errors.New("vm.Spec.Image is required"))}
+		}
+		vmiName := vm.Spec.Image.Name
+		var imageStatus vmopv1.VirtualMachineImageStatus
+
+		switch vm.Spec.Image.Kind {
```
dilyar85, line 161: Add a default case to return `http.StatusBadRequest` for other invalid image kinds?

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
```diff
@@ -0,0 +1,332 @@
...
+				}
+			}
+		}
+	}
+
+	return CapacityResponse{
+		RequestedCapacity: RequestedCapacity{
+			Capacity:         *capacity,
+			StorageClassName: scName,
+			StoragePolicyID:  sc.Parameters[scParamStoragePolicyID],
```
dilyar85, line 207: Add a check on `sc.Parameters` to avoid potential NPE (since it's an [optional field](https://github.com/kubernetes/api/blob/master/storage/v1/types.go#L46-L48))?

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
```diff
@@ -0,0 +1,332 @@
...
+				return CapacityResponse{Response: webhook.Errored(http.StatusNotFound, err)}
+			}
+			return CapacityResponse{Response: webhook.Errored(http.StatusInternalServerError, err)}
+		}
+
+		return CapacityResponse{
+			RequestedCapacity: RequestedCapacity{
+				Capacity:         *vm.Spec.Advanced.BootDiskCapacity,
+				StorageClassName: vm.Spec.StorageClass,
+				StoragePolicyID:  sc.Parameters[scParamStoragePolicyID],
```
dilyar85, line 244: Ditto.

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	oldVM := &vmopv1.VirtualMachine{}
+	if err := h.Converter.FromUnstructured(ctx.OldObj.UnstructuredContent(), oldVM); err != nil {
+		return CapacityResponse{Response: webhook.Errored(http.StatusBadRequest, err)}
+	}
+
+	if vm.Spec.Advanced != nil && vm.Spec.Advanced.BootDiskCapacity != nil &&
+		oldVM.Spec.Advanced != nil && oldVM.Spec.Advanced.BootDiskCapacity != nil &&
+		vm.Spec.Advanced.BootDiskCapacity.Cmp(*oldVM.Spec.Advanced.BootDiskCapacity) != 0 {
```
dilyar85, line 229: Should we also address the case when the capacity is added or removed in update requests (i.e. `vm.Spec.Advanced.BootDiskCapacity == nil` or `oldVM.Spec.Advanced.BootDiskCapacity == nil`)?

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
```diff
@@ -0,0 +1,332 @@
...
+		Obj:            obj,
+		OldObj:         oldObj,
+		UserInfo:       req.UserInfo,
+		Logger:         h.WebhookContext.Logger.WithName(obj.GetNamespace()).WithName(obj.GetName()),
+	}
+
+	return handleRequest(webhookRequestContext)
+}
+
+func (h *RequestedCapacityHandler) HandleCreate(ctx *pkgctx.WebhookRequestContext) CapacityResponse {
```
dilyar85, line 136: Can we add a brief comment here and the `HandleUpdate()` function for a better documentation?

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
abaruni: @akutz can you comment on this?

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
abaruni: `sc.Parameters` is a map, so this wouldn't cause a panic for nil pointer dereference, right? this would just return an empty string. so the question is what is the desired behavior in this situation? what would be the reason this is not populated, and is it better to error in this situation or to return as is and let the Storage Policy Quota webhook fail if this is indeed an error condition?

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
abaruni: see above

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
abaruni: reworked the logic here

webhooks/unifiedstoragequota/validation/unifiedstoragequota_validator.go
dilyar85: Ah, I got mixed up about reading and writing a nil map in Go, and only the latter would cause a NPE. I am fine with returning the empty value as is in that case. Thanks for adding a comment there!

dilyar85 (approved):
LGTM, thanks @abaruni for addressing all the comments!

akutz (approved):
Lgtm @abaruni , thank you for the updated tests!

PR #717, opened 2024-09-18
🐛 Only CPU/Memory resize when actually needed

bryanv:

Guard the CPU/Memory limited resize by the ResizeNeeded() check. Note that we're only covering CPU/Memory specific resizing here: the other UpdateFoo() were previously there, and this seems the best compromise when having to switch from full resize to hacking in the CPU/Memory limited resize in the old Session update path.

Reorg the resize specific vm_provider tests to better group related tests, and reduce some magic numbers.

Fixes #

```release-note
NONE
```


dilyar85 (approved):
LGTM!

PR #719, opened 2024-09-19
✨Support for reserved VM Classes

akutz:

This patch adds support for reserving N slots for VM Classes in a given namespace.
    
Please note that while there are multiple zones in a namespace, the current design does not allow a VM using a reserved VM Class to be scheduled in a zone with no reserved slots available.

Fixes `NA`

@bryanv, et. al, there is no special support in this PR for resize because it is already handled by virtue of `CreateConfigSpec` and then later the existing resize logic that handles `ExtraConfig`.

@mikekim, can you also please take a look at this PR?

```release-note
Support for reserved VM Classes
```


pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -51,6 +51,16 @@ func CreateConfigSpec(
 		},
 	)
 
+	// Ensure ExtraConfig contains the VM Class's reservation profile ID if set.
+	if id := vmClassSpec.ReservedProfileID; id != "" {
```
sreyasn, line 55: How will we be using ReservedSlots in the VM configSpec?

sreyasn (approved):
Change looks good to me otherwise!.

pkg/providers/vsphere/virtualmachine/configspec.go
akutz: We are not. It's there for future use so I can report on how many free slots there are in the VM Class status. I just have not done that yet. But since Michal is going to push down the number of slots with the changes he is making, I wanted to give him a place to store it.

bryanv (approved):

PR #724, opened 2024-09-25
:seedling: Do not fall back to using filenames for PVC disk backup

aruneshpa:

When recording PVC disk data for the backup, we added code that first checks disks by their filenames first before falling back to using UUIDs.  This was done to support partial (disk only restore).

Since partial restore no longer relies on the PVC disk data, we no longer need to do this.  The current code also has a side effect of not handling disks restored with a new PVC name.

This effectively reverts the commit: c3a0c36.

```release-note
Do not fall back to using filenames for PVC disk backup
```


dougm (approved):
lgtm

dilyar85 (approved):

PR #728, opened 2024-10-02
:seedling: Add ValidatingWebhookConfiguration for storage quota webhook

abaruni:

This PR adds a ValidatingWebhookConfiguration to allow the storage quota webhook to be called upon create/update of virtual machines

Fixes #

```release-note
NONE
```

### DIFF
--- a/config/webhook/kustomization.yaml
+++ b/config/webhook/kustomization.yaml
@@ -4,6 +4,7 @@ kind: Kustomization
 resources:
 - service.yaml
 - manifests.yaml
+- storage_quota_webhook_configuration.yaml
 
 patchesStrategicMerge:
 - manifests_label_patch.yaml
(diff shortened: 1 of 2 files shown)


config/webhook/kustomization.yaml
```diff
@@ -4,6 +4,7 @@ kind: Kustomization
 resources:
 - service.yaml
 - manifests.yaml
+- storage_quota_webhook_configuration.yaml
```
dilyar85, line 7: Is there any reason for not generating this in the `manifests.yaml` file using a kubebuilder tag as other existing webhook configs?

abaruni: we don't want to generate with a kubebuilder tag, because this is for the storage quota webhook, not for our webhook. we want the storage quota webhook to be called for create/update of vm, not any of our webhooks. the storage quota webhook will then call our "webhook" to get the capacity for storage policy usage. so, it's not included in `manifests.yaml` because we don't want it generated, and if it's hard-coded there, then it'll get removed upon generation

dilyar85: Got it. In that case it makes sense to keep it separate file to have its CA annotation and other fields correctly.

dilyar85 (approved):
LGTM!

sreyasn:
Looks good to me!. are there more changes required for this PR?.

sreyasn (approved):

PR #730, opened 2024-10-02
🐛 Use DeepCopy() on down convert for vApp Templating

sreyasn:

This change fixes the vApp templating logic to use a copy of the VMCtx.VM for down-convertion. This removes potential issues caused by conversion (ie) modifying the actual VM on next reconcile. 

Fixes # N/A

This fixes the bug that was specific to using a v1a1 VM with transport: OvfEnv and a ConfigMap. 
- dropping the respective v1a3 annotations 
- appending the v1a1 conversion annotation onto v1a3.

```
Fixes bug with dropping v1a3 annotations when v1a1 VM with OvfEnv and a ConfigMap is used.
```


pkg/providers/vsphere/vmlifecycle/bootstrap_templatedata.go
```diff
@@ -45,12 +45,14 @@ func GetTemplateRenderFunc(
 		Nameservers: bsArgs.DNSServers,
 	}
 
-	// Oh dear. The VM itself really should not have been included here.
+	// Use deepCopy of the VM to prevent issues due to down-converting the actual VM on
+	// next reconcile
+	copyVM := vmCtx.VM.DeepCopy()
```
bryanv, line 50: Do we want to give each `ConvertFrom` their own copy instead? Or at least restore the annotations after the first until we fix the aliasing in the conversion code

sreyasn: Using separate copies. Let me know if you have a better suggestion for naming these copies. updated

bryanv: Oh I guess I wouldn't even give them names: `ConvertFrom(vmCtx.VM.DeepCopy())`? . And leave your comment as to why we're doing that.

bryanv (approved):

PR #733, opened 2024-10-04
✨ Reconcile VMs on async signal

akutz:

This patch adds support for reconciling VMs when their state has changed on the underlying platform, ex. vSphere. There are three primary components:

* the watcher -- emits results on a channel monitored by...
* the service -- receives results and enqueues reconcile requests
* the zone controller -- adds/removes zone folders to/from the watcher

Please note, the environment variable `ASYNC_SIGNAL_DISABLED` may be set to a truth-y string value, ex. "true", to completely disable the async signal logic, regardless of the feature state switch.

### Watcher

The watcher is located in `pkg/util/vsphere/watcher` and watches one or more vSphere entities that can contain VMs, ex. a `Folder`, `ClusterComputeResource`, `HostSystem`, etc.. The watcher is initialized with a set of these entities and creates a `ContainerView` for each. These are added to the watcher's `ListView`, which enables entities to be added/removed later while the watcher is running. The watcher is signaled when a VM enters the view of the watcher or when a VM has a change to one of the on the following properties:

- `config.extraConfig` -- Signal when the guest changes something in `guestinfo`.
- `guest.ipStack` -- Signal on changes to the network state.
- `guest.net` -- Signal on changes to the network state.
- `summary.config.name` -- A way to detect when the VM enters the scope of the watch.
- `summary.guest` -- Signal on any changes to the guest.
- `summary.overallStatus` -- Signal on changes to the VM's status.
- `summary.runtime.host` -- Signal when the host on which a VM is running has been changed.
- `summary.runtime.powerState` -- Signal when the VM's power state is changed.
- `summary.storage.timestamp` -- Signal when the VM's storage information (capacity, used, etc.) has changed.

Not all changes to extraConfig cause the watcher to emit a result. The following extraConfig keys are ignored:

- backupapi.AdditionalResourcesYAMLExtraConfigKey
- backupapi.BackupVersionExtraConfigKey
- backupapi.ClassicDiskDataExtraConfigKey
- backupapi.DisableAutoRegistrationExtraConfigKey
- backupapi.EnableAutoRegistrationExtraConfigKey
- backupapi.PVCDiskDataExtraConfigKey
- backupapi.VMResourceYAMLExtraConfigKey
- "govcsim"
- "guestinfo.metadata"
- "guestinfo.metadata.encoding"
- "guestinfo.userdata"
- "guestinfo.userdata.encoding"
- "guestinfo.vendordata"
- "guestinfo.vendordata.encoding"
- "vmservice.namespacedName"

Additional keys may be ignored as well, but these are ignored by default in order to prevent an infinite loop:

- VM controller receives reconcile request -->
- VM controller reconfigures VM -->
- VM Op reconfigures VM --> 
- watcher sees change -->
- watcher emits result -->
- service sees result -->
- service enqueues reconcile request -->
- VM controller receives reconcile request, causing a loop.

When the watcher notices a VM enter its view or with a change, the watcher must get the namespace and name for the VM. This happens one of three ways:

1. The change itself includes the namespace and name in the extraConfig
2. The watcher looks up the namespace and name from the manager's cache, where the field `status.uniqueID` is now an indexed field.
3. Finally, the watcher retrieves the property `config.extraConfig["vmservice.namespacedName"]` from the vSphere server.

If the namespace and name can be determined, the watcher checks if the VM already exists in Kubernetes with a `status.uniqueID` field and if the update type was the VM entering the view of the watcher. If these conditions are met, no result is emitted for this VM. This prevents double-reconciling VMs when the Controller-Manager starts up for the first time. During start-up, the Controller-Manager automatically reconciles all objects watched by controllers. Since all VMs would also be entering the view of watchers, this would cause a large-scale double-reconcile. Therefore, this logic skips emitting results on startup for VMs that are already deployed.

If the namespace and name are non-empty, the update types was `Enter` and the VM has an empty `status.uniqueID` field or the update type was `Modify`, the watcher emits a result on a channel watched by the next component, the service.

### Service

The service is located in `services/vm-watcher` and is responsible for:

* keeping an instance of the watcher running
* receiving results from the watcher and enqueuing reconcile requests

The service will always start a new instance of the watcher as long as the reason the previous instance failed was due to a login/auth error. This is to handle the case of credential rotation.

The service starts a watcher with an initial set of entities to watch that includes the ManagedObject ID for each `Folder` that can contain VM Service VMs. These folder IDs are gathered by listing all `Zone` resources on the cluster and collecting the value of `spec.managedVMs.folderMoID`.

The service monitors results from the watcher. Upon receiving a result, the service determines if the reported VM is valid, and if so, a reconcile request is enqueued.

### Zone controller

The zone controller is located in `controllers/infra/zone` and reconciles `topologyv1.Zone` resources.

When a zone resource without a deletion timestamp is reconciled, the controller adds a finalizer to it and adds the zone's vm service folder to the list of the entities monitored by the watcher.

When a zone resource with a non-zero deletion timestamp is reconciled, the controller removes the zone's vm service folder from the list of the entities monitored by the watcher and removes the finalizer.

Fixes `NA`

* We may want to remove [`summary.storage.timestamp`](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.vm.Summary.StorageSummary.html) as this may result in too many reconciles. I will look into how often the storage summary is updated.
* @bryanv Please note, the watcher prevents double-reconciling VMs when the Controller-Manager starts. The documentation for the watcher component up above describes how.

```release-note
Reconcile VMs when their state changes on underlying platform
```

### DIFF
--- a/controllers/infra/zone/zone_controller.go
+++ b/controllers/infra/zone/zone_controller.go
@@ -1,0 +1,162 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package zone
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/go-logr/logr"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	topologyv1 "github.com/vmware-tanzu/vm-operator/external/tanzu-topology/api/v1alpha1"
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/patch"
+	"github.com/vmware-tanzu/vm-operator/pkg/record"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/vsphere/watcher"
+)
+
+// SkipNameValidation is used for testing to allow multiple controllers with the
+// same name since Controller-Runtime has a global singleton registry to
+// prevent controllers with the same name, even if attached to different
+// managers.
+var SkipNameValidation *bool
+
+// AddToManager adds this package's controller to the provided manager.
+func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) error {
+	var (
+		controlledType     = &topologyv1.Zone{}
+		controlledTypeName = reflect.TypeOf(controlledType).Elem().Name()
+
+		controllerNameShort = fmt.Sprintf("%s-controller", strings.ToLower(controlledTypeName))
+		controllerNameLong  = fmt.Sprintf("%s/%s/%s", ctx.Namespace, ctx.Name, controllerNameShort)
+	)
+
+	r := NewReconciler(
+		ctx,
+		mgr.GetClient(),
+		ctrl.Log.WithName("controllers").WithName(controlledTypeName),
+		record.New(mgr.GetEventRecorderFor(controllerNameLong)),
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(controlledType).
+		WithOptions(controller.Options{
+			SkipNameValidation: SkipNameValidation,
+		}).
+		Complete(r)
+}
+
+func NewReconciler(
+	ctx context.Context,
+	client client.Client,
+	logger logr.Logger,
+	recorder record.Recorder) *Reconciler {
+
+	return &Reconciler{
+		Context:  ctx,
+		Client:   client,
+		Logger:   logger,
+		Recorder: recorder,
+	}
+}
+
+// Finalizer is the finalizer placed on Zone objects by VM Operator.
+const Finalizer = "vmoperator.vmware.com/finalizer"
+
+// Reconciler reconciles a StoragePolicyQuota object.
+type Reconciler struct {
+	client.Client
+	Context  context.Context
+	Logger   logr.Logger
+	Recorder record.Recorder
+}
+
+// +kubebuilder:rbac:groups=topology.tanzu.vmware.com,resources=zones,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=topology.tanzu.vmware.com,resources=zones/status,verbs=get
+
+func (r *Reconciler) Reconcile(
+	ctx context.Context,
+	req ctrl.Request) (_ ctrl.Result, reterr error) {
+
+	ctx = pkgcfg.JoinContext(ctx, r.Context)
+	ctx = watcher.JoinContext(ctx, r.Context)
+
+	var obj topologyv1.Zone
+	if err := r.Get(ctx, req.NamespacedName, &obj); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	patchHelper, err := patch.NewHelper(&obj, r.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	defer func() {
+		if err := patchHelper.Patch(ctx, &obj); err != nil {
+			if reterr == nil {
+				reterr = err
+			} else {
+				reterr = fmt.Errorf("%w,%w", err, reterr)
+			}
+		}
+	}()
+
+	if !obj.DeletionTimestamp.IsZero() {
+		return r.ReconcileDelete(ctx, &obj)
+	}
+
+	return r.ReconcileNormal(ctx, &obj)
+}
+
+func (r *Reconciler) ReconcileDelete(
+	ctx context.Context,
+	obj *topologyv1.Zone) (ctrl.Result, error) {
+
+	if controllerutil.ContainsFinalizer(obj, Finalizer) {
+		if val := obj.Spec.ManagedVMs.FolderMoID; val != "" {
+			if err := watcher.Remove(
+				ctx,
+				vimtypes.ManagedObjectReference{
+					Type:  "Folder",
+					Value: val,
+				},
+				fmt.Sprintf("%s/%s", obj.Namespace, obj.Name)); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+		controllerutil.RemoveFinalizer(obj, Finalizer)
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *Reconciler) ReconcileNormal(
+	ctx context.Context,
+	obj *topologyv1.Zone) (ctrl.Result, error) {
+
+	if !controllerutil.ContainsFinalizer(obj, Finalizer) {
+		// The finalizer is not added until we are able to successfully
+		// add a watch to the zone's VM Service folder.
+		if val := obj.Spec.ManagedVMs.FolderMoID; val != "" {
+			if err := watcher.Add(
+				ctx,
+				vimtypes.ManagedObjectReference{
+					Type:  "Folder",
+					Value: val,
+				},
+				fmt.Sprintf("%s/%s", obj.Namespace, obj.Name)); err != nil {
+				return ctrl.Result{}, err
+			}
+			controllerutil.AddFinalizer(obj, Finalizer)
+		}
+	}
+	return ctrl.Result{}, nil
+}
--- a/controllers/infra/zone/zone_controller_test.go
+++ b/controllers/infra/zone/zone_controller_test.go
@@ -1,0 +1,306 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package zone_test
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/vmware/govmomi/object"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha3"
+	"github.com/vmware-tanzu/vm-operator/controllers/infra/zone"
+	topologyv1 "github.com/vmware-tanzu/vm-operator/external/tanzu-topology/api/v1alpha1"
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	"github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	providerfake "github.com/vmware-tanzu/vm-operator/pkg/providers/fake"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/kube/cource"
+	vsclient "github.com/vmware-tanzu/vm-operator/pkg/util/vsphere/client"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/vsphere/watcher"
+	vmwatcher "github.com/vmware-tanzu/vm-operator/services/vm-watcher"
+	"github.com/vmware-tanzu/vm-operator/test/builder"
+	"github.com/vmware-tanzu/vm-operator/test/testutil"
+)
+
+var _ = Describe(
+	"Reconcile",
+	Label(
+		testlabels.Controller,
+		testlabels.EnvTest,
+		testlabels.V1Alpha3,
+	),
+	func() {
+
+		var (
+			ctx       context.Context
+			vcSimCtx  *builder.IntegrationTestContextForVCSim
+			provider  *providerfake.VMProvider
+			initEnvFn builder.InitVCSimEnvFn
+		)
+
+		BeforeEach(func() {
+			ctx = context.Background()
+			ctx = logr.NewContext(ctx, testutil.GinkgoLogr(4))
+		})
+
+		JustBeforeEach(func() {
+			ctx = pkgcfg.WithContext(ctx, pkgcfg.Default())
+			ctx = pkgcfg.UpdateContext(
+				ctx,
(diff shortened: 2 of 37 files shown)


akutz:
@dougm I am going to add a flake allowance to the tests that depend on vC Sim. See https://github.com/vmware-tanzu/vm-operator/actions/runs/11206817769/job/31148110424?pr=733 -- there's a race that pops up on occasion.

dougm:
> @dougm I am going to add a flake allowance to the tests that depend on vC Sim. See https://github.com/vmware-tanzu/vm-operator/actions/runs/11206817769/job/31148110424?pr=733#step:5:783 -- there's a race that pops up on occasion.

I've not reproduced myself yet, but looks like this should fix: https://github.com/vmware/govmomi/pull/3584

akutz:
> > @dougm I am going to add a flake allowance to the tests that depend on vC Sim. See https://github.com/vmware-tanzu/vm-operator/actions/runs/11206817769/job/31148110424?pr=733#step:5:783 -- there's a race that pops up on occasion.
> 
> I've not reproduced myself yet, but looks like this should fix: [vmware/govmomi#3584](https://github.com/vmware/govmomi/pull/3584)

Thanks @dougm , it did in fact fix it. I already rebased this PR after pulling your patch. Thanks again!

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,672 @@
...
+}
+
+func (w *Watcher) add(ctx context.Context, ref moRef) error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.pc == nil {
+		return nil
+	}
+
```
sreyasn, line 482: Nit: Probably add a comment here 
// return early if containerView is already present

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,672 @@
...
+}
+
+func (w *Watcher) remove(_ context.Context, ref moRef) error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.pc == nil {
+		return nil
+	}
+
```
sreyasn, line 521: same Nit to add comment here for already removed (or) not present.

sreyasn (commented):
Did a first pass. Will look again later today. 

Also, Thank you for this change. learned quite a bit :)

pkg/util/vsphere/watcher/watcher.go
akutz: It returns early if the watcher is closed, but sure, will add comment.

sreyasn: was referring to L483 : 
`if _, ok := w.cv[ref]; ok {`

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,676 @@
...
+				ignore = ignoredOptVals == len(aov.OptionValue)
+			}
+		}
+		if ignore {
+			ignoredChanges++
+		}
+	}
+	if ignoredChanges == len(update.changes) {
+		return nil
+	}
```
dilyar85, line 395: Instead of counting all the ignored EC keys, we could also look for a non-ignored EC key to break the loop earlier if it's found and namespace & name are set. Something like:

```suggestion
func (w *Watcher) onObject(
	...

	// This update will be skipped if after removing all of the changes for
	// the ignoredExtraConfigKeys there is nothing left.
	var ignoredChanges int
	for i := range update.changes {
		c := update.changes[i]
		ignore := false

		if c.Name == extraConfigPropPath {
			if aov, ok := c.Val.(vimtypes.ArrayOfOptionValue); ok {
				ignore, namespace, name = checkEC(aov, w.ignoredExtraConfigKeys)
			}
		}
		if ignore {
			ignoredChanges++
		}
	}
	if ignoredChanges == len(update.changes) {
		return nil
	}
	
	...
}

func checkEC(
	aov vimtypes.ArrayOfOptionValue,
	ignoredKeys map[string]struct{}) (ignore bool, namespace, name string) {

	hasNonIgnoredECKey := false

	for j := range aov.OptionValue {
		if ov := aov.OptionValue[j].GetOptionValue(); ov != nil {

			// Get the namespace and name of the VM from the changes
			// if they are present there.
			if ov.Key == extraConfigNamespacedNameKey {
				if namespace == "" || name == "" {
					if s, ok := ov.Value.(string); ok {
						namespace, name = namespacedNameFromString(s)
					}
				}
			}

			// Note if the key cannot be ignored.
			if _, ok := ignoredKeys[ov.Key]; !ok {
				hasNonIgnoredECKey = true
			}

			if hasNonIgnoredECKey && (namespace != "" && name != "") {
				break
			}
		}
	}

	return !hasNonIgnoredECKey, namespace, name
}
```

dilyar85 (commented):
Thanks @akutz for taking VMOP to the next level!
Left some questions and comments (also learned a lot while reviewing this change).

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,672 @@
...
+		"guest.ipStack",
+		"guest.net",
+		"summary.config.name",
+		"summary.guest",
+		"summary.overallStatus",
+		"summary.runtime.host",
+		"summary.runtime.powerState",
+		"summary.storage.timestamp",
+	}
+}
```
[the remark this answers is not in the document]
akutz: So it cannot be modified.

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,673 @@
...
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
+	w.cancel = cancel
+
+	go func(chanRef chan moRef) {
+		for {
+			select {
+			case ref := <-chanRef:
+				if err := w.add(ctx, ref); err != nil {
+					logger.Error(err, "failed to add ref to watcher")
```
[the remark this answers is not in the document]
akutz: This code is no longer relevant.

controllers/infra/zone/zone_controller.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (r *Reconciler) ReconcileNormal(
+	ctx context.Context,
+	logger logr.Logger,
+	obj *topologyv1.Zone) error {
+
+	if !controllerutil.ContainsFinalizer(obj, Finalizer) {
+		if val := obj.Spec.ManagedVMs.FolderMoID; val != "" {
+			watcher.Add(
```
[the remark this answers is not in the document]
akutz: No, it won't, because the old implementation has a goroutine inside of it. However, this code is changed and now returns immediately sans a goroutine.

controllers/infra/zone/zone_controller_test.go
```diff
@@ -0,0 +1,307 @@
...
+			)
+
+			JustBeforeEach(func() {
+				nsInfo = vcSimCtx.CreateWorkloadNamespace()
+
+				By("ensure all zones have finalizers", func() {
+					Eventually(func(g Gomega) {
+						var obj topologyv1.ZoneList
+						g.Expect(vcSimCtx.Client.List(ctx, &obj, ctrlclient.InNamespace(nsInfo.Namespace))).To(Succeed())
+						g.Expect(obj.Items).ToNot(BeEmpty())
```
bryanv, line 113: Also assert the number of expected zones - `g.Expect(obj.Items).To(HaveLen(vcSimCtx.))`

controllers/infra/zone/zone_controller_test.go
```diff
@@ -0,0 +1,307 @@
...
+						for i := range obj.Items {
+							g.Expect(obj.Items[i].Finalizers).To(HaveLen(1))
```
bryanv, line 115: Can use `ConsistOf()` to combine these assertions  - https://onsi.github.io/gomega/#consistofelement-interface

main.go
```diff
@@ -202,9 +212,12 @@ func main() {
...
 
 	ctx := pkgcfg.WithConfig(defaultConfig)
-	if defaultConfig.Features.UnifiedStorageQuota {
+	if defaultConfig.Features.UnifiedStorageQuota || defaultConfig.Features.WorkloadDomainIsolation {
 		ctx = cource.WithContext(ctx)
 	}
+	if defaultConfig.Features.WorkloadDomainIsolation {
+		ctx = watcher.WithContext(ctx)
+	}
 	initFeaturesFromCapabilities(ctx, setupLog)
```
bryanv, line 221: I think we'd want to update the features w/ the capabilities before decided on what is enabled.

That said, it might just be easier to always init the `cource` and `watcher` in the context here so it is always there, and for a world where these things are flipped at runtime. It should be pretty low risk.

services/vm-watcher/vm_watcher_service.go
```diff
@@ -0,0 +1,220 @@
...
+			if !result.Verified {
+				logger.V(4).Info(
+					"received result but unable to validate vm",
+					"result", result)
+				continue
+			}
+
+			// Enqueue a reconcile request for the VM.
+			logger.V(4).Info("received result", "result", result)
+			go func(result watcher.Result) {
```
bryanv, line 184: Do we lose any useful backpressure on by doing this in a go routine? I guess I'm a little worried we could have these stackup.

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,674 @@
...
+		close(w.chanDone)
+
+		_ = w.pf.Destroy(context.Background())
+		_ = w.pc.Destroy(context.Background())
+		_ = w.lv.Destroy(context.Background())
+		for _, cv := range w.cv {
+			_ = cv.Destroy(context.Background())
+		}
+
+		w.pf = nil
```
dougm, line 211: Using `once.Do` here, a Watcher can't be used again after calling `Close()`, right? Just curious about setting these fields to nil in that case, is it for a gc hint or something else?
Also doesn't look Close would be called more than once, is `sync.Once` needed here?

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,674 @@
...
+		cvr:                    cvr,
+		ignoredExtraConfigKeys: toSet(ignoredExtraConfigKeys),
+		lookupNamespacedName:   lookupNamespacedName,
+	}, nil
+}
+
+func (w *Watcher) Close() {
+	w.once.Do(func() {
+		w.mu.Lock()
+		defer w.mu.Unlock()
```
dougm, line 199: May not matter in this case so feel free to ignore, but in general don't want to hold a mutex lock over network calls (Destroys below).

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,674 @@
...
+		// result when the object is entering the scope of the watcher and the
+		// corresponding Kubernetes object already exists with a matching
+		// status.uniqueID field.
+		return nil
+	}
+
+	if namespace == "" || name == "" {
+		rep, err := property.DefaultCollector(w.client).RetrieveProperties(
+			ctx,
+			vimtypes.RetrieveProperties{
```
dougm, line 383: You can use `RetrieveOne` (or `Retrieve`) with a `dst` of `*[]types.ObjectContent`. That avoids reflection and having to construct the PropertyFilterSpec:
```go
		var content []types.ObjectContent
		err := property.DefaultCollector(w.client).RetrieveOne(
			ctx,
			obj,
			[]string{extraConfigNamespaceNamePropPath},
			&content,
		)
		if err != nil {
			return err
		}
		namespace, name = namespacedNameFromObjContent(content)
```

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	// Only remove the container from the list view if it has a ref count of
+	// one.
+	if len(w.cvr[ref]) > 1 {
+		delete(w.cvr[ref], id)
+		return nil
+	}
+
+	_, err := w.lv.Remove(context.Background(), []moRef{cv.Reference()})
```
dougm, line 524: Why not use the Context param passed to the function instead of `context.Background()`?

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,674 @@
...
+		}
+		cvMap[containerRefs[i]] = cv
+		cvRefsMap[containerRefs[i]] = map[string]struct{}{}
+	}
+
+	if resultErr != nil {
+		// There was an error creating container views, so make sure to clean up
+		// any views that *were* created before returning.
+		for _, cv := range cvMap {
+			if err := cv.Destroy(context.Background()); err != nil {
```
dougm, line 576: Use `ctx` here?

services/vm-watcher/vm_watcher_service.go
akutz: Stacking is fine. It's the same as a work queue.

pkg/util/vsphere/watcher/watcher.go
akutz: Since `Close` is exported, I used the `sync.Once`. I made `Close` public in case I want to close it directly.

pkg/util/vsphere/watcher/watcher.go
akutz: Because the reason the earlier one failed could be due to the context being cancelled.

pkg/util/vsphere/watcher/watcher.go
akutz: See https://github.com/vmware-tanzu/vm-operator/pull/733#discussion_r1793981226

pkg/util/vsphere/watcher/watcher.go
akutz: The reason the param has a context is in case I use a logger.

dougm (approved):
Awesome, lgtm

bryanv (approved):
🥳

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -0,0 +1,656 @@
...
+	w.errMu.Lock()
+	w.err = err
+	w.errMu.Unlock()
+}
+
+func newWatcher(
+	ctx context.Context,
+	client *vim25.Client,
+	watchedPropertyPaths []string,
+	additionalIgnoredExtraConfigKeys []string,
```
sidharthsurana, line 138: I see that we have this parameter to augment the list of ignored extraconfig keys. However, when we instantiate this watcher we are passing nil to it.

Suggestion: Does it make sense to define a additional source (like a predefined named configmap or secret or something else as well) of this list that can be set in a customer env if need be that could on the fly augment this? Case I am thinking about is say the user has some third party tool or their own orchestration that uses extraConfig keys as well, in such cases they might want to skip those  as well. (although technically that may very well lead to a no-op reconcile, but just waste a reconcile cycle that could have been avoided). 
I am almost certain you would have considered this case already :) and if so, was there a specific reasoning of not to do that? 

Also, I just literally looked at this MR so if you have covered this somewhere else, please feel free to point me to that.

akutz: Allowing an external source of keys to ignore makes a ton of sense. Could you please file an issue for this idea? Thanks!

PR #737, opened 2024-10-07
Honor PVC requested zonal constraints for VM placement

bryanv:

For volumes that are specified at VM create time, honor the PVC's requested topology, since the VM must be created in the zone(s) that the PVCs allow.

Later work will handle already bound PVCs, and eventually we'll create dummy disks for the PVCs in the placement ConfigSpec.

Fixes #

```release-note
For VM Spec.Volumes that are specified at create time, the PVC's "csi.vsphere.volume-requested-topology" annotation, if specified, will be used to limit possible placement candidates.
```


zyiyi11 (approved):
LGTM, just minor naming suggestions.

PR #739, opened 2024-10-10
🌱 Remove lock on vm watcher

akutz:

This patch removes the lock from the vm watcher, making it more performant and safer since the lock is no longer held when closing the watcher and/or adding/removing references, during which network requests can be made. There is no need for a lock anymore since the vim objects are not set to nil when the watcher is closed. They will still be garbage collected when the watcher itself goes out of scope.

This patch also explicitly defines the `ASYNC_SIGNAL_DISABLED` env var in the deployment YAML to make it clear how to disable this feature.

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/config/local/vmoperator/local_env_var_patch.yaml
+++ b/config/local/vmoperator/local_env_var_patch.yaml
@@ -9,6 +9,8 @@ spec:
       containers:
       - name: manager
         env:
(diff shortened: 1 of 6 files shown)


dougm (approved):
lgtm

PR #743, opened 2024-10-14
🌱 Handle unbound PVCs Immediate SC during placement

bryanv:

We must wait for PVCs with an immediate StorageClass to be bound, so we'll know the accessible zones.

This is first pass intended to address the lowest hanging PVC placement work, with a lot of reworking of placement to be done later. Specifically, we should be smarter about whether we'll actually do placement or not, based on the cluster and VM config.

Fixes #

There is still a lot of work remaining for better volume aware placement, such as including the unbound WFFC PVC in the placement ConfigSpec. 

```release-note
For bound PVCs, their accessible zones are taking into consideration for placement candidates.
```


pkg/util/kube/storage.go
```diff
@@ -44,6 +45,15 @@ func getPVCRequestedZones(pvc corev1.PersistentVolumeClaim) sets.Set[string] {
...
 
+func getPVCRequestedZones(pvc corev1.PersistentVolumeClaim) sets.Set[string] {
+	return getPVCZones(pvc, csiRequestedTopologyAnnotation)
+}
+
+func getPVCAccessibleZones(pvc corev1.PersistentVolumeClaim) sets.Set[string] {
+	return getPVCZones(pvc, csiAccessibleTopologyAnnotation)
+}
+
+/*
```
sreyasn, line 56: did you wanna keep this around?. reminder to remove otherwise.

sreyasn (approved):

PR #747, opened 2024-10-15
📖 Clarify storage class requirement for crypto.encryptionClassName API

dilyar85:

This PR updates the API doc for the `crypto.encryptionClassName` field to reflect the validation logic in the VM webhook, which requires an encrypted storage class when this field is set. It also adds a comment in the webhook code explaining why the error is returned on `encryptionClassName` even though the validation checks the storage class.

Fixes N/A.

We should probably return both fields in the validating webhook result once K8s supports combining field errors.

```release-note
Clarify storage class requirement for crypto.encryptionClassName API
```

----
📚 Documentation preview 📚: https://vm-operator--747.org.readthedocs.build/en/747/


akutz (approved):
Lgtm @dilyar85 , thank you!

PR #753, opened 2024-10-17
🌱 GitHub action to verify WCP manifests

akutz:

This patch adds support for a GitHub action that verifies the WCP manifests by deploying a Kind cluster and ensuring the WCP manifests can be applied to the cluster successfully.

Fixes `NA`

Verified on my fork at https://github.com/akutz/vm-operator/actions/runs/11386366887/job/31678406635.

```release-note
Add GitHub action to verify the WCP manifests
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,30 @@ jobs:
     - name: Verify codegen
       run: make verify-codegen
 
+  verify-manifests:
+    needs:
+    - verify-go-modules
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v4
+    - name: Install Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: ${{ env.GO_VERSION }}
+        cache: true
+        cache-dependency-path: '**/go.sum'
+    - name: Create kind cluster
+      id: create_kind_cluster
+      run: make kind-up
+      env:
+        KIND_IMAGE: kindest/node:v1.31.1
+    - name: Verify the manifests
+      run: make verify-wcp-manifests
+    - name: Destroy kind cluster
+      run: make kind-down
+      if: steps.create_kind_cluster.outcome == 'success'
+
   lint-go:
     needs:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -90,6 +90,28 @@ jobs:
...
+        cache: true
+        cache-dependency-path: '**/go.sum'
+    - name: Create kind cluster
+      run: make kind-up
+      env:
+        KIND_IMAGE: kindest/node:v1.31.1
+    - name: Verify the WCP manifests
+      run: make verify-wcp-manifests
+    - name: Destroy kind cluster
+      run: make kind-down
```
dilyar85, line 113: Should we run this step regardless of whether the previous `Verify the WCP manifests` step fails, to ensure the kind cluster is destroyed? Or even better, we could run it as long as the `Create kind cluster` step succeeds, something like `if: ${{ steps.create_kind_cluster.outcome == 'success' }}`.

Makefile
```diff
@@ -617,9 +649,19 @@ undeploy-local:  ## Un-Deploy controller in the configured Kubernetes cluster in
 ## --------------------------------------
 
 .PHONY: deploy-local-vcsim
-deploy-local-vcsim: prereqs kustomize-local-vcsim  ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy-local-vcsim: prereqs kustomize-local-vcsim | $(KUBECTL)
+deploy-local-vcsim: ## Deploy local-vcsim YAML into kind cluster
 	KUBECONFIG=$(KUBECONFIG) hack/deploy-local.sh $(LOCAL_YAML)
 
+## --------------------------------------
+## Development - wco
```
dilyar85, line 657: Typo `s/wco/wcp`

dilyar85 (approved):
Some minor comments, overall LGTM!

.github/workflows/ci.yml
akutz: Technically the step is not required at all since the agent goes away right after, but I'll add your suggestion so things are explicit.

PR #756, opened 2024-10-17
🌱 Fixes for Bring Your Own Key (BYOK) provider

akutz:
 
This patch adds several fixes/updates for the BYOK feature:
    
* Support for deploying unencrypted VMs with default key provider

    This patch adds support for deploying unencrypted VMs even when there is a default key provider present on the system. This is a change from the previous behavior that would disallow deploying a VM on a non-encryption storage class or sans vTPM if there was a default key provider present. This would in effect enforce the need for encryption and its requirements. This patch treats new VMs as special and allows it to skip encryption IFF:
    
    - the discovered provider is the default
    - there is no vTPM
    - the storage class does not support encryption
    
* Fixes a bug where a VM encrypted with the default key provider would get rekeyed over and over and over again.
* Fixes a bug where the wrong ConfigSpec was being updated when creating a VM.

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -1097,7 +1097,7 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpec(
 				vs.vcClient.VimClient(),
 				vmCtx.VM,
 				vmCtx.MoVM,
-				&configSpec); err != nil {
+				&createArgs.ConfigSpec); err != nil {
 
 				return err
 			}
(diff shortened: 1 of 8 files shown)


dougm (approved):
lgtm

dougm (approved):
lgtm, thanks @akutz !
I also deployed your branch and validated:
- Re-key/Reconfigure loop bug is fixed
- Can create a vm w/ EncryptionClass that specifies `keyID`
- Passes the in-review gce2e tests (which includes enabling a test for the above)

PR #757, opened 2024-10-18
✨ Distinct webhook deployments

akutz:

This path introduces distinct webhook deployments for the admission and conversion webhooks. This is to provide a more stable and performant experience. The webhook pods will not require nearly as much memory as the controller-manager pods. Additionally, if the webhooks ever go down, it is a bad time. Admission webhooks being unavailable means no one can apply any changes to the VM Op APIs. The conversion webhooks being offline means no one can even *read* the VM Op APIs.

This patch also removes the vestigial use of kustomizeconfig files since they are no longer needed, having been replaced by, well, the replacements feature.

Fixes `NA`

We need to:

- [ ] Consider the impact to upgrades.
- [ ] Validate and reserve the new ports on Supervisor.

```release-note
Move the admission and conversion webhooks into their own pods
```


JoyceXiaojun:
Overall looks good to me from upgrade perspective. Shouldn't be any port conflict since we don't explicitly delete any deployment and re-deploy with the same port. We should test upgrade path from last release to main with this change for verification.

One minor thing: should add two new deployment names here when validating component health during upgrade: https://gitlab-vmw.devops.broadcom.net/core-build/cayman_vm-operator/-/blob/vmware-main/vm-operator/wcp_patch/upgrade/gc_component_upgrade.py#L20

Thanks for the change.

PR #760, opened 2024-10-22
:seedling: Refactor unifiedstoragequota validating webhook unit tests to use Ginkgo

abaruni:

The PR refactors the unit tests in `webhooks/unifiedstoragequota/validation` to use Ginkgo in order to maintain consistency with the other tests in this project.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/infra/validatingwebhookconfiguration/validatingwebhookconfiguration_controller_intg_test.go
+++ b/controllers/infra/validatingwebhookconfiguration/validatingwebhookconfiguration_controller_intg_test.go
@@ -141,7 +141,7 @@ func intgTestsReconcile() {
 			})
 		})
 
-		When("ValidatingWebhookConfiguration is updated", func() {
+		When("ValidatingWebhookConfiguration is updated", FlakeAttempts(5), func() {
 
 			BeforeEach(func() {
 				// Create ValidatingWebhookConfiguration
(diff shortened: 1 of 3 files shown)


controllers/infra/validatingwebhookconfiguration/validatingwebhookconfiguration_controller_intg_test.go
```diff
@@ -141,7 +141,7 @@ func intgTestsReconcile() {
 			})
 		})
 
-		When("ValidatingWebhookConfiguration is updated", func() {
+		When("ValidatingWebhookConfiguration is updated", FlakeAttempts(5), func() {
```
bryanv, line 144: What's flakey on this test? If it is the `Eventually()` you can specify a custom timeout via an optional param after the `func()` or via the recent'ish added `WithTimeout()` method

abaruni: this test failed on a recent PR job https://github.com/vmware-tanzu/vm-operator/actions/runs/11460568651/job/31887603167 but i'm not able to recreate the failure locally

bryanv: Ah, I think the reason that is racing is that the `StoragePolicyUsage` created during test setup hasn't made it into the client cache yet, so the `client.List()` doesn't include it. And because this controller only Watches the `ValidatingWebhookConfiguration`, there is nothing that will cause another reconcile before the `Eventually()` times out.

I'll talk with you how to fix it

bryanv (approved):

PR #763, opened 2024-10-23
🌱 Disallow PVCs with WaitForFirstConsumer StorageClass

bryanv:

CSI does not support this for VMs at the moment since it does not wait for VMs to be placed. CSI doesn't produce a very helpful error message for this condition, and there is an ask to provide immediate feedback so put this best-effort check in the validation webhook.

This uses field. Forbidden instead of field.NotSupported since the later does not take a string to provide a useful error message.

Fixes #

Temp/quick fix - I'm not otherwise a huge fan of this being here

```release-note
PVCs with a WaitForFirstConsumer StorageClass are not supported for VirtualMachines.
```

### DIFF
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -24,7 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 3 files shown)


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -855,14 +857,36 @@ func (v validator) validateVolumeWithPVC(
...
-		allErrs = append(allErrs, field.NotSupported(pvcPath.Child("readOnly"), true, []string{"false"}))
+	if vol.PersistentVolumeClaim.InstanceVolumeClaim != nil {
+		return allErrs
+	}
+
+	// For now, make this check best effort. There's an ask to provide immediate feedback but
+	// ideally the validation webhooks should just deal with internal consistency.
+	pvc := &corev1.PersistentVolumeClaim{}
+	if err := v.client.Get(ctx, ctrlclient.ObjectKey{Name: claimName, Namespace: vm.Namespace}, pvc); err != nil {
+		return allErrs
```
dilyar85, line 877: Ignore the not found error here in case the PVC is created after VM creation?

dilyar85 (approved):
LGTM!

webhooks/virtualmachine/validation/virtualmachine_validator.go
dilyar85: Nvm, I guess we cannot really enforce this if we ignore the not-found error here.

PR #764, opened 2024-10-23
🐛 Fix how crypto detects device sans policy

akutz:

This patch fixes how the crypto logic detects whether or not a device is being added/edited sans policy that supports encryption.

Previously this happened via a direct IOFILTERS check, but that may not be set in the profile on the device. Instead the profile ID must be used to query whether or not the profile supports encryption.

Fixes `NA`

```release-note
NONE
```


dougm (approved):
lgtm

PR #765, opened 2024-10-23
WiP: ✨ Rekey disks for encrypted VMs

akutz:

This patch supports rekeying disks when an encrypted VM is rekeyed.

Fixes `NA`

Not yet ready, just throwing up so @dougm can take a look.

```release-note
Support rekeying disks when an encrypted VM is rekeyed.
```


pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -571,17 +578,108 @@ func onRecrypt(
...
+
+// getOrCreateDeviceChangeForDisk returns the device change for the specified
+// disk. If there is an existing Edit change, that is returned. If there is an
+// existing Add/Remove change, nil is returned. Otherwise a new Edit change is
+// returned.
+func getOrCreateDeviceChangeForDisk(
+	args reconcileArgs,
+	disk *vimtypes.VirtualDisk) *vimtypes.VirtualDeviceConfigSpec {
+
+	for i := range args.configSpec.DeviceChange {
```
dougm, line 655: nit: Some of these should never be nil, right? If they happen to be, seems it would be better to panic than silently ignore? Easier to read without too imho.

```go
	for _, dc := range args.configSpec.DeviceChange {
		ds := dc.GetVirtualDeviceConfigSpec()
		vd := ds.Device.GetVirtualDevice()
		if vd.Key == disk.Key {
			if ds.Operation == vimtypes.VirtualDeviceConfigSpecOperationEdit {
				return ds
			}
			return nil
		}
	}
```

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -571,17 +578,108 @@ func onRecrypt(
...
+		"newProviderIsDefault", args.newKey.isDefaultProvider,
+		"recryptedDisks", recryptedDisks)
 
 	return 0, nil, nil
 }
 
+func onRecryptDisks(args reconcileArgs) []string {
+	var fileNames []string
+	for i := range args.moVM.Config.Hardware.Device {
+		baseDev := args.moVM.Config.Hardware.Device[i]
```
dougm, line 598: nit: looks like this line can be dropped with:
```go
for _, baseDev := range args.moVM.Config.Hardware.Device {
```

dougm (approved):
lgtm @akutz

pkg/vmconfig/crypto/crypto_reconciler_pre.go
dougm: - `args.configSpec.DeviceChange[]` - our code should not be adding a nil to this list
- `dc.GetVirtualDeviceConfigSpec()` - in this case just converts `vimtypes.BaseVirtualDeviceConfigSpec` to `*vimtypes.VirtualDeviceConfigSpec`
- `ds.Device` - can't have a DeviceChange w/o this field. If this is nil, it's a bug in our code
- `ds.Device.GetVirtualDevice()`- returns a pointer to a field value, can't be nil if we've made it this far

PR #769, opened 2024-10-24
🌱  Avoid endless retry if StorageClass does not have policy ID

bryanv:

If a StorageClass does not have a storagePolicyID, don't return an error from the Reconcile so we don't go into an endless error retry because an update to the StorageClass will cause a reconcile anyways.

Fixes #

```release-note
NONE
```


akutz (approved):
Thanks @bryanv !

PR #772, opened 2024-10-25
📖 Doc updates

akutz:

This patch updates the vm documentation with information about:

* Encryption
* Storage quota
* Identifiers

The patch also rewrites a bunch of the docs to be more consistent with the overall patterns/conventions of the material.

Fixes `NA`

```release-note
Documentation on encryption, storage, identifiers
```

----
📚 Documentation preview 📚: https://vm-operator--772.org.readthedocs.build/en/772/

### DIFF
--- a/docs/concepts/workloads/vm.md
+++ b/docs/concepts/workloads/vm.md
@@ -30,45 +30,59 @@ kubectl apply -n <NAMESPACE> -f ${{ config.repo_url_raw }}/main/docs/concepts/wo
 
 ## Working with VMs
 
-When a VM gets created, it is scheduled to run on a [`Node`](https://kubernetes.io/docs/concepts/architecture/nodes/) in your cluster. Regardless where VM Operator is running, the `Node` represents an underlying hypervisor. For example, on vSphere Supervisor a `Node` represents an ESXi host.
+VMs are scheduled to run on a [`Node`](https://kubernetes.io/docs/concepts/architecture/nodes/) in the cluster. Regardless where VM Operator is running, the `Node` represents an underlying hypervisor. For example, on vSphere Supervisor a `Node` represents an ESXi host.
 
-The VM remains on that `Node` until the VM resource is deleted or, if supported by the underlying platform, the VM is rescheduled on another `Node` via live-migration (ex. vMotion) due to an unexpected failure of planned maintenance resulting in a `Node`'s underlying hypervisor going offline. There may also be situations where a VM cannot be rescheduled with live-migration due to some hardware dependency. In these instances the VM is powered off and then is powered back onto a compatible `Node` if one is available. If no other `Nodes` can support the VM, its controller will continue to try powering on the VM until such time a compatible `Node` becomes available.
+The VM remains on that `Node` until the VM resource is deleted or, if supported by the underlying platform, the VM is rescheduled on another `Node` via live-migration (ex. vMotion). There may also be situations where a VM cannot be rescheduled with live-migration due to some hardware dependency. In these instances, depending on how the infrastructure administrator has configured the platform, the VM may be automatically or manually powered off and then powered on again on a compatible `Node` if one is available. If no other `Nodes` can support the VM, its controller will continue to try powering on the VM until such time a compatible `Node` becomes available.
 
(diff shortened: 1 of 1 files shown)


docs/concepts/workloads/vm.md
```diff
@@ -130,6 +144,126 @@ The VM must either be powered off or transitioning from powered off to powered o
...
+  className:    my-vm-class
+  imageName:    vmi-0a0044d7c690bcbea
+  storageClass: my-encrypted-storage-class
+  crypto:
+    encryptionClassName: my-encryption-class
+```
+
+### Rekeying a VM
+
+Users can rekey VMs by switching to a different `EncryptionClass` or setting a currently non-empty `spec.crypto.encryptionClassName` to an empty value. This will result in the VM being rekeyed using the default key provider if it exists. If it does not, then the VM could be left in a state where it cannot be powered on.
```
dilyar85, line 194: > This will result in the VM being rekeyed using the default key provider if it exists.

I assume "this" above refers to the latter specifically (i.e. resetting `encryptionClassName` to empty)? Or does "this" also include "switching to a different EncryptionClass" mentioned in the previous sentence that would result using a default key provider as well?

docs/concepts/workloads/vm.md
```diff
@@ -130,6 +144,126 @@ The VM must either be powered off or transitioning from powered off to powered o
...
+    encrypted:
+    - config
+    - disks
+    providerID: my-key-provider-id
+    keyID: my-key-id
+```
+
+#### Encryption Type
+
+The type of encryption used by the VM is reported in the list `status.crypto.encrypted`. The list may contain the values `Config` and/or `Disks`, depending on the storage class and hardware present in the VM as the chart below illustrates:
```
dilyar85, line 224: > The list may contain the values `Config` and/or `Disks`

"`Config` and/or `Disks`" don't match the `config` and `disks` values shown in the YAML above.

docs/concepts/workloads/vm.md
```diff
@@ -315,7 +449,74 @@ The above data does _not_ represent the _observed_ network configuration of the
...
+There are two types of volumes: _Managed_ and _Classic_:
+
+* **Managed** volumes refer to storage that is provided by [PersistentVolumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes).
+* **Classic** volumes refer to the disk(s) that came from the `VirtualMachineImage` or `ClusterVirtualMachineImage` from which the `VirtualMachine` was deployed.
+
+#### Adding Volumes
+
+##### Adding Classic Volumes
+
+There is no support for adding classic vSphere disks directly to `VirtualMachine` resources. Additional storage _must_ be provided by managed volumes, i.e. [PersistentVolumeClaims](https://kubernetes.io/docs/concepts/storage/persistent-volumes).
```
dilyar85, line 515: Use https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims to link the PVCs?

docs/concepts/workloads/vm.md
```diff
@@ -360,9 +561,75 @@ Deployed VMs inherit the storage defined in the `VirtualMachineImage`. To provid
...
+=== "Volume Properties"
+
+    | Name | Description |
+    |------|-------------|
+    | `attached` | Whether or not the volume has been successfully attached to the `VirtualMachine`. |
+    | `crypto` | An optional field set only if the volume is encrypted. |
+    | `diskUUID` | The unique identifier of the volume's underlying disk. |
+    | `limit` | The maximum amount of space that may be used by this volume. |
+    | `name` | The name of the volume. For managed disks this is the name from `spec.volumes` and for classic disks this is the name of the underlying disk. |
+    | `type` | The total storage space occupied by this VirtualMachine that is not shared with any other `VirtualMachine`. |
```
dilyar85, line 577: This seems wrong and should be https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha3/virtualmachine_storage_types.go#L111?

Also, should we include the [Error](https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha3/virtualmachine_storage_types.go#L145-L147) field here as well?

dilyar85 (approved):
LGTM, thanks @akutz!!

PR #777, opened 2024-10-27
🌱 Simplify main.go / use k8s common logging

akutz:

This patch simplifies main.go by moving a lot of the overloaded code from the main() function into distinct init functions.

This patch also updates VM Op to use k8s.io common logging config, fixing some issues with certain parts of the project not correctly inheriting the log level.

Fixes `NA`

I will make a note once I've validated this on a testbed.

```release-note
Simplify `main.go` and use k8s.io/component-base for logging configuration.
```


main.go
```diff
@@ -11,15 +11,14 @@ import (
 	"path"
 	"time"
 
-	"github.com/go-logr/logr"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
dilyar85, line 14: Is this really needed? If it's for future proofing, should we import `"k8s.io/client-go/plugin/pkg/client/auth"` to support more comprehensive cloud providers authentication?

dilyar85 (approved):
LGTM with a minor comment left. Thanks @akutz for optimizing this!

main.go
akutz: It's not new, it's a relocation from lower in the file. It just got re-ordered. I agree we should create an issue to investigate whether we really need to load this package, but I don't think that should block _this_ PR.

PR #785, opened 2024-10-29
🐛 byok: validate encryption class keyID if specified

dougm:
56a5c8a2 removed keyID validation, as 'IsValidKey' used the ListKeys API which only returns keys cached in vCenter. The 'IsValidKey' impl now uses 'QueryCryptoKeyStatus' which checks keyID against the KMS.

### DIFF
--- a/pkg/vmconfig/crypto/crypto_reconciler_pre.go
+++ b/pkg/vmconfig/crypto/crypto_reconciler_pre.go
@@ -70,6 +70,10 @@ var (
 	// ErrInvalidKeyProvider is returned if the key provider specified by an
 	// EncryptionClass is invalid.
 	ErrInvalidKeyProvider = errors.New("invalid key provider")
+
+	// ErrInvalidKeyID is returned if the key id specified by an
+	// EncryptionClass is invalid.
+	ErrInvalidKeyID = errors.New("invalid key id")
 )
 
 func (r reconciler) Reconcile(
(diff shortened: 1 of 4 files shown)


pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -546,6 +550,12 @@ func getCryptoKeyFromEncryptionClass(
 		return cryptoKey{}, ErrInvalidKeyProvider
 	}
 
+	if obj.Spec.KeyID != "" {
+		if ok, _ := m.IsValidKey(ctx, obj.Spec.KeyProvider, obj.Spec.KeyID); !ok {
```
dougm, line 554: Note that this method of key validation is not supported with the Native key provider. The Native key provider does not support directly generating keys either. So it seems specifying a `keyID` with native `keyProvider` is not a valid use-case? If it is considered a valid use-case, will need to update this to skip calling `IsValidKey` if the provider is native.

akutz: Earlier in development, I tracked if the provider was native. We can reintroduce that for this.

akutz: Earlier in development, I tracked if the provider was native. We can reintroduce that for this.

akutz (approved):
Thanks @dougm , looks good to me!

PR #791, opened 2024-11-01
🐛 Update VM API spec.bootstrap.sysprep fields

### LINKED ISSUE #790: UserData should not be optional for Sysprep / should throw error sooner

https://github.com/vmware-tanzu/vm-operator/blob/284063573075c66e7b068a81cfdb064752519053/api/v1alpha3/sysprep/sysprep.go#L45
I left out userData and the request was accepted, but during the customize request to VC it returned InvalidArgument("spec.identity.userData.fullName") as these are all required fields.

I just assumed the optional + omitempty meant it was, and possibly customization would be skipped, or this portion would be.

I'd be happy if the webhook rejected the request so it fails fast.

I'm just a proxy for this. Replace all the "I" with a teammate's name.

akutz:

This patch includes several changes to the VirtualMachine API's sysprep/linuxprep bootstrap provider, including:

* The field `spec.bootstrap.sysprep.userData` is now required. This is only true if `spec.bootstrap.sysprep` is set, so it does not prevent someone from omitting a bootstrap provider. This is aligned with the VMODL used by VM Operator to customize the guest.

* The sysprep bootstrap provider now defaults to the UTC timezone by setting the default value of the field `spec.bootstrap.sysprep.guiUnattended.timeZone` to be 85. This is because the VMODL says [`guiUnattended` is required](https://dp-downloads.broadcom.com/api-content/apis/API_VWSA_001/8.0U3/html/ReferenceGuides/vim.vm.customization.Sysprep.html), but we got away with it being optional in VM Op since all the contained fields default to their primitive type defaults. Except `timeZone` defaulted to `0`, which s GMT-12 per [the reference](https://learn.microsoft.com/en-us/previous-versions/windows/embedded/ms912391(v=winembedded.11)), which is not a sane default, whereas UTC is.

Fixes #790

```release-note
Treat spec.bootstrap.sysprep.userData as required; default to UTC for both sysprep and linuxPrep bootstrap providers.
```


dilyar85 (approved):
LGTM, thanks @akutz!

PR #794, opened 2024-11-07
🌱 Minimum version selection

akutz:

This patch updates all of the go.mod files in this repository, *other than the one for building the manager binary*, to specify the minimum required version of Go and minimum required versions for dependencies.

This change allows downstream consumers of things like VM Op's API module to use a version of Go older than the one used by this repo. It turns out that [Go 1.21 changed the meaning of the Go version in `go.mod`](https://go.dev/doc/modules/gomod-ref#go-notes):

> The go directive sets the minimum version of Go required to use this module. Before Go 1.21, the directive was advisory only; now it is a mandatory requirement: Go toolchains refuse to use modules declaring newer Go versions.

Prior to 1.21, the version of Go in go.mod was informational. With Go 1.21 and after, it's required. That means if someone depends on VM Op's API module, and the module specifies Go 1.23.2, then the project depending on us requires 1.23.2.

Of course our main binary still specifies the latest Go version and newest dependencies for handling CVEs. However, our sub-modules that are made to be pulled into *other* targets, such as our own manager, have no reason to specify the latest version of Go or dependencies.

Fixes `NA`

cc @prziborowski 

```release-note
Since Go 1.21 changes the version of Go in go.mod from informative to required, all VM Op sub-modules now use the oldest version of Go and oldest versions of dependencies to allow minimum version selection.
```


zyiyi11 (approved):
Thanks Andrew for going through the whole directory.

PR #796, opened 2024-11-08
🌱Remove BackupUpToDate Condition from the backup data

sreyasn:

This changes is purely cosmetic and for debugging ease. The backupData will always end up having a condition pointing to the previous successful backup while the annotation would be the newer one. This removes the ambiguity.

Fixes #
N/A

N/A

```
NONE
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/backup_test.go
+++ b/pkg/providers/vsphere/virtualmachine/backup_test.go
@@ -193,6 +193,9 @@ func backupTests() {
 					JustBeforeEach(func() {
 						if IncrementalRestore {
 							vmCtx.VM.Annotations[vmopv1.VirtualMachineBackupVersionAnnotation] = vT2
+							vmCtx.VM.Status.Conditions = []metav1.Condition{
+								{Type: vmopv1.VirtualMachineBackupUpToDateCondition, Status: "True"},
+							}
 						}
 
 						oldVM := vmCtx.VM.DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/virtualmachine/backup_test.go
```diff
@@ -234,6 +237,7 @@ func backupTests() {
 						}
 
 						Expect(virtualmachine.BackupVirtualMachine(backupOpts)).To(Succeed())
+						// getExpectedBackupObjectYaml empties the expected VM yaml's status
```
aruneshpa, line 240: Do you want to add an expect matcher for this to confirm that the condition does not exist?

aruneshpa (approved):

pkg/providers/vsphere/virtualmachine/backup_test.go
sreyasn: can do. yep

sreyasn: hmm.. looked at it and this is probably the max we can do with the string matching that's already happening in verifyBackupDataInExtraConfig unless we unmarshal the backupData back to the VM object (which feels like overkill) 

Let me explain. 

So the expected yaml doesn't have a status (ie no backup condition too). The expect string matcher comparing the decoded backupData and expected Yaml (string) is in verifyBackupDataInExtraConfig (ie) no condition is present.

PR #801, opened 2024-11-13
✨ Deprecate ProbeManager for non-TCP probes

akutz:

This patch deprecates the use of ProbeManager for non-TCP probes when the AsyncSignal feature is enabled. This is because VMs will be reconciled as soon as their heartbeat or guestinfo information is updated, and thus readiness can be determined in the normal VM reconcile loop.

Fixes `NA`

```release-note
The ProbeManager framework is deprecated for non-TCP probes when async signal is enabled.
```

### DIFF
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -280,6 +280,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 
 	ctx = ctxop.WithContext(ctx)
 
+	ctx = record.WithContext(ctx, r.Recorder)
+
 	vm := &vmopv1.VirtualMachine{}
 	if err := r.Get(ctx, req.NamespacedName, vm); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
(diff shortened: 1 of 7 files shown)


controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -451,9 +453,15 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineContext) (reterr
...
-	r.Prober.AddToProberManager(ctx.VM)
+	if !pkgcfg.FromContext(ctx).Features.WorkloadDomainIsolation ||
+		pkgcfg.FromContext(ctx).AsyncSignalDisabled {
+
+		// Add the VM to the probe manager. This is idempotent.
+		r.Prober.AddToProberManager(ctx.VM)
+	} else if p := ctx.VM.Spec.ReadinessProbe; p != nil && p.TCPSocket != nil {
+		// TCP probes still use the probe manager.
+		r.Prober.AddToProberManager(ctx.VM)
+	}
```
bryanv, line 464: I think we may need an `else { r.Prober.RemoveFromProberManager(ctx.VM) } ` here for if the probe is removed. I think the existing code is mostly a noop when the worker finds the `Spec.ReadinessProbe` empty but I don't think anything else will remove it from the vmReadinessProbeList until the VM is deleted.

akutz: Ah, in case the probe type changes, got it.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -880,3 +889,162 @@ const byteToGiB = 1 /* B */ * 1024 /* KiB */ * 1024 /* MiB */ * 1024 /* GiB */
...
+// readiness probes.
+// Please note, this function returns early if there are no non-TCP readiness
+// probes.
+func updateProbeStatus(
+	ctx context.Context,
+	vm *vmopv1.VirtualMachine,
+	moVM mo.VirtualMachine) {
+
+	p := vm.Spec.ReadinessProbe
+	if p == nil || p.TCPSocket != nil {
```
zyiyi11, line 932: iiuc, should be `p.TCPSocket == nil`?

zyiyi11 (approved):
LGTM. Thanks Andrew!

bryanv (approved):

PR #802, opened 2024-11-14
🌱 Use VAPI ActivationID with DeployOVF

akutz:

This patch updates the DeployOVF call to use the VAPI activationID header. The value is set to the VM's instance UUID. This means the vim.Task that Content Library spawns to track the actual DeployOVF operation will have its activationId field set to the value of vm.spec.instanceUUID. Thus VM Operator can find and track the task, even though the DeployOVF call is blocking and does not return a task ID.

Fixes `NA`

```release-note
Use vm.spec.instanceUUID as the VAPI activation ID when calling ContentLibrary's DeployOVF API.
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/publish.go
+++ b/pkg/providers/vsphere/virtualmachine/publish.go
@@ -5,7 +5,6 @@ package virtualmachine
 
 import (
 	"fmt"
-	"net/http"
 
 	"github.com/vmware/govmomi/vapi/rest"
 	"github.com/vmware/govmomi/vapi/vcenter"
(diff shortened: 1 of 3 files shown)


bryanv (approved):
Just in case you haven't seen it, there is some unfortunate behavior with the ActivationID, like IIRC there is some window after the original submit before you can actually find the task w/ it: the publish controller has more details on that behavior.

akutz:
> Just in case you haven't seen it, there is some unfortunate behavior with the ActivationID, like IIRC there is some window after the original submit before you can actually find the task w/ it: the publish controller has more details on that behavior.

Thanks. I remember. The period of time depends on the API implementation as some create the task immediate and others do not.

PR #803, opened 2024-11-15
:seedling: Allow SRM to change network of a registered VM for test failover

aruneshpa:

SRM (Site Reliability Manager) needs to be able to change a VM's network interface to connect to the test failover network.  This change skips part of the validation webhook if a predefined test failover label is present on the VM resource.  This change introduces a label (as opposed to an annotation) because we want to allow server side indexing so clients can filter and operate on all VMs that are part of a test failover.  This will help with debuggability, and cleanup of test failed over VMs. 

A test failover does not need to add new interfaces, so keep that check as is.

**Testing Done**
- Existing unit tests as well as newly added tests
- Manually tested by loading VM operator in an env 
  - Without the label
```
root@4214a043bbff661bcddbde32adee61b3 [ ~ ]# kubectl patch vm -n parunesh-ns srm-vm --type='json' -p='[{"op": "replace", "path": "/spec/network/interfaces/1/network/name", "value": "my-subnet-2"}]' 
Error from server (spec.network.interfaces[1].network: Forbidden: field is immutable): admission webhook "default.validating.virtualmachine.v1alpha3.vmoperator.vmware.com" denied the request: spec.network.interfaces[1].network: Forbidden: field is immutable
```

  - Edit the VM to apply the label 
```
root@4214a043bbff661bcddbde32adee61b3 [ ~ ]# k edit vm -n parunesh-ns srm-vm virtualmachine.vmoperator.vmware.com/srm-vm edited
```

  - Verify that edits are allowed 
```
root@4214a043bbff661bcddbde32adee61b3 [ ~ ]# kubectl patch vm -n parunesh-ns srm-vm --type='json' -p='[{"op": "replace", "path": "/spec/network/interfaces/1/network/name", "value": "my-subnet-2"}]' virtualmachine.vmoperator.vmware.com/srm-vm patched
```

This is needed specifically for SRM (and possibly other vendors exercising a test failover).  As such, I have called out in the API docs that it is not an otherwise supported workflow.

```release-note
Allow SRM to change network of a registered VM for test failover
```

### DIFF
--- a/pkg/backup/api/backup_types.go
+++ b/pkg/backup/api/backup_types.go
@@ -3,6 +3,9 @@
 
 package api
 
+// GroupName specifies the group name for VM operator.
+const GroupName = "vmoperator.vmware.com"
+
 // ToPersistentVolumeAccessModes returns a string slice from a slice of T.
 // This is useful when converting a slice of corev1.PersistentVolumeAccessMode
 // to a string slice.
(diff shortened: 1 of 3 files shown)


akutz:
Why a label and not annotation? Unless the intent is to find resources with this label, then please change to an annotation.

aruneshpa:
> Why a label and not annotation? Unless the intent is to find resources with this label, then please change to an annotation.

Andrew, the label here is intentional.  This allows a user (admin) to identify the VMs that are test failed over.  This can specially be helpful if we want to round up these VMs for cleanup after a test failover is complete.  I will add this in the MR description for posterity.

akutz (approved):
Lgtm, thank you sir!

PR #805, opened 2024-11-21
✨ Ensure NICs are connected / start connected

akutz:

This patch adds support for VM Op ensuring NICs are always connected and start connected for VMs that are not paused either by an Admin or DevOps user.

Fixes `NA`

@bryanv, please note that I ensured all the actual meat and potatoes logic is in the new resize location. 

```release-note
Ensure NICs are connected for VMs not paused by admin or DevOps users.
```

### DIFF
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -520,6 +520,8 @@ func (s *Session) prePowerOnVMConfigSpec(
 	}
 	configSpec.DeviceChange = append(configSpec.DeviceChange, ethCardDeviceChanges...)
 
+	vmopv1util.ReconcileNetworkDeviceConnectionState(config, configSpec)
+
 	var expectedPCIDevices []vimtypes.BaseVirtualDevice
 	if configSpecDevs := pkgutil.DevicesFromConfigSpec(&updateArgs.ConfigSpec); len(configSpecDevs) > 0 {
 		pciPassthruFromConfigSpec := pkgutil.SelectVirtualPCIPassthrough(configSpecDevs)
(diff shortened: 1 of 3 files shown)


bryanv:
This will change the behavior of when using GOSC in that the VM will boot with the NICs already connected, instead of that not happening until customization is complete. IDK if that's a necessarily a breaking change though.

akutz:
> This will change the behavior of when using GOSC in that the VM will boot with the NICs already connected, instead of that not happening until customization is complete. IDK if that's a necessarily a breaking change though.

cc @PengpengSun

PengpengSun:
> > This will change the behavior of when using GOSC in that the VM will boot with the NICs already connected, instead of that not happening until customization is complete. IDK if that's a necessarily a breaking change though.
> 
> cc @PengpengSun

Agree with Byran, please do not re-connect NICs before customization finishes in guest. When customizing network, connected NICs will create conflicts. If GOSC happens, GOSC will handle NICs re-connection.

akutz:
> > > This will change the behavior of when using GOSC in that the VM will boot with the NICs already connected, instead of that not happening until customization is complete. IDK if that's a necessarily a breaking change though.
> > 
> > 
> > cc @PengpengSun
> 
> Agree with Byran, please do not re-connect NICs before customization finishes in guest. When customizing network, connected NICs will create conflicts. If GOSC happens, GOSC will handle NICs re-connection.

Hmmm. This is problematic. Ideally VM Operator *should* try to ensure the NICs are in the connected state. We should figure out how to do this even if GOSC is running or has run. Likely I'll do it by first checking the current status of a possible customization.

PengpengSun:
> > > > This will change the behavior of when using GOSC in that the VM will boot with the NICs already connected, instead of that not happening until customization is complete. IDK if that's a necessarily a breaking change though.
> > > 
> > > 
> > > cc @PengpengSun
> > 
> > 
> > Agree with Byran, please do not re-connect NICs before customization finishes in guest. When customizing network, connected NICs will create conflicts. If GOSC happens, GOSC will handle NICs re-connection.
> 
> Hmmm. This is problematic. Ideally VM Operator _should_ try to ensure the NICs are in the connected state. We should figure out how to do this even if GOSC is running or has run. Likely I'll do it by first checking the current status of a possible customization.

GOSC won't re-connect NICs only if something wrong happened, at this point, even re-connect NICs by VM Operator,  the status of the VM is unpredictable(network might/might not work).
And the connection status of the template/base vm needs to be considered too,  if it's disconnected originally, GOSC will not re-connect NICs either.

If VM Operator wants to check customization status, this [vAPI](https://vmware.github.io/vsphere-automation-sdk-python/vsphere/8.0.3.0/com.vmware.vcenter.vm.html#com.vmware.vcenter.vm.guest_client.Customization.get) can be used.  IMHO, no need re-connect NICs when customization fails, but VM Operator could have a different story.

PR #811, opened 2024-12-03
🌱 OVF Cache Package

akutz:

This patch refactors the OVF cache into an context-based package so the cache may be shared across the project and not limited to the vSphere provider. This also simplifies a future refactor to transition the cache to using Secret/ ConfigMap resources in Kube for caching the OVF envelopes as opposed to an in-memory store.

Fixes `NA`

This is required for the Fast Deploy work in order to efficiently retrieve the OVF envelope for parsing its controllers/disks.

```release-note
Context-based OVF cache
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider.go
+++ b/pkg/providers/vsphere/vmprovider.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/ovf"
@@ -35,7 +34,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/vcenter"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
 	"github.com/vmware-tanzu/vm-operator/pkg/topology"
-	"github.com/vmware-tanzu/vm-operator/pkg/util"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/ovfcache"
 	vsclient "github.com/vmware-tanzu/vm-operator/pkg/util/vsphere/client"
 )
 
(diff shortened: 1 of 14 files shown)


pkg/providers/vsphere/vmprovider.go
```diff
@@ -44,10 +43,6 @@ const (
 
 	// taskHistoryCollectorPageSize represents the max count to read from task manager in one iteration.
 	taskHistoryCollectorPageSize = 10
-
-	ovfCacheMaxItem                 = 100
-	ovfCacheItemExpiration          = 30 * time.Minute
-	ovfCacheExpirationCheckInterval = 5 * time.Minute
```
dilyar85, line 50: Closest place GitHub allows me to comment: the `type VersionedOVFEnvelope struct ...` block below can also be deleted from this file.

pkg/util/ovfcache/ovfcache.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/vmware/govmomi/ovf"
+
+	"github.com/vmware-tanzu/vm-operator/pkg/util/ovfcache/internal"
+)
+
+// VersionedOVFEnvelope is an OVF envelope with an associated content version.
+type VersionedOVFEnvelope = internal.VersionedOVFEnvelope
+
+// GetterFn returns an OVF envelope for a provided library item ID.
+type GetterFn = internal.GetterFn
```
dilyar85, line 19: I don't see these types being referenced by any external callers. Is there any reason to declare them here instead of keeping them in the internal file?

pkg/util/ovfcache/ovfcache.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	maxItems            = 100
+	expireAfter         = 30 * time.Minute
+	expireCheckInterval = 5 * time.Minute
+)
+
+// ErrNoGetter is returned from GetOVFEnvelope if there is no getter function.
+var ErrNoGetter = internal.ErrNoGetter
```
dilyar85, line 28: Ditto.

akutz: This may not be referenced anywhere at the moment, but we leave it public to ensure it can be compared if necessary.

pkg/util/ovfcache/ovfcache.go
akutz: The getter function is as it is used in SetGetter.

pkg/util/ovfcache/internal/ovfcache_internal.go
```diff
@@ -0,0 +1,167 @@
...
+		func(curVal contextValueType) sync.Locker {
+			return curVal.locks.Get(itemID)
+		})
+}
+
+func SetGetter(parent context.Context, getter GetterFn) {
+	ctxgen.SetContext(
+		parent,
+		contextKeyValue,
+		func(curVal contextValueType) contextValueType {
```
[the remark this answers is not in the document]
akutz: A function is a pointer, so it's fine. Also, the new value is reflected appropriately. It's the way the generic context package works.

dilyar85 (approved):
LGTM, thanks @akutz!

PR #813, opened 2024-12-05
🐛 Fix panic in controllers missing ovfcache

akutz:

This patch fixes the issue where controllers getting the OVF cache from the context would panic because the original commit did not include the JoinContext call for the OVF cache in the controllers.

Fixes `NA`

```release-note
Fix panic when OVF cache is missing from context
```


zyiyi11 (approved):
Thanks for the quick fix!

PR #814, opened 2024-12-05
:seedling: Allow imported and registered VMs to skip image validation checks on Create/Update

aruneshpa:

VMs that are imported, or registered already have the virtual machine created on the infrastructure.  As such, we don't need to apply the same strict validations for them such as making sure the image is not null / valid.  To that end, this change introduces an annotation that will be applied on the Imported, or Registered VMs.  This will allow our webhooks to skip these checks if the annotation is present.

For now, maintain the existing mechanism which allows _all_ VM creations to be imageless if the import or the incremental restore feature is enabled.  Once this change is merged, we will yank out that code to make that check stricter (only annotation based).

There;s also a change to the UT method that validates messages.  Instead of splitting that into a string array and matching elements, we are now comparing substrings.  This is because the comma delimiter used to split the reasons results in false splits when a reason contains an object.

```release-note
Allow imported and registered VMs to skip image validation checks on Create/Update
```

### DIFF
--- a/api/v1alpha3/virtualmachine_types.go
+++ b/api/v1alpha3/virtualmachine_types.go
@@ -197,6 +197,18 @@ const (
 	// ManagerID on a VirtualMachine contains the UUID of the
 	// VMware vCenter (VC) that is managing this virtual machine.
 	ManagerID = GroupName + "/manager-id"
+
+	// RegisteredVMAnnotation on a VirtualMachine represents that a virtual machine has
+	// been registered using the RegisterVM API after a restore, or a fail-over operation by
+	// a vendor. The presence of this annotation is used to bypass some validation checks
+	// that are otherwise applicable to all VirtualMachine create/update requests.
+	RegisteredVMAnnotation = GroupName + "/registered-vm"
+
+	// ImportedVMAnnotation on a VirtualMachine represents that a traditional virtual
+	// machine has been imported into Supervisor using the ImportVM API. The presence of this
+	// annotation is used to bypass some validation checks that are otherwise applicable
+	// to all VirtualMachine create/update requests.
+	ImportedVMAnnotation = GroupName + "/imported-vm"
 )
 
 const (
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -41,6 +41,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/config"
 	"github.com/vmware-tanzu/vm-operator/pkg/topology"
 	"github.com/vmware-tanzu/vm-operator/pkg/util"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/annotations"
 	cloudinitvalidate "github.com/vmware-tanzu/vm-operator/pkg/util/cloudinit/validate"
 	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 	vmopv1util "github.com/vmware-tanzu/vm-operator/pkg/util/vmopv1"
@@ -151,6 +152,37 @@ func (v validator) ValidateCreate(ctx *pkgctx.WebhookRequestContext) admission.R
 
 func (v validator) ValidateDelete(*pkgctx.WebhookRequestContext) admission.Response {
 	return admission.Allowed("")
+}
+
+// Updates to VM's image are only allowed if it is a Registered VM (used for failover
+// in disaster recovery).
+func (v validator) validateImageOnUpdate(ctx *pkgctx.WebhookRequestContext, vm, oldVM *vmopv1.VirtualMachine) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if vmopv1util.IsImagelessVM(*vm) && pkgcfg.FromContext(ctx).Features.VMIncrementalRestore {
+		if annotations.HasRegisterVM(vm) {
+			if !vmopv1util.ImageRefsEqual(vm.Spec.Image, oldVM.Spec.Image) {
+				if !ctx.IsPrivilegedAccount {
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "image"), restrictedToPrivUsers))
+				}
+			}
+
+			if oldVM.Spec.ImageName != vm.Spec.ImageName {
+				if !ctx.IsPrivilegedAccount {
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "imageName"), restrictedToPrivUsers))
+				}
+			}
+
+			return allErrs
+		}
+	}
+
+	allErrs = append(allErrs,
+		validation.ValidateImmutableField(vm.Spec.ImageName, oldVM.Spec.ImageName, field.NewPath("spec", "imageName"))...)
+	allErrs = append(allErrs,
+		validation.ValidateImmutableField(vm.Spec.Image, oldVM.Spec.Image, field.NewPath("spec", "image"))...)
+
+	return allErrs
 }
 
 // ValidateUpdate validates if the given VirtualMachineSpec update is valid.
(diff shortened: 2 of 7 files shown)


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -153,6 +153,35 @@ func (v validator) ValidateDelete(*pkgctx.WebhookRequestContext) admission.Respo
 	return admission.Allowed("")
 }
 
+// Updates to VM's image are only allowed if it is a Registered VM (used for failover
+// in disaster recovery).
+func (v validator) validateImageOnUpdate(ctx *pkgctx.WebhookRequestContext, vm, oldVM *vmopv1.VirtualMachine) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if f := pkgcfg.FromContext(ctx).Features; f.VMIncrementalRestore && vmopv1util.IsRegisteredVM(*vm) {
```
sreyasn, line 161: If we wanna just allow unset of image/imageName. can we just check for IsImagelessVM(*vm) along with the IsRegisteredVM() check up top instead of the deepEqual() inside?. Because the deepEqual will let unset and reset now.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -197,6 +197,19 @@ const (
 	// ManagerID on a VirtualMachine contains the UUID of the
 	// VMware vCenter (VC) that is managing this virtual machine.
 	ManagerID = GroupName + "/manager-id"
+
+	// +kubebuilder:validation:Enum=ImportVM;RegisterVM
+	//
+	// SourceAPI on a VirtualMachine contains the API responsible for creating / updating
+	// 	the VirtualMachine custom resource for a virtual machine that already exists on the
```
akutz, line 204: Please fix the formatting here.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -197,6 +197,19 @@ const (
 	// ManagerID on a VirtualMachine contains the UUID of the
 	// VMware vCenter (VC) that is managing this virtual machine.
 	ManagerID = GroupName + "/manager-id"
+
+	// +kubebuilder:validation:Enum=ImportVM;RegisterVM
+	//
```
akutz, line 202: Remove the `//` between the kubebuilder annotation and the comment block. This prevents the kubebuilder annotation from becoming part of the GoDoc.

api/v1alpha3/virtualmachine_types.go
```diff
@@ -197,6 +197,19 @@ const (
...
+	// SourceAPI on a VirtualMachine contains the API responsible for creating / updating
+	// 	the VirtualMachine custom resource for a virtual machine that already exists on the
+	// infrastructure. The ImportVM API is used to import a traditional virtual machine in
```
akutz, line 205: `s/The ImportVM API is used to import a traditional virtual machine in/The ImportVM API is used to import a traditional virtual machine into/`

api/v1alpha3/virtualmachine_types.go
```diff
@@ -197,6 +197,19 @@ const (
 	// ManagerID on a VirtualMachine contains the UUID of the
 	// VMware vCenter (VC) that is managing this virtual machine.
 	ManagerID = GroupName + "/manager-id"
+
+	// +kubebuilder:validation:Enum=ImportVM;RegisterVM
```
akutz, line 201: Please do the following to make it consistent with our other enums.

```go
+kubebuilder:validation:Enum=ImportVM;RegisterVM
type SourceAPI string

const (
    SourceAPIImportVM SourceAPI = "ImportVM"

    SourceAPIRegisterVM SourceAPI = "RegisterVM"
)
```

Then you can use the above instead of defining strings in the constants package.

webhooks/virtualmachine/validation/virtualmachine_validator.go
akutz, line 161: nit: Consider rewriting as follows:

```go
if f := pkgcfg.FromContext(ctx).Features; f.VMIncrementalRestore {
    if vmopv1util.IsRegisteredVM(*vm) {
        // ...
    }
}
```

This makes it easier to remove the feature gate once the FSS is enabled.

aruneshpa: I was going with the idea that if it is a failover, or import, we don't want to do _any_ validations (not necessarily check for unset/reset).

sreyasn: Something to consider here. This will open up resetting image fields to new values when any privileged account modifies our VM that was imported/registered before (not just failover, right?). 

Might be safer just to allow unset (value -> "" , "" -> "") like we discussed?.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -153,6 +154,37 @@ func (v validator) ValidateDelete(*pkgctx.WebhookRequestContext) admission.Respo
...
+
+	if f := pkgcfg.FromContext(ctx).Features; f.VMIncrementalRestore {
+		if annotations.HasRegisterVM(vm) {
+			if !equality.Semantic.DeepEqual(oldVM.Spec.Image, vm.Spec.Image) {
+				if !ctx.IsPrivilegedAccount {
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "image"), restrictedToPrivUsers))
+				}
+			}
+
+			if !equality.Semantic.DeepEqual(oldVM.Spec.ImageName, vm.Spec.ImageName) {
```
akutz, line 170: Please change to a string comp to be more efficient.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -153,6 +154,37 @@ func (v validator) ValidateDelete(*pkgctx.WebhookRequestContext) admission.Respo
...
 }
 
+// Updates to VM's image are only allowed if it is a Registered VM (used for failover
+// in disaster recovery).
+func (v validator) validateImageOnUpdate(ctx *pkgctx.WebhookRequestContext, vm, oldVM *vmopv1.VirtualMachine) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if f := pkgcfg.FromContext(ctx).Features; f.VMIncrementalRestore {
+		if annotations.HasRegisterVM(vm) {
+			if !equality.Semantic.DeepEqual(oldVM.Spec.Image, vm.Spec.Image) {
```
akutz, line 164: The image ref is simple, so maybe just add a helper func to compare two of them instead of using a DeepEqual?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -369,8 +401,19 @@ func (v validator) validateImageOnCreate(ctx *pkgctx.WebhookRequestContext, vm *
 	case vmopv1util.IsImagelessVM(*vm) &&
 		(pkgcfg.FromContext(ctx).Features.VMImportNewNet ||
 			pkgcfg.FromContext(ctx).Features.VMIncrementalRestore):
+		// TODO: Simplify this once mobility operator starts creating VMs with correct annotation.
```
akutz, line 404: I am fine just removing the FSS checks.  However, you need to add a check to ensure only privileged users can set the annotations for Register/Import.

akutz (changes requested):
Left a few comments. Also, where are we restricting setting the new annotations to privileged users?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1295,6 +1337,14 @@ func (v validator) validateAnnotation(ctx *pkgctx.WebhookRequestContext, vm, old
 		allErrs = append(allErrs, field.Forbidden(annotationPath.Child(vmopv1.FirstBootDoneAnnotation), modifyAnnotationNotAllowedForNonAdmin))
 	}
 
+	if vm.Annotations[vmopv1.RegisteredVMAnnotation] != oldVM.Annotations[vmopv1.RegisteredVMAnnotation] {
+		allErrs = append(allErrs, field.Forbidden(annotationPath.Child(vmopv1.RegisteredVMAnnotation), modifyAnnotationNotAllowedForNonAdmin))
```
bryanv, line 1341: `annotationPath.Key()` is better to use for an annotation

bryanv (approved):

PR #818, opened 2024-12-09
🐛 Remove Inventory path fallback for GetVM

bryanv:

Originally, we'd try to find a VM via its MoID, with a fallback to the Inventory by path. Since then, we've added the Instance and Bios UUIDs as preferred ways to find a VM. VMs are now created with a known Instance ID - that of the k8s VM UID - that we can use for lookup.

This removes lookup depending on k8s objects to determine if the VM exists that have a separate lifecycle than the VM.

Fixes #

```release-note
The lookup VM by the Inventory path final fallback is removed. VM lookup will be via the Instance UUID, BiosUUID, or the MoID.
```


akutz (approved):
Lgtm, thanks @bryanv !

sreyasn (approved):

PR #822, opened 2024-12-11
:bug: Do not invoke StoragePolicyQuota Validating Webhook for imageless VMs

abaruni:

As part of StoragePolicyQuota Validating Webhook that runs inside VM Operator there is a check to ensure that the VM image is provided since this is where we obtain the image capacity from. This has the unintended side-effect of blocking vm creation for imageless VMs.

This patch modifies the ValidatingWebhookConfiguration used in the StoragePolicyQuota validation flow to ignore VM creation request where the image is not provided. This will prevent this webhook from blocking imageless VM creations, while allowing the VirtualMachine ValidatingWebhook to continue to perform image validation. 

Fixes #

```release-note
Do not invoke StoragePolicyQuota Validating Webhook for imageless VMs
```

### DIFF
--- a/config/webhook/storage_quota_webhook_configuration.yaml
+++ b/config/webhook/storage_quota_webhook_configuration.yaml
@@ -27,6 +27,9 @@ webhooks:
     resources:
     - virtualmachines
   sideEffects: None
+  matchConditions:
+  - expression: has(object.spec.image)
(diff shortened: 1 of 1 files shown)


akutz (approved):
Lgtm, thanks @abaruni !

PR #824, opened 2024-12-13
🐛 Update GoVmomi to b1027ba34d97 from main

akutz:

This patch updates the GoVmomi dependency to b1027ba34d97 to eagerly handle some issues it presents:

* the OVF package had some breaking changes
* there was a change in GoVmomi that resulted in the vC Sim datacenter MoID changing from datacenter-2 to datacenter-1. The simulated MoIDs are not stable, so it should not be hard-coded in VM Op's tests.
* the fallout from the unstable datacenter MoID also impacted the backup/restore tests since the UUID for the PVC disk was based on the datacenter MoID. This has been addressed as well, albeit with another hard-coded value (for now). A fix is needed in GoVmomi before it can be used to lookup the actual value.

Fixes `NA`

```release-note
Update GoVmomi to b1027ba34d97
```


dougm (approved):
Thanks @akutz !

PR #826, opened 2024-12-13
✨ Add option for CloudInit to default to global NS and SearchDomains

bryanv:

The CloudInit bootstrap only supports per-interface Nameservers and SearchDomains while others bootstrap like LinuxPrep support global. To simplify the UX and to make it easier to switch between bootstrap providers, add two *bool fields to the CloudInit spec:
 - UseGlobalNameserversAsDefault
 - UseGlobalSearchDomainsAsDefault
 
 that when either nil or true, use the global value when the corresponding per-interface field is unset.

In v1a2, the validation wehbook would have rejected an attempt to use the global Nameservers or SearchDomains with CloudInit, so this change should not result it any behavior change for an existing VM definition.

Fixes #

```release-note
When using the CloudInit bootstrap provider, unless UseGlobalNameserversAsDefault or UseGlobalSearchDomainsAsDefault is false, the per-interface nameservers and searchDomains will default to the corresponding global values in the NetworkSpec when the per-interface fields are not specified. 
```

----
📚 Documentation preview 📚: https://vm-operator--826.org.readthedocs.build/en/826/


akutz (approved):
Thank you @bryanv !

PR #827, opened 2024-12-13
🐛  Fix potential VM validation webhook NPE for Spec.Network

bryanv:

If the VM Spec.Network was changed to/from nil, the validation webhook would NPE. 

Add a few basic update tests for VM Spec.Network.

Fixes #

I think the stems from late in v1a2 we decided to make most of the top level Spec/Status fields pointers and this was just missed.

```release-note
NONE
```


akutz (approved):
Thanks BV!

zyiyi11 (approved):

PR #828, opened 2024-12-16
🌱 Fix race in WebConsoleRequest controller tests

dilyar85:

This PR fixes a potential race condition in unit tests by changing `v1a2ProviderCaller bool` type to `atomic.Bool`.  The race could happen if the WebConsoleRequest CR is reconciled multiple times, causing concurrent reads and writes.

Fixes https://github.com/vmware-tanzu/vm-operator/actions/runs/12355794076/job/34480234221?pr=823

N/A

```release-note
Fix race in WebConsoleRequest controller tests.
```

### DIFF
--- a/controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_intg_test.go
+++ b/controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_intg_test.go
@@ -5,6 +5,7 @@ package v1alpha1_test
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


dilyar85:
> Thanks @dilyar85 !
> 
> Could you could also use the atomic package to read/write `v1a2ProviderCalled`?

Done. I also removed the mutex when reading the var since it's not needed with `v1a2ProviderCalled` being atomic.

controllers/virtualmachinewebconsolerequest/v1alpha1/webconsolerequest_intg_test.go
```diff
@@ -97,19 +98,17 @@ func intgTestsReconcile() {
 			},
 		}
 
-		intgFakeVMProvider.Lock()
 		intgFakeVMProvider.GetVirtualMachineWebMKSTicketFn = func(ctx context.Context, vm *vmopv1.VirtualMachine, pubKey string) (string, error) {
```
bryanv, line 101: I think the fake FooFn assignments are protected by the lock because in the past the race detector has tripped with the Reconciler reading the field.

PR #834, opened 2024-12-18
🌱 Use errors.Is() to silence lint warnings

bryanv:

In case these become wrapped errors later.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -143,11 +143,16 @@ linters-settings:
           desc: "replaced by internal packages like pkg/util/ptr"
         - pkg: github.com/onsi/ginkgo$
           desc: "replaced by github.com/onsi/ginkgo/v2"
-
+  errorlint:
+    # Check whether fmt.Errorf uses the %w verb for formatting errors.
+    errorf: false
+    # Check for plain type assertions and type switches.
+    asserts: false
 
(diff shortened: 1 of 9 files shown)


akutz:
I'm not opposed to this, but what IDE is giving these errors? VSCode does not, and our lint target does not. I wonder if we can do something in our linter to catch these. I agree we should use `errors.Is` (or `.As` if we need the value) in case things are ever wrapped.

.golangci.yml
```diff
@@ -143,11 +143,16 @@ linters-settings:
           desc: "replaced by internal packages like pkg/util/ptr"
         - pkg: github.com/onsi/ginkgo$
           desc: "replaced by github.com/onsi/ginkgo/v2"
-
+  errorlint:
+    # Check whether fmt.Errorf uses the %w verb for formatting errors.
```
akutz, line 147: I _want_ `Errorf` to use `%w` to format errors. Why would we wish to disable that?

bryanv: That disables the lint for not using %w. There's lots of places where we're not and something that can be addressed later (along with the other disabled check)

akutz (approved):
Sorry for not getting back to this.

PR #838, opened 2024-12-23
:seedling: Use annotation helpers from upstream metav1 instead of custom ones

aruneshpa:

This change removes our helpers to check if an annotation is present. Instead, rely on the upstream (metav1) one.

```release-note
Use annotation helpers from upstream metav1 instead of custom ones
```


akutz (approved):
Lgtm, thank you!

bryanv (approved):

PR #839, opened 2024-12-23
   🌱 Enable all errorlint checks

bryanv:

Use %w instead of %v to wrap errors

Use errors Is() and As() and the govmomi equivalents instead of type casting/checking

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/kube/spq/spq.go
+++ b/pkg/util/kube/spq/spq.go
@@ -6,6 +6,7 @@ package spq
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -67,7 +68,8 @@ func GetStorageClassesForPolicy(
 				namespace,
 				o.Name)
 			if err != nil {
-				if _, ok := err.(NotFoundInNamespace); !ok {
+				var notFoundInNamespace NotFoundInNamespace
+				if !errors.As(err, &notFoundInNamespace) {
 					return nil, err
 				}
 			}
--- a/test/testutil/util.go
+++ b/test/testutil/util.go
@@ -6,6 +6,7 @@ package testutil
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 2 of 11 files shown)


test/testutil/util.go
```diff
@@ -102,6 +103,8 @@ func ContainsError(err error, message string) bool {
 					return true
 				}
 			}
+		default:
```
aruneshpa, line 106: Out of curiosity, why did you have to add the nolint and why this default block?

aruneshpa (approved):

pkg/util/kube/spq/spq.go
```diff
@@ -66,7 +67,8 @@ func GetStorageClassesForPolicy(
 				namespace,
 				o.Name)
 			if err != nil {
-				if _, ok := err.(NotFoundInNamespace); !ok {
+				var notFoundInNamespace NotFoundInNamespace
+				if !errors.As(err, &notFoundInNamespace) {
```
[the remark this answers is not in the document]
akutz: It should be `errors.Is`, yes. We only use `As` if we need something about the error, like a struct field.

test/testutil/util.go
akutz: He probably has the `//nolint:errorlint` because of the `.(type)`. I imagine the linter thinks this is type-asserting the error and tripping the same linter he's trying to handle with this change.

I imagine the `default` block was added because it was missing in the first place. It should behave similar to what I did at https://github.com/vmware/govmomi/blob/8eb362fe04b17b9922fb9f3a698533011e243778/fault/fault.go#L210-L211.

pkg/util/kube/spq/spq.go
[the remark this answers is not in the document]
bryanv: It needs to be `As()` here. `Is()` is the wrapped replacement for `==`, not just a type comparison ie `IsA[]()`. This is super confusingly named and way too easy to mess up.

test/testutil/util.go
bryanv: Rewrote this to not need the nolint anymore. The default is added just in case the error doesn't match and doesn't implement those unwrap interfaces and we'd otherwise go into an infinite loop.

PR #840, opened 2024-12-24
✨ Conditions helper & refactor

akutz:

This patch introduces the new type `Conditions` from `pkg/conditions`. This new type provides all of the same helper functions the package provides, but now for any slice of `metav1.Condition` objects. The previous helpers have been rewritten to use the new type.

Perhaps more importantly, the package has been updated to no longer require that the `Getter` and `Setter` interfaces refer to Kubernetes objects. Now the interfaces are valid for *any* Golang type that implements the get/set functions.

Finally, the `pkg/patch` package has also been updated to account for the changes to the `pkg/conditions` interfaces `Getter` and `Setter`.

Fixes `NA`

```release-note
The `pkg/conditions` package now works with non-Kubernetes objects.
```

### DIFF
--- a/pkg/conditions/conditions.go
+++ b/pkg/conditions/conditions.go
@@ -1,0 +1,229 @@
+// Copyright (c) 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package conditions
+
+import (
+	"sort"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// ReadyConditionType indicates a resource is ready.
+	ReadyConditionType = "Ready"
+)
+
+// Conditions is an alias for a slice of metav1.Condition objects and provides
+// helpful functions.
+type Conditions []metav1.Condition
+
+// c4g returns a Conditions type from a Getter. This is used throughout this
+// package to reduce the need to write "Conditions(obj.GetConditions())".
+func c4g(g Getter) Conditions {
+	return g.GetConditions()
+}
+
+// GetConditions allows the Conditions type to implement the Getter interface.
+func (l Conditions) GetConditions() []metav1.Condition {
+	return l
+}
+
+// Get returns the condition with the given type, otherwise nil is returned.
+func (l Conditions) Get(t string) *metav1.Condition {
+	for _, c := range l {
+		if c.Type == t {
+			return &c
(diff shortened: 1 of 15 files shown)


pkg/conditions/conditions.go
```diff
@@ -0,0 +1,229 @@
...
+		if existingCondition.Type == c.Type {
+			exists = true
+			if !hasSameState(&existingCondition, c) {
+				c.LastTransitionTime = metav1.NewTime(
+					time.Now().UTC().Truncate(time.Second))
+				l[i] = *c
+				break
+			}
+			c.LastTransitionTime = existingCondition.LastTransitionTime
+			break
```
dilyar85, line 126: Is this `break` necessary given it's already end of the for loop?

pkg/conditions/conditions.go
```diff
@@ -0,0 +1,229 @@
...
+	return false
+}
+
+// IsUnknown returns true if the condition with the given type exists and is
+// Unknown, otherwise false is returned.
+func (l Conditions) IsUnknown(t string) bool {
+	if c := l.Get(t); c != nil {
+		return c.Status == metav1.ConditionUnknown
+	}
+	return true
```
dilyar85, line 72: This should be `return false`?

pkg/conditions/conditions.go
```diff
@@ -0,0 +1,229 @@
...
+	// ReadyConditionType indicates a resource is ready.
+	ReadyConditionType = "Ready"
+)
+
+// Conditions is an alias for a slice of metav1.Condition objects and provides
+// helpful functions.
+type Conditions []metav1.Condition
+
+// c4g returns a Conditions type from a Getter. This is used throughout this
+// package to reduce the need to write "Conditions(obj.GetConditions())"".
```
dilyar85, line 23: Nit: extra `"` at the end :)

dilyar85 (approved):
LGTM, thanks @akutz!

pkg/conditions/conditions.go
akutz: Yes, because it breaks out of the loop.

pkg/conditions/conditions.go
akutz: The old function returned `true`, and there's even a test that I did not touch that shows this. Perhaps it *should* return `false`, but changing it here and now could break other things. Will revisit in separate change if required.

PR #854, opened 2025-01-09
🐛 Reconcile VM SetResourcePolicy because of Zone change

bryanv:

When a Zone is added, the VirtualMachineSetResourcePolicies in the namespace need to be reconciled because we need to create the child ResourcePool and ClusterModules for that new CCR.

While at the moment we only care when a Zone is added just watch for all events since we eventually will need to handle Zone removal.

Without this, the Zone won't be ready to use until the next resync time.

Fixes #

```release-note
NONE
```


dougm (approved):
lgtm

PR #858, opened 2025-01-14
🐛  VM Watcher Service improvements

bryanv:

When building the initial container MoID list, include the IDs so that we can match the Add() and Remove() of the same ID. We had a bug here in that if the Zone was initially added to a NS when at start up time, any later removal of the Zone from a NS would remove the watch of that Zone across all namespaces. This allows the watcher Add() and Remove() to be idempotent which is needed for some infra Zone controller described later.

Tweak how the infra Zone controller handles the finalizer. The VC watcher itself is ephemeral since it will go away when the VC client is closed. But we need the Zone finalizer so that we can remove it from the watcher when the Zone goes away. Therefore, call Add() or Remove() on the Zone regardless if the finalizer was already present or not.

Better handle a Zone with a stale FolderMoID: if a folder does not exist on VC, the watcher will fail to start. We filter out invalid container MoIDs to the watcher; the Zone controller will later keep retrying to add that FolderMoID to the watcher.

Much of change is fallout from 14c1ac32b when we changed the VM watcher service to not os.Exit() whenever the watcher could not be started.

Fixes #

```release-note
NONE
```


dougm (approved):
Nice, lgtm

controllers/infra/zone/zone_controller.go
```diff
@@ -123,48 +123,52 @@ func (r *Reconciler) ReconcileDelete(
...
+				Type:  "Folder",
+				Value: val,
+			},
+			fmt.Sprintf("%s/%s", obj.Namespace, obj.Name)); err != nil {
+
+			if !errors.Is(err, watcher.ErrAsyncSignalDisabled) {
+				// We don't ignore watcher.ErrNoWatcher here to interlock with the vm watcher
+				// service that is in the process of restarting the watcher. This does mean
+				// that if watcher cannot start like because of invalid VC creds the finalizer
+				// won't be removed.
```
akutz, line 139: Sure, the finalizer won't be removed, but the zone will get requeued. 

We could have the same discussion about the inability to delete vSphere VMs if their Kube object has a deletion timestamp and the creds are bad.

services/vm-watcher/vm_watcher_service.go
```diff
@@ -98,39 +104,107 @@ func (s Service) Start(ctx context.Context) error {
...
 	for i := range zones.Items {
 		z := zones.Items[i]
-		if v := z.Spec.ManagedVMs.FolderMoID; v != "" {
-			if _, ok := moids[v]; !ok {
-				moids[v] = struct{}{}
-				frefs = append(frefs, vimtypes.ManagedObjectReference{
+
+		v := z.Spec.ManagedVMs.FolderMoID
+		if v == "" {
+			continue
```
akutz, line 136: FWIW, I tend to avoid `continue` when I cannot force it account for coverage.

akutz (approved):
Thanks @bryanv . I also read `zPatch` out loud in my head as if it was French, i.e. "zeeeee patch!" :)

PR #864, opened 2025-01-22
:seedling: Include UUID of disk backing in disk backup

aruneshpa:

This change modifies the disk backup to also include disk UUIDs.  This allows us to look up and identify disks using UUIDs to differentiate classic disks from disks that should be FCDs.

For now, the disk UUIDs are only used in failover workflows since vSphere Replication guarantees that the disks will be restored with the same UUID as the primary/source.

This change is specific for failover of VMs.

```release-note
Include UUID of disk backing in disk backup
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/backup_test.go
+++ b/pkg/providers/vsphere/virtualmachine/backup_test.go
@@ -692,6 +692,7 @@ func backupTests() {
 								FileName:    vcSimDiskFileName,
 								PVCName:     dummyPVC.Name,
 								AccessModes: backupapi.ToPersistentVolumeAccessModes(dummyPVC.Spec.AccessModes),
+								UUID:        vcSimDiskUUID,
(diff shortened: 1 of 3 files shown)


sreyasn (approved):

pkg/providers/vsphere/virtualmachine/backup_test.go
```diff
@@ -689,9 +689,10 @@ func backupTests() {
 
 						diskData := []backupapi.PVCDiskData{
 							{
-								FileName:    vcSimDiskFileName,
+								FileName:    vcSimDiskFileName, //
```
akutz, line 692: Nit, why the empty comment? No reason to re-spin the PR, just curious.

aruneshpa: That was a leftover.  Removed.

PR #866, opened 2025-01-24
🐛 Handle invalid VC Session in GetVirtualMachine()

bryanv:

If the VC session is invalid - like because the VC creds had rotated - GetVirtualMachine() could incorrectly return that the VM does not exist. If the lookup is being done to in response to the k8s VM being deleted, we would not actually delete the VC VM.

While here, improve how looking up by the MoID was done: we don't need to use the Finder instead just use the PropertyCollector. The old Finder lookup would always return an error if the MoID doesn't exist, which means that we could get stuck if the VM actually didn't exist. This would really only be an issue of we removed the VC VM but failed to remove our finalizer.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vcenter/getvm.go
+++ b/pkg/providers/vsphere/vcenter/getvm.go
@@ -5,27 +5,37 @@
 package vcenter
 
 import (
+	"errors"
 	"fmt"
 
-	"github.com/vmware/govmomi/find"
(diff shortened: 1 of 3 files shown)


dougm (approved):
lgtm, thanks @bryanv !

akutz (approved):

PR #870, opened 2025-01-28
🌱 Do not requeue VMIC for empty Spec fields

bryanv:

If these empty Spec fields are updated, the Watch() will cause the controller to reconcile the object so we don't need to do a requeue.

Add NoRequeueError that returns the controller-runtime TerminalError that does not do a requeue but is logged and counted as an error.

While here, update the VMIC ReconcileNormal() to use the logger from the context instead of parent reconciler so the object name is still there.

Fixes #

There are a few other ways to do this:
 - Just have the controller return the TerminalError directly but this would cause "terminal error: " to be prepended to the condition message when I'd prefer to keep it as-is (and not complicate the defer())
 - Have the existing RequeueError with a negative After denote this instead but but I'd rather have a different name so the intention is explicit.  If we wanted to later, we could use a negative After to have ResultFromError() return a ctrl.Result{}, nil for an error that doesn't need to be logged or counted as an error.

Also TerminalError has a confusing name: [it just causes](https://github.com/kubernetes-sigs/controller-runtime/blob/990f2eddd615fa4f8cbee1e176b2b6096f112ac8/pkg/internal/controller/controller.go#L331-L332) the object to not be requeued, but can be later queue due to an event from like a watch.

```release-note
NONE
```

### DIFF
--- a/controllers/virtualmachineimagecache/virtualmachineimagecache_controller.go
+++ b/controllers/virtualmachineimagecache/virtualmachineimagecache_controller.go
@@ -6,7 +6,6 @@ package virtualmachineimagecache
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"path"
 	"reflect"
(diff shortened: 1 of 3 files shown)


akutz (approved):
Good work @bryanv , thanks!

PR #874, opened 2025-01-30
✨ Introduce v1alpha4

akutz:

This patch introduces support for the v1alpha4 VM Operator API.

Fixes `NA`

Please note that at the moment the v1alpha4 APIs are a one-to-one match with v1alpha3. The new version has been created in order to allow for changes to the `API` on `main`.

```release-note
Introduce the v1alpha4 API.
```

----
📚 Documentation preview 📚: https://vm-operator--874.org.readthedocs.build/en/874/


.golangci.yml
```diff
@@ -60,14 +63,24 @@ linters-settings:
...
         pkg: github.com/vmware-tanzu/vm-operator/api/v1alpha3
-      - alias: vmopv1cloudinit
+      - alias: vmopv1a3cloudinit
         pkg: github.com/vmware-tanzu/vm-operator/api/v1alpha3/cloudinit
-      - alias: vmopv1common
+      - alias: vmopv1a3common
         pkg: github.com/vmware-tanzu/vm-operator/api/v1alpha3/common
-      - alias: vmopv1sysprep
+      - alias: vmopv1a3sysprep
         pkg: github.com/vmware-tanzu/vm-operator/api/v1alpha3/sysprep
```
dilyar85, line 73: Seems like these aliases are not enforced as we have the following in the `api/v1alpha3/virtualmachine_bootstrap_types.go` file:
```
import (
	vmopv1cloudinit "github.com/vmware-tanzu/vm-operator/api/v1alpha3/cloudinit"
	vmopv1common "github.com/vmware-tanzu/vm-operator/api/v1alpha3/common"
	vmopv1sysprep "github.com/vmware-tanzu/vm-operator/api/v1alpha3/sysprep"
)
```

akutz: I've noticed in the past that aliases are not always enforced everywhere, and I've never been able to figure out why. My guess is maybe it's because of the API package being a sub-module?

akutz: And I just confirmed it's because it does not cross Go module boundaries. Unfortunately fixing that now would bring in a ton of new issues, so I will follow up on this in a separate commit.

dilyar85 (approved):
LGTM, thanks @akutz!

.golangci.yml
akutz: FWIW, there is no native support for loading partial configs or overwriting configs. The best I found is https://github.com/golangci/golangci-lint/discussions/2001. This matters because we may want rules that are specific to the `api` sub-module and not for the rest of the repo.

PR #875, opened 2025-01-30
🐛 Fix duplicate append of Image.Status fields

bryanv:

Add back the content version check and Disks not empty hack. This will prevent duplicate CL task from being created when the CL item has not changed.

Clear arrays in the Image Status before updating them. This bug had been mostly masked by the short circuit check above but without clearing it the duplicate appends would eventually cause the object to exceed the allowed limit.

Fixes #

```release-note
Restore content library item content version short circuit check.
```


akutz (approved):
Thanks @bryanv !

PR #877, opened 2025-02-04
🐛 Avoid needlessly VC client logout when creds haven't changed

bryanv:

Only logout an existing VC client when the creds have actually changed. Previously, when the Infra Secret controller first starts it would have always logged the VC client if present. While this race has been present for a long time, the addition of the VM Watcher Service made this more likely now since that will usually create a VC client before the Infra secret has started.

Move the VC creds secret into a config param. Previously, the Infra Secret controller was hardcoding the secret name, while we'd otherwise obtain the name from the VMOP ConfigMap. This secret name hasn't changed since day 1 so just move it into our config.

In the fake VM provider, add gets of the config from the context since we now expect and require the config to be there.

Fixes #

Effectively deprecating the VcCredsSecretName is what @akutz and I settled on for this

```release-note
VcCredsSecretName item in the VMOP ConfigMap is no longer referenced
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider.go
+++ b/pkg/providers/vsphere/vmprovider.go
@@ -39,6 +39,7 @@ import (
 	vcconfig "github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/config"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/constants"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/contentlibrary"
+	vccreds "github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/credentials"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/vcenter"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
 	"github.com/vmware-tanzu/vm-operator/pkg/topology"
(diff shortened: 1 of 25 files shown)


pkg/providers/vsphere/vmprovider.go
```diff
@@ -133,25 +134,51 @@ func (vs *vSphereVMProvider) UpdateVcPNID(ctx context.Context, vcPNID, vcPort st
...
-	vcClient := vs.vcClient
-	vs.vcClient = nil
-	vs.vcClientLock.Unlock()
+	oldVcClient := func() *vcclient.Client {
+		vs.vcClientLock.Lock()
+		defer vs.vcClientLock.Unlock()
+
+		if vcClient := vs.vcClient; vcClient != nil && vcClient.Config().VcCreds != newVcCreds {
+			// Clear and logout existing clean so new client will be created next call to getVcClient().
+			fmt.Printf("### BOOM\n")
```
dougm, line 169: Remove or use a logger instead?

dougm (approved):
lgtm, just noticed a debug line got left behind

pkg/providers/vsphere/vmprovider.go
bryanv: Whoops. Thanks.

akutz (approved):

PR #878, opened 2025-02-04
🐛 Fix duplicate append of Image.Status fields

bryanv:

Clear arrays in the Image Status before updating them. This bug had been mostly masked by the short circuit check above but without clearing it the duplicate appends would eventually cause the object to exceed the allowed limit.

Partial cherry-pick of bb1c8810 since this branch still has the content version check. In practice, that only means an image would end up with duplicated items after version upgrade when we need to populate in the disks field.

Fixes #

```release-note
NONE
```


akutz (approved):
LGTM, thank you!

PR #880, opened 2025-02-04
🌱 Remove IsVirtualMachineSetResourcePolicyReady()

bryanv:

Long ago, this method was called from the VirtualMachine controller and used to determine if we could create the (TKG) VM. Now, what constituted the RP ready checks are done individually during the VM creation flow, leaving this method only being used for test assertions.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_resourcepolicy_test.go
+++ b/pkg/providers/vsphere/vmprovider_resourcepolicy_test.go
@@ -6,18 +6,18 @@ package vsphere_test
 
 import (
 	"fmt"
-	"path"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vapi/cluster"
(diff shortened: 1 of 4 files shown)


pkg/providers/vsphere/vmprovider_resourcepolicy_test.go
```diff
@@ -99,58 +156,34 @@ func resourcePolicyTests() {
 			)
 
 			JustBeforeEach(func() {
-				testPolicyName := "test-policy"
-
-				resourcePolicy = getVirtualMachineSetResourcePolicy(testPolicyName, nsInfo.Namespace)
+				resourcePolicy = getVirtualMachineSetResourcePolicy(" test-policy", nsInfo.Namespace)
```
dougm, line 159: Is leading space here intentional?

dougm (approved):
lgtm

pkg/providers/vsphere/vmprovider_resourcepolicy_test.go
bryanv: Sigh. Harmless but no.

PR #885, opened 2025-02-06
⚠️ Remove UpdateConfigSpecFirmware()

bryanv:

This is from the early days when we could only update the VM's Firmware - and most things for that matter - after the VM was created since Content Library did not support creating from a ConfigSpec. Since now CL supports that, and some time back I added support in CreateConfigSpec() for the firmware annotation, so the VM is just created with the specified firmware.

Changing the firmware of existing VM is generally not going to result in a booting VM, so at the moment I didn't implement similar support in resize. My preference would be to remove this.

Fixes #

I want to start to reworking the entire VM create/update unit tests so they can be run with resize enabled and disabled. This is one the deltas. It is easy enough to add this as a dev ops resize overwrite if we want to keep things the same instead.

```release-note
If set, the VM vmoperator.vmware.com/firmware annotation is only used to create the VM.
```

### DIFF
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -438,18 +438,6 @@ func UpdateConfigSpecGuestID(
 	}
 }
 
-func UpdateConfigSpecFirmware(
-	config *vimtypes.VirtualMachineConfigInfo,
-	configSpec *vimtypes.VirtualMachineConfigSpec,
-	vm *vmopv1.VirtualMachine) {
-
-	if val, ok := vm.Annotations[constants.FirmwareOverrideAnnotation]; ok {
-		if (val == "efi" || val == "bios") && config.Firmware != val {
-			configSpec.Firmware = val
-		}
-	}
-}
-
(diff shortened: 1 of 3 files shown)


dilyar85 (approved):

aruneshpa (approved):

akutz:
~~@bryanv We support specifying the firmware in the spec anyway, right?~~

Huh, we do not. I know I wanted to add it. It will likely come in via the parity DevOps settings, and that may involve allowing a user to reconfigure the firmware if they want. We'd certainly need to allow it during a resize. Therefore we will need to consider firmware during update eventually, but since it would be different enough than the current code I see no reason not to proceed with this PR as-is today.

akutz (approved):
~~@bryanv We support specifying the firmware in the spec anyway, right?~~

Huh, we do not. I know I wanted to add it. It will likely come in via the parity DevOps settings, and that may involve allowing a user to reconfigure the firmware if they want. We'd certainly need to allow it during a resize. Therefore we will need to consider firmware during update eventually, but since it would be different enough than the current code I see no reason not to proceed with this PR as-is today.

PR #896, opened 2025-02-18
🌱 Only reconcile connected VMs

akutz:

This patch ensures only connected VMs are reconciled. If a VM is not in a connected state, it will not be requeued for reconciliation. Instead the watcher service will enqueue the VM for reconciliation once the VM's connection state has changed.

Fixes `NA`

```release-note
Reconcile VMs with connectionState == connected or reconcile when it changes to connected
```


aruneshpa (approved):
LGTM

PR #902, opened 2025-02-20
🌱 Defaults for ConfigSpec CPU/mem alloc

akutz:

This patch ensures there are defaults for the ConfigSpec CPU/mem allocation fields for limits and shares.

The documentation for a ResourceAllocation object states that all (VM) fields must be set for certain APIs, including ImportVApp. A part of that API is used by the vpxd placement engine to construct fake VMs in order to support assignable hardware. Therefore:

CPU allocation:
- Ensure limit is set to -1 if there is a reservation.

Memory allocation:
- Ensure shares is not nil.
- Ensure limit is set to -1 if there is a reservation.

Fixes `NA`

```release-note
Defaults for CPU/memory allocations in ConfigSpec to fix bug in placement engine
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/configspec.go
+++ b/pkg/providers/vsphere/virtualmachine/configspec.go
@@ -156,6 +156,33 @@ func CreateConfigSpec(
 		}
 	}
 
+	// The documentation for a ResourceAllocation object states that all (VM)
+	// fields must be set for certain APIs, including ImportVApp. A part of
+	// that API is used by the vpxd placement engine to construct fake VMs in
+	// order to support assignable hardware. Therefore:
+	//
+	// CPU allocation:
+	// - Ensure limit is set to -1 if there is a reservation.
+	//
+	// Memory allocation:
+	// - Ensure shares is not nil.
+	// - Ensure limit is set to -1 if there is a reservation.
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -156,6 +156,33 @@ func CreateConfigSpec(
...
+	// that API is used by the vpxd placement engine to construct fake VMs in
+	// order to support assignable hardware. Therefore:
+	//
+	// CPU allocation:
+	// - Ensure limit is set to -1 if there is a reservation.
+	//
+	// Memory allocation:
+	// - Ensure shares is not nil.
+	// - Ensure limit is set to -1 if there is a reservation.
+	if a := configSpec.CpuAllocation; a != nil {
```
bryanv, line 170: Does it make more sense to have these as a final `} else {` clause in the CPU or memory blocks above, since we're otherwise kind of spreading our defaulting logic out

bryanv (approved):
Left one comment but otherwise gtg

pkg/providers/vsphere/virtualmachine/configspec.go
akutz: This way it handles it no matter how they started. That was intentional.

PR #903, opened 2025-02-21
✨ virtualmachinesetresourcepolicy: allow mutability of .spec.clusterModuleGroups

chrischdi:

* Allows modification of VirtualMachinesetResourcePolicy.spec.clusterModuleGroups
* Adjust the reconciler for VirtualMachinesetResourcePolicy to create or delete ClusterModuleGroups on changes to `.spec.clusterModuleGroups`.

Fixes #

I implemented some changes for CAPV and did take a quick look into this for vm-operator and seemed not to be too hard so I went ahead and implemented this feature.

```release-note
Allow mutability of ClusterModuleGroups in VirtualMachinesetResourcePolicy
```


akutz:
It looks like you did not include the change to the actual validation webhook?

akutz:
@chrischdi I think you will need to also update the webhook's unit/integ tests.

pkg/providers/vsphere/vmprovider_resourcepolicy.go
```diff
@@ -58,7 +59,7 @@ func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachineSetResourcePolicy(
 			clusterRef, err := vcenter.GetResourcePoolOwnerMoRef(ctx, vimClient, rpMoID)
 			if err == nil {
 				clusterModuleProvider := clustermodules.NewProvider(client.RestClient())
-				err = vs.createClusterModules(ctx, clusterModuleProvider, clusterRef.Reference(), resourcePolicy)
+				err = vs.reconcileClusterModules(ctx, clusterModuleProvider, clusterRef.Reference(), resourcePolicy)
```
bryanv, line 62: This will miss the case when all the cluster modules are removed, no?

pkg/providers/vsphere/vmprovider_resourcepolicy.go
```diff
@@ -190,6 +191,24 @@ func (vs *vSphereVMProvider) createClusterModules(
 		}
 	}
 
+	desiredGroups := sets.New(resourcePolicy.Spec.ClusterModuleGroups...)
```
bryanv, line 194: This isn't the right place for this - this function is handle the cluster modules within just a CCR so this is duplicate work.

One existing issue with this code that I was planning to fix under this work is that when a zone is removed from a NS the cluster modules for that CCR should also be removed. Right now that won't happen until the VMSetRP is deleted.

pkg/providers/vsphere/vmprovider_resourcepolicy_test.go
```diff
@@ -178,6 +178,22 @@ func resourcePolicyTests() {
 					Expect(resourcePolicy.Status.ClusterModules).To(HaveExactElements(status.ClusterModules))
 					assertSetResourcePolicy(resourcePolicy, true)
 				})
+
+				It("should create additional cluster modules", func() {
+					assertSetResourcePolicy(resourcePolicy, true)
+					resourcePolicy.Spec.ClusterModuleGroups = append(resourcePolicy.Spec.ClusterModuleGroups, "another-NodeGroup2")
+					Expect(vmProvider.CreateOrUpdateVirtualMachineSetResourcePolicy(ctx, resourcePolicy)).To(Succeed())
+					Expect(resourcePolicy.Status.ClusterModules).To(HaveLen(len(resourcePolicy.Spec.ClusterModuleGroups) * len(ctx.ZoneNames)))
```
bryanv, line 186: Doesn't assertSetResourcePolicy() already assert this (it also takes into account for when ClustersPerZone > 1)?

pkg/providers/vsphere/vmprovider_resourcepolicy_test.go
```diff
@@ -178,6 +178,22 @@ func resourcePolicyTests() {
...
+					assertSetResourcePolicy(resourcePolicy, true)
+				})
+
+				It("should delete removed cluster modules", func() {
+					assertSetResourcePolicy(resourcePolicy, true)
+					resourcePolicy.Spec.ClusterModuleGroups = resourcePolicy.Spec.ClusterModuleGroups[:1]
+					Expect(vmProvider.CreateOrUpdateVirtualMachineSetResourcePolicy(ctx, resourcePolicy)).To(Succeed())
+					Expect(resourcePolicy.Status.ClusterModules).To(HaveLen(len(ctx.ZoneNames)))
```
bryanv, line 194: Again, doesn't assertSetResourcePolicy() already assert this?

I'd prefer for this test we also preserve the cluster modules UUIDs that we expect to be deleted and assert that they DNE after the second CreateOrUpdateVirtualMachineSetResourcePolicy() call.

chrischdi:
Thanks for the prompt responses, I hope this does not make too much noise to you folks. I had it around and thought it might be worth to not double the work / help you out :-) .

I will take a look at the tests and the comments next week.

pkg/providers/vsphere/vmprovider_resourcepolicy.go
chrischdi: Refactored, removed the if and also moved it out of this loop (see below)

pkg/providers/vsphere/vmprovider_resourcepolicy.go
chrischdi: Ah got it, I agree.

Refactored the code to handle compute cluster resources in its separate loop inside the reconcile function itself.

pkg/providers/vsphere/vmprovider_resourcepolicy_test.go
chrischdi: Yeah, you are right. I was thinking about this to make sure we see the creation, but this already happens in that assert function.

chrischdi:
I tried to address the findings :-)

chrischdi:
Please don't hestitate let me know if you currently have no time for this and want to follow up yourself on it. I'm also okay to help bring this over the finish line :-)

pkg/providers/vsphere/vmprovider_resourcepolicy.go
```diff
@@ -54,16 +56,18 @@ func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachineSetResourcePolicy(
...
-			if err == nil {
-				clusterModuleProvider := clustermodules.NewProvider(client.RestClient())
-				err = vs.createClusterModules(ctx, clusterModuleProvider, clusterRef.Reference(), resourcePolicy)
-			}
-			if err != nil {
-				errs = append(errs, err)
-			}
+		clusterRef, err := vcenter.GetResourcePoolOwnerMoRef(ctx, vimClient, rpMoID)
+		if err != nil {
+			errs = append(errs, err)
```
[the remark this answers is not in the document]
chrischdi: That's right, if
* there's an error getting the resource pool itself
* or the properties collector failed to get the owner property of the resource pool

The clusterRef may be missing.

I'll adjust to in that case don't do deletion based on clusterrefs.
Alternatively we could skip reconcilng clustermodules if we had errors.

pkg/providers/vsphere/vmprovider_resourcepolicy_unit_test.go
```diff
@@ -0,0 +1,243 @@
...
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+)
+
+func Test_vSphereVMProvider_reconcileClusterModules(t *testing.T) {
+	ctx := context.Background()
+
+	scheme := runtime.NewScheme()
+	_ = vmopv1.AddToScheme(scheme)
+
+	tests := []struct {
```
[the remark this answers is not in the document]
chrischdi: Integrated the cases into the existing integration tests.

With the suite its not possible to directly test `reconcileClusterModules` because its a private function and the integration tests have a different package name, which is because otherwise there'd be import cycles.

pkg/providers/vsphere/vmprovider_resourcepolicy_unit_test.go
```diff
@@ -0,0 +1,243 @@
...
+			}
+
+			g.Expect(vs.reconcileClusterModules(ctx, clusterModProvider, tt.computeClusterReferences, tt.resourcePolicy)).ToNot(HaveOccurred())
+			g.Expect(tt.resourcePolicy.Status.ClusterModules).To(BeEquivalentTo(tt.wantClusterModuleStatus))
+			g.Expect(clusterModProvider.modules).To(HaveLen(len(tt.resourcePolicy.Status.ClusterModules)))
+		})
+	}
+}
+
+type fakeClusterModule struct {
```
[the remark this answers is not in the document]
chrischdi: see below, tests got obsolete because I integrated the missing cases into the integration tests.

chrischdi:
/close

Will not have bandwidth to come back to this (if this even makes sense)

PR #908, opened 2025-02-27
🌱 Fix expected error message in VM network validation

bryanv:

This message was split over multiple lines. This might have been a left over from when doValidateWithMsg() was splitting the validation response by "," (since that is what is used to join the individual failure reasons) but that falls apart for messages that have "," in them so we just match by substring instead.

Fixes #

```release-note
NONE
```


aruneshpa (approved):
Pretty sure I hit this at some point :(  Thanks for fixing this.

PR #912, opened 2025-03-04
✨ Support PVCs with WFFC StorageClass

bryanv:

For PVCs that are using a WaitForFirstConsumer StorageClass, set the PVC selected-node annotation to one of the Node names that is in the same Zone as the VM (any Node is sufficient).

Fixes #

```release-note
For VM Volumes that reference a PVC with a WaitForFirstConsumer StorageClass, set the volume.kubernetes.io/selected-node annotation to that of a Node in the VM's Zone.
```

### DIFF
--- a/controllers/virtualmachine/volume/volume_controller.go
+++ b/controllers/virtualmachine/volume/volume_controller.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apierrorsutil "k8s.io/apimachinery/pkg/util/errors"
@@ -39,6 +40,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/constants"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
+	"github.com/vmware-tanzu/vm-operator/pkg/topology"
 	"github.com/vmware-tanzu/vm-operator/pkg/util"
 	vmopv1util "github.com/vmware-tanzu/vm-operator/pkg/util/vmopv1"
 )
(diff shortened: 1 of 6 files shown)


dougm (approved):
lgtm

controllers/virtualmachine/volume/volume_controller.go
```diff
@@ -785,6 +793,110 @@ func (r *Reconciler) createCNSAttachment(
...
+		// trying to resolve that here now and just continue the prior flow.
+		return nil
+	}
+
+	sc := &storagev1.StorageClass{}
+	if err := r.Get(ctx, client.ObjectKey{Name: *scName}, sc); err != nil {
+		return fmt.Errorf("cannot get StorageClass for PVC %s: %w", pvc.Name, err)
+	}
+
+	if mode := sc.VolumeBindingMode; mode == nil || *mode != storagev1.VolumeBindingWaitForFirstConsumer {
```
aruneshpa, line 841: Any value in using this: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-helpers/storage/volume/pv_helpers.go#L96?  Or do you think this will increase the dep graph significantly?

controllers/virtualmachine/volume/volume_controller.go
```diff
@@ -785,6 +793,110 @@ func (r *Reconciler) createCNSAttachment(
...
+		nodes,
+		client.MatchingLabels{topology.KubernetesTopologyZoneLabelKey: zoneName}); err != nil {
+		return "", err
+	}
+
+	if len(nodes.Items) == 0 {
+		return "", fmt.Errorf("no Nodes for zone %s", zoneName)
+	}
+
+	for i := range nodes.Items {
```
aruneshpa, line 890: Out of curiosity, any reason why you are using iterators instead of a range scoped local var?

controllers/virtualmachine/volume/volume_controller.go
```diff
@@ -785,6 +793,110 @@ func (r *Reconciler) createCNSAttachment(
...
+		if nodes.Items[i].DeletionTimestamp.IsZero() {
+			return nodes.Items[i].Name, nil
+		}
+	}
+
+	// Otherwise just return the first one.
+	return nodes.Items[0].Name, nil
```
aruneshpa, line 897: Is there a point at all of returning a node that is being deleted?  Can a PVC be even bound to an object on such a node?

aruneshpa (approved):
LGTM, with a few questions.

controllers/virtualmachine/volume/volume_controller.go
bryanv: It really doesn't matter either way here

controllers/virtualmachine/volume/volume_controller.go
bryanv: Not really. We'd have to go implement another interface too

controllers/virtualmachine/volume/volume_controller.go
bryanv: Nothing is being done to the specific node here - we just need a node that's in the zone. Might as well try avoid nodes that might go away before CSI processes the PVC but this specifically is an not expected edge case since the VM exists.

PR #913, opened 2025-03-04
:bug: Do not patch a zone with finalizer if it is being deleted

aruneshpa:

API server rejects Patch requests when the zone controller is attempting to add a finalizer to the zone, causing the watcher to never start.  And since the watcher is the entity removing the finalizer, the zone stays stuck in deletion forever.  This change skips a zone from being patched if the zone is being deleted.

Additionally, do not watch folders backing a zone if the zone is marked for deletion and the zone does not contain the finalizer.  This can happen in situations where the zone controller has already removed the finalizer after stopping the watcher, but the zone is somehow not removed.

```release-note
Do not patch a zone with finalizer if it is being deleted
```

### DIFF
--- a/services/vm-watcher/vm_watcher_service.go
+++ b/services/vm-watcher/vm_watcher_service.go
@@ -132,6 +132,17 @@ func (s Service) vmFolderMoRefWithIDs(
 		z := zones.Items[i]
 
 		if v := z.Spec.ManagedVMs.FolderMoID; v != "" {
+
+			// If a zone is being deleted and it does not have any
+			// finalizer, the zone controller has already stopped the
+			// watcher and removed the finalizer. Or, a zone is being
+			// deleted even before a watcher could start. In either
+			// case, no need to watch this folder.
+			if !z.DeletionTimestamp.IsZero() &&
+				!controllerutil.ContainsFinalizer(&z, zonectrl.Finalizer) {
+				continue
+			}
+
 			if _, ok := moids[v]; !ok {
 				moids[v] = make([]string, 0, 1)
 
@@ -144,6 +155,7 @@ func (s Service) vmFolderMoRefWithIDs(
 			}
 			moids[v] = append(moids[v], fmt.Sprintf("%s/%s", z.Namespace, z.Name))
 
+			// Ensure we add a finalizer to zones that the watcher is watching.
 			if !controllerutil.ContainsFinalizer(&z, zonectrl.Finalizer) {
 				zonesWithoutFinalizer = append(zonesWithoutFinalizer, z)
 			}
--- a/services/vm-watcher/vm_watcher_service_test.go
+++ b/services/vm-watcher/vm_watcher_service_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/vmware/govmomi/object"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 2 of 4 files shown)


dougm (approved):
lgtm, thanks @aruneshpa

services/vm-watcher/vm_watcher_service.go
```diff
@@ -132,6 +132,17 @@ func (s Service) vmFolderMoRefWithIDs(
...
 		if v := z.Spec.ManagedVMs.FolderMoID; v != "" {
+
+			// If a zone is being deleted and it does not have any
+			// finalizer, the zone controller has already stopped the
+			// watcher and removed the finalizer. Or, a zone is being
+			// deleted even before a watcher could start. In either
+			// case, no need to watch this folder.
+			if !z.DeletionTimestamp.IsZero() &&
+				!controllerutil.ContainsFinalizer(&z, zonectrl.Finalizer) {
+				continue
```
akutz, line 143: Same nit I give to others, by avoiding the use of continue you can avoid having to produce coverage that can be difficult to cause on purpose. In this case though I think you are able to do it without much effort.

akutz (approved):
Lgtm, thank you.

services/vm-watcher/vm_watcher_service_test.go
```diff
@@ -356,6 +356,64 @@ var _ = Describe(
...
+								},
+								Spec: topologyv1.ZoneSpec{},
+							}
+							Expect(ctx.Client.Create(ctx, zone)).To(Succeed())
+							// Delete the Zone so the deletion timestamp is set.
+							Expect(ctx.Client.Delete(ctx, zone)).To(Succeed())
+						}
+					})
+
+					When("the zone has the zone finalizer", func() {
```
dilyar85, line 382: By `zone finalizer` do you mean the specific `zonectrl.Finalizer`? If so, I don't see it's being added anywhere in this block?

services/vm-watcher/vm_watcher_service_test.go
```diff
@@ -356,6 +356,64 @@ var _ = Describe(
...
+									ObjectMeta: metav1.ObjectMeta{
+										Namespace: vcSimCtx.NSInfo.Namespace,
+										Name:      vmName,
+									},
+								},
+							})))
+						})
+					})
+
+					When("the zone does not have the finalizer", func() {
```
dilyar85, line 397: I think we should also clarify which "finalizer" here. If it's just a general one, then this is not the case since the zone has a fake finalizer added in parent's BeforeEach().

services/vm-watcher/vm_watcher_service.go
```diff
@@ -144,7 +155,10 @@ func (s Service) vmFolderMoRefWithIDs(
 			}
 			moids[v] = append(moids[v], fmt.Sprintf("%s/%s", z.Namespace, z.Name))
 
-			if !controllerutil.ContainsFinalizer(&z, zonectrl.Finalizer) {
+			// API server prevents adding finalizers to resources that
+			// are in Deleting state, so skip such zones.
+			if z.DeletionTimestamp.IsZero() &&
```
bryanv, line 160: This second `IsZero()` check shouldn't be needed, right? With out it makes it clearer that for us to watch something - that is, its in the `moids` maps - the finalizer must be present.

aruneshpa: For posterity: without this, we are saying that regardless of whether the zone is in deleting or not, we will add the finalizer if it is not present.
If the zone is deleting, and it doesn't have the finalizer, we would have returned from L143.  And if it is not deleting, we want to add the finalizer.

So yeah, this seems redundant.  Removing.

services/vm-watcher/vm_watcher_service_test.go
aruneshpa: Fixed this.  Can you PTAL?

services/vm-watcher/vm_watcher_service.go
```diff
@@ -144,6 +155,8 @@ func (s Service) vmFolderMoRefWithIDs(
 			}
 			moids[v] = append(moids[v], fmt.Sprintf("%s/%s", z.Namespace, z.Name))
 
+			// API server prevents adding finalizers to resources that
+			// are in Deleting state, so skip such zones.
```
bryanv, line 159: This comment seems a little out of place - that check & skip is done up above. Here, we must make sure that the finalizer is there for any Zones that are apart of the `moids` map

services/vm-watcher/vm_watcher_service_test.go
```diff
@@ -356,6 +357,178 @@ var _ = Describe(
...
+							By("verify that the zone has the finalizer added", func() {
+								Eventually(func(g Gomega) {
+									zoneList := topologyv1.ZoneList{}
+									err := vcSimCtx.Client.List(vcSimCtx, &zoneList, ctrlclient.InNamespace(vcSimCtx.NSInfo.Namespace))
+									g.Expect(err).NotTo(HaveOccurred())
+									g.Expect(zoneList.Items).ToNot(BeEmpty())
+									for _, zone := range zoneList.Items {
+										g.Expect(zone.Finalizers).To(ContainElement(zonectrl.Finalizer))
+									}
+								})
```
dilyar85, line 372: Missing `.Should(Succeed())` at the end?

services/vm-watcher/vm_watcher_service_test.go
```diff
@@ -356,6 +357,178 @@ var _ = Describe(
...
+								chanSource := cource.FromContext(ctx, "VirtualMachine")
+								var e event.GenericEvent
+								Eventually(chanSource).Should(Receive(&e, Equal(event.GenericEvent{
+									Object: &vmopv1.VirtualMachine{
+										ObjectMeta: metav1.ObjectMeta{
+											Namespace: vcSimCtx.NSInfo.Namespace,
+											Name:      vmName,
+										},
+									},
+								})))
```
dilyar85, line 385: Ditto.

services/vm-watcher/vm_watcher_service_test.go
```diff
@@ -356,6 +357,178 @@ var _ = Describe(
...
```
dilyar85, line 439: Ditto.

bryanv (approved):

dilyar85 (approved):

services/vm-watcher/vm_watcher_service_test.go
aruneshpa: This is `Eventually().Should(Receive())`

services/vm-watcher/vm_watcher_service_test.go
aruneshpa: This is `Eventually().Should(Receive())`

services/vm-watcher/vm_watcher_service_test.go
aruneshpa: Done.

PR #916, opened 2025-03-07
🐛 Handle resize for VMs that were classless

bryanv:

For a classless VM, we'll store an empty class name in the LastResizedAnnotation value. Update ResizeNeeded() to instead compare that empty class name with the existing - now not empty class name - so it will now return that the VM needs a resize.

Fixes #

```release-note
Handle resize of VMs that were originally classless.
```


aruneshpa (approved):
LGTM.

PR #917, opened 2025-03-10
🐛 (x-port vc-9.0.0) Do not patch a zone with finalizer if it is being deleted

aruneshpa:

API server rejects Patch requests when the zone controller is attempting to add a finalizer to the zone, causing the watcher to never start.  And since the watcher is the entity removing the finalizer, the zone stays stuck in deletion forever.  This change skips a zone from being patched if the zone is being deleted.

Additionally, do not watch folders backing a zone if the zone is marked for deletion and the zone does not contain the finalizer.  This can happen in situations where the zone controller has already removed the finalizer after stopping the watcher, but the zone is somehow not removed.

This is a crossport of: #913 

```release-note
Do not patch a zone with finalizer if it is being deleted
```

### DIFF
--- a/pkg/util/vsphere/watcher/watcher.go
+++ b/pkg/util/vsphere/watcher/watcher.go
@@ -128,6 +128,8 @@ type Watcher struct {
 
 	ignoredExtraConfigKeys map[string]struct{}
 	lookupNamespacedName   lookupNamespacedNameFn
+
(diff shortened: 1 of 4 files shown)


akutz (approved):

dilyar85 (approved):
LGTM!

PR #918, opened 2025-03-13
🌱 Golang 1.24.1 / Golangci-lint 1.64.7

akutz:
    
This patch updates VM Operator to use Golang 1.24.1 and Golangci-lint 1.64.7 (as it supports Golang 1.24.1).
    
This patch also addresses several new linter errors:
    
* Loop variables no longer need to be captured due to Go 1.22
* There are a massive number of false positives regarding int overflows. See https://github.com/securego/gosec/issues/1212 for more information. The linter directive G115 is used surgically to disable the false-positives but also places where we know there is not a chance for an overflow.
    
    There is also a good blog on the issue at https://dev.to/ccoveille/about-the-gosec-g115-drama-or-how-i-faced-back-integer-conversion-overflow-in-go-1302.

Fixes `NA`

```release-note
Golang 1.24.1 and Golangci-lint 1.64.7
```


dilyar85 (approved):
Looks like need to run `go -C hack/tools mod tidy` to update that sub module.

dilyar85 (approved):
LGTM, thanks @akutz!

PR #919, opened 2025-03-25
🐛 Do not list all StoragePolicyQuotas during SPQ reconcile

bryanv:

When reconciling a SQP, we would later end up listing all the SQP in the namespace to determine what storage classes are associated with the policy quota. We don't need to list all the SQPs since each one will be reconciled.

While we can't - or should not - have the same storage policy ID for multiple SQP in a namespace, but if we could this cause a failure on the reconcile of later SQPs since we'd try to set multiple controller owner references.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/storagepolicyquota/storagepolicyquota_controller.go
+++ b/controllers/storagepolicyquota/storagepolicyquota_controller.go
@@ -126,12 +126,11 @@ func (r *Reconciler) ReconcileNormal(
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


akutz (approved):
Thanks @bryanv !

bryanv:
Superseded by https://github.com/vmware-tanzu/vm-operator/pull/920

PR #921, opened 2025-03-26
WiP: 🐛 Do not watch child folders

akutz:

This patch changes the behavior of the async watcher service to no longer watch child folders under the Namespace's folder. Otherwise PodVMs are caught in the mix.

To ensure VM Op still watches the folders created for VKS clusters, these folders are explicitly added to the watcher.

Fixes `NA`

This is a WiP that needs tests. It's also unclear if I will have time to take it over the finish line or if this is even the correct solution. I'm opening this as a WiP PR primarily so folks can more easily add comments and thoughts.

```release-note
Do not watch child folders in async watcher.
```


dougm (approved):
lgtm

PR #922, opened 2025-03-28
🌱 Fix new go vet non const printf warnings

bryanv:

Since conditions.MarkFalse(..., err.Error()) is so common add MarkError() to simplify that pattern.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/conditions/setter.go
+++ b/pkg/conditions/setter.go
@@ -92,6 +92,11 @@ func MarkFalse(to Setter, t string, reason string, messageFormat string, message
 	to.SetConditions(c4g(to).MarkFalse(t, reason, messageFormat, messageArgs...))
 }
 
+// MarkError sets Status=False and the error message for the condition with the given type.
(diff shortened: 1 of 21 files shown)


pkg/conditions/setter.go
```diff
@@ -92,6 +92,11 @@ func MarkFalse(to Setter, t string, reason string, messageFormat string, message
 	to.SetConditions(c4g(to).MarkFalse(t, reason, messageFormat, messageArgs...))
 }
 
+// MarkFalseErr sets Status=False for the condition with the given type.
+func MarkFalseErr(to Setter, t string, reason string, err error) {
```
dougm, line 96: Could just name it `MarkError` ?

dougm (approved):
lgtm

pkg/conditions/setter.go
akutz: Agreed. I don't think we will ever have `MarkTrueError` or `MarkUnknownError`.

PR #924, opened 2025-04-01
 🌱 Create StoragePolicyQuota in vcsim test env

bryanv:

Only create the old ResourceQuota when the relevant FFS is not enabled to closer match the real env in case we want to start enforcing the SC namespace association during reconcile.

Fixes #

```release-note
NONE
```


dilyar85 (approved):
LGTM, thanks @bryanv!

PR #936, opened 2025-04-11
🐛 Require VM ResourcePolicyName when specifying a cluster module

bryanv:

The VM SetResourcePolicy contains the cluster module info, so if the VM has the special cluster module annotation, enforce that the resource policy is set.  Post create, we only need the VM's resource policy to set the VM cluster module. We really only need if it is not already set but that is not easy to check as this point.

Fixes #

The only user of this is well behaved but otherwise this would have caused a panic.

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -37,6 +37,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/api/v1alpha4/common"
 	pkgcnd "github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	pkgconst "github.com/vmware-tanzu/vm-operator/pkg/constants"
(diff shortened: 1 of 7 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1786,9 +1787,16 @@ func (vs *vSphereVMProvider) vmUpdateGetArgs(
 		return nil, err
 	}
 
-	resourcePolicy, err := GetVMSetResourcePolicy(vmCtx, vs.k8sClient)
-	if err != nil {
-		return nil, err
+	var resourcePolicy *vmopv1.VirtualMachineSetResourcePolicy
+	if vmCtx.VM.Annotations[pkg.ClusterModuleNameKey] != "" {
```
dilyar85, line 1791: Can we move `ClusterModuleNameKey` into `pkg/constants/constants.go` to avoid importing the `pkg` package here and also the webhook code below?

dilyar85 (approved):

PR #938, opened 2025-04-14
 🌱 Set VM as controller owner ref on network interfaces

bryanv:

This was an oversight from the original code. Do not attempt to use an existing network interface that is currently controller owned by another VM. For any existing situations where the interface has an controller owner that is not a VM add just an owner ref like before to keep things working.

Fixes #

```release-note
Set the VirtualMachine as the controller owner ref on the network interface custom resources.
```

### DIFF
--- a/pkg/providers/vsphere/network/network.go
+++ b/pkg/providers/vsphere/network/network.go
@@ -7,6 +7,7 @@ package network
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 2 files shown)


akutz:
If we are doing this in CreateOrPatch then this will set up ownership of external resources that could be managed independently of the VM, right? We should avoid that so these resources are not deleted along with the VM. VM Op should only own the resources it creates.

bryanv:
> If we are doing this in CreateOrPatch then this will set up ownership of external resources that could be managed independently of the VM, right? We should avoid that so these resources are not deleted along with the VM. VM Op should only own the resources it creates.

Yes, but I don't view that we have really have external resources here until we add reference field. It is easy to to do something a little different based on the RV or whatever though. It is unfortunate this wasn't using controller references from the start, and I want to avoid reusing a not yet GC interface when quickly recreating a VM.

akutz (approved):
Approved with a minor nit.

PR #942, opened 2025-04-23
🌱 Use OpaqueNetwork backing for NSX for VM create

bryanv:

For NSXT/VPC we determine the DVPG from the LogicalSwitchUID until we know the CCR, so we had some code that would go fix up the ConfigSpec post placement but before the create. Instead, we can just use the OpaqueNetwork backing with the NSX type and LSUUID to skip that.

The backing gets swizzled to the actual DVPG so we still need to be aware of that to later compare the VM's desired and current state.

Fixes #

```release-note
NONE
```


dilyar85 (approved):
LGTM

PR #944, opened 2025-04-25
🌱 Allow 'default' for routes To value

bryanv:

Fixes #

```release-note
VirtualMachine Spec.Network.Interfaces[].Routes[].To value to be set to "default" to specify the default route
```

### DIFF
--- a/api/v1alpha2/virtualmachine_network_types.go
+++ b/api/v1alpha2/virtualmachine_network_types.go
@@ -12,7 +12,7 @@ import (
 
 // VirtualMachineNetworkRouteSpec defines a static route for a guest.
 type VirtualMachineNetworkRouteSpec struct {
-	// To is an IP4 or IP6 address.
+	// To is either "default", or an IP4 or IP6 address.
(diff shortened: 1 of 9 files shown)


api/v1alpha2/virtualmachine_network_types.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // VirtualMachineNetworkRouteSpec defines a static route for a guest.
 type VirtualMachineNetworkRouteSpec struct {
-	// To is an IP4 or IP6 address.
+	// To is either "default", or an IP4 or IP6 address.
```
dilyar85, line 15: Can we use an empty value as the default instead, so users don't have to explicitly set "default" here? IOW, webhook only calls `net.ParseCIDR(r.To)` if the field is not empty.

bryanv: I think that makes the output look too incomplete and confusing to see just a `via` field

akutz (approved):

PR #946, opened 2025-04-29
✨ Support poweron.check and delete.check annotations

akutz:

This patch adds support for the following annotations:

* `poweron.check.vmoperator.vmware.com`

    This annotation prevents a VM from being powered on until the annotation is removed. Unlike the generic pause annotation, this annotation:

    * Does not prevent the VM from being reconciled.
    * Can exist multiple times in the format `poweron.check.vmoperator.vmware.com/<COMPONENT>: <REASON>`

    This annotation can be applied to new VMs by non-privileged or privileged users. Only privileged users can modify or remove this annotation.

*  `delete.check.vmoperator.vmware.com`

    This annotation prevents a VM from being deleted until the annotation is removed. Unlike the generic pause annotation, this annotation:

    * Does not prevent the VM from being reconciled.
    * Can exist multiple times in the format `delete.check.vmoperator.vmware.com/<COMPONENT>: <REASON>`

    This annotation can only be added, modified, or removed on new/existing VMs by privileged users.

The purpose of both annotations is to enable external components to participate in the lifecycle of a VM.

Fixes `NA`

```release-note
Support annotations poweron.check and delete.check to allow external components to participate in VM lifecycle events.
```

----
📚 Documentation preview 📚: https://vm-operator--946.org.readthedocs.build/en/946/

### DIFF
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -77,6 +77,9 @@ const (
 	invalidImageKind                         = "supported: " + vmiKind + "; " + cvmiKind
 	invalidZone                              = "cannot use zone that is being deleted"
 	restrictedToPrivUsers                    = "restricted to privileged users"
+	addRestrictedAnnotation                  = "adding this annotation is restricted to privileged users"
+	delRestrictedAnnotation                  = "removing this annotation is restricted to privileged users"
+	modRestrictedAnnotation                  = "modifying this annotation is restricted to privileged users"
 )
 
 // +kubebuilder:webhook:verbs=create;update,path=/default-validate-vmoperator-vmware-com-v1alpha4-virtualmachine,mutating=false,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachines,versions=v1alpha4,name=default.validating.virtualmachine.v1alpha4.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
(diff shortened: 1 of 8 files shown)


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1568,3 +1572,59 @@ var capvDefaultServiceAccount = regexp.MustCompile("^system:serviceaccount:svc-t
...
+
+	for k := range newVM.Annotations {
+		if strings.HasPrefix(k, prefix) {
+			f := field.NewPath("metadata", "annotations", k)
+			oldVal, onOldVM := oldVM.Annotations[k]
+			newVal, onNewVM := newVM.Annotations[k]
+
+			switch {
+			case onOldVM && !onNewVM:
+				allErrs = append(allErrs, field.Forbidden(f, delRestrictedAnnotation))
```
dilyar85, line 1620: Same here. I don't think this would ever be the case because `onNewVM` is always true as you're looping through new VM in this loop?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1568,3 +1572,59 @@ var capvDefaultServiceAccount = regexp.MustCompile("^system:serviceaccount:svc-t
...
+	var allErrs field.ErrorList
+
+	if ctx.IsPrivilegedAccount {
+		return allErrs
+	}
+
+	for k := range oldVM.Annotations {
+		if strings.HasPrefix(k, prefix) {
+			f := field.NewPath("metadata", "annotations", k)
+			_, onOldVM := oldVM.Annotations[k]
```
dilyar85, line 1603: Is this line needed since you're already looping through oldVM.Annotations (i.e. `onOldVM` is always true)?

webhooks/virtualmachine/validation/virtualmachine_validator.go
dilyar85, line 1620: Same here. I don't think this would ever be the case because onNewVM is always true as you're looping through new VM in this loop?

webhooks/virtualmachine/validation/virtualmachine_validator.go
akutz: Good catch, fixed!

webhooks/virtualmachine/validation/virtualmachine_validator.go
akutz: Good catch, fixed!

dilyar85 (approved):
LGTM, thanks @akutz!

PR #949, opened 2025-05-01
✨ FastDeploy: Add DiskPromotion API and reconciler

dougm:
Depends on https://github.com/vmware/govmomi/pull/3778


api/v1alpha4/virtualmachine_types.go
```diff
@@ -705,6 +732,32 @@ type VirtualMachineSpec struct {
...
+	// +kubebuilder:default=Linked
+
+	// DeployMode describes the mode in which the VM is deployed. The available
+	// modes are:
+	//
+	// - Direct              -- a full clone of the VM image
+	// - Linked              -- a linked clone of the VM image
+	//
+	// Defaults to Linked.
+	// DeployMode VirtualMachineDeployMode `json:"deployMode,omitempty"`
```
akutz, line 746: This can be removed.

api/v1alpha4/virtualmachine_types.go
```diff
@@ -883,6 +936,12 @@ type VirtualMachineStatus struct {
 
 	// Storage describes the observed state of the VirtualMachine's storage.
 	Storage *VirtualMachineStorageStatus `json:"storage,omitempty"`
+
+	// +optional
+
+	// TaskID describes the ID of a task created by VM Operator to modify this
```
akutz, line 942: Suggest:

```
// TaskID describes the observed ID of the task created by VM Operator to
// perform some long-running operation on the VM.
```

I do not want to say "modify" because this ID might also surface the ID of a snapshot task for instance.

akutz (commented):
Thanks @dougm . I will look more in-depth later, this was just a first-pass of the API, thanks!

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -5,21 +5,70 @@
...
+
+	// Manually restore data.
+	restored := &vmopv1.VirtualMachine{}
+	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
+		return err
+	}
+
+	// BEGIN RESTORE
+
+	restore_v1alpha4_VirtualMachinePromoteDisksMode(dst, restored)
```
bryanv, line 54: Don't we need to do this restore v1a1 and v1a2 as well?

dougm: Ah yes, adding those. Adding v1alpha3/virtualmachine_conversion_test.go, which is copy-n-paste of   v1alpha2/virtualmachine_conversion_test.go's `VirtualMachine hub-spoke-hub` test, with `PromoteDisksMode` added to the spec. Not sure if I should prune some of the copied `hub` fields or leave as-is?

dougm: Rebasing with Sai's change merged.. I will prune that to just PromoteDisksMode

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -288,8 +289,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
 	ctx = cource.JoinContext(ctx, r.Context)
 
-	if pkgcfg.FromContext(ctx).Features.BringYourOwnEncryptionKey {
+	if pkgcfg.FromContext(ctx).Features.BringYourOwnEncryptionKey ||
+		pkgcfg.FromContext(ctx).Features.FastDeploy {
 		ctx = vmconfig.WithContext(ctx)
+		ctx = vmconfig.Register(ctx, diskpromo.New())
 		ctx = vmconfig.Register(ctx, crypto.New())
```
bryanv, line 296: Do  we want to guard the Register() calls based on their corresponding feature flag?

bryanv (approved):

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
dougm: Yes, done

PR #954, opened 2025-05-05
Introduce VirtualMachineGroup API and controller for reconciling members

dilyar85:

This PR adds a new `VirtualMachineGroup` type in v1alpha4, along with a new controller that handles reconciliation of group members.

To keep this PR focused and manageable, follow-up PRs will implement remaining TODOs comments, including `reconcilePowerState()` and `reconcilePlacement()` after this PR is reviewed and merged.

Credits to @akutz for contributing all the API field changes!

Fixes N/A.

N/A.

```release-note
Introduce VirtualMachineGroup API in v1alpha4.
```

----
📚 Documentation preview 📚: https://vm-operator--954.org.readthedocs.build/en/954/


akutz:
@dilyar85 We need to figure out how to model the placement results for groups of groups. Please think about that and propose something and add it to this PR after we chat about it.

api/v1alpha4/virtualmachinegroup_types.go
```diff
@@ -0,0 +1,321 @@
...
+
+	// Delay is the amount of time to wait before performing the power
+	// operation.
+	//
+	// If omitted, the power operation will be applied immediately.
+	Delay *metav1.Duration `json:"delay,omitempty"`
+}
+
+// VirtualMachineGroupSpec defines the desired state of VirtualMachineGroup.
+type VirtualMachineGroupSpec struct {
```
zyiyi11, line 78: Does this VMGroup need a name spec?

api/v1alpha4/virtualmachinegroup_types.go
```diff
@@ -0,0 +1,338 @@
...
+)
+
+const (
+	// LastUpdatedPowerStateTimeAnnotation is the annotation key for the last
+	// updated time of the power state.
+	LastUpdatedPowerStateTimeAnnotation = GroupName + "/last-updated-power-state-time"
+
+	// ScheduledPowerStateTimeAnnotation is the annotation key for the scheduled
+	// time when a power state change should be applied.
+	ScheduledPowerStateTimeAnnotation = GroupName + "/scheduled-power-state-time"
```
akutz, line 28: What are these? Why do we have "scheduled" anything? We should not be scheduling anything.

api/v1alpha4/virtualmachinegroup_types.go
```diff
@@ -0,0 +1,338 @@
...
+
+	// Kind is the kind of this member, which can be either VirtualMachine or
+	// VirtualMachineGroup.
+	Kind string `json:"kind"`
+
+	// +optional
+
+	// Placement describes the placement results for this member.
+	//
+	// Please note this field is only set for VirtualMachine kind members.
```
akutz, line 244: Just say `VirtualMachine members`.

api/v1alpha4/virtualmachinegroup_types.go
akutz: I _think_ I remember why you did this. Pinging you offline.

api/v1alpha4/virtualmachinegroup_types.go
dilyar85: Updated.

api/v1alpha4/virtualmachinegroup_types.go
dilyar85: Discussed offline. Going to remove these annotations from this /api package since they're not user facing.

controllers/virtualmachine/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -0,0 +1,384 @@
...
+	client.Client
+	Context  context.Context
+	Logger   logr.Logger
+	Recorder record.Recorder
+}
+
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinegroups,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinegroups/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines/status,verbs=get;update;patch
```
akutz: The VM Group controller does not need to update or patch a VM's status.

controllers/virtualmachine/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -0,0 +1,384 @@
...
+func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineGroupContext) (ctrl.Result, error) {
+	if !controllerutil.ContainsFinalizer(ctx.VMGroup, finalizerName) {
+		controllerutil.AddFinalizer(ctx.VMGroup, finalizerName)
+		return ctrl.Result{}, nil
+	}
+
+	ctx.Logger.Info("Reconciling VirtualMachineGroup")
+
+	defer func(beforeVMGroupStatus *vmopv1.VirtualMachineGroupStatus) {
+		if !apiequality.Semantic.DeepEqual(beforeVMGroupStatus, &ctx.VMGroup.Status) {
```
akutz, line 177: What is the purpose of this bit? Not opposed, just curious since we do not do this anywhere of which I am aware. Plus, maybe use `cmp.Diff` if they are not equal to print the differences?

controllers/virtualmachine/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -0,0 +1,384 @@
...
+
+// setReadyCondition sets the Ready condition for the group based on the
+// status of its members.
+func setReadyCondition(ctx *pkgctx.VirtualMachineGroupContext) {
+	if len(ctx.VMGroup.Status.Members) == 0 {
+		conditions.MarkTrue(ctx.VMGroup, vmopv1.ReadyConditionType)
+		return
+	}
+
+	var readyCount int
```
akutz: Please change to:

```go
var (
	readyCount     int
	failureReasons []string
	totalCount     = len(ctx.VMGroup.Status.Members)
)
```

akutz (changes requested):
A few nits, but other than those, looks good!

controllers/virtualmachine/virtualmachinegroup/virtualmachinegroup_controller.go
dilyar85: This is copy-pasted from the VM Controller: https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go#L502

I think it's better visibility to see when a VM Group CR has finally reached its desires state from looking at controller logs. Adding the `cmp.Diff` output might spam the logs though with all the status updates.

That said, I'm also fine removing it since reconciliation of VM Groups are fairly straightforward (compared to VM).

akutz: No, it's fine. We can improve it in both places later. Thanks!

akutz (approved):
Approving as @dilyar85 will make the requested changes.

PR #956, opened 2025-05-06
✨ Enhanced Storage APIs

akutz:

This patch introduces enhanced storage APIs to VM Operator.

Changes include:

* The ability to specify IDE, NVME, SATA, and SCSI controllers directly as part of the VM spec.
* The ability to specify the controller to which to attach a volume.
* The ability to specify the disk mode used to attach a volume.

Fixes `NA`

```release-note
Update the storage APIs for VMs to support explicit disk controllers and disk mode / unit numbers for volumes.
```


api/v1alpha4/virtualmachine_types.go
```diff
@@ -674,9 +674,48 @@ type VirtualMachineSpec struct {
...
-	// Volumes describes a list of volumes that can be mounted to the VM.
+	// Volumes describes the desired list of volumes to attach to the VM.
 	Volumes []VirtualMachineVolume `json:"volumes,omitempty"`
 
+	// +optional
+	// +listType=map
+	// +listMapKey=busNumber
+	// +kubebuilder:validation:MaxItems=2
+
+	// IDEControllers describes the desired list of IDE controllers for the VM.
```
bryanv, line 685: We talked about this in person but leaving a comment for others to chime in - maybe move these under a new `Spec.Hardware` would be good for if we have to later add more controller types or other HW'y things (like possibility moving `Spec.Cdrom` under this new container too)

akutz: Great point. Done and one.

api/v1alpha2/virtualmachine_conversion.go
```diff
@@ -307,6 +309,37 @@ func restore_v1alpha4_VirtualMachineBootOptions(dst, src *vmopv1.VirtualMachine)
...
+					dstPvc.UnitNumber = srcPvc.UnitNumber
+				}
+			}
+		}
+	}
+}
+
+func restore_v1alpha4_VirtualMachineHardware(dst, src *vmopv1.VirtualMachine) {
+	if src.Spec.Hardware != nil {
+		dst.Spec.Hardware = src.Spec.Hardware.DeepCopy()
```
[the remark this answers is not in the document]
akutz: We've seen issues in the past where we just copy objects with pointers during conversion, so I just did it to be safe.

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -52,6 +114,28 @@ func Convert_v1alpha3_VirtualMachineStatus_To_v1alpha4_VirtualMachineStatus(
...
 
+func Convert_v1alpha3_VirtualMachineSpec_To_v1alpha4_VirtualMachineSpec(
+	in *VirtualMachineSpec, out *vmopv1.VirtualMachineSpec, s apiconversion.Scope) error {
+
+	if err := autoConvert_v1alpha3_VirtualMachineSpec_To_v1alpha4_VirtualMachineSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if len(in.Cdrom) > 0 {
+		if out.Hardware == nil {
```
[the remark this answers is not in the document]
akutz: It took me FOREVER to understand how what I had was any different than what you wrote. It finally clicked :)

api/v1alpha2/virtualmachine_conversion.go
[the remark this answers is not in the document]
akutz: Ohh, but this is during restore and should be fine. Yeah, I'll remove the deep copy.

api/v1alpha4/virtualmachine_storage_types.go
```diff
@@ -56,6 +100,114 @@ type PersistentVolumeClaimVolumeSource struct {
...
+	//                       diskMode=IndependentPersistent and
+	//                       sharingMode=MultiWriter and attached to the first
+	//                       SCSI controller with an available slot and
+	//                       sharingMode=None. If no such controller exists,
+	//                       a new ParaVirtual SCSI controller will be created
+	//                       with sharingMode=None long as there are currently
+	//                       three or fewer SCSI controllers.
+	//   - MicrosoftWSFC  -- The volume is configured with
+	//                       diskMode=IndependentPersistent and attached to a
+	//                       SCSI controller with sharingMode=Physical.
```
skogta: @akutz what is meant by sharimgMode as Physical? We only have 2 options, multiwriter or none.

akutz: I think you may be conflating a controller's sharing mode with a volume's sharing mode.


### VirtualController SharingMode

Please note that the enum `VirtualControllerSharingMode` is specified on two types:

* [`VirtualSCSIController`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualSCSIController.html), which has a [`sharedBus`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualSCSIController.html#sharedBus) field that can take the following [values](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualSCSIController.Sharing.html):
    * `noSharing`
    * `physicalSharing`
    * `virtualSharing`
* [`VirtualNVMEController`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualNVMEController.html), which has a [`sharedBus`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualNVMEController.html#sharedBus) field that can take the following [values](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualNVMEController.Sharing.html):
    * `noSharing`
    * `physicalSharing`

For our Kubernetes APIs, I defined the type [`VirtualControllerSharingMode`](https://github.com/akutz/vm-operator/blob/8e1fda1d06e1c633e28b47571318b9b1169a9fe8/api/v1alpha4/virtualmachine_hardware_types.go#L18-L26) which has the following values:
* `None`
* `Physical`
* `Virtual`

We will map these to the underlying values accordingly.


### VolumeSharingMode

This PR introduces a new Kubernetes API enum [`VolumeSharingMode`](https://github.com/akutz/vm-operator/blob/8e1fda1d06e1c633e28b47571318b9b1169a9fe8/api/v1alpha4/virtualmachine_storage_types.go#L52-L59) with the following values:
* `MultiWriter`
* `None`

Does that answer your question @skogta ?

skogta: Thanks Andrew for explaining.
I was trying to figure out what all values I need to expose on the CnsNodeVmBatchAttachment CRD.

Here is what I was thinking:
1. DiskMode
3.  SharingMode
4. Unit controller
5. Slot number

Do I need to expose VirtualControllerSharingMode also?

akutz: No, we create the controllers. Please note that you need to accept the DeviceKey of the controller to which we want the disk connected. Your API should accept the following:

* DiskMode
* SharingMode
* UnitNumber
* ControllerKey (the device key of the controller to which the disk should be connected)

The slot number makes no sense as that is for PCI devices, such as controllers. AFAIK, the slot number has nothing to do with disks. I could be wrong though. How did you picture the slot number being used?

akutz: @skogta Please see https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualDisk.html and https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.device.VirtualDevice.html to learn about these concepts.

You need to take:

* DiskMode
* SharingMode
* UnitNumber
* ControllerKey (the device key of the controller to which the disk should be connected)

The `ControllerKey` is not in our spec because it is not something the user specifies. We are sending you the DeviceKey of the `VirtualController` to which to attach the disk. 

The `UnitNumber` is what you keep referring to in our Google chat as the "slot number."

skogta: Thanks Andrew. I understand this now. Appreciate your help.

bryanv (approved):

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -71,10 +109,20 @@ func Convert_v1alpha4_VirtualMachineStorageStatus_To_v1alpha3_VirtualMachineStor
...
-	in *vmopv1.VirtualMachineVolumeStatus, out *VirtualMachineVolumeStatus, s apiconversion.Scope) error {
+func Convert_v1alpha4_VirtualMachineCdromSpec_To_v1alpha3_VirtualMachineCdromSpec(
+	in *vmopv1.VirtualMachineCdromSpec, out *VirtualMachineCdromSpec, s apiconversion.Scope) error {
 
-	return autoConvert_v1alpha4_VirtualMachineVolumeStatus_To_v1alpha3_VirtualMachineVolumeStatus(in, out, s)
+	if err := autoConvert_v1alpha4_VirtualMachineCdromSpec_To_v1alpha3_VirtualMachineCdromSpec(in, out, s); err != nil {
+		return err
+	}
+
+	out.Name = fmt.Sprintf(
```
bryanv: Slightly worried here that changing out the name may confuse things. Is there a need for v1a3 users to be able to specify the bus/unit numbers?

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -129,6 +211,37 @@ func restore_v1alpha4_VirtualMachineAffinitySpec(dst, src *vmopv1.VirtualMachine
...
+					dstPvc.DiskMode = srcPvc.DiskMode
+					dstPvc.SharingMode = srcPvc.SharingMode
+					dstPvc.UnitNumber = srcPvc.UnitNumber
+				}
+			}
+		}
+	}
+}
+
+func restore_v1alpha4_VirtualMachineHardware(dst, src *vmopv1.VirtualMachine) {
```
bryanv: The behavior here seems like a breaking in the sense that you cannot add/remove CDROM devices anymore at v1a3, because of the wholesale overriden of what was previously converted?

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -123,10 +162,116 @@ func restore_v1alpha5_VirtualMachinePromoteDisksMode(dst, src *vmopv1.VirtualMac
...
+	}
+
+	// Restore the missing hardware.
+	if src.Spec.Hardware == nil {
+		if len(newCDROMs) == 0 {
+			dst.Spec.Hardware = nil
+		}
+	} else {
+		dst.Spec.Hardware = src.Spec.Hardware.DeepCopy()
+		dst.Spec.Hardware.Cdrom = slices.Clone(newCDROMs)
```
bryanv: Does this need to be a clone of the clone?

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -123,10 +162,116 @@ func restore_v1alpha5_VirtualMachinePromoteDisksMode(dst, src *vmopv1.VirtualMac
...
+	// Create a copy of the CD-ROMs currently stored on the dst VM.
+	var newCDROMs []vmopv1.VirtualMachineCdromSpec
+	if dst.Spec.Hardware != nil && len(dst.Spec.Hardware.Cdrom) > 0 {
+		newCDROMs = slices.Clone(dst.Spec.Hardware.Cdrom)
+	}
+
+	// Restore the missing hardware.
+	if src.Spec.Hardware == nil {
+		if len(newCDROMs) == 0 {
+			dst.Spec.Hardware = nil
```
bryanv: Is the purpose of this to clear out the controllers, although if the `src.Spec.Hardware` was nil, I'm not sure how would ever be any

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -123,10 +162,107 @@ func restore_v1alpha5_VirtualMachinePromoteDisksMode(dst, src *vmopv1.VirtualMac
...
+					dstPvc.UnitNumber = srcPvc.UnitNumber
+				}
+			}
+		}
+	}
+}
+
+func restore_v1alpha5_VirtualMachineHardware(dst, src *vmopv1.VirtualMachine) {
+	// Create a copy of the CD-ROMs currently stored on the dst VM.
+	var cdromChanges []vmopv1.VirtualMachineCdromSpec
```
bryanv, line 223: Maybe just check `if src.Spec.Hardware == nil { return } ` up front since without that there is nothing to restore. 

Or move this into the `if src.Spec.Hardware != nil { ...` block to scope `cdromChanges` to just that (and don't think we'd need the Clone since we're keeping it for the dst object)

PR #960, opened 2025-05-08
🌱 Move utilconversion test helpers into pkg

akutz:

This patch moves the utilconversion fuzz helpers into a sub-package in order to prevent those helpers from being imported into the dependency graph when building non-test packages.

This helps downstream projects like CSI that import our API module from being impacted by any versioning issues that occur out of the fuzz helper logic.

Fixes `NA`

```release-note
Move API fuzz helpers into sub-package.
```


dilyar85 (approved):
LGTM!

PR #966, opened 2025-05-09
🌱 Move PCI passthrough comparison into resize

bryanv:

Remove the unused PCI passthrough add/remove devices from the old session into resize, and add formal backing comparison functions.

Fixes #

Backing comparison functions are based on govmoni's SelectByBackingInfo

```release-note
NONE
```


pkg/util/resize/configspec_pci_devices.go
```diff
@@ -26,34 +24,20 @@ func comparePCIDevices(
...
+			switch a := expectedPci.Backing.(type) {
 			case *vimtypes.VirtualPCIPassthroughVmiopBackingInfo:
-				b := expectedBacking.(*vimtypes.VirtualPCIPassthroughVmiopBackingInfo)
-				backingMatch = a.Vgpu == b.Vgpu
-
+				backingMatch = MatchVirtualPCIPassthroughVmiopBackingInfo(a, curBacking)
 			case *vimtypes.VirtualPCIPassthroughDynamicBackingInfo:
-				currAllowedDevs := a.AllowedDevice
-				b := expectedBacking.(*vimtypes.VirtualPCIPassthroughDynamicBackingInfo)
-				if a.CustomLabel == b.CustomLabel && len(b.AllowedDevice) > 0 {
```
aruneshpa, line 48: So, if the labels matched, but the allowed devices didn't we skip the comparison?  Shouldn't the labels be the authority on identifying the correct device that is being updated?

aruneshpa (approved):
Everything looks good to me, except the comparison of allowed devices bit.

pkg/util/resize/configspec_pci_devices.go
bryanv: Talked offline - I cribbed this govmoni's SelectByBackingInfo() and makes sense that the label, if set, is sufficient in this case. I really only want to fix the blind `b.AllowedDevice[0]` assumption and can revisit later if needed.

PR #969, opened 2025-05-12
🌱 Initialize VM interface Network reference on update

bryanv:

When network interfaces are added, initialize the interface spec Network reference field based on network provider like we do on create.

Fixes #

```release-note
NONE
```

### DIFF
--- a/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
+++ b/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
@@ -156,6 +156,12 @@ func (m mutator) Mutate(ctx *pkgctx.WebhookRequestContext) admission.Response {
 			wasMutated = true
 		}
 
+		if pkgcfg.FromContext(ctx).Features.MutableNetworks {
+			if ok := SetDefaultNetworkOnUpdate(ctx, m.client, modified); ok {
+				wasMutated = true
+			}
+		}
+
 		if ok := SetDefaultCdromImgKindOnUpdate(ctx, modified, oldVM); ok {
 			wasMutated = true
 		}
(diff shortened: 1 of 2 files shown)


webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -216,22 +222,51 @@ func SetNextRestartTime(
...
-	if _, ok := vm.Annotations[v1alpha1.NoDefaultNicAnnotation]; ok {
-		return false
-	}
+func setDefaultNetworkInterfaceNetwork(
+	vm *vmopv1.VirtualMachine,
+	ifaceIdx int,
+	networkRef common.PartialObjectRef,
+) bool {
+	ifaceNetwork := vm.Spec.Network.Interfaces[ifaceIdx].Network
+	var updated bool
```
dougm, line 231: nit: looks you could drop this var and just `return true` instead of `updated = true`. And with those early returns you can drop the `else {` below

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -216,22 +222,51 @@ func SetNextRestartTime(
...
+			}
+			vm.Spec.Network.Interfaces[ifaceIdx].Network = ifaceNetwork
+			updated = true
+
+		} else if ifaceNetwork.Kind == "" && ifaceNetwork.APIVersion == "" {
+			ifaceNetwork.Kind = networkRef.Kind
+			ifaceNetwork.APIVersion = networkRef.APIVersion
+			updated = true
+		}
+	} else {
```
dougm, line 249: Referring to this `else {`

dougm (approved):
lgtm

PR #970, opened 2025-05-13
api: Add VirtualMachineClass DisplayName field

dougm:
In support of upcoming Immutable VM Classes feature.


dougm:
We'll be using an annotation rather than adding a new field.

PR #971, opened 2025-05-14
Make VM Route Metric field optional

bryanv:

This field is not required in the Netplan so also only set it in the generated Netplan if non-zero.

Use Gomega's HaveValue matcher to better assert Netplan fields that are pointers.

Fixes #

```release-note
The Metric field in the VirtualMachineNetworkRouteSpec is optional. Add a validation that if set, the metric must be greater than zero.
```

### DIFF
--- a/api/v1alpha2/virtualmachine_network_types.go
+++ b/api/v1alpha2/virtualmachine_network_types.go
@@ -19,7 +19,10 @@ type VirtualMachineNetworkRouteSpec struct {
 	Via string `json:"via"`
 
 	// Metric is the weight/priority of the route.
-	Metric int32 `json:"metric"`
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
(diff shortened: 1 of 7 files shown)


api/v1alpha2/virtualmachine_network_types.go
```diff
@@ -19,7 +19,9 @@ type VirtualMachineNetworkRouteSpec struct {
 	Via string `json:"via"`
 
 	// Metric is the weight/priority of the route.
-	Metric int32 `json:"metric"`
+	//
+	// +optional
+	Metric int32 `json:"metric,omitempty"`
```
akutz, line 24: Should this not be nillable if we want it to be optional?

akutz: BV told me offline the netplan docs state this must be a positive int, so we'll add a min value of 1.

akutz (approved):
A nit BV is going to fix.

PR #976, opened 2025-05-21
✨ Introduce VirtualMachineSnapshot API

aruneshpa:

This change introduces the APIs to take a snapshot of a VM Service VM. A new custom resource definition is added that contains the desired specifications and the observed status of the snapshot.

VirtualMachine custom resource definition has been modified to enable workflows for creating, deleting a snapshot as well as reverting a VM to a snapshot.

```release-note
Introduce VirtualMachineSnapshot API
```

----
📚 Documentation preview 📚: https://vm-operator--976.org.readthedocs.build/en/976/


api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,142 @@
...
+	QuiesceSpec *QuiesceSpec `json:"quiesce,omitempty"`
+
+	// +optional
+	//
+	// Description represents a description of the snapshot.
+	Description string `json:"description,omitempty"`
+
+	// VMRef represents the name of the virtual machine for which the
+	// snapshot is requested.
+	VMRef corev1.TypedLocalObjectReference `json:"vmRef:omitempty"`
```
bryanv, line 50: Several `...:omitempty` instead of `, omitempty` in this file.

api/v1alpha4/virtualmachine_types.go
```diff
@@ -977,6 +1001,18 @@ type VirtualMachineStatus struct {
 	// TaskID describes the observed ID of the task created by VM Operator to
 	// perform some long-running operation on the VM.
 	TaskID string `json:"taskID,omitempty"`
+
+	// +optional
+
+	// CurrentSnapshot describes the observed working snapshot of the VirtualMachine.
+	CurrentSnapshot *corev1.TypedObjectReference `json:"currentSnapshot,omitempty"`
```
sreyasn, line 1008: do you want to use TypedLocalObjectReference here too ? any specific reason for using for non-namespaced (non-local) reference

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,142 @@
...
+	// that the virtual machine snapshot is ready.
+	VirtualMachineSnapshotReadyCondition = "VirtualMachineSnapshotReady"
+)
+
+// VirtualMachineSnapshotStatus defines the observed state of VirtualMachineSnapshot.
+type VirtualMachineSnapshotStatus struct {
+	// +optional
+
+	// Phase represents the phase of a VM snapshot operation.
+	Phase *VMSnapshotPhase `json:"phase:omitempty"`
```
sreyasn, line 87: I think we can do without a pointer here?. I referred to definitions for Phase in Persistent Volumes.

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,142 @@
...
+	// create.
+	// The virtual machine must support this capability.
+	// When set to false, the power state of the snapshot is set to
+	// false.
+	// For a VM in suspended state, memory is always included
+	// in the snashot.
+	Memory *bool `json:"memory,omitempty"`
+
+	// +optional
+	// +kubebuilder:default=false
```
sreyasn, line 32: This kubebuilder validation (default=false) will err out for non-boolean types..

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,142 @@
...
+
+// QuiesceSpec represents specifications that will be used to quiesce
+// the guest when taking a snapshot.
+type QuiesceSpec struct {
+	// +optional
+
+	// Timeout represents the maximum time in minutes for snapshot
+	// operation to be performed on the virtual machine. The timeout
+	// can not be less than 5 minutes or more than 240 minutes.
+	Timeout *time.Duration `json:"timeout,omitempty"`
```
sreyasn, line 61: I was thinking if we should keep this as int32 here also?. Since the snapshotAPI only seems to support minutes and int32.

api/v1alpha4/virtualmachine_types.go
```diff
@@ -793,6 +794,29 @@ type VirtualMachineSpec struct {
...
+	//
+	// If the value of this field is set to an existing snapshot that
+	// is different from the status.currentSnapshot, the virtual machine is
+	// reverted to the requested snapshot. A revert operation results
+	// in the virtual machine's configuration and data being rolled
+	// back to the state of the snapshot. When other fields of the
+	// VirtualMachineSpec (e.g., PowerState) are updated along with
+	// the revert operation, then those are applied once the revert is
+	// completed.
+	CurrentSnapshot *corev1.LocalObjectReference `json:"currentSnapshot,omitempty"`
```
sreyasn, line 819: Do we wanna make the spec.CurrentSnapshot and status.CurrentSnapshot have same types? for easier comparison.

api/v1alpha4/virtualmachinesnapshot_types.go
aruneshpa: This is duration on purpose since that allows a user to specify values like "10m", or "2m45s".

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// +kubebuilder:object:root=true
+
+// VirtualMachineSnapshotList contains a list of VirtualMachineSnapshot.
+type VirtualMachineSnapshotList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []VirtualMachineSnapshot `json:"items"`
+}
```
sreyasn, line 140: reminder to init() objectTypes() 

```
func init() {
	objectTypes = append(objectTypes, &VirtualMachineSnapshot{}, &VirtualMachineSnapshotList{})
}
```

api/v1alpha4/virtualmachine_types.go
```diff
@@ -895,6 +896,29 @@ type VirtualMachineSpec struct {
...
+	//
+	// If the value of this field is set to an existing snapshot that
+	// is different from the status.currentSnapshot, the virtual machine is
+	// reverted to the requested snapshot. A revert operation results
+	// in the virtual machine's configuration and data being rolled
+	// back to the state of the snapshot. When other fields of the
+	// VirtualMachineSpec (e.g., PowerState) are updated along with
+	// the revert operation, then those are applied once the revert is
+	// completed.
+	CurrentSnapshot *corev1.TypedLocalObjectReference `json:"currentSnapshot,omitempty"`
```
dougm: When should we use `corev1.TypedLocalObjectReference` vs `vmopv1common.LocalObjectRef`? (Asking as I'm using the latter in #989 ) They both have the same Kind+Name fields, only diff is `APIGroup *string` vs `APIVersion string`.

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,140 @@
...
+	// +kubebuilder:validation:format:=duration
+
+	// Timeout represents the maximum time in minutes for snapshot
+	// operation to be performed on the virtual machine. The timeout
+	// can not be less than 5 minutes or more than 240 minutes.
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
+}
+
+// VMSnapshotPhase represents the phase of a VM snapshot.
+type VMSnapshotPhase string
```
dougm: We use `VirtualMachine` prefix rather than `VM` in all other type names, consider that here?

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,140 @@
...
+
+const (
```
dougm: Sounds like these map to [vim.TaskInfo.State](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.TaskInfo.State.html) ? Maybe we should keep the mapping more direct, even keep the type name generic `TaskState` rather than `VMSnapshotPhase` ?

dougm (commented):
lgtm, maps well to the vSphere snapshot model

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,140 @@
...
+	// quiesce details. If quiesceSpec is set and the virtual machine
+	// is powered on when the snapshot is taken, VMware Tools is used
+	// to quiesce the file system in the virtual machine. This assures
+	// that a disk snapshot represents a consistent state of the guest
+	// file systems. If the virtual machine is powered off or VMware
+	// Tools are not available, the quiesce spec is ignored.
+	QuiesceSpec *QuiesceSpec `json:"quiesce,omitempty"`
+
+	// +optional
+	//
```
akutz: Remove this `//`

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -0,0 +1,140 @@
...
+	// +optional
+
+	// Quiesce represents the spec used for granular control over
+	// quiesce details. If quiesceSpec is set and the virtual machine
+	// is powered on when the snapshot is taken, VMware Tools is used
+	// to quiesce the file system in the virtual machine. This assures
+	// that a disk snapshot represents a consistent state of the guest
+	// file systems. If the virtual machine is powered off or VMware
+	// Tools are not available, the quiesce spec is ignored.
+	QuiesceSpec *QuiesceSpec `json:"quiesce,omitempty"`
```
akutz: The field name should be `Quiesce`

api/v1alpha4/virtualmachinesnapshot_types.go
aruneshpa: Removed this entirely because the new upstream API conventions don't recommend using "Phase": https://github.com/kubernetes-sigs/kube-api-linter?tab=readme-ov-file#nophase

We can add other fields later if/when needed.

api/v1alpha4/virtualmachine_types.go
aruneshpa: Yeah, so apparently there was a somewhat recent guidance recommending *not* to use the corev1 types.
https://github.com/kubernetes/api/commit/1aa9f66df7d5e3c79de1b0c157882dac343838ea

Until now, this warning was only present on corev1.ObjectReference but upstream has not added it on _all_ types (localref, localobjectref etc).

Let's use our internal one.

api/v1alpha4/virtualmachinesnapshot_types.go
aruneshpa: Removed the phase entirely.

api/v1alpha4/virtualmachine_types.go
```diff
@@ -1084,6 +1117,18 @@ type VirtualMachineStatus struct {
 	// TaskID describes the observed ID of the task created by VM Operator to
 	// perform some long-running operation on the VM.
 	TaskID string `json:"taskID,omitempty"`
+
+	// +optional
+
+	// CurrentSnapshot describes the observed working snapshot of the VirtualMachine.
+	CurrentSnapshot *corev1.TypedLocalObjectReference `json:"currentSnapshot,omitempty"`
```
sreyasn: change type to match spec.CurrentSnapshot ?. (*vmopv1common.LocalObjectRef)

PR #979, opened 2025-05-22
🐛 Prefer cloudinit local-ipv4/6 for VM's Primary IPs

bryanv:

The cloudinit VMware datasource sets these values in the ExtraConfig, so use those fields if set as the VM's primary IPv4/6 over what is reported by VM Tools. Apply the same validity checks on EC fields as what we were already doing.

Since we were already both walking and making a map of the ExtraConfig convert it to a map once and pass that down.

Fixes #

```release-note
When using CloudInit bootstrap, the local-ipv4 and local-ipv6 keys in the ExtraConfig, if set, will be preferred as the VM's PrimaryIP4 and PrimaryIP6 addresses
```


aruneshpa (approved):
LGTM.

PR #983, opened 2025-05-27
🐛 Only set Netplan MTU if a positive value

bryanv:

It appears there was an upstream change that started to render an MTU of zero as "null" in the netplan configuration which is not valid.

Fixes #

First noticed in Ubuntu 24.04

```release-note
When using cloudinit bootstrap, only set the netplan MTU when greater than zero.
```


dilyar85 (approved):

akutz:
Related to https://github.com/canonical/cloud-init/issues/6239

PR #984, opened 2025-05-28
✨ Report disk capacity for usage

akutz:

This patch updates the way VM Op reports VM disk usage against the namespace's storage quota. Previously the usage was reported via the actual usage on disk, i.e. even if a disk was provisioned with a capacity of 50GiB, and it was only actually taking up 20GiB on disk, then only the 20GiB would be reported against quota. Now the disk's potential usage, its total capacity, is reported for quota accounting.

However, a user can still determine a disk's usage vs. capacity as well as the same for the entire VM via status.storage. This field has been updated such that:

* `status.storage.used.other` is the total amount of storage used by all of the VM's non disks.
* `status.storage.used.disks` is the total amount of storage used by all of the VM's non-PVC disks.
* `status.storage.requested.disks` is the total amount of storage requested by all of the VM's non-PVC disks.
* `status.storage.total` is a sum of `status.storage.used.other` and `status.storage.requested.disks`.

The value of `status.storage.total` is what is reported against the namespace's storage quota.

Fixes `NA`

```release-note
Report disk capacity against namespace quota.
```

----
📚 Documentation preview 📚: https://vm-operator--984.org.readthedocs.build/en/984/


docs/concepts/workloads/vm.md
```diff
@@ -566,18 +566,24 @@ The following fields summarize a `VirtualMachine` resource's overall storage usa
 
 | Name | Description |
 |------|-------------|
-| `status.storage.committed` | The total storage space committed to this `VirtualMachine`. |
-| `status.storage.uncommitted` | The total storage space potentially used by this `VirtualMachine` |
-| `status.storage.unshared` | The total storage space occupied by this VirtualMachine that is not shared with any other `VirtualMachine`. |
+| `status.storage.other.used` | The total storage space used by all of the VM's non-disk files, ex. snapshots, logs, etc. |
```
xing-yang, line 569: Are the "snapshots" here refer to VM snapshots or something else?

akutz: Yes. Obviously this would be updated once we introduce VM snapshots as first-class objects.

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -12,6 +12,72 @@ import (
...
+	in *vmopv1.VirtualMachineStorageStatus, out *VirtualMachineStorageStatus, s apiconversion.Scope) error {
+
+	if err := autoConvert_v1alpha4_VirtualMachineStorageStatus_To_v1alpha3_VirtualMachineStorageStatus(in, out, s); err != nil {
+		return err
+	}
+
+	if in.Total != nil {
+		if out.Usage == nil {
+			out.Usage = &VirtualMachineStorageStatusUsage{}
+			out.Usage.Total = in.Total
```
bryanv, line 51: Is this really only suppose to be set when `out.Usage` is nil, instead of the usual pattern of

```
if out.Usage == nil {
  out.Usage = &VirtualMachineStorageStatusUsage{}
}
out.Usage.Total = in.Total
```

The later ones that just check for `if out.Usage == nil ` won't set their fields

dougm (approved):
lgtm

PR #987, opened 2025-05-30
✨ Move network interface matching into network

bryanv:

Also try to preserve ethernet card device type when an interface spec is updated to just point to a new network.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/network/reconcile.go
+++ b/pkg/providers/vsphere/network/reconcile.go
@@ -1,0 +1,151 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package network
+
+import (
+	"context"
+	"strings"
+
+	"github.com/vmware/govmomi/object"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	netopv1alpha1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
+	vpcv1alpha1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
+	ncpv1alpha1 "github.com/vmware-tanzu/vm-operator/external/ncp/api/v1alpha1"
+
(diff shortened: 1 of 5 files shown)


pkg/providers/vsphere/network/reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+	var deviceChanges []vimtypes.BaseVirtualDeviceConfigSpec
+
+	for idx, r := range results.Results {
+		matchingIdx := FindMatchingEthCard(currentEthCards, r.Device.(vimtypes.BaseVirtualEthernetCard))
+		if matchingIdx >= 0 {
+			// Exact match. Claim it by removing the device from the current ethernet cards.
+			matchDev := currentEthCards[matchingIdx].(vimtypes.BaseVirtualEthernetCard).GetVirtualEthernetCard()
+			results.Results[idx].DeviceKey = matchDev.Key
+			results.Results[idx].MacAddress = matchDev.MacAddress
+			currentEthCards = append(currentEthCards[:matchingIdx], currentEthCards[matchingIdx+1:]...)
```
dougm: Use [slices.Delete](https://pkg.go.dev/slices#Delete) instead? (golangci-lint has been nagging me to use the slices package)

pkg/providers/vsphere/network/reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+
+			results.UpdatedEthCards = true
+		}
+	}
+
+	// Remove any unmatched existing interfaces.
+	removeDeviceChanges := make([]vimtypes.BaseVirtualDeviceConfigSpec, 0, len(currentEthCards))
+	for _, dev := range currentEthCards {
+		removeDeviceChanges = append(removeDeviceChanges, &vimtypes.VirtualDeviceConfigSpec{
```
dougm, line 70: Used `make` w/ cap, but appending here. Using `removeDeviceChanges[i] = ...` instead?

dougm (approved):
lgtm, couple minor suggestions

pkg/providers/vsphere/network/reconcile.go
bryanv: This ends up tripping another lint warning wrt the append in the return statement so I think I'll just leave it as-is

PR #989, opened 2025-06-02
✨ Add VirtualMachineClassInstance API

dougm:


api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,124 @@
...
+	vmopv1common "github.com/vmware-tanzu/vm-operator/api/v1alpha4/common"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// VirtualMachineClassInstanceSpec defines the desired state of VirtualMachineClassInstance.
+type VirtualMachineClassInstanceSpec struct {
+	// Class is a reference to the VirtualMachineClass source of this
+	// VirtualMachineClassInstance.
+	Class vmopv1common.LocalObjectRef `json:"class"`
```
aruneshpa: Why not use the upstream variant? https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/local-object-reference/

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Active is true when a VirtualMachineClassInstance is in sync with its Class source.
+	// Active is false when the Class source is modified or removed from the namespace.
+	Active bool `json:"active"`
```
aruneshpa: WDYT about calling it "latest", or "current"?  I know I proposed active in the design, but this leaks the implementation a little.  I don't have a strong preference.  Open to other suggestions as well.

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,124 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// VirtualMachineClassInstanceSpec defines the desired state of VirtualMachineClassInstance.
+type VirtualMachineClassInstanceSpec struct {
+	// Class is a reference to the VirtualMachineClass source of this
+	// VirtualMachineClassInstance.
+	Class vmopv1common.LocalObjectRef `json:"class"`
+
+	// Active is true when a VirtualMachineClassInstance is in sync with its Class source.
```
aruneshpa: Could you also add Kubebuilder defaulting and validation markers?

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// Active is false when the Class source is modified or removed from the namespace.
+	Active bool `json:"active"`
+
+	// +optional
+
+	// ControllerName describes the name of the controller responsible for
```
aruneshpa: Hmm.  Fields like `controllerName` are not even used.   `Hardware` and `Policies` will also be removed once the oldest builds support ConfigSpec.  So, we either include the fields that we care about (like ConfigSpec, and reservation), or, we inline the VMClassSpec type here.  What do you think?

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,51 @@
...
+package v1alpha4
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// VirtualMachineClassInstanceSpec defines the desired state of VirtualMachineClassInstance.
+// It is a composite of VirtualMachineClassSpec.
+type VirtualMachineClassInstanceSpec struct {
+	VirtualMachineClassSpec `json:",inline"`
```
aruneshpa, line 14: We would need something in the instance spec to indicate the hash.  During each VM class reconcile:
1. The controller would need to calculate the hash from the spec of the class
2. Get all existing instances for the class (using ownerrefs)
3. If an instance for this hash exists, mark it as active by applying the label.  If it is already active, then no-op.
4. If any other instance has the active label, that is now removed.

For step 3, we can get the class spec from the instance spec and re-calculate the hash.  But it will be massively more performant to just have the configspec hash in the spec -- specially since it is never supposed to change.

Another approach would be to rely on the naming (vm-class-display-name + SHA), but that seems brittle.

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,51 @@
...
+type VirtualMachineClassInstanceStatus struct {
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=vmclassinstance
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="CPU",type="string",JSONPath=".spec.hardware.cpus"
+// +kubebuilder:printcolumn:name="Memory",type="string",JSONPath=".spec.hardware.memory"
+// +kubebuilder:printcolumn:name="Capabilities",type="string",priority=1,JSONPath=".status.capabilities"
```
aruneshpa: VM classes don't have capabilities _yet_.  So this can be removed.  If you want, you can also remove it from the virtualmachineclass_types.go

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,51 @@
...
+
+// VirtualMachineClassInstance is the schema for the virtualmachineclasses API and
```
aruneshpa:
```suggestion
// VirtualMachineClassInstance is the schema for the virtualmachineclassinstances API and
```

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -0,0 +1,51 @@
...
+// represents the desired state and observed status of a virtualmachineclasses
```
aruneshpa:
```suggestion
// represents the desired state and observed status of a virtualmachineclassinstances
```

api/v1alpha4/virtualmachineclassinstance_types.go
aruneshpa: Resolving this.  It seems that upstream no longer recommends using their types.

api/v1alpha4/virtualmachineclassinstance_types.go
dougm: As discussed in zoom, we'll use an annotation for now, rather than added a spec field.

aruneshpa (approved):

api/v1alpha4/virtualmachineclassinstance_types.go
dougm: Field removed, as we agreed to use a label instead.

api/v1alpha4/virtualmachineclassinstance_types.go
dougm: Ah, removed. Will look at removing from existing separately.

PR #992, opened 2025-06-03
✨ Skip delete platform resource annotation

akutz:

This patch adds support for the privileged annotation `vmoperator.vmware.com.protected/skip-delete-platform-resource` that may be used to skip the deletion of an underlying platform resource when the corresponding Kubernetes object is deleted.

This supports use cases such as rollback of import where the underlying VM should not be deleted even though the associated Kubernetes object is.

Fixes `NA`

```release-note
Support vmoperator.vmware.com.protected/skip-delete-platform-resource for skipping deletion of platform resources.
```

----
📚 Documentation preview 📚: https://vm-operator--992.org.readthedocs.build/en/992/

### DIFF
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -453,18 +453,26 @@ func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineContext) (reterr 
 
 	if controllerutil.ContainsFinalizer(ctx.VM, finalizerName) ||
 		controllerutil.ContainsFinalizer(ctx.VM, deprecatedFinalizerName) {
-		defer func() {
-			r.Recorder.EmitEvent(
-				ctx.VM, "Delete", filterNoRequeueErr(reterr), false)
-		}()
 
-		if err := r.VMProvider.DeleteVirtualMachine(ctx, ctx.VM); err != nil {
-			return err
+		_, skip := ctx.VM.Annotations[pkgconst.SkipDeletePlatformResourceKey]
+		if skip {
+			ctx.Logger.Info("Skipping deletion of underlying VM",
+				"reason", pkgconst.SkipDeletePlatformResourceKey,
+				"managedObjectID", ctx.VM.Status.UniqueID,
+				"biosUUID", ctx.VM.Status.BiosUUID,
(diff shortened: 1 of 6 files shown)


controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -452,17 +452,32 @@ func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineContext) (reterr
...
+
+		if skip {
+			ctx.Logger.Info("Skipping deletion of underlying VM",
+				"reason", pkgconst.SkipDeletePlatformResourceKey,
+				"managedObjectID", ctx.VM.Status.UniqueID,
+				"biosUUID", ctx.VM.Status.BiosUUID,
+				"instanceUUID", ctx.VM.Status.InstanceUUID)
+		} else {
+			defer func() {
+				r.Recorder.EmitEvent(ctx.VM, "Delete", reterr, false)
```
akutz, line 466: @bryanv / @ericvmw / @yanleizhao-vmware / et. al,

Thoughts on skipping the event when the underlying resource is not deleted? That's my take on it, but I could see wanting to emit the event regardless. Still, this made sense to me. I am curious as to what y'all think.

yanleizhao-vmware: The original event seems to be tight to the deletion of the underlying resource, skipping makes sense to me.

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -452,17 +452,32 @@ func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineContext) (reterr
...
+				return err
+			}
 		}
 
 		controllerutil.RemoveFinalizer(ctx.VM, finalizerName)
 		controllerutil.RemoveFinalizer(ctx.VM, deprecatedFinalizerName)
-		ctx.Logger.Info("Provider Completed deleting Virtual Machine", "time", time.Now().Format(time.RFC3339))
+
+		if !skip {
+			ctx.Logger.Info("Provider Completed deleting Virtual Machine",
```
bryanv: Can this be moved to L471.5? Also, the time is there is for VMC LINT that we don't care about any more so that can be removed.

bryanv (approved):

PR #1003, opened 2025-06-06
✨VirtualMachineSnapshot Controller and VirtualMachine Provider Implementation

sreyasn:

Snapshot controller
    -  This change introduces the snapshot controller to reconcile the
    VirtualMachineSnapshot resource and patch the VirtualMachine resource with
    the spec.CurrentSnapshot field
    
Snapshots: VirtualMachine Provider implementation
    - This change handles reconciliation of spec.CurrentSnapshot field
    in the VirtualMachine resource to create snapshots of the VM
    and update the Snapshot resource with the respective status fields.
    

Fixes # N/A

 Testing Done:
- Tested create snapshots of a VM on a Patched WCP testbed.

Screenshot 2025-06-05 at 2 55 42 PM

```
Introduce Snapshots controller and Create Snapshots feature for VMs

```

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -74,6 +74,7 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		}
 	}
 
+	// TODO: guard with a capability
 	if err := virtualmachinesnapshot.AddToManager(ctx, mgr); err != nil {
 		return fmt.Errorf("failed to initialize VirtualMachineSnapshot controller: %w", err)
 	}
--- a/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
+++ b/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
@@ -74,38 +74,38 @@ type Reconciler struct {
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
 
-	vmSnapShot := &vmopv1.VirtualMachineSnapshot{}
-	if err := r.Get(ctx, req.NamespacedName, vmSnapShot); err != nil {
+	vmSnapshot := &vmopv1.VirtualMachineSnapshot{}
+	if err := r.Get(ctx, req.NamespacedName, vmSnapshot); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	vmSnapShotCtx := &pkgctx.VirtualMachineSnapshotContext{
+	vmSnapshotCtx := &pkgctx.VirtualMachineSnapshotContext{
 		Context:                ctx,
 		Logger:                 ctrl.Log.WithName("VirtualMachineSnapShot").WithValues("name", req.Name),
-		VirtualMachineSnapshot: vmSnapShot,
+		VirtualMachineSnapshot: vmSnapshot,
 	}
 
-	patchHelper, err := patch.NewHelper(vmSnapShot, r.Client)
+	patchHelper, err := patch.NewHelper(vmSnapshot, r.Client)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to init patch helper for %s: %w", vmSnapShotCtx, err)
+		return ctrl.Result{}, fmt.Errorf("failed to init patch helper for %s: %w", vmSnapshotCtx, err)
 	}
 	defer func() {
-		vmSnapShotCtx.Logger.Info("Patching VirtualMachineSnapShot", "snap", vmSnapShot)
-		if err := patchHelper.Patch(ctx, vmSnapShot); err != nil {
+		vmSnapshotCtx.Logger.Info("Patching VirtualMachineSnapShot", "snap", vmSnapshot)
+		if err := patchHelper.Patch(ctx, vmSnapshot); err != nil {
 			if reterr == nil {
 				reterr = err
 			}
-			vmSnapShotCtx.Logger.Error(err, "patch failed")
+			vmSnapshotCtx.Logger.Error(err, "patch failed")
 		}
 	}()
 
-	if !vmSnapShot.DeletionTimestamp.IsZero() {
-		// Noop.
+	if !vmSnapshot.DeletionTimestamp.IsZero() {
+		// TODO: Handle snapshot deletion here.
 		return ctrl.Result{}, nil
 	}
 
-	if err := r.ReconcileNormal(vmSnapShotCtx); err != nil {
-		vmSnapShotCtx.Logger.Error(err, "Failed to reconcile VirtualMachineSnapShot")
+	if err := r.ReconcileNormal(vmSnapshotCtx); err != nil {
+		vmSnapshotCtx.Logger.Error(err, "Failed to reconcile VirtualMachineSnapShot")
 		return ctrl.Result{}, err
 	}
 
@@ -114,43 +114,35 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 
 func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) error {
 	ctx.Logger.Info("Reconciling VirtualMachineSnapshot")
-	vmSnapShot := ctx.VirtualMachineSnapshot
+	vmSnapshot := ctx.VirtualMachineSnapshot
 
 	// return early if snapshot is ready; nothing to do
 	if conditions.IsTrue(ctx.VirtualMachineSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition) {
 		return nil
 	}
 
-	ctx.Logger.Info("Fetching VirtualMachine from snapshot object", "vmSnapshot", vmSnapShot.Name)
+	ctx.Logger.Info("Fetching VirtualMachine from snapshot object", "vmSnapshot", vmSnapshot.Name)
 	vm := &vmopv1.VirtualMachine{}
-	objKey := client.ObjectKey{Name: vmSnapShot.Spec.VMRef.Name, Namespace: vmSnapShot.Namespace}
-	err := r.Get(ctx, objKey, vm)
-	if err != nil {
+	objKey := client.ObjectKey{Name: vmSnapshot.Spec.VMRef.Name, Namespace: vmSnapshot.Namespace}
+	if err := r.Get(ctx, objKey, vm); err != nil {
 		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
 		return err
 	}
-	ctx.VM = vm
 
+	ctx.VM = vm
 	if vm.Status.UniqueID == "" {
-		err = errors.New("VM hasn't been created and has no uniqueID")
-		return err
+		return errors.New("VM hasn't been created and has no uniqueID")
 	}
 
 	// vm object already set with snapshot reference
 	if vm.Spec.CurrentSnapshot != nil && vm.Spec.CurrentSnapshot.Name == ctx.VirtualMachineSnapshot.Name {
+		ctx.Logger.Info("VirtualMachine current snapshot already up to date", "spec.currentSnapshot", vm.Spec.CurrentSnapshot.Name)
 		return nil
 	}
 
-	// get VM again to ensure it's up-to-date.
-	err = r.Get(ctx, objKey, vm)
-	if err != nil {
-		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
-		return err
-	}
-
 	objRef := &vmopv1common.LocalObjectRef{
-		APIVersion: vmSnapShot.APIVersion,
-		Kind:       vmSnapShot.Kind,
+		APIVersion: vmSnapshot.APIVersion,
+		Kind:       vmSnapshot.Kind,
 		Name:       ctx.VirtualMachineSnapshot.Name,
 	}
 
@@ -163,5 +155,6 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) 
 			ctx.VirtualMachineSnapshot.Name, err)
 	}
 
+	ctx.Logger.Info("Successfully patched VirtualMachine's current snapshot reference", "vm.Name", vm.Name, "spec.currentSnapshot", vm.Spec.CurrentSnapshot.Name)
 	return nil
 }
--- a/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go
+++ b/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go
@@ -128,8 +128,33 @@ func unitTestsReconcile() {
 			})
 		})
 
-		When("vm ready with different/empty current snapshot ", func() {
+		When("vm ready with empty current snapshot ", func() {
 			BeforeEach(func() {
+				vm.Status.UniqueID = dummyVMUUID
+				initObjects = append(initObjects, vm)
+			})
+
+			It("returns success", func() {
+				Expect(err).ToNot(HaveOccurred())
+				objKey := types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace}
+				vmObj := &vmopv1.VirtualMachine{}
+				Expect(ctx.Client.Get(ctx, objKey, vmObj)).To(Succeed())
+
+				Expect(vmObj.Spec.CurrentSnapshot).To(Equal(&vmopv1common.LocalObjectRef{
+					APIVersion: vmSnapshot.APIVersion,
+					Kind:       vmSnapshot.Kind,
+					Name:       vmSnapshot.Name,
+				}))
+			})
+		})
+
+		When("vm ready with different current snapshot", func() {
+			BeforeEach(func() {
+				vm.Spec.CurrentSnapshot = &vmopv1common.LocalObjectRef{
+					APIVersion: vmSnapshot.APIVersion,
+					Kind:       vmSnapshot.Kind,
+					Name:       "dummy-diff-snapshot",
+				}
 				vm.Status.UniqueID = dummyVMUUID
 				initObjects = append(initObjects, vm)
 			})
--- a/pkg/providers/vsphere/virtualmachine/snapshot.go
+++ b/pkg/providers/vsphere/virtualmachine/snapshot.go
@@ -1,0 +1,89 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package virtualmachine
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+	vmopv1common "github.com/vmware-tanzu/vm-operator/api/v1alpha4/common"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+)
+
+// SnapshotArgs contains the options for createSnapshot.
+type SnapshotArgs struct {
+	VMCtx      pkgctx.VirtualMachineContext
+	VcVM       *object.VirtualMachine
+	VMSnapshot vmopv1.VirtualMachineSnapshot
+}
+
+func SnapshotVirtualMachine(args SnapshotArgs) (*types.ManagedObjectReference, error) {
+	obj := args.VMSnapshot
+	vm := args.VcVM
+	// Find snapshot by name
+	snapMoRef, _ := vm.FindSnapshot(args.VMCtx, obj.Name)
+	if snapMoRef != nil {
+		// TODO: Handle revert to snapshot. Need a way to compare currentSnapshot's moID
+		// 	with spec.currentSnap
+		//
+		// Update vm.status with currentSnapshot
+		updateVMStatusCurrentSnapshot(args.VMCtx, obj)
+		// Return early, snapshot found
+		return snapMoRef, nil
+	}
+
+	// If no snapshot was found, create it
+	args.VMCtx.Logger.Info("Creating Snapshot of VirtualMachine", "snapshot name", obj.Name)
+	snapMoRef, err := CreateSnapshot(args)
+	if err != nil {
+		args.VMCtx.Logger.Error(err, "failed to create snapshot for VM", "snapshot", obj.Name)
+		return nil, err
+	}
+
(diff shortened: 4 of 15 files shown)


pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -0,0 +1,87 @@
...
+	VcVM       *object.VirtualMachine
+	VMSnapshot vmopv1.VirtualMachineSnapshot
+}
+
+func SnapshotVirtualMachine(args SnapshotArgs) error {
+	obj := args.VMSnapshot
+	vm := args.VcVM
+	// find snapshot by name
+	snap, _ := vm.FindSnapshot(args.VMCtx, obj.Name)
+	if snap != nil {
```
aruneshpa: At this point, you have confirmed that spec.currentSnapshot != status.currentSnapshot.
So, this can be either a "take snapshot" or a "revert snapshot".  And you just found that spec.Snapshot is a valid, legit snapshot.
We need to check for restore here.  Drop a TODO here since you are not handling restore in this MR.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -0,0 +1,87 @@
...
+		args.VMCtx.Logger.Error(err, "failed to create snapshot for VM", "snapshot", obj.Name)
+		return err
+	}
+
+	// update vm.status with currentSnapshot
+	updateVMStatusCurrentSnapshot(args.VMCtx, obj)
+	return nil
+}
+
+func CreateSnapshot(vmCtx pkgctx.VirtualMachineContext, vcVM *object.VirtualMachine, name string, description string,
```
aruneshpa: Why not pass the snapshot context instead of passing all these parameters individually?

pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -740,3 +759,35 @@ func isVMPaused(vmCtx pkgctx.VirtualMachineContext) bool {
...
+	err := k8sClient.Get(vmCtx, objKey, snapShot)
+	if err != nil {
+		vmCtx.Logger.Error(err, "failed to get snapshot resource", "snapshot", objKey)
+		return err
+	}
+
+	snapPatch := ctrlclient.MergeFrom(snapShot.DeepCopy())
+	vmStatus := vmCtx.VM.Status
+
+	snapShot.Status.UniqueID = vmCtx.VM.Status.UniqueID
```
aruneshpa: The snapshot unique ID is the moref of the snapshot.

pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -740,3 +759,35 @@ func isVMPaused(vmCtx pkgctx.VirtualMachineContext) bool {
...
+	snapShot.Status = vmopv1.VirtualMachineSnapshotStatus{
+		UniqueID:   vmStatus.UniqueID,
+		PowerState: vmStatus.PowerState,
```
aruneshpa: This is the powerstate reported by the snapshot on the VC.  For example, if you take a snapshot of a VM that is powered on, but you exclude the memory, then the power state of the snapshot is set to "powered off".
I recommend adding a function to extract all VC details out of the snapshot managed object.  That has children etc as well.

pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -740,3 +759,35 @@ func isVMPaused(vmCtx pkgctx.VirtualMachineContext) bool {
...
+		Quiesced:   snapShot.Spec.Quiesce != nil,
+		// TODO: populate children
+	}
+
```
aruneshpa: Can you also add a TODO here to
- Signal CSI to sync their volume snapshot quota?
- Update our storage quota usage
- Update the VM storage quota usage

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return err
+	}
+
+	// wait for task to finish
+	taskInfo, err := t.WaitForResult(vmCtx)
+	if err != nil {
+		if taskInfo != nil {
+			vmCtx.Logger.V(5).Error(err, "create snapshot task failed", "taskInfo", taskInfo)
+		}
```
bryanv: missing return?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	ctx.VM = vm
+
+	if vm.Status.UniqueID == "" {
+		err = errors.New("VM hasn't been created and has no uniqueID")
+		return err
+	}
+
+	objRef := &vmopv1common.LocalObjectRef{
+		APIVersion: vmSnapShot.APIVersion,
```
bryanv: With the entire local object ref being used for the DeepEquals, is the desired behavior really to take another snapshot when we rev our API version?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,170 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) error {
+	ctx.Logger.Info("Reconciling VirtualMachineSnapshot")
+	vmSnapShot := ctx.VirtualMachineSnapshot
+
+	// return early if phase was already set; nothing to do
+	if conditions := ctx.VirtualMachineSnapshot.Status.Conditions; len(conditions) != 0 {
```
bryanv: Don't need the len check here. There is a conditions Get() helper too. Should this be checking that the condition is true?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
sreyasn: makes sense. only checking the Name now

pkg/providers/vsphere/vmprovider_vm_utils.go
sreyasn: SG. added the TODO in vmprovider_vm.go after the PatchStatus() method call.

pkg/providers/vsphere/vmprovider_vm_utils.go
sreyasn: Wasn't able to find the PowerState field yet. I think you mean we'll parse the VM configInfo from the Snapshot MO and get this info?. 

Anyways I've made it a TODO along with the child snapshot population. Can be handled in a separate PR.

pkg/providers/vsphere/vmprovider_vm_utils.go
sreyasn: done

controllers/controllers.go
```diff
@@ -73,5 +74,9 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		}
 	}
 
+	if err := virtualmachinesnapshot.AddToManager(ctx, mgr); err != nil {
```
aruneshpa, line 78: Can you add a TODO here to guard this with a capability?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,167 @@
...
+
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinesnapshots,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinesnapshots/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines/status,verbs=get;update;patch
+
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+	ctx = pkgcfg.JoinContext(ctx, r.Context)
+
+	vmSnapShot := &vmopv1.VirtualMachineSnapshot{}
```
aruneshpa: Here and elsewhere: r/snapShot/snapshot/g

```suggestion
	vmSnapshot := &vmopv1.VirtualMachineSnapshot{}
```
(casing)

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,167 @@
...
+		if err := patchHelper.Patch(ctx, vmSnapShot); err != nil {
+			if reterr == nil {
+				reterr = err
+			}
+			vmSnapShotCtx.Logger.Error(err, "patch failed")
+		}
+	}()
+
+	if !vmSnapShot.DeletionTimestamp.IsZero() {
+		// Noop.
```
aruneshpa: This is not a no-op.  TODO here to handle VM snapshot deletion (tracked by VMSVC-2706)

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// return early if snapshot is ready; nothing to do
+	if conditions.IsTrue(ctx.VirtualMachineSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition) {
+		return nil
+	}
+
+	ctx.Logger.Info("Fetching VirtualMachine from snapshot object", "vmSnapshot", vmSnapShot.Name)
+	vm := &vmopv1.VirtualMachine{}
+	objKey := client.ObjectKey{Name: vmSnapShot.Spec.VMRef.Name, Namespace: vmSnapShot.Namespace}
+	err := r.Get(ctx, objKey, vm)
```
aruneshpa: Nit: Here and elsewhere, combine these two lines to `if err := <>; err != nil {}`

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	ctx.VM = vm
+
+	if vm.Status.UniqueID == "" {
+		err = errors.New("VM hasn't been created and has no uniqueID")
+		return err
+	}
+
+	// vm object already set with snapshot reference
+	if vm.Spec.CurrentSnapshot != nil && vm.Spec.CurrentSnapshot.Name == ctx.VirtualMachineSnapshot.Name {
+		return nil
```
aruneshpa, line 140: Log something here?
Because this is a snapshot resource with not-ready condition and the VM spec points to this snapshot already.  This won't be a regular case.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	// get VM again to ensure it's up-to-date.
+	err = r.Get(ctx, objKey, vm)
```
aruneshpa: Hmm.  We only need to patch one field of the VM, so there's little chance of a conflict when patching.  Was there another reason to re-fetch the VM?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	// patch vm resource with the spec.currentSnapshot
+	vmPatch := client.MergeFrom(vm.DeepCopy())
+	vm.Spec.CurrentSnapshot = objRef
+	if err := r.Patch(ctx, vm, vmPatch); err != nil {
+		return fmt.Errorf(
+			"failed to patch VM resource %s with current snapshot %s: %w", objKey,
+			ctx.VirtualMachineSnapshot.Name, err)
+	}
```
aruneshpa, line 156: Log something here to indicate that VM snapshot controller has finished its job of patching the VM spec?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go
```diff
@@ -0,0 +1,168 @@
...
+				initObjects = append(initObjects, vm)
+			})
+
+			It("returns failure", func() {
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("VM hasn't been created and has no uniqueID"))
+			})
+		})
+
+		When("vm ready with different/empty current snapshot ", func() {
```
aruneshpa: Consider splitting this in two tests, one for empty, and other one for replacing the spec.currentSnapshot?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1102,6 +1111,47 @@ func (vs *vSphereVMProvider) reconcilePowerState(
 	return nil
 }
 
+func (vs *vSphereVMProvider) reconcileSnapshot(
+	vmCtx pkgctx.VirtualMachineContext,
+	vcVM *object.VirtualMachine) error {
+	// no need for snapshot or current snapshot is up-to-date
```
aruneshpa: Something along the lines of..
```suggestion
	// Skip creating a new snapshot if the VM already points to the desired snapshot.
```
?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -0,0 +1,89 @@
...
+	VMSnapshot vmopv1.VirtualMachineSnapshot
+}
+
+func SnapshotVirtualMachine(args SnapshotArgs) (*types.ManagedObjectReference, error) {
+	obj := args.VMSnapshot
+	vm := args.VcVM
+	// find snapshot by name
+	snapMoRef, _ := vm.FindSnapshot(args.VMCtx, obj.Name)
+	if snapMoRef != nil {
+		// TODO: Handle revert to snapshot
```
aruneshpa: We also need to handle the scenario where the spec.currentSnapshot exists on VC VM, but because we only compare it with `vm.Status.CurrentSnapshot` in vmprovider_vm.go L1118, we fell through in this code path.
```
if vmCtx.VM.Spec.CurrentSnapshot == nil || reflect.DeepEqual(vmCtx.VM.Spec.CurrentSnapshot, vmCtx.VM.Status.CurrentSnapshot) {
	return nil
}
```
So, basically compare VM's [currentSnapshot](https://developer.broadcom.com/xapis/vsphere-web-services-api/8.0u3/vim.vm.SnapshotInfo.html?scrollString=currentSnapshot) with the spec.currentSnapshot.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -0,0 +1,89 @@
...
+		// update vm.status with currentSnapshot
+		updateVMStatusCurrentSnapshot(args.VMCtx, obj)
+		// return early, snapshot found
+		return snapMoRef, nil
+	}
+
+	// if no snapshot was found, create it
```
aruneshpa: Here and elsewhere, consider capitalizing the comments.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// wait for task to finish
+	taskInfo, err := t.WaitForResult(args.VMCtx)
+	if err != nil {
+		if taskInfo != nil {
+			args.VMCtx.Logger.V(5).Error(err, "create snapshot task failed", "taskInfo", taskInfo)
+		}
```
aruneshpa: Not sure when the taskInfo would be nil and `err` will be not nil, but to handle that maybe log this line unconditionally?

pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -740,3 +759,30 @@ func isVMPaused(vmCtx pkgctx.VirtualMachineContext) bool {
 	delete(vmCtx.VM.Labels, vmopv1.PausedVMLabelKey)
 	return false
 }
+
+func PatchSnapshotStatus(vmCtx pkgctx.VirtualMachineContext, k8sClient ctrlclient.Client,
+	objName string, objNs string, snapMoRef *vimtypes.ManagedObjectReference) error {
+	snapShot := &vmopv1.VirtualMachineSnapshot{}
+	objKey := ctrlclient.ObjectKey{Name: objName, Namespace: objNs}
+	// get snapshot again to ensure it's up-to-date.
+	err := k8sClient.Get(vmCtx, objKey, snapShot)
```
aruneshpa: Again, just curious, do we need this?  The Status.uniqueID, and status.Quiesced are not supposed to be patched/modified by any one else other than this controller.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
sreyasn: Any changes to the VM object/CR during this timeframe (127-144) can cause patch conflicts. Better to refetch here to prevent issues.

pkg/providers/vsphere/virtualmachine/snapshot.go
sreyasn: Hmm. The VM.SnapshotInfo.CurrentSnapshot MoRef can give us the MOID of the snapshot but our spec.currentSnapshot would be still the name (not the MOID). So, We don't have a way to compare the unique vcenter MOIDs. 

Currently the way this is handled since revert is a TODO. When the snapshot is already present, it just patches the status of the VM. I agree though, The problem you mentioned would arise, when someone deletes the exists snapshot CR and creates a new one with the same name. This might have to be punted for when we handle RevertToSnapshot() here. Let me add a TODO if that sounds alright.

pkg/providers/vsphere/vmprovider_vm_utils.go
sreyasn: That's true. I can get rid of it but I'm coding defensively for any future changes that might affect this object. Let me know

pkg/providers/vsphere/virtualmachine/snapshot.go
sreyasn: done

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -0,0 +1,166 @@
...
+		return errors.New("VM hasn't been created and has no uniqueID")
+	}
+
+	// vm object already set with snapshot reference
+	if vm.Spec.CurrentSnapshot != nil && vm.Spec.CurrentSnapshot.Name == ctx.VirtualMachineSnapshot.Name {
+		ctx.Logger.Info("VirtualMachine current snapshot already up to date", "spec.currentSnapshot", vm.Spec.CurrentSnapshot.Name)
+		return nil
+	}
+
+	// get VM again to ensure it's up-to-date.
```
aruneshpa: Like I said in the other comment, is this necessary given that we are patching just the snapshot.vmRef?  There's a tiny tiny chance that two Snapshot CRs are created for a VM simultaneously and we have two reconciler threads trying to modify the VM's spec.currentSnapshot.  But re-fetching the VM doesn't guarantee against that anyway.
To mitigate this, we will add a validating webhook that will reject any Snapshot CRs if there's a pending snapshot for a VM (snapshot CR in not-ready state).  The race _still_ doesn't go away, but the chances of it happening will be greatly reduced.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -823,6 +824,14 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 		reconcileErr = getReconcileErr("power state", reconcileErr, err)
 	}
 
+	// create snapshots of VM if necessary
```
aruneshpa:
```suggestion
	// Reconcile the current snapshot of this VM.
```

Because this can essentially be:
- take a snapshot
- don't do anything (just patch status)
- revert

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -823,6 +824,14 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+	if err := vs.reconcileSnapshot(vmCtx, vcVM); err != nil {
+		if pkgerr.IsNoRequeueError(err) {
+			return err
+		}
+		return fmt.Errorf("failed to reconcile snapshots: %w", err)
```
aruneshpa:
```suggestion
		return fmt.Errorf("failed to reconcile the current snapshot: %w", err)
```

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1102,6 +1111,47 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+		return err
+	}
+
+	snapArgs := virtualmachine.SnapshotArgs{
+		VMCtx:      vmCtx,
+		VcVM:       vcVM,
+		VMSnapshot: vmSnapshot,
+	}
+
+	vmCtx.Logger.Info("Taking a snapshot of VM Service managed VM")
```
aruneshpa: vvNit: ALL VMs that we are reconciling are VM Service managed VMs.  Just say `Creating a new snapshot of the virtual machine` or something like that.

pkg/providers/vsphere/vmprovider_vm_utils.go
aruneshpa: I think it is okay to remove this.  We can always add it if we start to see conflicts.  I don't expect multiple concurrent snapshot operations on VMs over and over anyway.

pkg/providers/vsphere/virtualmachine/snapshot.go
aruneshpa: Yeah, TODO is fine for now.  But it would be simple.  You can get the name of the snapshot from the Moref of the snapshot and then compare that with the name of the snapshot object's name.

aruneshpa (approved):
Approved with nits.  We can circle back and handle the TODOs later.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
sreyasn: ack. removed

pkg/providers/vsphere/vmprovider_vm.go
sreyasn: ack

pkg/providers/vsphere/virtualmachine/snapshot.go
sreyasn: > You can get the name of the snapshot from the Moref of the snapshot and then compare that with the name of the snapshot object's name

ah. I thought I could get the name too. But from what I found, the MoRef gives access to 
- a ref.Value() (the MOID "snap-01" or similar) 
- the ref.Type() ("VirtualMachineSnapshot") and 
- the ref.ServerGUID() . 

I wasn't sure we can fetch the name from here?. I've added TODO to handle this as a follow-up.

PR #1014, opened 2025-06-10
✨ Reconcile VM schema upgrade

akutz:

This patch provides an officially supported path for upgrading a VM Op VM's schema to match the latest version supported by VM Op.

Importantly, this change also disallows all updates to the VM Op VM object until such time the VM object is updated. This is critical as it prevents any changes that might interrupt or be in conflict with the changes occurring as part of the upgrade.

Fixes `NA`

I ran this by @zjs for VKS consideration as well as Brandon Nelson, and they both agree this change makes sense and are good with it.

```release-note
Do not allow modifying VM objects that are not upgraded.
```


akutz:
The vulncheck-go failed due to the go version. Merging anyway.

PR #1015, opened 2025-06-12
:sparkles: Handle VM Class instances during VM class reconcile

aruneshpa:

Reconcile VM classes to create VM class instance.  An instance is created with an annotation containing the hash of the ConfigSpec.  Additionally, the instance object also has an OwnerReference to the corresponding class.  The controller ensures that at a time, only the instance mapping to the latest ConfigSpec hash is marked as active.  The active / inactive property is represented by a label to allow clients to query efficiently.

```release-note
Handle VM Class instances during VM class reconcile
```

### DIFF
--- a/api/v1alpha4/virtualmachineclassinstance_types.go
+++ b/api/v1alpha4/virtualmachineclassinstance_types.go
@@ -8,6 +8,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// VMClassInstanceActiveLabelKey represents if a
+	// VirtualMachineClassInstance can be used to deploy a new virtual
+	// machine, or resize an existing one. The presence of this label
+	// on a VirtualMachineClassInstance marks the instance as active.
+	// The value of this key does not matter.
+	VMClassInstanceActiveLabelKey = GroupName + "/active"
+)
+
 // VirtualMachineClassInstanceSpec defines the desired state of VirtualMachineClassInstance.
 // It is a composite of VirtualMachineClassSpec.
 type VirtualMachineClassInstanceSpec struct {
@@ -24,6 +33,7 @@ type VirtualMachineClassInstanceStatus struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="CPU",type="string",JSONPath=".spec.hardware.cpus"
 // +kubebuilder:printcolumn:name="Memory",type="string",JSONPath=".spec.hardware.memory"
+// +kubebuilder:printcolumn:name="Active",type="string",JSONPath=".metadata.labels['vmoperator\\.vmware\\.com/active']"
 
 // VirtualMachineClassInstance is the schema for the virtualmachineclassinstances API and
 // represents the desired state and observed status of a virtualmachineclassinstance
--- a/controllers/virtualmachineclass/virtualmachineclass_controller.go
+++ b/controllers/virtualmachineclass/virtualmachineclass_controller.go
@@ -6,20 +6,27 @@ package virtualmachineclass
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
 
+	xxhash "github.com/cespare/xxhash/v2"
 	"github.com/go-logr/logr"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	pkgconst "github.com/vmware-tanzu/vm-operator/pkg/constants"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
+	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
 )
 
@@ -69,6 +76,8 @@ type Reconciler struct {
 
 // +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachineclasses,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachineclasses/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachineclassinstances,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachineclassinstances/status,verbs=get;update;patch
 
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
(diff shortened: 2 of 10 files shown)


controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 }
 
 func (r *Reconciler) ReconcileNormal(vmClassCtx *pkgctx.VirtualMachineClassContext) error {
+	vmClassCtx.Logger.Info("Arunesh: reconcilenormal",
```
bryanv: Debug leftover

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		class.Spec.ReservedProfileID != "",
+		spec,
+	}
+
+	return getVimTypeHash(data)
+}
+
+// reconcileInstance creates a VirtualMachineClassInstance for the
+// given VirtualMachineClass. The name of the instance object contains
+// the a hash of SHA256 digest of the ConfigSpec (and other relevant
```
bryanv: sha256 in the comment here looks stale

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+	_, err = ctrlutil.CreateOrPatch(ctx, r.Client, vmClassInstance, func() error {
+		if err := ctrlutil.SetControllerReference(
+			ctx.VMClass, vmClassInstance, r.Scheme()); err != nil {
+
+			return err
+		}
+
+		vmClassInstance.Spec.VirtualMachineClassSpec = ctx.VMClass.Spec
+
+		vmClassInstance.Annotations = map[string]string{
```
bryanv: Really want to remove all other annotations here?

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+			pkgconst.VirtualMachineClassHashAnnotationKey: hash,
+		}
+
+		vmClassInstance.Labels = map[string]string{
```
bryanv: Ditto for labels here

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+			return fmt.Errorf("failed to check if VM class instance is owned by the class. err: %w", err)
+		}
+
+		if !owned {
+			ctx.Logger.Info("Skipping this instance since it is not owned by VM class")
+			continue
+		}
+
+		// Create patch helper before modifying the instance
+		patchHelper, err := patch.NewHelper(&instance, r.Client)
```
bryanv: I don't think you need all of what the patcher helper does here - just use Client.Patch instead

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+				if instance.Labels == nil {
+					instance.Labels = make(map[string]string)
+				}
+				instance.Labels[vmopv1.VMClassInstanceActiveLabelKey] = ""
+			}
+		} else {
+			// Mark the instance as inactive since its hash does not
+			// match the hash of the current version of VM class.
+			ctx.Logger.Info("Marking instance as inactive")
+			if instance.Labels == nil {
```
bryanv: If there are no labels then there is nothing needed to be done - we don't need to assign an empty map

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+				instance.Labels = make(map[string]string)
+			} else {
+				// Create a new map without the active label
+				newLabels := make(map[string]string)
+				for k, v := range instance.Labels {
+					if k != vmopv1.VMClassInstanceActiveLabelKey {
```
bryanv: Just `delete()` instead? If you're going to stick with the patch helper there is no need to make a labels copy

controllers/virtualmachineclass/virtualmachineclass_controller.go
aruneshpa: Fixed.

controllers/virtualmachineclass/virtualmachineclass_controller.go
aruneshpa: Fixed.

api/v1alpha4/virtualmachineclassinstance_types.go
```diff
@@ -8,6 +8,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// VMClassInstanceActiveLabelKey represents if a
+	// VirtualMachineClassInstance can be used to deploy a new virtual
+	// machine, or resize an existing one. The presence of this label
+	// on a VirtualMachineClassInstance marks the instance as active.
+	// The value of this key does not matter.
+	VMClassInstanceActiveLabelKey = GroupName + "/active"
```
dilyar85, line 17: Is this label user facing? If not, I believe the guidance is to avoid defining it in the API folder (move it to the `pkg/constants.go` file instead)?

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -6,20 +6,28 @@ package virtualmachineclass
...
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
 
+	xxhash "github.com/cespare/xxhash/v2"
 	"github.com/go-logr/logr"
+
```
dilyar85: Nit: extra empty line

controllers/virtualmachineclass/virtualmachineclass_controller.go
```diff
@@ -111,5 +121,176 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		if val, ok := instance.Annotations[pkgconst.VirtualMachineClassHashAnnotationKey]; ok && val == hash {
+			// Hash value on the VM class instance matches with the
+			// hash we just calculated. Mark this class as active.
+			if _, ok := instance.Labels[vmopv1.VMClassInstanceActiveLabelKey]; !ok {
+				ctx.Logger.Info("Marking instance as active")
+				if instance.Labels == nil {
+					instance.Labels = make(map[string]string)
+				}
+				instance.Labels[vmopv1.VMClassInstanceActiveLabelKey] = ""
+			}
```
dilyar85, line 267: Just to clarify, is it possible for multiple instances to match the hash? If not, this block seems unnecessary since we always create or patch the matching instance above.

controllers/virtualmachineclass/virtualmachineclass_controller_intg_test.go
```diff
@@ -78,7 +87,72 @@ func intgTestsReconcile() {
...
+				g.Expect(instance.Labels[vmopv1.VMClassInstanceActiveLabelKey]).To(Equal(""))
+
+				// Ensure owner reference is set
+				g.Expect(instance.OwnerReferences).To(HaveLen(1))
+				g.Expect(instance.OwnerReferences[0].Name).To(Equal(vmClass.Name))
+
+				// Ensure that the class has the ConfigHash condition set.
+				class := &vmopv1.VirtualMachineClass{}
+				err = ctx.Client.Get(ctx, client.ObjectKeyFromObject(vmClass), class)
+				g.Expect(err).ToNot(HaveOccurred())
```
dilyar85: Looks like the condition check part is missing here?

controllers/virtualmachineclass/virtualmachineclass_controller_intg_test.go
```diff
@@ -78,7 +87,72 @@ func intgTestsReconcile() {
 			Expect(err == nil || apierrors.IsNotFound(err)).To(BeTrue())
 		})
 
-		It("noop", func() {
+		It("Should reconcile VirtualMachineClassInstance", func() {
+			By("validating that the instance was created")
+			Eventually(func(g Gomega) {
+				var list vmopv1.VirtualMachineClassInstanceList
+				err := ctx.Client.List(ctx, &list, client.InNamespace(vmClass.Namespace))
```
dilyar85: Suggestion (here and below): `g.Expect(ctx.Client.List(ctx, &list, client.InNamespace(vmClass.Namespace))).To(Succeed())`

controllers/virtualmachineclass/virtualmachineclass_controller_intg_test.go
aruneshpa: Yeah, I decided to yank out the Condition setting and validation fro this PR.  Updated the tests.

api/v1alpha4/virtualmachineclassinstance_types.go
aruneshpa: So, this is indeed user facing.  This is the only way for a user to know if a class is active.  In fact, this is part of the printer columns for a VM class.  CCI will also use this label to filter out the instances that are active.

controllers/virtualmachineclass/virtualmachineclass_controller.go
aruneshpa: There shouldn't be multiple such instances _ideally_ since the name of the instance is VM class name + hash -- and class name is namespace scoped.  However, I don't see any harm in defensively trying to handle this since this can basically mean that the name of the instance doesn't matter at all.  The hash annotation is the contract.

api/v1alpha4/virtualmachineclassinstance_types.go
dilyar85: Thanks for the context, marked this one resolved.

dilyar85 (approved):
LGTM!

PR #1017, opened 2025-06-13
🐛 Fix PCI DVX devs on update

akutz:

This patch fixes how PCI Passthrough devices are handled when their backing is VirtualPCIPassthroughDvxBackingInfo. Previously they would be removed/added during a pre-poweron reconfigure due to the lack of proper device matching. This patch ensures if nothing has changed, the existing devices are not removed/added.

Fixes `NA`

```release-note
Do not remove/add PCI DVX devices on update if nothing has changed.
```

### DIFF
--- a/pkg/util/resize/backings.go
+++ b/pkg/util/resize/backings.go
@@ -6,6 +6,7 @@ package resize
 
 import (
 	vimtypes "github.com/vmware/govmomi/vim25/types"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 )
 
 func MatchVirtualDeviceDeviceBackingInfo(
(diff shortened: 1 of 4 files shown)


pkg/util/resize/backings.go
```diff
@@ -309,3 +310,29 @@ func MatchVirtualPCIPassthroughDynamicBackingInfo(
...
+	}
+
+	if a.DeviceClass == b.DeviceClass {
+		if len(a.ConfigParams) == len(b.ConfigParams) {
+			am := make(map[string]any, len(a.ConfigParams))
+			bm := make(map[string]any, len(b.ConfigParams))
+			for i := 0; i < len(a.ConfigParams); i++ {
+				aov := a.ConfigParams[i].GetOptionValue()
+				bov := b.ConfigParams[i].GetOptionValue()
+				am[aov.Key] = aov.Value
```
bryanv, line 330: Check aov/bov for `nil` just in case?

akutz: Doug told me in the past it cannot be `nil` without something contrived, but sure.

bryanv: OK. I guess we check for nil in most places but good to know

PR #1023, opened 2025-06-18
:seedling: Mutating and validating webhook changes for spec.class and spec.className

aruneshpa:

This change introduces the webhook mutations and validations for spec.className and spec.class of a VM.

The mutating webhook changes include:
- If spec.class is specified, but spec.className is omitted, the mutating webhook populates the spec.className with the class that owns the instance
- If spec.className is specified but spec.class is omitted, the mutating webhook picks the active instance and populates the spec.class with the instance.
- If both are specified, then mutating webhook is a pass-through.  The validating webhook will validate the request.
- If neither of them are specified, then mutating webhook is a pass-through.

The validating webhook changes include:
- If spec.className and spec.class both have been specified, they must point to the same underlying class.

Additionally, the mutating webhook also applies an annotation containing the last resized annotation name so later in the resize workflow we can accurately evaluate if a resize has been requested. This is the same behavior with className.

```release-note
Mutating and validating webhook changes for spec.class and spec.className
```

### DIFF
--- a/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
+++ b/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
@@ -46,6 +46,8 @@ const (
 	defaultInterfaceName   = "eth0"
 	defaultNamedNetwork    = "VM Network"
 	defaultCdromNamePrefix = "cdrom"
+
+	vmclassKind = "VirtualMachineClass"
 )
 
 // +kubebuilder:webhook:path=/default-mutate-vmoperator-vmware-com-v1alpha4-virtualmachine,mutating=true,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachines,verbs=create;update,versions=v1alpha4,name=default.mutating.virtualmachine.v1alpha4.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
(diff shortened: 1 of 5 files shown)


abaruni (approved):

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -103,6 +105,110 @@ type mutator struct {
...
+			instanceList := &vmopv1.VirtualMachineClassInstanceList{}
+			if err := c.List(ctx, instanceList, ctrlclient.InNamespace(vm.Namespace)); err != nil {
+				return false, err
+			}
+
+			instancesForClass := []vmopv1.VirtualMachineClassInstance{}
+			for _, instance := range instanceList.Items {
+				for _, ownerClass := range instance.OwnerReferences {
+					if ownerClass.Name == className {
+						instancesForClass = append(instancesForClass, instance)
```
bryanv: Is having this be a slice of either 0 or 1 elements for some later work?

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -103,6 +105,110 @@ type mutator struct {
...
+	if className != "" {
+		// If a class is specified, but no instance has been
+		// specified, pick the active instance.
+		if classInstance == nil {
+
+			// TODO: AKP: This is inefficient. Each instance should
+			// have a link to its parent class.
+			// Fetch all the instances and filter the one owned by this class
+			instanceList := &vmopv1.VirtualMachineClassInstanceList{}
+			if err := c.List(ctx, instanceList, ctrlclient.InNamespace(vm.Namespace)); err != nil {
```
bryanv: Include the label selector for `VMClassInstanceActiveLabelKey` here?

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
aruneshpa: By default, at most only one VM class instance for a given class will be active.  So, when we find one, we don't have to keep going.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -103,6 +105,104 @@ type mutator struct {
...
+				instanceList,
+				ctrlclient.InNamespace(vm.Namespace),
+				ctrlclient.MatchingLabels{vmopv1.VMClassInstanceActiveLabelKey: ""},
+			); err != nil {
+				return false, err
+			}
+
+			for _, instance := range instanceList.Items {
+				for _, ownerClass := range instance.OwnerReferences {
+					if ownerClass.Name == className {
```
bryanv: Check the Kind here too?

bryanv (approved):

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -103,6 +105,106 @@ type mutator struct {
...
+	// If class and class instance both are set, then pass through to the
+	// validation webhook. Nothing to mutate.
+	if className != "" && classInstance != nil {
+		return false, nil
+	}
+
+	if className != "" {
+		// If a class is specified, but no instance has been
+		// specified, pick the active instance.
+		if classInstance == nil {
```
bryanv, line 145: `classInstance` will always be `nil` at this point because of the check on L138

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -103,6 +105,106 @@ type mutator struct {
...
+						return true, nil
+					}
+				}
+			}
+		}
+	} else {
+		// If an instance is specified in spec.class, but className is
+		// omitted, fetch the VM class pointed by the instance and
+		// populate spec.className.
+		if vm.Spec.Class != nil {
```
bryanv, line 180: Similarly, I don't think check is strictly needed because of L127. If you're going to keep this, `classInstance` is used before instead of spelling out `vm.Spec.Class` (and maybe don't shadow classInstance below?)

bryanv (approved):

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
aruneshpa: I am going to handle these in a followup.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
aruneshpa: I am going to handle these in a followup.

PR #1029, opened 2025-06-21
🌱 Ensure ctr load image with correct image bundle path

lubronzhan:

By default, the image bundle created by `make image-build-amd64` is stored in `artifacts/vmoperator-controller-linux_amd64.tar`

Then running `./hack/deploy-wcp.sh` command would like
```
./hack/deploy-wcp.sh -v $VC_URL -V $VC_ROOT_PASSWORD artifacts/vmoperator-controller-linux_amd64.tar
```
Then it will fail with below error
```
ctr: open artifacts/vmoperator-controller-linux_amd64.tar: no such file or directory
```
This PR use absolute image bundle when running ctr command

Fixes #

```release-note

```

### DIFF
--- a/hack/deploy-wcp.sh
+++ b/hack/deploy-wcp.sh
@@ -5,7 +5,7 @@ set -o nounset
 set -o pipefail
 
 USAGE="
-Usage: ${0} [[-s SV_IP,SV_IP,SV_IP [-S SV_PASSWORD]] | [-v VC_IP [-V VC_SSH_PASSWORD]] | [-T testbedInfo.json]] [-c cluster] image.tar
+Usage: ${0} [[-s SV_IP,SV_IP,SV_IP [-S SV_PASSWORD]] | [-v VC_IP [-V VC_SSH_PASSWORD]] | [-T testbedInfo.json]] [-c cluster] [-t build version] image.tar
 
 Loads the VM Operator container image to the Supervisor control plane VMs,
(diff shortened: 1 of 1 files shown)


hack/deploy-wcp.sh
```diff
@@ -176,8 +176,9 @@ function sv_get_cp_ips() {
 function sv_copy_and_load_image() {
     local svip=$1 image=$2
 
+    remote_image_path="/root/$(basename $image)"
     sv_cp_scp_cmd "$svip" "$image"
-    sv_cp_ssh_cmd "$svip" "ctr -n=k8s.io images import '$image'"
```
bryanv, line 180: Something like `${image##*/}` instead?

This standalone change is largely moot - the `image-build-amd64` target doesn't use the tag that this script assumes: see the comment around IMAGE_REF. I usually just podman save with the expected ref but cool if we want to start to extract the ref from the image.

lubronzhan: IC, didn't know that flow.
BTW, I added a flag to specify the build_number to this script.

lubronzhan: If you think if it's too much I can try just pick up the latest image from `podman image list` output

lubronzhan:
Superceded by https://github.com/vmware-tanzu/vm-operator/pull/1036

PR #1032, opened 2025-06-25
Support for deleting snapshot

lubronzhan:

Fixes #

```release-note

```


pkg/providers/vm_provider_interface.go
```diff
@@ -63,4 +63,8 @@ type VirtualMachineProviderInterface interface {
 
 	// VSphereClient returns the provider's vSphere client.
 	VSphereClient(context.Context) (*client.Client, error)
+
+	// DeleteSnapshot deletes a snapshot from a virtual machine.
+	DeleteSnapshot(ctx context.Context, snapshot *vmopv1.VirtualMachineSnapshot,
```
lubronzhan: Not using virtualmachine.SnapshotArgs as an input here, otherwise will result as import cycle

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +176,78 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+func (r *Reconciler) deleteSnapshotFromVSphere(ctx *pkgctx.VirtualMachineSnapshotContext) error {
+	snapshot := ctx.VirtualMachineSnapshot
+	vm := &vmopv1.VirtualMachine{}
+	objKey := client.ObjectKey{Name: snapshot.Spec.VMRef.Name, Namespace: snapshot.Namespace}
+	if err := r.Get(ctx, objKey, vm); err != nil {
+		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
+		return err
+	}
+	// TODO: set removeChildren and consolidate to false by default for now.
+	if err := r.VMProvider.DeleteSnapshot(ctx, snapshot, vm, false, nil); err != nil {
```
lubronzhan: Set `removeChildren ` and `consolidate ` to false for now

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -114,8 +136,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 
 func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) error {
 	ctx.Logger.Info("Reconciling VirtualMachineSnapshot")
-	vmSnapshot := ctx.VirtualMachineSnapshot
 
+	vmSnapshot := ctx.VirtualMachineSnapshot
```
bryanv: Move this down later since that is when this var name is actually used, or update things around here to use this var

Also, the shortname var name for `ctx.VirtualMachineSnapshot` is `vmSnapshot` and `snapshot` in other places in this MR. It would be nice to consistently use one or the other, or always the ctx field instead.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_intg_test.go
```diff
@@ -116,3 +106,142 @@ func intgTestsReconcile() {
...
+			textlogger.NewLogger(textlogger.NewConfig(
+				textlogger.Verbosity(2),
+				textlogger.Output(GinkgoWriter),
+			)))
+
+		ctx = pkgcfg.WithContext(ctx, pkgcfg.Default())
+		provider = providerfake.NewVMProvider()
+		provider.Lock()
+
+		provider.DeleteSnapshotFn = func(_ context.Context, _ *vmopv1.VirtualMachineSnapshot, _ *vmopv1.VirtualMachine, _ bool, _ *bool) error {
```
bryanv: This would already be the default, no?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go
```diff
@@ -188,6 +162,214 @@ func unitTestsReconcile() {
...
+		When("A Snapshot is marked for deletion", func() {
+			BeforeEach(func() {
+				now = metav1.Now()
+				vmSnapshot.DeletionTimestamp = &now
+				initObjects = append(initObjects, vm)
+			})
+
+			It("returns success", func() {
+				fakeVMProvider.Lock()
+				fakeVMProvider.DeleteSnapshotFn = func(_ context.Context, _ *vmopv1.VirtualMachineSnapshot, _ *vmopv1.VirtualMachine, _ bool, _ *bool) error {
```
bryanv: It isn't clear to me why this is done in the body of an It() - Reconcile() has already been called by this point

For a unit test with a controller, you typically shouldn't need to Lock/Unlock the fake provider

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +181,121 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+		ctx.Logger.V(5).Info("parent snapshot not found")
+		return nil, nil
+	}
+	ctx.Logger.V(5).Info("parent snapshot found", "parent", parent.Name)
+
+	// remove current snapshot from parent's children
+	parentPatch := client.MergeFrom(parent.DeepCopy())
+	for i, child := range parent.Status.Children {
+		if child.Name == vmSnapshot.Name {
+			parent.Status.Children = append(parent.Status.Children[:i], parent.Status.Children[i+1:]...)
```
bryanv: Sometimes a little nicer to use the slices package - https://pkg.go.dev/slices#Delete

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +182,132 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+				controllerutil.RemoveFinalizer(snapshot, Finalizer)
+				return nil
+			}
+			ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
+			return err
+		}
+
+		ctx.VM = vm
+		if err := r.deleteSnapshotFromVSphere(ctx); err != nil {
+			if strings.Contains(err.Error(), fmt.Sprintf(vsphere.VirtualMachineNotFoundErrorf, vm.Name)) {
```
bryanv: Please define a sentinel error and use errors.Is() (or like encode this as a bool param return value from the provider interface)

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +182,132 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+				ctx.Logger.Info("VirtualMachine not found in VC, assuming the VM is deleted, remove finalizer")
+				controllerutil.RemoveFinalizer(snapshot, Finalizer)
```
bryanv: Shouldn't this still update the parent, etc below? If like VMOP crashes after deleting the snapshot, is the next reconcile going to leave things in an inconscient state?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +182,132 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+	if err := r.Status().Patch(ctx, parent, parentPatch); err != nil {
+		return nil, fmt.Errorf("failed to patch parent snapshot %s with children: %w", parent.Name, err)
+	}
+	return parent, nil
+}
+
+func (r *Reconciler) getParentSnapshot(ctx *pkgctx.VirtualMachineSnapshotContext, snapshot *vmopv1.VirtualMachineSnapshot) (*vmopv1.VirtualMachineSnapshot, error) {
+	allSnapshots := &vmopv1.VirtualMachineSnapshotList{}
+	if err := r.List(ctx, allSnapshots, &client.ListOptions{
+		Namespace: snapshot.Namespace,
```
bryanv: @aruneshpa Do we expect this to potentially grow large enough that we'll want to be able to select via label?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +182,132 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+			parent.Status.Children = append(parent.Status.Children[:i], parent.Status.Children[i+1:]...)
+			break
+		}
+	}
+	children := vmSnapshot.Status.Children
+	if children != nil {
+		ctx.Logger.V(5).Info("Add children snapshots of current snapshot to parent's children")
+		// merge current's children to parent's children.
+		// no duplicates check here since we assume each snapshot can only have one parent.
+		parent.Status.Children = append(parent.Status.Children, children...)
```
bryanv: If you fix to make the deleteSnapshotFromVSphere() idempotent, then this isn't quite right: if we crash after the patch below but have not removed the finalizer, the next time through will append the same children.

lubronzhan:
Ok gonna move part of DeleteSnapshot logic from https://github.com/vmware-tanzu/vm-operator/pull/1032/files#diff-762a752437a2e38bc94742d215595fb345b1e96bf8572c6b22dcfe3dcecadbe9R532 to https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/virtualmachine/snapshot.go

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go
```diff
@@ -188,6 +165,248 @@ func unitTestsReconcile() {
...
+					parent = vmSnapshotL1
+					initObjects = append(initObjects, vm, vmSnapshotL1, vmSnapshotL2, vmSnapshotL3Node1, vmSnapshotL3Node2)
+				})
+				When("it's the current snapshot", func() {
+					It("returns success, vm current snapshot is updated to root", func() {
+						Expect(err).ToNot(HaveOccurred())
+						Expect(vm.Spec.CurrentSnapshot).To(Equal(vmSnapshotCRToLocalObjectRef(vmSnapshotL1)))
+						By("check parent snapshot's children should be updated")
+						Expect(parent).To(Not(BeNil()))
+						Expect(parent.Status.Children).To(ContainElement(*vmSnapshotCRToLocalObjectRef(vmSnapshotL3Node1)))
```
bryanv: Assert expected length wit HaveLen() as well

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go
```diff
@@ -188,6 +165,248 @@ func unitTestsReconcile() {
...
+					parent = vmSnapshotL2
+					initObjects = append(initObjects, vm, vmSnapshotL1, vmSnapshotL2, vmSnapshotL3Node1, vmSnapshotL3Node2)
+				})
+				When("it's the current snapshot", func() {
+					It("returns success, vm current snapshot is updated to parent", func() {
+						Expect(err).ToNot(HaveOccurred())
+						Expect(vm.Spec.CurrentSnapshot).To(Equal(vmSnapshotCRToLocalObjectRef(vmSnapshotL2)))
+						By("check parent snapshot's children should be updated")
+						Expect(parent).To(Not(BeNil()))
+						Expect(parent.Status.Children).ToNot(ContainElement(*vmSnapshotCRToLocalObjectRef(vmSnapshotL3Node1)))
```
bryanv: Assert length here too (is it expected to BeEmpty())?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +182,132 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+				ctx.Logger.Info("VirtualMachine not found, assuming the VM is deleted, remove finalizer")
+				controllerutil.RemoveFinalizer(snapshot, Finalizer)
+				return nil
+			}
+			ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
+			return err
+		}
+
+		ctx.VM = vm
+		if err := r.deleteSnapshotFromVSphere(ctx); err != nil {
```
aruneshpa: Nit: just call it `deleteVMSnapshot` or something generic?  Most things we do are on vSphere, so slapping the name on every method is not necessary.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +182,132 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+		controllerutil.RemoveFinalizer(snapshot, Finalizer)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) deleteSnapshotFromVSphere(ctx *pkgctx.VirtualMachineSnapshotContext) error {
+	snapshot := ctx.VirtualMachineSnapshot
+	// TODO: set removeChildren and consolidate to false by default for now.
+	if err := r.VMProvider.DeleteSnapshot(ctx, snapshot, ctx.VM, false, nil); err != nil {
```
aruneshpa: Hmm.  Why not just pass in the snapshot context which already has the snapshot and the VM?

pkg/providers/vsphere/vmprovider.go
```diff
@@ -525,3 +528,47 @@ func (vs *vSphereVMProvider) VSphereClient(
...
+	if err != nil {
+		return err
+	}
+
+	vcVM, err := vs.getVM(vmCtx, client, true)
+	if err != nil {
+		return fmt.Errorf("failed to get VirtualMachine %q: %w", vmCtx.VM.Name, err)
+	}
+
+	t, err := vcVM.RemoveSnapshot(vmCtx, snapshot.Name, removeChildren, consolidate)
```
aruneshpa: RemoveSnapshot is calling FindSnapshot internally to find the snapshot:
https://github.com/vmware/govmomi/blob/v0.51.0/object/virtual_machine.go#L794-L797

Which returns an error if the snapshot does not exist:
https://github.com/vmware/govmomi/blob/v0.51.0/object/virtual_machine.go#L783-L784

I think we need to gracefully handle that here and return a no-op?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +182,132 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+			return err
+		}
+
+		controllerutil.RemoveFinalizer(snapshot, Finalizer)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) deleteSnapshotFromVSphere(ctx *pkgctx.VirtualMachineSnapshotContext) error {
```
aruneshpa: This needs to be idempotent.  We need to catch the "snapshot doesn't exist" error and return a no-op -- followed by still patching the status of the VM, and the snapshots.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
aruneshpa: Yes, it could grow large.  With the proposed scale of 25k VM Service VMs, assuming worst case we could get to > 50k with just 2 snapshots per VM.  This is not realistic because not all VMs will be in same namespace, but still.

Since each VM snapshot in vCenter maps to a custom resource, what if we just got the parent snapshot of the VM before we deleted the snapshot, found out the custom resource, and then fixed up the references that way?  This won't be truly idempotent but I think is much more efficient.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Got it. Thought VC is idempotent but looks like can't rely on it

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: >Since each VM snapshot in vCenter maps to a custom resource, what if we just got the parent snapshot of the VM before we deleted the snapshot, found out the custom resource, and then fixed up the references that way? This won't be truly idempotent but I think is much more efficient.

You mean find the snapshot from VC first right, then infer the CR. Yeah that's a one way to do it. If we don't have a special label for specifying which VM does the snapshot belong to or a `Parent` snapshot field in the snapshot spec.

Feel like selecting through label would be more k8s way of finding things. But yeah it will fetch all VMSnapshot in that namespace first then filter using the label. Definitely not as scalable as finding the name from VC first then infer the CR name. Especially considering the size of snapshot tree should be small. 

If you all agree, I can just do what Arunesh suggested

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go
lubronzhan: Ack, will remove the lock

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: It will result as an import cycle, that's why I didn't use `SnapshotArg` either https://github.com/vmware-tanzu/vm-operator/pull/1032/files#r2167139528

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: I miss the case that when the snapshot is not found here. I think that's what you are referring to?
If VM is not found, I should still just remove the finalizer since all snapshot that created from this VM can be assumed as being in a deletion state.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: The DeleteSnapshot accept context.Context type. let me pass ctx.Context to avoid confusion

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Oh just found out that VirtualMachine has `RootSnapshots`, I can just recursively loop through that tree then
https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha4/virtualmachine_types.go#L1254
The downside of it would be it need to send API server (O(number of child)) calls  until finding the parent. Which increase number of roundtrips compared to directly finding the parent from the VC with single call

lubronzhan: Ok added a PR in govc to help find the parent snapshot
https://github.com/vmware/govmomi/pull/3812

aruneshpa: I will be honest.  I don't see the "find parent snapshot" to be a helpful enough helper to land in Govmomi.  The use-cases for that are not that common.  For now, I would recommend moving it here so we don't have to go through merging the change in Govmomi, cutting a new release and then spec bumping.

aruneshpa: > Oh just found out that VirtualMachine has `RootSnapshots`, I can just recursively loop through that tree then https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha4/virtualmachine_types.go#L1254 The downside of it would be it need to send API server (O(number of child)) calls until finding the parent. Which increase number of roundtrips compared to directly finding the parent from the VC with single call

That's the issue though.  When you fetch a snapshot using `vm.FindSnapshot`, then you will only have MoRef of the snapshot.  You _cannot_ fetch the _name_ and description of the snapshot from that.  So, there is no way to avoid recursively looping over all the snapshots of a VM to find a snapshot from the tree.
Plus, I never expect a VM to have exponential number of snapshots.  Since the max snapshots are capped at like 128(?), this is constant complexity.

lubronzhan: Oh yeah you are right, the function I added is not useful for fetching parent's name... Still need to check CR recursively to compare the reference().Value. So I guess maybe this logic of finding parent snapshot is not necessary against VC API. Let me just call agaisnt k8s apiserver to recursively fetch CRs then

lubronzhan:
Turn out race condition could happen on fake.DeleteSnapshot
```
2025-07-02T00:13:52.8133308Z ==================
2025-07-02T00:13:52.8133979Z WARNING: DATA RACE
2025-07-02T00:13:52.8134558Z Read at 0x00c001b96be0 by goroutine 1328:
2025-07-02T00:13:52.8135728Z   github.com/vmware-tanzu/vm-operator/pkg/providers/fake.(*VMProvider).DeleteSnapshot()
2025-07-02T00:13:52.8137387Z       /home/runner/work/vm-operator/vm-operator/pkg/providers/fake/fake_vm_provider.go:374 +0x115
2025-07-02T00:13:52.8138690Z   github.com/vmware-tanzu/vm-operator/controllers/virtualmachinesnapshot.(*Reconciler).deleteSnapshot()
2025-07-02T00:13:52.8140490Z       /home/runner/work/vm-operator/vm-operator/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go:235 +0x128
2025-07-02T00:13:52.8142179Z   github.com/vmware-tanzu/vm-operator/controllers/virtualmachinesnapshot.(*Reconciler).ReconcileDelete()
2025-07-02T00:13:52.8143340Z       /home/runner/work/vm-operator/vm-operator/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go:205 +0x5f3
2025-07-02T00:13:52.8144182Z   github.com/vmware-tanzu/vm-operator/controllers/virtualmachinesnapshot.(*Reconciler).Reconcile()
2025-07-02T00:13:52.8145138Z       /home/runner/work/vm-operator/vm-operator/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go:125 +0xc51
2025-07-02T00:13:52.8146152Z   sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[go.shape.struct { k8s.io/apimachinery/pkg/types.NamespacedName }]).Reconcile()
2025-07-02T00:13:52.8147453Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:119 +0x1bd
2025-07-02T00:13:52.8148439Z   sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[go.shape.struct { k8s.io/apimachinery/pkg/types.NamespacedName }]).reconcileHandler()
2025-07-02T00:13:52.8149530Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:340 +0x52f
2025-07-02T00:13:52.8150506Z   sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[go.shape.struct { k8s.io/apimachinery/pkg/types.NamespacedName }]).processNextWorkItem()
2025-07-02T00:13:52.8151743Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:300 +0x3b2
2025-07-02T00:13:52.8152889Z   sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[go.shape.struct { k8s.io/apimachinery/pkg/types.NamespacedName }]).Start.func2.1()
2025-07-02T00:13:52.8153919Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:202 +0xe4
2025-07-02T00:13:52.8154316Z 
2025-07-02T00:13:52.8154451Z Previous write at 0x00c001b96be0 by goroutine 1346:
2025-07-02T00:13:52.8155353Z   github.com/vmware-tanzu/vm-operator/controllers/virtualmachinesnapshot_test.intgTestsReconcileDelete.func4.1.4.2.1()
2025-07-02T00:13:52.8156591Z       /home/runner/work/vm-operator/vm-operator/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_intg_test.go:221 +0x48
2025-07-02T00:13:52.8157306Z   github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3()
2025-07-02T00:13:52.8158120Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/node.go:475 +0x2e
2025-07-02T00:13:52.8158623Z   github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
2025-07-02T00:13:52.8159210Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/suite.go:894 +0x128
2025-07-02T00:13:52.8159545Z 
2025-07-02T00:13:52.8159644Z Goroutine 1328 (running) created at:
2025-07-02T00:13:52.8160348Z   sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[go.shape.struct { k8s.io/apimachinery/pkg/types.NamespacedName }]).Start.func2()
2025-07-02T00:13:52.8161495Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:198 +0x494
2025-07-02T00:13:52.8162402Z   sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[go.shape.struct { k8s.io/apimachinery/pkg/types.NamespacedName }]).Start()
2025-07-02T00:13:52.8163617Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:209 +0x4ef
2025-07-02T00:13:52.8164716Z   sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[sigs.k8s.io/controller-runtime/pkg/reconcile.Request]).Start()
2025-07-02T00:13:52.8165707Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:148 +0x4a
2025-07-02T00:13:52.8166349Z   sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.func1()
2025-07-02T00:13:52.8167086Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/manager/runnable_group.go:226 +0x1bd
2025-07-02T00:13:52.8167715Z   sigs.k8s.io/controller-runtime/pkg/manager.(*runnableGroup).reconcile.gowrap1()
2025-07-02T00:13:52.8168465Z       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/manager/runnable_group.go:229 +0x41
2025-07-02T00:13:52.8169115Z 
2025-07-02T00:13:52.8169223Z Goroutine 1346 (finished) created at:
2025-07-02T00:13:52.8169555Z   github.com/onsi/ginkgo/v2/internal.(*Suite).runNode()
2025-07-02T00:13:52.8170149Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/suite.go:881 +0x149c
2025-07-02T00:13:52.8170637Z   github.com/onsi/ginkgo/v2/internal.(*group).attemptSpec()
2025-07-02T00:13:52.8171344Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/group.go:199 +0x1104
2025-07-02T00:13:52.8171804Z   github.com/onsi/ginkgo/v2/internal.(*group).run()
2025-07-02T00:13:52.8172355Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/group.go:349 +0x1166
2025-07-02T00:13:52.8172840Z   github.com/onsi/ginkgo/v2/internal.(*Suite).runSpecs()
2025-07-02T00:13:52.8173466Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/suite.go:489 +0x1271
2025-07-02T00:13:52.8174000Z   github.com/onsi/ginkgo/v2/internal.(*Suite).Run()
2025-07-02T00:13:52.8174550Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/internal/suite.go:130 +0x5ce
2025-07-02T00:13:52.8174975Z   github.com/onsi/ginkgo/v2.RunSpecs()
2025-07-02T00:13:52.8175481Z       /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.22.0/core_dsl.go:300 +0xe8a
2025-07-02T00:13:52.8175993Z   github.com/vmware-tanzu/vm-operator/test/builder.(*TestSuite).Register()
2025-07-02T00:13:52.8176604Z       /home/runner/work/vm-operator/vm-operator/test/builder/test_suite.go:350 +0x1c4
2025-07-02T00:13:52.8177275Z   github.com/vmware-tanzu/vm-operator/controllers/virtualmachinesnapshot_test.TestVirtualMachineSnapShot()
2025-07-02T00:13:52.8178317Z       /home/runner/work/vm-operator/vm-operator/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_suite_test.go:25 +0x53
2025-07-02T00:13:52.8178910Z   testing.tRunner()
2025-07-02T00:13:52.8179491Z       /opt/hostedtoolcache/go/1.24.4/x64/src/testing/testing.go:1792 +0x225
2025-07-02T00:13:52.8180246Z   testing.(*T).Run.gowrap1()
2025-07-02T00:13:52.8180685Z       /opt/hostedtoolcache/go/1.24.4/x64/src/testing/testing.go:1851 +0x44
2025-07-02T00:13:52.8181000Z ==================

```

lubronzhan:
https://knowledge.broadcom.com/external/article?legacyId=1025279 found this, it said the support limit of a chain of snapshot is 32

lubronzhan:
Just found out that the lock in unit tests is unnecessary. It shouldn't impact the tests. It passed yesterday

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -114,13 +139,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+	}
 	// return early if snapshot is ready; nothing to do
 	if conditions.IsTrue(ctx.VirtualMachineSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition) {
 		return nil
 	}
 
+	vmSnapshot := ctx.VirtualMachineSnapshot
 	ctx.Logger.Info("Fetching VirtualMachine from snapshot object", "vmSnapshot", vmSnapshot.Name)
 	vm := &vmopv1.VirtualMachine{}
 	objKey := client.ObjectKey{Name: vmSnapshot.Spec.VMRef.Name, Namespace: vmSnapshot.Namespace}
```
bryanv, line 162: @aruneshpa Here and elsewhere should we be checking VMRef for nil?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +184,213 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+				controllerutil.RemoveFinalizer(vmSnapshot, Finalizer)
+				return nil
+			}
+			ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
+			return err
+		}
+
+		ctx.VM = vm
+		if err := r.deleteSnapshot(ctx); err != nil {
+			if errors.Is(err, vcenter.ErrVMNotFound) { //nolint:gocritic
```
bryanv: If we're check these underlying errors IMO they should be a part of the provider interface instead of leaking the underlying packages here. It might be easier to encode as a additional bool return value instead.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +184,213 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+				ctx.Logger.Info("VirtualMachine not found, assuming the snapshot is deleted along with moVM, remove finalizer")
+				controllerutil.RemoveFinalizer(vmSnapshot, Finalizer)
+				return nil
+			} else if errors.Is(err, virtualmachine.ErrVMSnapshotNotFound) {
+				ctx.Logger.Info("vmSnapshot not found in VC, continue reconciling")
```
bryanv: For this one we could log closer to the actual error?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +184,213 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+}
+
+// getParentSnapshotName returns the parent snapshot name of the given snapshot.
+// It first checks if the snapshot is a root snapshot, if not, it will traverse the root snapshots to find the parent snapshot.
+// The upper limit of snapshot count of a VM is 128. The overhead here could be considered acceptable.
+func (r *Reconciler) getParentSnapshot(ctx *pkgctx.VirtualMachineSnapshotContext, vmSnapshot *vmopv1.VirtualMachineSnapshot) (*vmopv1.VirtualMachineSnapshot, error) {
+	rootSnapshots := ctx.VM.Status.RootSnapshots
+	if len(rootSnapshots) == 0 {
+		ctx.Logger.V(5).Info("no root snapshots found for VM")
+		return nil, virtualmachine.ErrParentVMSnapshotNotFound
```
bryanv: I'm not a fan of importing this package for its error type. For how this error is being used in its callstack I think we can define its own error type private to the controller.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: ~Ah yeah you are right. The ref could be empty when the vmSnapshot was just created before the controller reconcile it for the first time. And reconcileNormal exit after adding Finalizer, then a deletion request comes in.~

Maybe we could check it in the validation webhook?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Then we would need 2 different boolean values, one for indicating whether the VM is gone and another one for whether the snapshot is gone. Would that seems too much? I can move those predefined error to vmprovider interface level as well.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Sorry, do you mean switch this line with error in line 206? https://github.com/vmware-tanzu/vm-operator/pull/1032/files/20c5e5b7b0eb3a8fc4f26f16738ebb94fce8891d#diff-f9f5e5e7af6ecc904d51b02cec19981963d80640ba483bab7f04c0ee157af5cdR206

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Well, I can just use one, and return error nil if the snapshot is gone

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
aruneshpa: We should handle that here as well since it is optional.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +184,215 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
 	ctx.Logger.Info("Successfully patched VirtualMachine's current snapshot reference", "vm.Name", vm.Name, "spec.currentSnapshot", vm.Spec.CurrentSnapshot.Name)
 	return nil
 }
+
+func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineSnapshotContext) error {
+	ctx.Logger.Info("Reconciling VirtualMachineSnapshot deletion")
+	vmSnapshot := ctx.VirtualMachineSnapshot
+	if controllerutil.ContainsFinalizer(vmSnapshot, Finalizer) {
```
aruneshpa: vvNit: consider returning early if the finalizer is not set.  That way you can reduce the indent of this entire block.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +184,215 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+		if err != nil {
+			return fmt.Errorf("failed to delete snapshot: %w", err)
+		}
+		if vmNotFound {
+			ctx.Logger.Info("VirtualMachine not found, assuming the snapshot is deleted along with moVM, remove finalizer", "error", err)
+			controllerutil.RemoveFinalizer(vmSnapshot, Finalizer)
+			return nil
+		}
+
+		parent, err := r.updateParentSnapshot(ctx)
```
aruneshpa: Just wondering, this controller, and the VM controller can modify the snapshot resource's status at the same time.  Do you see any issues with that if the parent of a snapshot was deleted _while_ the snapshot status was being populated by the VM controller?

aruneshpa (approved):
Approved with nits.  We can iterate and fix any other issues that we find.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Ohk, just saw this
```
	// +optional

	// VMRef represents the name of the virtual machine for which the
	// snapshot is requested.
	VMRef *vmopv1common.LocalObjectRef `json:"vmRef,omitempty"`
```
Do you have example about when the vmRef will be empty? I thought it's always generated based on  a VM

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Then the `updateParentSnapshot` will fail, then it will trigger the reconcile again, and next time the parent snapshot won't be found and it shouldn't reach this code, will exit earlier when finding the parent snapshot.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Since the createSnapshotEx and DeleteSnapshot are all based on the VM. How should I handle the case when the vmRef is empty? Just assume the underling VM is gone, so does the snapshot? So remove the finalizer directly?

aruneshpa: So, I wasn't super worried about VMRef being nil _in practice_ because we will have webhooks enforcing that.  All  I meant is to make sure you are guarding for nil checks so this doesn't blow over if the webhook checks are skipped somehow.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Ohk, just reviewing your revert snapshot PR, IC you are also trying to reconcile the VM.status.currentSnapshot in the vmstatus. Maybe I shouldn't reconcile VM.status.currentSnapshot, and rootSnapshots here, and just let the vmStatus reconciler update that information?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Then the behavior would just to return err? Let me change that

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +184,215 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+			return err
+		}
+
+		ctx.VM = vm
+		vmNotFound, err := r.deleteSnapshot(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to delete snapshot: %w", err)
+		}
+		if vmNotFound {
+			ctx.Logger.Info("VirtualMachine not found, assuming the snapshot is deleted along with moVM, remove finalizer", "error", err)
```
bryanv: No need to log error since it will always be nil

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -158,3 +184,215 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+
+func (r *Reconciler) updateVMRootSnapshots(ctx *pkgctx.VirtualMachineSnapshotContext) error {
+	ctx.Logger.Info("Updating VM root snapshots")
+	vm := ctx.VM
+	vmSnapshot := ctx.VirtualMachineSnapshot
+	if len(vm.Status.RootSnapshots) == 0 {
+		ctx.Logger.V(5).Info("No root snapshots found for VM, skipping update")
+		return nil
+	}
+	if !slices.Contains(vm.Status.RootSnapshots, *vmSnapshotCRToLocalObjectRef(vmSnapshot)) {
```
bryanv: Should this just be checking by Name, like the removal done later? When we rev the next API version, it isn't totally obvious to me that we'll go update the Status with new APIVersion so this check would fail.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
bryanv: Does it need to be a pointer in the first place?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
aruneshpa: Well, you will need to handle currentSnapshot, rootSnapshot here because it is possible that the snapshot that is being deleted _is_ the only snapshot (so it is current, and is the root).
My PR handles the steady state update of the status whereas your PR handles one time fixup of these properties related to snapshot deletion.  I think we can keep this.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Agree, maybe we could change it to required.
Added the check

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: OK, let see if we could observe conflict scenarios in live env. Can't think of good one now.

lubronzhan:
Resolved all comments. Will merge once the test passed.
Gonna create a follow up PR to refactor the findParentSnapshot func. Just found out from the revert Snapshot PR that `VirtualMachineSnapshotTree` object has everything I need, so one call to VC API is enough

lubronzhan:
Also missed to update the vm.Status.CurrentSnapshot in `updateVMCurrentSnapshot`. Only updated the spec.

PR #1041, opened 2025-07-03
:sparkles: Add support for reverting a VM to a given snapshot

aruneshpa:

This change adds the capability to revert a VM from a given snapshot. This can be done by the user by modifying the currentSnapshot field of the VM's spec to point to the desired snapshot.

Along with reverting the VM to the given snapshot, we also roll back the VM's Kubernetes state (Spec, Labels, Annotations) to the state that was recorded in the Snapshot.  This is done by leveraging the state recorded for backup / restore.

Other changes include:
- Update the create snapshot workflow to be conditioned on the VM snapshot capability.  Also move the tests to be conditioned on the capability.  Added a test to ensure that if the capability is not enabled, no snapshots are created on the VM.
- A new annotation is added on the VM to indicate when a VM is being reverted from a snapshot.  It is possible that the VC VM is successfully reverted, but the VM's spec/metadata reverting fails.  In that case, we would want to indicate to the user that this VM is _in_ restore from snapshot phase.  In a followup change, we will modify the VM reconcile workflow to bail out if the VM is being reverted.
- Add a method to update the status of the VM with the current snapshot details.
- A couple of helper methods to find the snapshot from the snapshot tree of the VM.

NA

```release-note
Add support for reverting a VM to a given snapshot
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/snapshot.go
+++ b/pkg/providers/vsphere/virtualmachine/snapshot.go
@@ -7,6 +7,7 @@ package virtualmachine
 import (
 	"errors"
 	"fmt"
+	"path"
 	"strings"
 	"time"
 
@@ -27,7 +28,12 @@ type SnapshotArgs struct {
 	Consolidate    *bool
 }
 
-var ErrVMSnapshotNotFound = errors.New("snapshot not found")
+// Snapshot related errors.
+var (
+	ErrNoSnapshots       = errors.New("no snapshots for this VM")
+	ErrSnapshotNotFound  = errors.New("snapshot not found")
+	ErrMultipleSnapshots = errors.New("multiple snapshots found")
+)
 
 func SnapshotVirtualMachine(args SnapshotArgs) (*types.ManagedObjectReference, error) {
 	obj := args.VMSnapshot
@@ -95,7 +101,7 @@ func DeleteSnapshot(args SnapshotArgs) error {
 		// https://github.com/vmware/govmomi/blob/v0.52.0-alpha.0/object/virtual_machine.go#L775
 		if strings.Contains(err.Error(), fmt.Sprintf("snapshot %q not found", args.VMSnapshot.Name)) ||
 			strings.Contains(err.Error(), "no snapshots for this VM") {
-			return ErrVMSnapshotNotFound
+			return ErrSnapshotNotFound
 		}
 		return err
 	}
@@ -114,5 +120,55 @@ func updateVMStatusCurrentSnapshot(vmCtx pkgctx.VirtualMachineContext, vmSnapsho
 		APIVersion: vmSnapshot.APIVersion,
 		Kind:       vmSnapshot.Kind,
 		Name:       vmSnapshot.Name,
+	}
+}
+
+// FindSnapshot returns the snapshot matching a given name from the
+// snapshots present on a VM. Much of this is taken from Govmomi, but
+// we maintain our version because we don't want to make another
+// property collector round trip to fetch those properties again.
+func FindSnapshot(
+	vmCtx pkgctx.VirtualMachineContext,
+	snapshotName string) (*types.ManagedObjectReference, error) {
+
+	o := vmCtx.MoVM
+	if o.Snapshot == nil || len(o.Snapshot.RootSnapshotList) == 0 {
+		return nil, ErrNoSnapshots
+	}
+
+	m := make(snapshotMap)
+	m.add("", o.Snapshot.RootSnapshotList)
+
+	s := m[snapshotName]
+	switch len(s) {
+	case 0:
+		return nil, fmt.Errorf("snapshot %q not found: %w", snapshotName, ErrSnapshotNotFound)
+	case 1:
+		return &s[0], nil
+	default:
+		return nil, fmt.Errorf("%q resolves to %d snapshots: %w", snapshotName, len(s), ErrMultipleSnapshots)
+	}
+}
+
+// snapshotMap is a custom type that traverses over the entire snapshot tree.
+type snapshotMap map[string][]types.ManagedObjectReference
+
+func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree) {
+	for i, st := range tree {
+		sname := st.Name
+		names := []string{sname, st.Snapshot.Value}
+
+		if parent != "" {
+			sname = path.Join(parent, sname)
+			// Add full path as an option to resolve duplicate names
+			names = append(names, sname)
+		}
+
+		for _, name := range names {
+			m[name] = append(m[name], tree[i].Snapshot)
+		}
+
+		m.add(sname, st.ChildSnapshotList)
+
 	}
 }
--- a/pkg/providers/vsphere/vmlifecycle/update_status.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status.go
@@ -1,5 +1,5 @@
 // © Broadcom. All Rights Reserved.
-// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
 // SPDX-License-Identifier: Apache-2.0
 
 package vmlifecycle
@@ -121,6 +121,12 @@ func ReconcileStatus(
 
 	if f := pkgcfg.FromContext(vmCtx).Features; f.VMResize || f.VMResizeCPUMemory {
 		MarkVMClassConfigurationSynced(vmCtx, vmCtx.VM, k8sClient)
+	}
+
+	if pkgcfg.FromContext(vmCtx).Features.VMSnapshots {
+		if err := updateCurrentSnapshotStatus(vmCtx, k8sClient); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
 	zoneName := vm.Labels[topology.KubernetesTopologyZoneLabelKey]
@@ -1158,4 +1164,111 @@ func updateProbeStatusGuestInfo(
 	}
 
 	return probeResultSuccess, ""
+}
+
+// updateCurrentSnapshotStatus updates the VM status to reflect the
+// current snapshot on the VM.
+func updateCurrentSnapshotStatus(
+	vmCtx pkgctx.VirtualMachineContext,
+	k8sClient ctrlclient.Client) error {
+
+	vmCtx.Logger.V(5).Info("Updating current snapshot in VM status")
+
+	vm := vmCtx.VM
+
+	// If VM has no snapshots, clear the status.
+	if vmCtx.MoVM.Snapshot == nil {
+		if vm.Status.CurrentSnapshot != nil {
+			vmCtx.Logger.V(4).Info("VM has no snapshots, clearing status.currentSnapshot")
+			vm.Status.CurrentSnapshot = nil
+		}
+		return nil
+	}
+
+	// If VM has no current snapshot, clear the status.
+	if vmCtx.MoVM.Snapshot.CurrentSnapshot == nil {
+		if vm.Status.CurrentSnapshot != nil {
+			vmCtx.Logger.V(4).Info("VM has no current snapshot, clearing status.currentSnapshot")
+			vm.Status.CurrentSnapshot = nil
+		}
+		return nil
+	}
+
+	// Get the current snapshot reference from vCenter.
+	currentSnapMoref := vmCtx.MoVM.Snapshot.CurrentSnapshot
+
+	// Find the snapshot name by traversing the snapshot tree.
+	snapshotName := FindSnapshotNameInTree(vmCtx.MoVM.Snapshot.RootSnapshotList, currentSnapMoref.Value)
+	if snapshotName == "" {
+		vmCtx.Logger.V(4).Info("Could not find snapshot name in tree",
+			"snapshotRef", currentSnapMoref.Value)
+		// Clear the status if we can't find the snapshot name.
+		if vm.Status.CurrentSnapshot != nil {
+			vm.Status.CurrentSnapshot = nil
+		}
+		return nil
+	}
+
+	vmCtx.Logger.V(5).Info("Found snapshot name in tree",
+		"snapshotName", snapshotName,
+		"snapshotRef", currentSnapMoref.Value)
+
+	// Check if there's a VirtualMachineSnapshot custom resource with this name.
+	vmSnapshot := &vmopv1.VirtualMachineSnapshot{}
+	objKey := ctrlclient.ObjectKey{
+		Name:      snapshotName,
+		Namespace: vm.Namespace,
+	}
+
+	if err := k8sClient.Get(vmCtx, objKey, vmSnapshot); err != nil {
+		if !apierrors.IsNotFound(err) {
+			vmCtx.Logger.Error(err, "Failed to get VirtualMachineSnapshot custom resource",
+				"snapshotName", snapshotName)
+			return err
+		}
+		// Snapshot custom resource doesn't exist, clear the status.
+		vmCtx.Logger.V(4).Info("VirtualMachineSnapshot custom resource not found, clearing status",
+			"snapshotName", snapshotName)
+		vm.Status.CurrentSnapshot = nil
+
+		return nil
+	}
+
+	// If the snapshot is being deleted, don't update the status.
+	if !vmSnapshot.DeletionTimestamp.IsZero() {
+		vmCtx.Logger.Error(nil, "VM points to a snapshot that is marked for deletion",
+			"snapshotName", snapshotName)
+		return nil
+	}
+
+	// Update the status to reflect the current snapshot custom resource.
+	currentSnapshot := &common.LocalObjectRef{
+		APIVersion: vmSnapshot.APIVersion,
+		Kind:       vmSnapshot.Kind,
+		Name:       vmSnapshot.Name,
+	}
+
+	// Update the status. Let patch helper figure out which fields
+	// need to be updated exactly.
+	vm.Status.CurrentSnapshot = currentSnapshot
+
+	return nil
+}
+
+// FindSnapshotNameInTree recursively searches the snapshot tree for a
+// snapshot with the given reference value.
+// TODO: AKP: Replace this with a Govmomi helper once merged.
+func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, targetRef string) string {
+	for _, snapshot := range snapshots {
+		// Check if this snapshot matches the target reference.
+		if snapshot.Snapshot.Value == targetRef {
+			return snapshot.Name
+		}
+
+		// Recursively search child snapshots.
+		if childName := FindSnapshotNameInTree(snapshot.ChildSnapshotList, targetRef); childName != "" {
+			return childName
+		}
+	}
+	return ""
 }
--- a/pkg/providers/vsphere/vmlifecycle/update_status_test.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status_test.go
@@ -1,10 +1,12 @@
 // © Broadcom. All Rights Reserved.
-// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
 // SPDX-License-Identifier: Apache-2.0
 
 package vmlifecycle_test
 
 import (
+	"context"
+	"fmt"
 	"slices"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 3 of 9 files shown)


pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1165,3 +1171,109 @@ func updateProbeStatusGuestInfo(
...
+	vmCtx pkgctx.VirtualMachineContext,
+	k8sClient ctrlclient.Client) error {
+
+	vmCtx.Logger.V(5).Info("Updating current snapshot in VM status")
+
+	vm := vmCtx.VM
+
+	// If VM has no snapshots, clear the status.
+	if vmCtx.MoVM.Snapshot == nil {
+		if vm.Status.CurrentSnapshot != nil {
```
lubronzhan, line 1181: nit: merge these two condition check? Or just set it to nil without the second check?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1165,3 +1171,109 @@ func updateProbeStatusGuestInfo(
...
+			vmCtx.Logger.V(4).Info("VM has no snapshots, clearing status.currentSnapshot")
+			vm.Status.CurrentSnapshot = nil
+		}
+		return nil
+	}
+
+	// If VM has no current snapshot, clear the status.
+	if vmCtx.MoVM.Snapshot.CurrentSnapshot == nil {
+		if vm.Status.CurrentSnapshot != nil {
```
lubronzhan, line 1190: Same as above

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1165,3 +1171,109 @@ func updateProbeStatusGuestInfo(
...
+	// Get the current snapshot reference from vCenter.
+	currentSnapMoref := vmCtx.MoVM.Snapshot.CurrentSnapshot
+
+	// Find the snapshot name by traversing the snapshot tree.
+	snapshotName := FindSnapshotNameInTree(vmCtx.MoVM.Snapshot.RootSnapshotList, currentSnapMoref.Value)
+	if snapshotName == "" {
+		vmCtx.Logger.V(4).Info("Could not find snapshot name in tree",
+			"snapshotRef", currentSnapMoref.Value)
+		// Clear the status if we can't find the snapshot name.
+		if vm.Status.CurrentSnapshot != nil {
```
lubronzhan, line 1206: Same as above

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1165,3 +1171,109 @@ func updateProbeStatusGuestInfo(
...
+	}
+
+	if err := k8sClient.Get(vmCtx, objKey, vmSnapshot); err != nil {
+		if !apierrors.IsNotFound(err) {
+			vmCtx.Logger.Error(err, "Failed to get VirtualMachineSnapshot custom resource",
+				"snapshotName", snapshotName)
+			return err
+		}
+		// Snapshot custom resource doesn't exist, clear the status.
+		vmCtx.Logger.V(4).Info("VirtualMachineSnapshot custom resource not found, clearing status",
```
lubronzhan, line 1230: The VC shows VM has snapshot but there is no Snapshot resource. How should we resolve this inconsistency? 
I think this only happen when the customer manually added a snapshot from VC UI, not from VMOP? Since a snapshot in VC will be created after a customer creates a VMSnapshot resource, and VMSnapshot resource won't be deleted until VC resources have been cleaned up.

I remember you discussed containerized permission in yesterday's demo which could give user the right to do above action. So in this case we just log this, should we give a big warning to the user?

pkg/providers/vsphere/vmlifecycle/update_status_test.go
```diff
@@ -2329,3 +2332,388 @@ var _ = Describe("UpdateGroupLinkedCondition", func() {
...
+		})
+
+		Context("when snapshots capability is not enabled", func() {
+			BeforeEach(func() {
+				pkgcfg.SetContext(ctx, func(config *pkgcfg.Config) {
+					config.Features.VMSnapshots = false
+				})
+			})
+
+			It("should update the current snapshot status to match the found snapshot", func() {
```
lubronzhan: It should not update the current snapshot in this case?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1165,3 +1171,109 @@ func updateProbeStatusGuestInfo(
...
+			"snapshotName", snapshotName)
+		if vm.Status.CurrentSnapshot != nil {
+			vm.Status.CurrentSnapshot = nil
+		}
+		return nil
+	}
+
+	// Update the status to reflect the current snapshot custom resource.
```
lubronzhan, line 1244: Should we skip setting this if the DeletionTimestamp is not zero?

pkg/providers/vsphere/vmlifecycle/update_status_test.go
lubronzhan: A note, there is a chance that vmsnapshot controller updates the currentSnapshot to the parent of this, but then got overwritten in this logic. Then need next VMStatus reconcile will correct the current Snapshot status

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,42 +1160,325 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+	// Perform the actual snapshot revert
+	vmCtx.Logger.V(4).Info("Starting vSphere snapshot revert operation")
+	if err := vs.performSnapshotRevert(vmCtx, vcVM, snapObj, desiredSnapshotName); err != nil {
+		vmCtx.Logger.Error(err, "vSphere snapshot revert failed")
+		return true, err
+	}
+	vmCtx.Logger.Info("vSphere snapshot revert completed successfully")
+
+	// TODO: AKP: We will modify the snapshot workflow to always skip the automatic
+	// registration.  Once we do that, we will have to remove that ExtraCoonfig key.
```
lubronzhan: nit: typo
```suggestion
	// registration.  Once we do that, we will have to remove that ExtraConfig key.
```

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,42 +1160,325 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+		// Only skip (return nil) for expected cases where we should not fail:
+		// 1. VM has no snapshots at all
+		// 2. Snapshot by name doesn't exist (this would be a different error message)
+		if err.Error() == ErrMsgNoSnapshots {
+			vmCtx.Logger.V(4).Info("VM has no snapshots, skipping",
+				"snapshotName", snapshotName)
+			return nil, nil
+		}
+
+		if err.Error() == ErrMsgSnapshotNotFound {
```
lubronzhan:
```suggestion
		if strings.Contains(err.Error(), fmt.Sprintf("snapshot %q not found", snapshotName)) {
```
https://github.com/vmware/govmomi/blob/v0.52.0-alpha.0/object/virtual_machine.go#L784

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,42 +1160,325 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+
+	// Fetch the config stored with the snapshot.
+	if err := vcVM.Properties(vmCtx, *snap, []string{"config"}, &moSnap); err != nil {
+		vmCtx.Logger.Error(err, "error fetching the snapshot config")
+		return "", err
+	}
+
+	vmCtx.Logger.V(5).Info("Successfully fetched snapshot properties")
+
+	if moSnap.Config.ExtraConfig == nil {
```
lubronzhan: newBie question: Config.ExtraConfig is required for VM?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,42 +1160,325 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
 	vcVM *object.VirtualMachine) error {
-	// Skip creating a new snapshot if the VM already points to the desired snapshot.
-	if vmCtx.VM.Spec.CurrentSnapshot == nil || reflect.DeepEqual(vmCtx.VM.Spec.CurrentSnapshot, vmCtx.VM.Status.CurrentSnapshot) {
+	// Skip creating a new snapshot if the VM doesn't specify a desired snapshot.
+	if vmCtx.VM.Spec.CurrentSnapshot == nil {
 		return nil
 	}
 
-	// get VirtualMachineSnapshot object
+	// Fetch the VirtualMachineSnapshot object so we have the
```
lubronzhan:
```suggestion
	// Fetch the VirtualMachineSnapshot object
```

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -2825,6 +2826,285 @@ func vmTests() {
...
+						Expect(err).ToNot(HaveOccurred())
+
+						// Status should reflect the actual current snapshot
+						Expect(vm.Status.CurrentSnapshot).ToNot(BeNil())
+						Expect(vm.Status.CurrentSnapshot.Name).To(Equal(vmSnapshot.Name))
+					})
+				})
+			})
+
+			Context("Create/Update/Delete ISO backed VirtualMachine", func() {
```
lubronzhan: This is unrelated to this PR I assume

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1165,3 +1171,109 @@ func updateProbeStatusGuestInfo(
...
+	if err := k8sClient.Get(vmCtx, objKey, vmSnapshot); err != nil {
+		if !apierrors.IsNotFound(err) {
+			vmCtx.Logger.Error(err, "Failed to get VirtualMachineSnapshot custom resource",
+				"snapshotName", snapshotName)
+			return err
+		}
+		// Snapshot custom resource doesn't exist, clear the status.
+		vmCtx.Logger.V(4).Info("VirtualMachineSnapshot custom resource not found, clearing status",
+			"snapshotName", snapshotName)
+		if vm.Status.CurrentSnapshot != nil {
```
bryanv: Prb just as clear to always clear the field

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -79,6 +79,8 @@ var (
 	ErrIsPaused               = pkgerr.NoRequeueNoErr("is paused")
 	ErrHasTask                = pkgerr.NoRequeueNoErr("has outstanding task")
 	ErrPromoteDisks           = vmconfdiskpromo.ErrPromoteDisks
+	ErrMsgNoSnapshots         = "no snapshots for this VM"
```
bryanv: It is kind of weird to have these here and prefixed with Err when they aren't actual error types

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1165,3 +1171,109 @@ func updateProbeStatusGuestInfo(
...
+	// Update the status to reflect the current snapshot custom resource.
+	currentSnapshot := &common.LocalObjectRef{
+		APIVersion: vmSnapshot.APIVersion,
+		Kind:       vmSnapshot.Kind,
+		Name:       vmSnapshot.Name,
+	}
+
+	// Only update if it's different to avoid unnecessary updates.
+	if vm.Status.CurrentSnapshot == nil ||
+		vm.Status.CurrentSnapshot.APIVersion != currentSnapshot.APIVersion ||
```
bryanv: Just compare the structs instead of each field individually? 

Or just assign the expected value and let the patch helper determine what's changed like we do for most stuff in status.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,42 +1160,325 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+	}
+
+	// Handle cases where the snapshot doesn't contain any VM spec and metadata.
+	if vmYAML == "" {
+		return fmt.Errorf("no VM YAML in snapshot config")
+	}
+
+	var vm vmopv1.VirtualMachine
+	if err := yaml.Unmarshal([]byte(vmYAML), &vm); err != nil {
+		vmCtx.Logger.Error(err, "failed to unmarshal the payload stored in the snapshot into a VM type")
```
bryanv, line 1362: JW is there worked planned to support this when the snapshot yaml is at a down version?

pkg/providers/vsphere/vmlifecycle/update_status.go
aruneshpa: Unclear what you mean.  Do you just mean roll the two `if` conditions into one?  That doesn't buy anything, right?

Having the second if condition as a separate one allows for the logging which helps figure out when exactly the transition is happening from not-nil to nil snapshot status.

pkg/providers/vsphere/vmlifecycle/update_status.go
aruneshpa: > I think this only happen when the customer manually added a snapshot from VC UI, not from VMOP?

This is _exactly_ the scenario where this could happen.  And I am okay emitting a log here.
We actually need to figure out a better way to handle this scenario otherwise the chain of snapshot custom resources might be broken.  I am going to file a story for this.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,42 +1160,325 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+			vmCtx.Logger.V(4).Info("Snapshot not found on the VM (snapshot doesn't exist)",
+				"snapshotName", snapshotName)
+			return nil, nil
+		}
+
+		// For all other errors (property collector error, multiple snapshots, etc.),
+		// return the actual error and requeue the reconcile request.
+		vmCtx.Logger.Error(err, "Error finding snapshot",
+			"snapshotName", snapshotName)
+		return nil, pkgerr.NoRequeueNoErr(err.Error())
```
bryanv: Is this really intended to be a `NoRequeueNoErr` error?

pkg/providers/vsphere/vmlifecycle/update_status.go
aruneshpa: Hmm.  So the argument would be that if a snapshot is being deleted, then there would be a race between the VM controller and the snapshot controller trying to modify the status of the VM status.  It makes sense to skip in that case.  Will fix.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,42 +1160,325 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+	return true, nil
+}
+
+// findDesiredSnapshot finds the snapshot object by name, returns nil if not found.
+func (vs *vSphereVMProvider) findDesiredSnapshot(
+	vmCtx pkgctx.VirtualMachineContext,
+	vcVM *object.VirtualMachine,
+	snapshotName string) (*vimtypes.ManagedObjectReference, error) {
+
+	snapObj, err := vcVM.FindSnapshot(vmCtx, snapshotName)
```
bryanv: TBH I'd like us to move away from using this govmoni helper. We've already fetched the VM's properties, but FindSnapshot will do that for the snapshots again, its error returns that we're trying to key off aren't very convenient, and its impl isn't very complicated.

pkg/providers/vsphere/vmlifecycle/update_status_test.go
aruneshpa: This is a unit test.  The snapshot controller is not running here so we don't need to handle that scenario.
In real env, yeah, the status will be eventually consistent.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: I had that error check originally, but then decided to use an exported string.  
There was an issue to address this but it never got fixed: https://github.com/vmware/govmomi/issues/2049

Let me switch back to exact matches with a comment to address this.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: No.  It is optional.  For all intents and purposes, it is never empty.  But it _can_ be, so we need to handle it.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Hmm.  I thought the `Msg` in the variable name was supposed to indicate that?

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Hmm.  Would you recommend we duplicate this in VM operator then?

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Nope.  Just changed it to make it a regular error so it is not a `NoRequeue` error.

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: Just to merge nested if check. Sometimes the linter doesn't like it. 2 levels should be fine.

>Having the second if condition as a separate one allows for the logging which helps figure out when exactly the transition is happening from not-nil to nil snapshot status.

Yeah that's true, I mentioned it since you put the logging under the nested if statement.  This is not a big deal.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Created a tracker for this.  But this is the same class of problems that are present in the backup / restore world.  We need to prioritize it.

pkg/providers/vsphere/vmprovider_vm.go
bryanv: It isn't very much logic, and we don't have to deal with funky error returns so that's my preference

pkg/providers/vsphere/vmprovider_vm.go
bryanv: Is it worth handling separately here, or just use the key doesn't exist check that's done next?

pkg/providers/vsphere/vmprovider_vm.go
bryanv: I'd prefer them be consts since they are just strings. But from other comments, I'm thinking these will just go away so its moot

pkg/providers/vsphere/vmprovider_vm_test.go
aruneshpa: Yes.  This got added from an incorrect merge after a rebase. :(
Fixed.

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -2825,6 +2826,262 @@ func vmTests() {
...
+					})
+				})
+
+				Context("when reverting to valid snapshot", func() {
+					var secondSnapshot *vmopv1.VirtualMachineSnapshot
+
+					It("should successfully revert to desired snapshot", func() {
+						// Create VM first
+						var err error
+						vcVM, err = createOrUpdateAndGetVcVM(ctx, vmProvider, vm)
```
bryanv: There is a mix of this and `vcVM, err := ...` in these tests. I don't think we need the outer `vcVM` and can use `:=` here & elsewhere

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -29,6 +30,13 @@ type SnapshotArgs struct {
 
 var ErrVMSnapshotNotFound = errors.New("snapshot not found")
 
+// Error message constants for snapshot operations
+const (
+	ErrMsgNoSnapshots       = "no snapshots for this VM"
```
bryanv: Use `var` and actual errors via `errors.New("foo")` so you can use `errors.Is()` for comparison

You can wrap the errors to include additional details for the formatted strings below

bryanv (approved):

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1133,15 +1158,288 @@ func (vs *vSphereVMProvider) reconcilePowerState(
...
+	}
+
+	vmCtx.Logger.Info("Starting snapshot revert operation",
+		"snapshotName", desiredSnapshotName,
+		"currentSnapshot", vmCtx.MoVM.Snapshot.CurrentSnapshot.Value,
+		"desiredSnapshot", snapObj.Reference().Value)
+
+	// Set the revert in progress annotation.
+	vmCtx.Logger.V(4).Info("Setting revert in progress annotation")
+	vs.setRevertInProgressAnnotation(vmCtx)
```
lubronzhan, line 1205: Just occur to me that does this annotation really taking effect?
Since we don't exit immediately to patch the VM object. The annotation will only be added to a VM CR if we added annotation and some error returns before we remove the annotation in restoreVMSpecFromSnapshot?

lubronzhan (approved):

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Exactly.
It will take affect if the revert fails for any reason.  In happy path, it is meant to never be shown to the user.
This will help with situations where if revert fails during, say, VM spec revert, we know that a revert is in progress.

Plus, I intend to use this annotation to skip reconciling a VM in the VM controller because something has clearly failed catastrophically so we should have the user inspect the spec and remove this annotation.

PR #1045, opened 2025-07-10
🌱 Standardized controller reconcile logging

bryanv:

Generally, we were using different loggers for the FooContext and the logger that is obtained from the reconcile context. This would lead to different logging depending on which one was being used.

Add a LogConstructor that makes the controller-runtime logger less verbose and more like what we have been logging in our FooContext loggers.

Use the first 8 char of the controller-runtime reconcile ID as part of the ID passed along to VC to easily correlate logs.

Fixes #

Example:

```
I0710 19:18:51.336694       1 virtualmachine_controller.go:287] "TEST" logger="virtualmachine-controller" VirtualMachine="test-bmv-ns/vm1" reconcileID="64ed1217-5afd-4d48-9e9e-9640cdee8789"
```

```release-note
NONE
```


pkg/providers/vsphere/vmprovider.go
```diff
@@ -367,16 +366,22 @@ func (vs *vSphereVMProvider) UpdateContentLibraryItem(ctx context.Context, itemI
...
-		id[i] = charset[idx]
+func (vs *vSphereVMProvider) getOpID(ctx context.Context, vm *vmopv1.VirtualMachine, operation string) string {
+	var id string
+
+	if recID := controller.ReconcileIDFromContext(ctx); recID != "" {
+		id = string(recID[:8])
+	} else {
+		buf := make([]byte, 8)
+		for i := range buf {
+			const charset = "0123456789abcdef"
```
lubronzhan: nit: this can be moved out of  the for loop

pkg/util/log_constructor.go
```diff
@@ -0,0 +1,51 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
```
lubronzhan: nit:
```suggestion
// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
```

lubronzhan (commented):
Thanks, will keep this in mind when using logger next time

controllers/contentlibrary/utils/controller_builder.go
```diff
@@ -126,7 +128,7 @@ func (r *Reconciler) Reconcile(
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
 	ctx = ovfcache.JoinContext(ctx, r.Context)
 
-	logger := ctrl.Log.WithName(r.Kind).WithValues("name", req.String())
+	logger := logr.FromContextOrDiscard(ctx)
```
akutz: Why not `pkgutil.FromContextOrDefault`?

controllers/infra/validatingwebhookconfiguration/validatingwebhookconfiguration_controller.go
```diff
@@ -101,14 +105,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
 
 	if req.Name != spqutil.ValidatingWebhookConfigName {
-		r.Logger.Error(nil, "Reconcile called for unexpected object", "name", req.Name)
+		logr.FromContextOrDiscard(ctx).Error(nil, "Reconcile called for unexpected object")
```
akutz: Same question as before, should we ever be using `logr.FromContextOrDiscard` anymore?

akutz (approved):
Thank you for doing this @bryanv . A few notes:

* Are the uses of `logr.FromContextOrDiscard` vestigial?
* Should we add the op ID as a key/value in the logger? That way we can match it to vpxd directly?

controllers/contentlibrary/utils/controller_builder.go
bryanv: Sure I'll change these too. Within the controllers it is safe to assume we'll have the logger there but nbd to change these. Might add a linter rule to prevent using FromContextOrDiscard depending on what over usages remain

bryanv:
> Thank you for doing this @bryanv . A few notes:
> 
> * Are the uses of `logr.FromContextOrDiscard` vestigial?
> * Should we add the op ID as a key/value in the logger? That way we can match it to vpxd directly?

* Removed from controllers
* Its the first 8 characters from the controller-runtime generated UUID so pretty easy to lookup (we were generated 8 char random string before - I think 8 chars again is good enough for uniqueness, and IDK if there is a limit)

PR #1047, opened 2025-07-11
🌱 Add VMSnapshot Validation Webhook logic

palnabarun:

Adds validation logic to webhooks for VMSnapshot resources

None

```release-note
NONE
```

### DIFF
--- a/webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
+++ b/webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
@@ -6,12 +6,16 @@ package validation
 
 import (
 	"fmt"
+	"net/http"
 	"reflect"
 
+	"k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
@@ -22,6 +26,8 @@ import (
 
 const (
 	webHookName = "default"
+
+	virtualMachineKind = "VirtualMachine"
 )
 
 // +kubebuilder:webhook:verbs=create;update,path=/default-validate-vmoperator-vmware-com-v1alpha4-virtualmachinesnapshot,mutating=false,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachinesnapshots,versions=v1alpha4,name=default.validating.virtualmachinesnapshot.v1alpha4.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
@@ -56,10 +62,45 @@ func (v validator) For() schema.GroupVersionKind {
 	return vmopv1.GroupVersion.WithKind(reflect.TypeOf(vmopv1.VirtualMachineSnapshot{}).Name())
 }
 
+// ValidateCreate makes sure the VirtualMachineSnapshot create request is valid.
 func (v validator) ValidateCreate(ctx *pkgctx.WebhookRequestContext) admission.Response {
-	// TODO: Implement validation for create
+	vmSnapshot, err := v.vmSnapshotFromUnstructured(ctx.Obj)
+	if err != nil {
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
+
+	var fieldErrs field.ErrorList
+
+	vmRefField := field.NewPath("spec", "vmRef")
+
+	if vmSnapshot.Spec.VMRef != nil {
+		if vmSnapshot.Spec.VMRef.Kind != virtualMachineKind {
+			fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("kind"),
+				vmSnapshot.Spec.VMRef.Kind, fmt.Sprintf("must be %q", virtualMachineKind)))
+		}
+
+		if vmSnapshot.Spec.VMRef.APIVersion != "" {
+			gv, err := schema.ParseGroupVersion(vmSnapshot.Spec.VMRef.APIVersion)
+			if err != nil {
+				fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("apiVersion"),
+					vmSnapshot.Spec.VMRef.APIVersion, "must be valid group version"))
+			} else if gv.Group != vmopv1.GroupName {
+				fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("apiVersion"),
+					vmSnapshot.Spec.VMRef.APIVersion, fmt.Sprintf("group must be %q", vmopv1.GroupName)))
+			}
+		}
+
+		if vmSnapshot.Spec.VMRef.Name == "" {
+			fieldErrs = append(fieldErrs, field.Required(vmRefField.Child("name"), "name must be provided"))
+		}
+	} else {
+		fieldErrs = append(fieldErrs, field.Required(vmRefField, "vmRef must be provided"))
+	}
 
 	validationErrs := make([]string, 0)
+	for _, fieldErr := range fieldErrs {
+		validationErrs = append(validationErrs, fieldErr.Error())
+	}
 
 	return common.BuildValidationResponse(ctx, nil, validationErrs, nil)
 }
@@ -68,10 +109,38 @@ func (v validator) ValidateDelete(*pkgctx.WebhookRequestContext) admission.Respo
 	return admission.Allowed("")
 }
 
+// ValidateUpdate validates if the VirtualMachineSnapshot update is valid
+// - Fields other than Description are not allowed to be changed.
 func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.Response {
-	// TODO: Implement validation for update
+	vmSnapshot, err := v.vmSnapshotFromUnstructured(ctx.Obj)
+	if err != nil {
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
 
-	validationErrs := make([]string, 0)
+	oldVMSnapshot, err := v.vmSnapshotFromUnstructured(ctx.OldObj)
+	if err != nil {
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
+
+	var fieldErrs field.ErrorList
+
+	fieldErrs = append(fieldErrs, validation.ValidateImmutableField(vmSnapshot.Spec.Memory, oldVMSnapshot.Spec.Memory, field.NewPath("spec", "memory"))...)
+	fieldErrs = append(fieldErrs, validation.ValidateImmutableField(vmSnapshot.Spec.Quiesce, oldVMSnapshot.Spec.Quiesce, field.NewPath("spec", "quiesce"))...)
+	fieldErrs = append(fieldErrs, validation.ValidateImmutableField(vmSnapshot.Spec.VMRef, oldVMSnapshot.Spec.VMRef, field.NewPath("spec", "vmRef"))...)
+
+	validationErrs := make([]string, 0, len(fieldErrs))
+	for _, fieldErr := range fieldErrs {
+		validationErrs = append(validationErrs, fieldErr.Error())
+	}
 
 	return common.BuildValidationResponse(ctx, nil, validationErrs, nil)
+}
+
+// vmSnapshotFromUnstructured returns the VirtualMachineSnapshot from the unstructured object.
+func (v validator) vmSnapshotFromUnstructured(obj runtime.Unstructured) (*vmopv1.VirtualMachineSnapshot, error) {
+	vmSnapshot := &vmopv1.VirtualMachineSnapshot{}
+	if err := v.converter.FromUnstructured(obj.UnstructuredContent(), vmSnapshot); err != nil {
+		return nil, err
+	}
+	return vmSnapshot, nil
 }
--- a/webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_intg_test.go
+++ b/webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_intg_test.go
@@ -7,7 +7,6 @@ package validation_test
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
 	"github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels"
 	"github.com/vmware-tanzu/vm-operator/test/builder"
@@ -60,6 +59,7 @@ func newIntgValidatingWebhookContext() *intgValidatingWebhookContext {
 	}
 
 	ctx.vmSnapshot = builder.DummyVirtualMachineSnapshot(ctx.Namespace, "dummy-vm-snapshot", "dummy-vm")
+
 	return ctx
 }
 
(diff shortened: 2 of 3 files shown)


webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +61,85 @@ func (v validator) For() schema.GroupVersionKind {
...
 
 	validationErrs := make([]string, 0)
 
+	if vmSnapshot.Spec.VMRef == nil {
+		validationErrs = append(validationErrs, "spec.vmRef must not be nil")
+	}
+
+	if vmSnapshot.Spec.VMRef != nil {
+		if vmSnapshot.Spec.VMRef.Kind != virtualMachineKind {
+			validationErrs = append(validationErrs, fmt.Sprintf("spec.vmRef.kind must be %q, got %q",
```
bryanv: PTAL at our other validation webhooks like the virtualmachine one, use the `field` package to accumulate errors in a `field.ErrorList`.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +61,85 @@ func (v validator) For() schema.GroupVersionKind {
...
+				virtualMachineKind, vmSnapshot.Spec.VMRef.Kind))
+		}
+
+		if vmSnapshot.Spec.VMRef.APIVersion == "" {
```
bryanv: Is this something that we really have to require? Can't we just implicitly default this (no need for a mutation webhook imo)

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +61,85 @@ func (v validator) For() schema.GroupVersionKind {
...
+			validationErrs = append(validationErrs, "spec.vmRef.apiVersion must be set to a non-empty value")
+		}
+
+		if vmSnapshot.Spec.VMRef.Name == "" {
+			validationErrs = append(validationErrs, "spec.vmRef.name must be set to a non-empty value")
+		} else {
+			// in addition to fetching the VM referred to in the snapshot, it also
```
bryanv: @aruneshpa Can we please have our validation webhooks just focus on internal consistency? There is no guarantee that any of this will be true when the snapshot is actually reconciled.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +61,85 @@ func (v validator) For() schema.GroupVersionKind {
...
+	}
+
+	oldVMSnapshot, err := v.vmSnapshotFromUnstructured(ctx.OldObj)
+	if err != nil {
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
 
 	validationErrs := make([]string, 0)
 
+	if vmSnapshot.Spec.Memory != oldVMSnapshot.Spec.Memory ||
```
bryanv: Use `validation.ValidateImmutableField()` for this

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
palnabarun: Default in the controller?

In future, if `v1alpha5` and further comes out and there's some API change that makes the controller behave differently, then this the APIVersion is important to be given by the users.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
palnabarun: True that the guarantee might not hold good when the object is reconciled.

But this protects against the case where a VM is powered off and the user is requesting for a snapshot with memory included. User gets an early warning and ensures that an unnecessary VMSnapshot resource isn't created.

_Unless_ the behavior we want is that users can create a VMSnapshot with memory=true even when VM=PoweredOff **and** if and when the VM is powered back on, the snapshot is taken.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +63,99 @@ func (v validator) For() schema.GroupVersionKind {
...
+
 	return common.BuildValidationResponse(ctx, nil, validationErrs, nil)
 }
 
 func (v validator) ValidateDelete(*pkgctx.WebhookRequestContext) admission.Response {
 	return admission.Allowed("")
 }
 
+// ValidateUpdate validates if the VirtualMachineSnapshot update is valid
+// - Fields other than Description are nota allowed to be changed.
```
lubronzhan: nit:
```suggestion
// - Fields other than Description are not allowed to be changed.
```

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_intg_test.go
```diff
@@ -76,57 +81,85 @@ func intgTestsValidateCreate() {
...
+			err = ctx.Client.Create(ctx, ctx.vmSnapshot)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("failed to get VirtualMachine %q", ctx.vm.Name)))
+		})
+	})
+
+	When("the VMSnapshot is created for an existing VM", func() {
+		BeforeEach(func() {
+			err = ctx.Client.Create(ctx, ctx.vm)
+			Expect(err).NotTo(HaveOccurred())
```
lubronzhan:
```suggestion
			Expect(ctx.Client.Create(ctx, ctx.vm)).To(Succeed())
```

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_intg_test.go
```diff
@@ -76,57 +81,85 @@ func intgTestsValidateCreate() {
...
+		})
+
+		It("should allow the request", func() {
+			err = ctx.Client.Create(ctx, ctx.vmSnapshot)
+			Expect(err).NotTo(HaveOccurred())
```
lubronzhan:
```suggestion
			Expect(ctx.Client.Create(ctx, ctx.vmSnapshot)).To(Succeed())
```

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_intg_test.go
```diff
@@ -76,57 +81,85 @@ func intgTestsValidateCreate() {
...
+			originalDescription = ctx.vmSnapshot.Spec.Description
+			ctx.vmSnapshot.Spec.Description = "new-description"
+		})
+
+		AfterEach(func() {
+			ctx.vmSnapshot.Spec.Description = originalDescription
+		})
+
+		It("should should allow the request", func() {
+			err = ctx.Client.Update(ctx, ctx.vmSnapshot)
```
lubronzhan: nit:
```suggestion
		Expect(ctx.Client.Update(ctx, ctx.vmSnapshot)).To(Succeed())
```

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_unit_test.go
```diff
@@ -136,23 +236,15 @@ func unitTestsValidateUpdate() {
...
-		})
-
-		It("should allow the request", func() {
-			Expect(response.Allowed).To(BeTrue())
-			Expect(response.Result).ToNot(BeNil())
-		})
-	})
+	DescribeTable("update table", validateUpdate,
+		Entry("should allow updating description", updateArgs{}, true, "", nil),
+		Entry("should not allow updating memory",
```
lubronzhan, line 239: Also verify updating other fields to increase coverage?

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
bryanv: It really doesn't protect against anything. Status can be stale now or later. Any errors need to be reported in a condition, users need to have a done/ready/whatever condition to wait/check for completion.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
bryanv: AFACT current we don't ever use the APIVersion value of this field. This validation doesn't really add much value - just not empty but could be a bogus value.

I think for most situation the APIVersion doesn't really matter. Maybe someday we'll use that as an indication of what version of the yaml to store or something but if unset it is reasonable to just assume whatever the current storage version is.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -6,10 +6,15 @@ package validation
 
 import (
 	"fmt"
+	"net/http"
 	"reflect"
 
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
```
aruneshpa: Nit: remove these to collapse in the same group.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
aruneshpa: I would modify the current check to only validate that if it is not empty, then it must be valid (e.g., GroupName + version).  You can just start with validating the version.

aruneshpa: I agree that it's a little weird that we are not using it.  But this is a good, extensible pattern to have an object ref.  We can revisit this decision if we feel strongly against it.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
aruneshpa: Let's remove this check for now. We can add it later.
I experimented with vSphere APIs and it turns out that vSphere doesn't _prevent_ users from taking a memory snapshot of a powered off VM.  It simply ignores that field and the power state of the snapshot is set to "PoweredOff".

aruneshpa (approved):
Other than the few comments, LGTM.  Thanks @palnabarun !

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +62,104 @@ func (v validator) For() schema.GroupVersionKind {
...
+		}
+
+		if vmSnapshot.Spec.VMRef.APIVersion != "" {
+			gv, err := schema.ParseGroupVersion(vmSnapshot.Spec.VMRef.APIVersion)
+			if err != nil {
+				fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("apiVersion"),
+					vmSnapshot.Spec.VMRef.APIVersion, "must be valid group version"))
+			}
+
+			if gv.Group != "vmoperator.vmware.com" {
```
bryanv: Use `vmopv1.GroupName`

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +62,104 @@ func (v validator) For() schema.GroupVersionKind {
...
+			fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("kind"),
+				vmSnapshot.Spec.VMRef.Kind, fmt.Sprintf("must be %q", virtualMachineKind)))
+		}
+
+		if vmSnapshot.Spec.VMRef.APIVersion != "" {
+			gv, err := schema.ParseGroupVersion(vmSnapshot.Spec.VMRef.APIVersion)
+			if err != nil {
+				fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("apiVersion"),
+					vmSnapshot.Spec.VMRef.APIVersion, "must be valid group version"))
+			}
```
bryanv: Should probably be `} else if gv.Group ...` as to not double up errors

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +62,104 @@ func (v validator) For() schema.GroupVersionKind {
...
+
+			if gv.Group != "vmoperator.vmware.com" {
+				fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("apiVersion"),
+					vmSnapshot.Spec.VMRef.APIVersion, "group must be vmoperator.vmware.com"))
+			}
+		}
+
+		if vmSnapshot.Spec.VMRef.Name == "" {
+			fieldErrs = append(fieldErrs, field.Invalid(vmRefField.Child("name"),
```
bryanv: Use `field.Required`

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,22 +62,104 @@ func (v validator) For() schema.GroupVersionKind {
...
+			}, vm); err != nil {
+				validationErrs = append(validationErrs, fmt.Sprintf("failed to get VirtualMachine %q: %v",
+					vmSnapshot.Spec.VMRef.Name, err))
+			}
+
+			// TODO(future): if needed, add more validation to ensure the VM is in a valid state for snapshotting.
+		}
+	}
+
+	for _, fieldErr := range fieldErrs {
```
bryanv, line 101: Move `validationErrs := make([]string, len(fieldErrs)))` down here

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -56,10 +62,47 @@ func (v validator) For() schema.GroupVersionKind {
...
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
+
+	var fieldErrs field.ErrorList
+
+	vmRefField := field.NewPath("spec", "vmRef")
+
+	if vmSnapshot.Spec.VMRef == nil {
+		fieldErrs = append(fieldErrs, field.Required(vmRefField, "vmRef must be provided"))
+	}
```
bryanv: You could just `} else {` this here to not check the same thing for not nil below

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_intg_test.go
```diff
@@ -52,14 +51,18 @@ func intgTests() {
 type intgValidatingWebhookContext struct {
 	builder.IntegrationTestContext
 	vmSnapshot *vmopv1.VirtualMachineSnapshot
+	vm         *vmopv1.VirtualMachine
```
bryanv: Is this still needed?

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_unit_test.go
```diff
@@ -94,9 +93,39 @@ func unitTestsValidateCreate() {
 	)
 
 	type createArgs struct {
+		invalidVMRef                bool
+		invalidVMRefKind            bool
+		invalidVMRefAPIVersion      bool
+		invalidVMRefAPIVersionGroup bool
+		emptyVMRefName              bool
+		nonExistingVM               bool
```
bryanv: Still needed anymore?

palnabarun: Thanks for the catch!

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator_unit_test.go
```diff
@@ -55,15 +59,12 @@ type unitValidatingWebhookContext struct {
 	builder.UnitTestContextForValidatingWebhook
 	vmSnapshot    *vmopv1.VirtualMachineSnapshot
 	oldVMSnapshot *vmopv1.VirtualMachineSnapshot
-	vm            *vmopv1.VirtualMachine
-	cl            *imgregv1a1.ContentLibrary
 }
 
 func newUnitTestContextForValidatingWebhook(isUpdate bool) *unitValidatingWebhookContext {
 	vm := builder.DummyVirtualMachine()
```
bryanv: Can't this vm be dropped too now?

bryanv (approved):
I think still one unneeded VM to clean up but otherwise LGTM

PR #1048, opened 2025-07-11
:seedling: Remove references to VM class controller name and predicate handling

aruneshpa:

We introduced VM class controller name when there was a need to have controllers other than VM operator reconcile VMs.  For example, a controller written to reconcile VM resources on Fusion.

This code has not been used since its inception, so there's no point maintaining it.  This change removes it from our codebase. 

I am leaving the controllerName field in the types as is for now.  It can be deprecated and removed in a separate change.

```release-note
Remove references to VM class controller name and predicate handling
```


bryanv:
Do you want to remove the `DefaultVMClassControllerName` feature knob as well?

bryanv (approved):

PR #1049, opened 2025-07-15
:sapling: Drive create snapshot workflow only through the custom resource

aruneshpa:

Initially, we were driving the snapshot workflow through the snapshot custom resource modifying the VM's spec.currentSnapshot field -- which will then be reconciled by the VM controller to take a snapshot.

This presents a problem that we can not revert a VM to the current snapshot -- which is one of the most common workflows.  To address that, we will not watch the VirtualMachineSnapshot type custom resources that are not in ready state and reconcile those from the VM controller.  Even though this is not ideal (a controller reconciling a resources for which there's a dedicated controller already exists), this is necessary to keep all the VM configuration changes in a single controller so we don't have to use distributed locking between controllers.

This change:
- Updates the APIs with correct commentary
- Modifies the snapshot controller to always set OwnerRef on the snapshot pointing to the VM.
- Modifies the snapshot controller to not patch the spec.currentSnapshot on the VM.
- Modifies the VM controller to enqueue reconcile requests to VMs for any events on a snapshot resource that is not ready.
- Implement an ordering in the snapshot workflow so that the oldest snapshot request is processed.
- Modifies the revert workflow to clear out the spec.currentStatus (if set) so that users can revert the VM to the current snapshot.

I would appreciate a closer look on the snapshot ordering and the requeue semantics in case of both snapshot, and revert operations.

```release-note
Drive create snapshot workflow only through the custom resource
```

### DIFF
--- a/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
+++ b/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
@@ -132,63 +132,50 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
-	if err := r.ReconcileNormal(vmSnapshotCtx); err != nil {
-		vmSnapshotCtx.Logger.Error(err, "Failed to reconcile VirtualMachineSnapShot")
-		return ctrl.Result{}, err
-	}
-
-	return ctrl.Result{}, nil
+	return r.ReconcileNormal(vmSnapshotCtx)
 }
 
-func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) error {
+func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) (ctrl.Result, error) {
 	ctx.Logger.Info("Reconciling VirtualMachineSnapshot")
 
-	if !controllerutil.ContainsFinalizer(ctx.VirtualMachineSnapshot, Finalizer) {
-		// Set the finalizer and return so the object is patched immediately.
-		controllerutil.AddFinalizer(ctx.VirtualMachineSnapshot, Finalizer)
-		return nil
-	}
-	// return early if snapshot is ready; nothing to do
-	if conditions.IsTrue(ctx.VirtualMachineSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition) {
-		return nil
+	// If the finalizer is not present, add it.  Return so the object is patched immediately.
+	if controllerutil.AddFinalizer(ctx.VirtualMachineSnapshot, Finalizer) {
+		ctx.Logger.Info("Added finalizer to snapshot", "snapshot", ctx.VirtualMachineSnapshot.Name)
+		return ctrl.Result{}, nil
 	}
 
 	vmSnapshot := ctx.VirtualMachineSnapshot
 	ctx.Logger.Info("Fetching VirtualMachine from snapshot object", "vmSnapshot", vmSnapshot.Name)
+
+	// TODO: AKP: These validations must be removed once they are added in the validation webhook.
 	if vmSnapshot.Spec.VMRef == nil {
-		return errVMRefNil
+		return ctrl.Result{}, errVMRefNil
 	}
+
 	vm := &vmopv1.VirtualMachine{}
 	objKey := client.ObjectKey{Name: vmSnapshot.Spec.VMRef.Name, Namespace: vmSnapshot.Namespace}
 	if err := r.Get(ctx, objKey, vm); err != nil {
 		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
-		return err
+		return ctrl.Result{}, err
 	}
 
 	ctx.VM = vm
-	if vm.Status.UniqueID == "" {
-		return errors.New("VM hasn't been created and has no uniqueID")
+
+	// The snapshot must be owned by a VM.  Set an owner reference to the VM.
+	if err := controllerutil.SetControllerReference(ctx.VM, ctx.VirtualMachineSnapshot, r.Scheme()); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to set owner reference to snapshot: %w", err)
 	}
 
-	// vm object already set with snapshot reference
-	if vm.Status.CurrentSnapshot != nil && vm.Status.CurrentSnapshot.Name == ctx.VirtualMachineSnapshot.Name {
-		ctx.Logger.Info("VirtualMachine current snapshot already up to date", "status.currentSnapshot", vm.Status.CurrentSnapshot.Name)
-		return nil
+	// Return early if snapshot is ready; nothing to do
+	if conditions.IsTrue(ctx.VirtualMachineSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition) {
+		// TODO: Patch the status of the snapshot to populate any other fields such as storage used.
+		return ctrl.Result{}, nil
 	}
 
-	objRef := vmSnapshotCRToLocalObjectRef(ctx.VirtualMachineSnapshot)
-
-	// patch vm resource with the status.currentSnapshot
-	vmPatch := client.MergeFrom(vm.DeepCopy())
-	vm.Status.CurrentSnapshot = objRef
-	if err := r.Status().Patch(ctx, vm, vmPatch); err != nil {
-		return fmt.Errorf(
-			"failed to patch VM resource %s with current snapshot %s: %w", objKey,
-			ctx.VirtualMachineSnapshot.Name, err)
-	}
-
-	ctx.Logger.Info("Successfully patched VirtualMachine status with current snapshot reference", "vm.Name", vm.Name, "status.currentSnapshot", vm.Status.CurrentSnapshot.Name)
-	return nil
+	// The snapshot is not ready yet. The VM controller will handle the actual snapshot creation
+	// when it detects this snapshot resource. We just need to requeue to check again later.
+	ctx.Logger.V(5).Info("snapshot not yet complete. Requeuing reconcile request")
+	return ctrl.Result{RequeueAfter: pkgcfg.FromContext(ctx).SnapshotInProgressRequeueDelay}, nil
 }
 
 func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineSnapshotContext) error {
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -11,6 +11,7 @@ import (
 	"maps"
 	"math/rand"
 	"path"
+	"sort"
 	"strings"
 	"sync"
 	"text/template"
@@ -1169,17 +1170,35 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 		vmCtx.Logger.V(5).Info("No current snapshot specified, skipping revert")
 		return false, nil
 	}
-
-	// Check if the VM has a snapshot by the desired name.
 	desiredSnapshotName := vmCtx.VM.Spec.CurrentSnapshot.Name
-	snapObj, err := vs.findDesiredSnapshot(vmCtx, desiredSnapshotName)
-	if err != nil {
-		vmCtx.Logger.Error(err, "Error finding desired snapshot to revert to")
+
+	// Check if the snapshot CR exists.
+	snopCR := &vmopv1.VirtualMachineSnapshot{}
+	if err := vs.k8sClient.Get(
+		vmCtx,
+		ctrlclient.ObjectKey{Name: desiredSnapshotName, Namespace: vmCtx.VM.Namespace},
+		snopCR); err != nil {
+		vmCtx.Logger.Error(err, "Error getting snapshot CR", "snapshotName", desiredSnapshotName)
 		return false, err
 	}
 
-	// Govmomi ensures that if FindSnapshot doesn't return an error, a
-	// snapshot managed object is returned. Handle it anyway.
+	// Check if the snapshot CR is ready.  This is to ensure that the snapshot
+	//  as reconciled via our snapshot controller and is not an out-of-band snapshot.
+	if !pkgcnd.IsTrue(snopCR, vmopv1.VirtualMachineSnapshotReadyCondition) {
+		vmCtx.Logger.V(5).Info("Snapshot CR is not ready, skipping revert", "snapshotName", desiredSnapshotName)
+		return false, fmt.Errorf("snapshot CR is not ready, skipping revert")
+	}
+
+	// Check if the VM has a snapshot by the desired name.
+	snapObj, err := virtualmachine.FindSnapshot(vmCtx, desiredSnapshotName)
+	if err != nil {
+		vmCtx.Logger.Error(err, "Error finding snapshot", "snapshotName", desiredSnapshotName)
+
+		return false, err
+	}
+
+	// If FindSnapshot doesn't return an error, a snapshot managed object is returned.
+	// But handle a scenario where the object is nil anyway.
 	if snapObj == nil {
 		vmCtx.Logger.V(4).Info("Snapshot not found, treating as new snapshot request",
 			"snapshotName", desiredSnapshotName)
@@ -1239,47 +1258,6 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 	// the updated VM spec, labels, and annotations. The VM status
 	// will be updated in the subsequent reconcile loop.
 	return true, nil
-}
-
-// findDesiredSnapshot finds the snapshot object by name, returns nil if not found.
-func (vs *vSphereVMProvider) findDesiredSnapshot(
-	vmCtx pkgctx.VirtualMachineContext,
-	snapshotName string) (*vimtypes.ManagedObjectReference, error) {
-
-	snapObj, err := virtualmachine.FindSnapshot(vmCtx, snapshotName)
-	if err != nil {
-		// Only skip (return nil) for expected cases where we should not fail:
-		// 1. VM has no snapshots at all
-		// 2. Snapshot by name doesn't exist (this would be a different error message)
-		if errors.Is(err, virtualmachine.ErrNoSnapshots) {
-			vmCtx.Logger.V(4).Info("VM has no snapshots, skipping",
-				"snapshotName", snapshotName)
-			return nil, nil
-		}
-
-		// Check if snapshot by name doesn't exist
-		if errors.Is(err, virtualmachine.ErrSnapshotNotFound) {
-			vmCtx.Logger.V(4).Info("Snapshot not found on the VM (snapshot doesn't exist)",
-				"snapshotName", snapshotName)
-			return nil, nil
-		}
-
-		// For all other errors (property collector error, multiple snapshots, etc.),
-		// return the actual error and requeue the reconcile request.
-		vmCtx.Logger.Error(err, "Error finding snapshot",
-			"snapshotName", snapshotName)
-		return nil, err
-	}
-
-	// snapObj is nil when VM has snapshots but the specific snapshot doesn't exist
-	// This is the normal case when FindSnapshot succeeds but doesn't find the named snapshot.
-	if snapObj == nil {
-		vmCtx.Logger.V(4).Info("Snapshot not found on the VM (snapshot doesn't exist)",
-			"snapshotName", snapshotName)
-		return nil, nil
-	}
-
-	return snapObj, nil
 }
 
 // needsSnapshotRevert checks if the VM's current snapshot differs
@@ -1367,6 +1345,16 @@ func (vs *vSphereVMProvider) restoreVMSpecFromSnapshot(
 	// about pruning fields here since the backup process already
 	// handles that.
 	vmCtx.VM.Spec = *(&vm.Spec).DeepCopy()
+
+	// VM spec should never have spec.currentSnapshot set because the
+	// currentSnapshot is only set during a revert operation. However,
+	// in the off-chance that a snapshot was taken _during_ a revert,
+	// clear it out.
+	vmCtx.VM.Spec.CurrentSnapshot = nil
+
+	// TODO: AKP: We need to merge (and skip) some labels so that we
+	// are not simply reverting to the previous labels. Otherwise, we
+	// risk the VM being impacted after a snapshot revert.
 	vmCtx.VM.Labels = maps.Clone(vm.Labels)
 	vmCtx.VM.Annotations = maps.Clone(vm.Annotations)
 	// Empty out the status.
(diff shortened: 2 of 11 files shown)


controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -132,63 +132,50 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+	// If the finalizer is not present, add it.  Return so the object is patched immediately.
+	if controllerutil.AddFinalizer(ctx.VirtualMachineSnapshot, Finalizer) {
+		ctx.Logger.Info("Added finalizer to snapshot", "snapshot", ctx.VirtualMachineSnapshot.Name)
+		return ctrl.Result{}, nil
 	}
 
 	vmSnapshot := ctx.VirtualMachineSnapshot
 	ctx.Logger.Info("Fetching VirtualMachine from snapshot object", "vmSnapshot", vmSnapshot.Name)
+
+	// TODO: AKP: These validations must be removed once they are added in the validation webhook.
```
bryanv: IMO cool to leave this check here too. I guess though why is this field a pointer?

It isn't a huge deal but the the `errVMRefNil` should be a `NoRequeue` (or I guess NoRequeueNoErr)

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -132,63 +132,50 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
-		// Set the finalizer and return so the object is patched immediately.
-		controllerutil.AddFinalizer(ctx.VirtualMachineSnapshot, Finalizer)
-		return nil
-	}
-	// return early if snapshot is ready; nothing to do
-	if conditions.IsTrue(ctx.VirtualMachineSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition) {
-		return nil
+	// If the finalizer is not present, add it.  Return so the object is patched immediately.
+	if controllerutil.AddFinalizer(ctx.VirtualMachineSnapshot, Finalizer) {
+		ctx.Logger.Info("Added finalizer to snapshot", "snapshot", ctx.VirtualMachineSnapshot.Name)
```
bryanv: I don't think this log adds a ton of useful info. Easy to infer from the logs, or by looking at the object.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -132,63 +132,50 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
-	if vm.Status.UniqueID == "" {
-		return errors.New("VM hasn't been created and has no uniqueID")
-	}
 
-	// vm object already set with snapshot reference
-	if vm.Status.CurrentSnapshot != nil && vm.Status.CurrentSnapshot.Name == ctx.VirtualMachineSnapshot.Name {
-		ctx.Logger.Info("VirtualMachine current snapshot already up to date", "status.currentSnapshot", vm.Status.CurrentSnapshot.Name)
-		return nil
+	// The snapshot must be owned by a VM.  Set an owner reference to the VM.
+	if err := controllerutil.SetControllerReference(ctx.VM, ctx.VirtualMachineSnapshot, r.Scheme()); err != nil {
```
bryanv: Does this have to be a controller owner ref though? I guess I could see some higher level orchestrator of the VirtualMachineSnapshot being created which I would reasonable expect it to be the controller owner ref since it created the snapshot CR.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1169,17 +1170,35 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 		vmCtx.Logger.V(5).Info("No current snapshot specified, skipping revert")
 		return false, nil
 	}
+	desiredSnapshotName := vmCtx.VM.Spec.CurrentSnapshot.Name
+
+	// Check if the snapshot CR exists.
+	snopCR := &vmopv1.VirtualMachineSnapshot{}
```
bryanv: `snop`- > `snap`?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2588,3 +2648,63 @@ func (vs *vSphereVMProvider) vmResizeGetArgs(
...
+	// List all VirtualMachineSnapshot objects in the VM's namespace
+	var snapshotList vmopv1.VirtualMachineSnapshotList
+	if err := vs.k8sClient.List(vmCtx, &snapshotList, ctrlclient.InNamespace(vmCtx.VM.Namespace)); err != nil {
+		return nil, fmt.Errorf("failed to list VirtualMachineSnapshot objects: %w", err)
+	}
+
+	// Filter snapshots that reference this VM
+	var vmSnapshots []vmopv1.VirtualMachineSnapshot
+	for _, snapshot := range snapshotList.Items {
+		if snapshot.Spec.VMRef != nil && snapshot.Spec.VMRef.Name == vmCtx.VM.Name {
```
bryanv, line 2703: Kind of nit - instead of just the k8s object name, check the owner ref so we do it by UID (or some other field that more uniquely identifies the VM). I think if the VM was deleted and recreated with the same name and we get here before GC has completed we could end with snapshots that aren't really for the new VM, no?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1434,41 +1422,113 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+	// Find all VirtualMachineSnapshot objects that reference this VM
+	vmSnapshots, err := vs.getVirtualMachineSnapshotsForVM(vmCtx)
 	if err != nil {
 		return err
 	}
 
-	// If it's being deleted, skip snapshot creation.
-	if !vmSnapshot.DeletionTimestamp.IsZero() {
-		vmCtx.Logger.Info("Skipping snapshot creation because the VirtualMachineSnapshot is being deleted")
+	vmCtx.Logger.Info("Found snapshots for VM", "snapshotCount", len(vmSnapshots), "snapshots", vmSnapshots)
```
bryanv: Kind of nit but just dumping the k8s objects`vmSnapshots` could be a pretty large.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1169,17 +1170,35 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
...
+	if err := vs.k8sClient.Get(
+		vmCtx,
+		ctrlclient.ObjectKey{Name: desiredSnapshotName, Namespace: vmCtx.VM.Namespace},
+		snopCR); err != nil {
+		vmCtx.Logger.Error(err, "Error getting snapshot CR", "snapshotName", desiredSnapshotName)
+		return false, err
+	}
+
+	// Check if the snapshot CR is ready.  This is to ensure that the snapshot
+	//  as reconciled via our snapshot controller and is not an out-of-band snapshot.
```
lubronzhan: VirtualMachineSnapshotReadyCondition was set by VM controller, not snapshot controller. So here it even it's ready it doesn't mean it was already reconciled by snapshot controller. Unless we have a different condition set by snapshot controller that shows it already correctly set the usage?
Otherwise might be hard for SPU controller to know whether a Snapshot CR is ready for exposing its usage?
Or we could calculate the usage as part of VM Controller logic. Or we just move the usage calculation to SPU controller

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1434,41 +1422,113 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+		}
+
+		if snapshot.DeletionTimestamp.IsZero() {
+			totalPendingSnapshots++
+		}
+
+		// Found the first snapshot that needs processing
+		if snapshotToProcess == nil {
+			snapshotToProcess = &snapshot
+		}
```
lubronzhan, line 1477: This logic could be part of DeletionTimestamp check?
Since if it's being deleted, no need to process either? Then you don't need check on line 1475

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1434,41 +1422,113 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
 	}
 
-	vmCtx.Logger.Info("Creating a new snapshot of the virtual machine")
+	vmCtx.Logger.Info("Creating snapshot on vSphere",
+		"snapshotName", snapshotToProcess.Name)
 	snapMoRef, err := virtualmachine.SnapshotVirtualMachine(snapArgs)
 	if err != nil {
-		vmCtx.Logger.Error(err, "failed to snapshot virtual machine")
+		// Mark the snapshot as failed and clear in-progress status
+		vs.markSnapshotFailed(vmCtx, snapshotToProcess, err)
```
lubronzhan, line 1540: If `marking failed` Patch call failed, there is a chance that the snapshot stuck in `inprogress` condition forever. Maybe we should add retry around the patch call?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1434,41 +1422,113 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+
+	// If the oldest snapshot is being deleted, wait for it to finish.
+	if !snapshotToProcess.DeletionTimestamp.IsZero() {
+		vmCtx.Logger.V(4).Info("Snapshot is being deleted, waiting for completion",
+			"snapshotName", snapshotToProcess.Name)
 		return nil
 	}
 
+	// If the oldest snapshot is in progress, wait for it to complete
+	if pkgcnd.IsTrue(snapshotToProcess, "VirtualMachineSnapshotInProgress") {
```
lubronzhan: Nit `VirtualMachineSnapshotInProgress ` could be declared as a constant

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1434,41 +1422,113 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+		// Mark the snapshot as failed and clear in-progress status
+		vs.markSnapshotFailed(vmCtx, snapshotToProcess, err)
+		vmCtx.Logger.Error(err, "Failed to create snapshot", "snapshotName", snapshotToProcess.Name)
 		return err
 	}
 
-	if err = PatchSnapshotStatus(vmCtx, vs.k8sClient,
-		&vmSnapshot, snapMoRef); err != nil {
+	// Update the snapshot status with the successful result
+	if err = PatchSnapshotStatus(vmCtx, vs.k8sClient, snapshotToProcess, snapMoRef); err != nil {
```
lubronzhan, line 1546: Here it should remove `VirtualMachineSnapshotInProgress` condition as well.

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: In that case, then maybe better to also add logic of `adding controllerRef to vmSnapshot` in VM Controller as well? Since vmsnapshot controller might get the wrong VM as well. And vmSnapshot should also check the UID of VM in controllerRef before proceeding.

test/builder/dummies.go
```diff
@@ -549,23 +550,21 @@ func DummyVirtualMachineWebConsoleRequest(namespace, wcrName, vmName, pubKey str
...
 func DummyVirtualMachineSnapshot(namespace, name, vmName string) *vmopv1.VirtualMachineSnapshot {
 	return &vmopv1.VirtualMachineSnapshot{
-		TypeMeta: metav1.TypeMeta{
-			Kind:       "VirtualMachineSnapshot",
-			APIVersion: "vmoperator.vmware.com/v1alpha4",
-		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: namespace,
-			Finalizers: []string{
```
lubronzhan: The finalizers are needed for other test

test/builder/dummies.go
```diff
@@ -549,23 +550,21 @@ func DummyVirtualMachineWebConsoleRequest(namespace, wcrName, vmName, pubKey str
 
 func DummyVirtualMachineSnapshot(namespace, name, vmName string) *vmopv1.VirtualMachineSnapshot {
 	return &vmopv1.VirtualMachineSnapshot{
-		TypeMeta: metav1.TypeMeta{
```
lubronzhan: I believe these are also needed for other tests

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: But that would just be a temporary workaround. If the APIserver is down for a long time, we'd better to have a timeout check on the in progress status, in case the retry doesn't cover that amount of time.
But that could be a future improvement

lubronzhan:
Also, question about the status.CurrentSnapshot, I assume it's still used to show the indication `You are here` in vsphere UI. Then we should always specify the value of it whenever a Snapshot is created for a VM?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
aruneshpa: Fair point.  Let me change this.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: This was a leftover from debugging.  I removed this log line since in the called method, we are logging the count of snapshots.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Not sure what you mean by "this logic",  could you clarify?

What I am trying to do is to filter out all the snapshots that are not ready -- including the ones that are in deletion.  Then, pick the oldest one.  If the oldest snapshot is being deleted, we should wait for it to be gone since the disk chain is still being modified.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: >  I think if the VM was deleted and recreated with the same name and we get here before GC has completed we could end with snapshots that aren't really for the new VM, no?

This shouldn't be a concern generally because if a VM is marked for deletion, all the snapshots will be marked for deletion as well.  And we skip those everywhere from processing.  I still like the idea of checking the owner ref, so will add that here.

test/builder/dummies.go
aruneshpa: TypeMeta should not be needed since we are defining these as a concrete type.  That's why other tests don't need specify that as well.  Did you have a particular test in mind that uses this?

test/builder/dummies.go
aruneshpa: I am generally trying to avoid creating the dummy object with the finalizer all the time.  A finalizer is something that we should verify that is added as part of a reconcile.

aruneshpa:
> Also, question about the status.CurrentSnapshot, I assume it's still used to show the indication `You are here` in vsphere UI. Then we should always specify the value of it whenever a Snapshot is created for a VM?

That's exactly the intention.  The `status.currentSnapshot` will always be set to the current snapshot of the VM (similar to a VM's ConfigInfo that always has this field set as long as the VM points to a snapshot).

test/builder/dummies.go
lubronzhan: Ok, I added it for the convenience  of testing reconcileDeletion. What you mentioned make sense for reconcileNormal. Then changes to deletion are needed

test/builder/dummies.go
lubronzhan: I see DummyBasicVirtualMachine also defines this https://github.com/vmware-tanzu/vm-operator/blob/a173780248ce5ffd1a41b5e2979fa703c8c9e56e/test/builder/dummies.go#L284-L372.
Defining this makes it convenient when we try to create LocalObjectRef directly from the object using function like `vmSnapshotCRToLocalObjectRefWithDefaultVersion` 
https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_unit_test.go#L436
Since I'm comparing the whole LocalObjectRef instead of just comparing the object name

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: I mean you come change it to below
```
		if snapshot.DeletionTimestamp.IsZero() {
			totalPendingSnapshots++
			// Found the first snapshot that needs processing
			if snapshotToProcess == nil {
				snapshotToProcess = &snapshot
			}
		}

```
Then there is no need for 
```
	if !snapshotToProcess.DeletionTimestamp.IsZero() {
		vmCtx.Logger.V(4).Info("Snapshot is being deleted, waiting for completion",
			"snapshotName", snapshotToProcess.Name)
		return nil
	}
```

lubronzhan:
> > Also, question about the status.CurrentSnapshot, I assume it's still used to show the indication `You are here` in vsphere UI. Then we should always specify the value of it whenever a Snapshot is created for a VM?
> 
> That's exactly the intention. The `status.currentSnapshot` will always be set to the current snapshot of the VM (similar to a VM's ConfigInfo that always has this field set as long as the VM points to a snapshot).

~Then I guess the best place for this logic would be in reconcileSnapshot in vmcontroller I assume, I don't see it in the PR just want to double check~
Ohk found it here https://github.com/aruneshpa/vm-operator/blob/features/snapshot/dont-use-spec-for-create-snapshot/pkg/providers/vsphere/virtualmachine/snapshot.go#L63

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -132,63 +132,50 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
-	if conditions.IsTrue(ctx.VirtualMachineSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition) {
-		return nil
+	// If the finalizer is not present, add it.  Return so the object is patched immediately.
+	if controllerutil.AddFinalizer(ctx.VirtualMachineSnapshot, Finalizer) {
+		ctx.Logger.Info("Added finalizer to snapshot", "snapshot", ctx.VirtualMachineSnapshot.Name)
+		return ctrl.Result{}, nil
 	}
 
 	vmSnapshot := ctx.VirtualMachineSnapshot
 	ctx.Logger.Info("Fetching VirtualMachine from snapshot object", "vmSnapshot", vmSnapshot.Name)
```
bryanv: The snapshot object name will already be a part of the logger

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: VirtualMachineSnapshotReadyCondition is the final ready condition of a snapshot.
If this condition is not true, that means the the snapshot was never processed by the VM controller -- or that it is in progress. In either of these cases, we can't revert the VM to this snapshot.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: No.  The current version is intentional :)

So, we are trying to find out all snapshots that are not _done_.  We want this to also include snapshots that are being deleted.  Then, if the oldest snapshot is being deleted, we want to requeue the reconcile so the deletion can finish.  In your suggested diff, we would end up skipping the deleting snapshots.  Let me know if this makes sense.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: I have filed a tracker to address this.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: I removed that condition and made it a reason under the ready condition to conform to upstream patterns.

test/builder/dummies.go
aruneshpa: Added this back.

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: Make sense, thanks

pkg/util/vmopv1/vm.go
```diff
@@ -19,8 +18,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/go-logr/logr"
```
lubronzhan, line 21: nit: is this intentional to put these in a different section

lubronzhan (approved):
LGTM!

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: I was asking about the the relationship between calculating the disk usage of Snapshot and readiness. Now since we calculate usage in VMSnapshot controller while marking VM as ready in VM controller, there is a chance that the VM is marked as ready, but it doesn't have usage calculated.
I now just skip the calculation of that snapshot without any error in my PR. We can continue the discussion there

pkg/util/vmopv1/vm.go
aruneshpa: Yes. Grouping the k8s imports together, and Github ones together.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2604,3 +2668,78 @@ func (vs *vSphereVMProvider) vmResizeGetArgs(
...
+			// VM. It is possible that the snapshot controller is
+			// still adding the owner reference.
+			owner, err := controllerutil.HasOwnerReference(snapshot.OwnerReferences, vmCtx.VM, vs.k8sClient.Scheme())
+			if err != nil {
+				vmCtx.Logger.Error(err, "error in checking if snapshot is owned by the VM")
+				return nil, err
+			}
+
+			if !owner {
+				return nil, fmt.Errorf("snapshot is not owned by the VM resource")
```
bryanv: The comment says we're skipping ones that aren't owned so is this really the desired behavior?

palnabarun: I think we should `continue` here?

aruneshpa: I will update the comment here.  But yeah, we don't want to "skip" these.  This is a snapshot that was create for a VM.  And even though a VM is still around (namespace + name) but this is not the same VM (snapshot not owned by this VM).  So we return an error and requeue.  If the owner ref didn't exist because the controller did not have a chance to add the owner ref, then the requeue will handle that.

lubronzhan: I guess it got confused since you already filter the VMSnapshot by its VMRef in func`getVirtualMachineSnapshotsForVM`, so the VMSnapshot at this stage should eventually have the ownerRef to this VM, not other VM.

aruneshpa: I updated the diff :)
Now I have moved the owner check down where I return if the oldest snapshot is still waiting for the owner ref.

lubronzhan: No wonder, I thought my memory got messed up

palnabarun: This makes sense now. :D

PR #1051, opened 2025-07-17
vm: add snapshot validations

palnabarun:

Adds validation for VirtualMachine currentSnapshot reference

None

```release-note
NONE
```

### DIFF
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -53,9 +53,10 @@ const (
 	isRestrictedNetworkKey               = "IsRestrictedNetwork"
 	allowedRestrictedNetworkTCPProbePort = 6443
 
-	vmiKind     = "VirtualMachineImage"
-	cvmiKind    = "ClusterVirtualMachineImage"
-	vmclassKind = "VirtualMachineClass"
+	vmiKind        = "VirtualMachineImage"
+	cvmiKind       = "ClusterVirtualMachineImage"
+	vmclassKind    = "VirtualMachineClass"
+	vmSnapshotKind = "VirtualMachineSnapshot"
 
 	readinessProbeOnlyOneAction                = "only one action can be specified"
 	tcpReadinessProbeNotAllowedVPC             = "VPC networking doesn't allow TCP readiness probe to be specified"
@@ -164,6 +165,7 @@ func (v validator) ValidateCreate(ctx *pkgctx.WebhookRequestContext) admission.R
 	fieldErrs = append(fieldErrs, v.validateChecks(ctx, vm, nil)...)
 	fieldErrs = append(fieldErrs, v.validateNextPowerStateChangeTimeFormat(ctx, vm)...)
 	fieldErrs = append(fieldErrs, v.validateBootOptions(ctx, vm)...)
+	fieldErrs = append(fieldErrs, v.validateSnapshot(ctx, vm, nil)...)
 
 	validationErrs := make([]string, 0, len(fieldErrs))
 	for _, fieldErr := range fieldErrs {
@@ -264,6 +266,7 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
 	fieldErrs = append(fieldErrs, v.validateChecks(ctx, vm, oldVM)...)
 	fieldErrs = append(fieldErrs, v.validateNextPowerStateChangeTimeFormat(ctx, vm)...)
 	fieldErrs = append(fieldErrs, v.validateBootOptions(ctx, vm)...)
+	fieldErrs = append(fieldErrs, v.validateSnapshot(ctx, vm, oldVM)...)
 
 	validationErrs := make([]string, 0, len(fieldErrs))
 	for _, fieldErr := range fieldErrs {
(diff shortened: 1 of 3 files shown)


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1852,3 +1854,57 @@ func (v validator) validateBootOptions(
...
+	if vm.Spec.CurrentSnapshot.Kind != vmSnapshotKind {
+		allErrs = append(allErrs, field.Invalid(
+			snapshotPath.Child("kind"),
+			vm.Spec.CurrentSnapshot.Kind,
+			fmt.Sprintf("must be %q", vmSnapshotKind),
+		))
+	}
+
+	if vm.Spec.CurrentSnapshot.Name == "" {
+		allErrs = append(allErrs, field.Invalid(
```
bryanv: `field.Required()` instead

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1852,3 +1854,57 @@ func (v validator) validateBootOptions(
...
+	snapshotPath := field.NewPath("spec", "currentSnapshot")
+
+	if vm.Spec.CurrentSnapshot.APIVersion != "" {
+		gv, err := schema.ParseGroupVersion(vm.Spec.CurrentSnapshot.APIVersion)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(snapshotPath.Child("apiVersion"),
+				vm.Spec.CurrentSnapshot.APIVersion, "must be valid group version"))
+		}
+
+		if gv.Group != "vmoperator.vmware.com" {
```
bryanv: Use `vmopv1.GroupName`

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1852,3 +1854,57 @@ func (v validator) validateBootOptions(
...
+			allErrs = append(allErrs, field.Invalid(snapshotPath.Child("apiVersion"),
+				vm.Spec.CurrentSnapshot.APIVersion, "group must be vmoperator.vmware.com"))
+		}
+	}
+
+	if vm.Spec.CurrentSnapshot.Kind != vmSnapshotKind {
+		allErrs = append(allErrs, field.Invalid(
```
bryanv: Do you think `field.NotSupported()` is more appropriate for this?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1852,3 +1854,57 @@ func (v validator) validateBootOptions(
...
+
+	if vm.Spec.CurrentSnapshot.Name == "" {
+		allErrs = append(allErrs, field.Invalid(
+			snapshotPath.Child("name"),
+			vm.Spec.CurrentSnapshot.Name,
+			"must not be empty",
+		))
+	}
+
+	vmSnapshot := &vmopv1.VirtualMachineSnapshot{}
```
bryanv: If the snapshot gets deleted for some reason - or even just hasn't made itno the client cache yet, all updates to the VM will fail. I don't think we want that behavior.

We deviate it from in some places but in general the validation webhook should only check for internal consistently. Anything else needs to be handle during the reconcile.

palnabarun: > or even just hasn't made itno the client cache yet, all updates to the VM will fail. I don't think we want that behavior.

This is a valid concern. Let's have this check in the reconciler instead. The check needs to be in the revert workflow.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -264,6 +265,7 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
 	fieldErrs = append(fieldErrs, v.validateChecks(ctx, vm, oldVM)...)
 	fieldErrs = append(fieldErrs, v.validateNextPowerStateChangeTimeFormat(ctx, vm)...)
 	fieldErrs = append(fieldErrs, v.validateBootOptions(ctx, vm)...)
+	fieldErrs = append(fieldErrs, v.validateSnapshot(ctx, vm)...)
```
lubronzhan: Question, do we need to do this in CreateRequest

webhooks/virtualmachine/validation/virtualmachine_validator_intg_test.go
```diff
@@ -249,6 +249,39 @@ func intgTestsValidateUpdate() {
...
+		When("snapshot ref is invalid", func() {
+			BeforeEach(func() {
+				ctx.vm.Spec.CurrentSnapshot = &common.LocalObjectRef{
+					Kind:       "VMSnapshot",
+					APIVersion: "",
+					Name:       "",
+				}
+			})
+
+			It("should allow the request", func() {
```
lubronzhan:
```suggestion
			It("should not allow the request", func() {
```

webhooks/virtualmachine/validation/virtualmachine_validator.go
palnabarun: No, when a VirtualMachine is created, there's no snapshot for it to begin with.

@aruneshpa what is the workflow for restoring a VM from snapshot? Will it involve creating a VM with spec.currentSnapshot set to some VirtualMachineSnapshot CR?

lubronzhan: >what is the workflow for restoring a VM from snapshot? Will it involve creating a VM with spec.currentSnapshot set to some VirtualMachineSnapshot CR?

No, the user will just set the VM CR's vm.spec.currentSnapshot to another snapshot CR

> No, when a VirtualMachine is created, there's no snapshot for it to begin with.

Got it, just curious whether user want to create a VM along with a snapshot

aruneshpa: You should add a validation to make sure spec.currentSnapshot is not set during creation.  Like you said, it does not make sense to have a VM be created _while_ trying to restore to a snapshot.

aruneshpa (approved):
LGTM.

webhooks/virtualmachine/validation/virtualmachine_validator.go
lubronzhan: Ohhh yeah that field is for revert. My head is not working

palnabarun: > You should add a validation to make sure spec.currentSnapshot is not set during creation. 

This is done.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1852,3 +1855,52 @@ func (v validator) validateBootOptions(
...
+	if vm.Spec.CurrentSnapshot == nil {
+		return allErrs
+	}
+
+	if vm.Spec.CurrentSnapshot.APIVersion != "" {
+		gv, err := schema.ParseGroupVersion(vm.Spec.CurrentSnapshot.APIVersion)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(snapshotPath.Child("apiVersion"),
+				vm.Spec.CurrentSnapshot.APIVersion, "must be valid group version"))
+		}
```
bryanv: Maybe '} else if gv.Group != ....` so we don't double up errors?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1852,3 +1855,52 @@ func (v validator) validateBootOptions(
...
+
+	if vm.Spec.CurrentSnapshot.Kind != vmSnapshotKind {
+		allErrs = append(allErrs, field.NotSupported(
+			snapshotPath.Child("kind"), vm.Spec.CurrentSnapshot.Kind, []string{vmSnapshotKind},
+		))
+	}
+
+	if vm.Spec.CurrentSnapshot.Name == "" {
+		allErrs = append(allErrs, field.Required(
+			snapshotPath.Child("name"), "must not be empty",
```
bryanv: I think the detail message can be omitted here - that's what we typically do and just the field be required is sufficient

PR #1055, opened 2025-07-21
🌱 VMICache location / disk promotion ignores snapshots & waits on guest customization

akutz:

This patch updates the VMICache controller to create a top-level directory per cached item. This ensures there is no single, top-level directory per datastore for all cached disks/files. On VSAN, a top-level datastore directory is a volume, and having multiple, concurrent threads read/write to this volume can lead to read/write lock errors.

This patch also updates the disk promotion reconciler to ignore any disks that participate in snapshots. This prevents a VM from constantly being tagged as out-of-sync for disk promotion if the VM has a snapshot.
    
Finally, this patch causes disk promotion to wait until there is not a pending or running guest customization.

Fixes `NA`

This fix has been manually verified on a VSAN-enabled testbed, and it is about to be put through gce2e.

```release-note
NONE
```

### DIFF
--- a/pkg/util/vsphere/library/item_cache.go
+++ b/pkg/util/vsphere/library/item_cache.go
@@ -64,8 +64,8 @@ func CacheStorageURIs(
 	ctx context.Context,
 	client CacheStorageURIsClient,
 	dstDatacenter, srcDatacenter *object.Datacenter,
-	dstDir string,
-	dstDisksFormat vimtypes.DatastoreSectorFormat,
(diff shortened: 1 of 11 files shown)


pkg/util/vsphere/library/item_cache.go
```diff
@@ -188,6 +197,9 @@ func copyFile(
 			dstDatacenter,
 			false)
 	}
+	if err != nil {
```
bryanv: Duplicate error check

PR #1057, opened 2025-07-21
✨ Add VM Group controller placement

bryanv:

Add the basic  plumbing for VM Group placement in the group controller, and a noop group placement implementation in the VM Provider.

Fixes #

```release-note
NONE
```


controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -481,11 +487,138 @@ patchMember:
...
+			isPlaced = conditions.IsTrue(&member, vmopv1.VirtualMachineGroupMemberConditionPlacementReady)
+			if !isPlaced {
+				isLinked = conditions.IsTrue(&member, vmopv1.VirtualMachineGroupMemberConditionGroupLinked)
+			}
+			break
+		}
+	}
+	if isPlaced {
+		return nil, nil
+	} else if !isLinked {
```
dilyar85: Nit: `else if` => `if`

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -481,11 +487,138 @@ patchMember:
...
+		return nil, fmt.Errorf("VM %q is not linked", vmName)
+	}
+
+	vm := &vmopv1.VirtualMachine{}
+	if err := r.Get(ctx, client.ObjectKey{Name: vmName, Namespace: vmGroup.Namespace}, vm); err != nil {
+		return nil, fmt.Errorf("failed to get group member VM %q: %w", vmName, err)
```
dilyar85, line 590: Should we ignore not found here if the VM was deleted right after we checked its condition in group member status? Could avoid recursively getting all members again in that case.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -481,11 +487,138 @@ patchMember:
...
+	parentVMGroup *vmopv1.VirtualMachineGroup,
+	groupName string) ([]providers.VMGroupPlacement, error) {
+
+	var isLinked bool
+	for _, member := range parentVMGroup.Status.Members {
+		if member.Name == groupName && member.Kind == vmgKind {
+			isLinked = conditions.IsTrue(&member, vmopv1.VirtualMachineGroupMemberConditionGroupLinked)
+			break
+		}
+	}
```
dilyar85: Might be worth making this a helper function that returns a member status from group to avoid duplicate code in `getVMForPlacement()`?

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
bryanv: I guess I'd rather wait for the later reconcile so the placement is always working with the group as-is

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
bryanv: Sure. But this is just code to get this "working" - my later plan is to refactor things a bit more so this uses a map to do the lookup, and that the map we've already created as a part of the power state reconcile is available here so we don't duplicate that for the root group

dilyar85 (approved):
LGTM, thanks @bryanv!

PR #1050, opened 2025-07-15
✨ Introduce VirtualMachineGroupPublishRequest API

akutz:
**What does this PR do, and why is it needed?**

This patch introduces the VirtualMachineGroupPublishRequest API used to publish a group of VMs. This API is largely syntactic sugar around a user creating multiple VirtualMachinePublishRequest objects individually.

Closes #975 

**Which issue(s) is/are addressed by this PR?** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes `NA`

**Are there any special notes for your reviewer**:

**Please add a release note if necessary**:

```release-note
Introduce the VirutalMachineGroupPublishRequest API (sans controller)
```

PR #1059, opened 2025-07-23
✨ Introduce VirtualMachineGroupPublishRequest controller

zhangheliu:

This PR implements the VirtualMachineGroupPublishRequest API that was introduced on https://github.com/vmware-tanzu/vm-operator/pull/1050. 

Fixes NA

Manual Testing:

- created VMs 
Screenshot 2025-07-28 at 8 34 29 PM

- created  the VM group publish request
Screenshot 2025-07-28 at 8 34 51 PM

- completed the VM group publish request
Screenshot 2025-07-28 at 8 35 22 PM

Webhook changes will be in a separate PR.  The mutating webhook is responsible for generating spec.virtualMachines list from the VM group. 

```release-note
Implement the VirutalMachineGroupPublishRequest API
```

### DIFF
--- a/api/v1alpha4/virtualmachinegroup_publishrequest_types.go
+++ b/api/v1alpha4/virtualmachinegroup_publishrequest_types.go
@@ -15,6 +15,10 @@ const (
 	// The condition's status is set to true only when all other conditions
 	// present on the resource have a truthy status.
 	VirtualMachineGroupPublishRequestConditionComplete = "Complete"
+
+	// VirtualMachineGroupPublishRequestConditionReasonPending indicates there are still pending
+	// VirtualMachinePublishRequest to be completed.
+	VirtualMachineGroupPublishRequestConditionReasonPending = "VirtualMachinePublishRequestsPending"
 )
 
 // VirtualMachineGroupPublishRequestSpec defines the desired state of a
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -16,6 +16,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachine"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachineclass"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinegroup"
+	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinegrouppublishrequest"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachineimagecache"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinepublishrequest"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinereplicaset"
@@ -84,6 +85,9 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 	if pkgcfg.FromContext(ctx).Features.VMGroups {
 		if err := virtualmachinegroup.AddToManager(ctx, mgr); err != nil {
 			return fmt.Errorf("failed to initialize VMG controller: %w", err)
+		}
+		if err := virtualmachinegrouppublishrequest.AddToManager(ctx, mgr); err != nil {
+			return fmt.Errorf("failed to initialize VirtualMachineGroupPublishRequest controller: %w", err)
 		}
 	}
 
--- a/controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
+++ b/controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
@@ -6,20 +6,29 @@ package virtualmachinegrouppublishrequest
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 	"strings"
+	"time"
 
+	"github.com/go-logr/logr"
+	imgregv1a1 "github.com/vmware-tanzu/image-registry-operator-api/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
-	"github.com/go-logr/logr"
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	pkgerr "github.com/vmware-tanzu/vm-operator/pkg/errors"
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
@@ -27,7 +36,10 @@ import (
 )
 
 const (
-	finalizerName = "vmoperator.vmware.com/virtualmachinegrouppublishrequest"
+	finalizerName         = "vmoperator.vmware.com/virtualmachinegrouppublishrequest"
+	undefinedSpecErrorMsg = "spec.%s is undefined"
+	invalidSpecErrorMsg   = "webhooks failed to mutate/validate spec. please delete and create again."
+	delTTLMsg             = "%s vm group publish request due to TTL expired"
 )
 
 // AddToManager adds this package's controller to the provided manager.
@@ -49,6 +61,7 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 	)
 	return ctrl.NewControllerManagedBy(mgr).
 		For(controlledType).
+		Owns(&vmopv1.VirtualMachinePublishRequest{}).
 		WithOptions(controller.Options{
 			MaxConcurrentReconciles: ctx.MaxConcurrentReconciles,
 			LogConstructor: pkgutil.ControllerLogConstructor(
@@ -59,7 +72,7 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		Complete(r)
 }
 
-// Reconciler reconciles a VirtualMachineGroupePublishRequest object.
+// Reconciler reconciles a VirtualMachineGroupPublishRequest object.
 type Reconciler struct {
 	client.Client
 	Context    context.Context
@@ -95,63 +108,308 @@ func NewReconciler(
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
 
-	obj := &vmopv1.VirtualMachineGroupPublishRequest{}
-	if err := r.apiReader.Get(ctx, req.NamespacedName, obj); err != nil {
+	vmGroupPublishRequest := &vmopv1.VirtualMachineGroupPublishRequest{}
+	if err := r.Get(ctx, req.NamespacedName, vmGroupPublishRequest); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	objCtx := &pkgctx.VirtualMachineGroupPublishRequestContext{
-		Context: ctx,
-		Logger:  pkgutil.FromContextOrDefault(ctx),
-		Obj:     obj,
+	vmGroupPublishRequestCtx := &pkgctx.VirtualMachineGroupPublishRequestContext{
+		Context:               ctx,
+		Logger:                pkgutil.FromContextOrDefault(ctx),
+		VMGroupPublishRequest: vmGroupPublishRequest,
 	}
-	patchHelper, err := patch.NewHelper(obj, r.Client)
+	patchHelper, err := patch.NewHelper(vmGroupPublishRequest, r.Client)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf(
 			"failed to init patch helper for %s/%s: %w",
-			obj.Namespace,
-			obj.Name, err)
+			vmGroupPublishRequest.Namespace,
+			vmGroupPublishRequest.Name,
+			err)
 	}
 
 	defer func() {
-		if err := patchHelper.Patch(ctx, obj); err != nil {
+		if err := patchHelper.Patch(ctx, vmGroupPublishRequest); err != nil {
 			if reterr == nil {
 				reterr = err
 			}
-			objCtx.Logger.Error(err, "patch failed")
+			vmGroupPublishRequestCtx.Logger.Error(err, "patch failed")
 		}
 	}()
 
-	if !obj.DeletionTimestamp.IsZero() {
-		return r.ReconcileDelete(objCtx)
+	if !vmGroupPublishRequest.DeletionTimestamp.IsZero() {
+		return r.ReconcileDelete(vmGroupPublishRequestCtx)
 	}
 
-	return r.ReconcileNormal(objCtx)
+	return pkgerr.ResultFromError(r.ReconcileNormal(vmGroupPublishRequestCtx))
 }
 
 func (r *Reconciler) ReconcileDelete(
 	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (ctrl.Result, error) {
-
 	ctx.Logger.Info("Reconciling VirtualMachineGroupPublishRequest Deletion")
 
-	if controllerutil.ContainsFinalizer(ctx.Obj, finalizerName) {
-		controllerutil.RemoveFinalizer(ctx.Obj, finalizerName)
+	if controllerutil.ContainsFinalizer(ctx.VMGroupPublishRequest, finalizerName) {
+		ctx.Logger.Info("Removing owned VirtualMachinePublishRequests")
+		if err := r.deleteVMPublishRequests(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		controllerutil.RemoveFinalizer(ctx.VMGroupPublishRequest, finalizerName)
 		ctx.Logger.Info("Removed finalizer for VirtualMachineGroupPublishRequest")
 	}
 	ctx.Logger.V(4).Info("Finished Reconciling VirtualMachineGroupPublishRequest Deletion")
 	return ctrl.Result{}, nil
 }
 
-func (r *Reconciler) ReconcileNormal(
-	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (_ ctrl.Result, reterr error) {
+func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineGroupPublishRequestContext) error {
+	ctx.Logger.Info("Reconciling VirtualMachineGroupPublishRequest")
+	vmGroupPublishReq := ctx.VMGroupPublishRequest
 
-	ctx.Logger.Info("Reconciling VirtualMachinePublishRequest")
-	vmGroupPublishReq := ctx.Obj
-
-	if !controllerutil.ContainsFinalizer(vmGroupPublishReq, finalizerName) {
-		controllerutil.AddFinalizer(vmGroupPublishReq, finalizerName)
-		return ctrl.Result{}, reterr
+	if controllerutil.AddFinalizer(vmGroupPublishReq, finalizerName) {
+		ctx.Logger.Info("Added finalizer for VirtualMachineGroupPublishRequest")
+		return nil
 	}
 
-	return ctrl.Result{}, nil
+	if conditions.IsTrue(vmGroupPublishReq, vmopv1.VirtualMachineGroupPublishRequestConditionComplete) {
+		return r.reconcileSpecTTL(ctx)
+	}
+
+	if err := r.verifySpec(vmGroupPublishReq); err != nil {
+		return err
+	}
+
+	return r.reconcileStatus(ctx)
+}
+
+func (r *Reconciler) reconcileStatus(ctx *pkgctx.VirtualMachineGroupPublishRequestContext) error {
+	if ctx.VMGroupPublishRequest.Status.StartTime.IsZero() {
+		ctx.VMGroupPublishRequest.Status.StartTime = metav1.NewTime(time.Now())
+	}
+
+	existReqsMap, pendingReqsSet, completedReqsSet, err := r.getVMPublishRequests(ctx)
+	if err != nil {
+		return err
+	}
+
+	r.reconcileStatusImages(ctx, existReqsMap)
+
+	desiredReqsSet := sets.New(ctx.VMGroupPublishRequest.Spec.VirtualMachines...)
+	if err := r.reconcileNewPublishRequests(ctx, sets.KeySet(existReqsMap), desiredReqsSet); err != nil {
+		return err
+	}
+	return r.reconcileStatusCompletedCondition(ctx, pendingReqsSet, completedReqsSet, desiredReqsSet)
+}
+
+// getVMPublishRequests returns a map of VMGroupPublishRequests own by the VirtualMachineGroupPublishRequest
+// along with a set of pending VMGroupPublishRequest and a set of completed VMGroupPublishRequest.
+// The key for those objects is the source VirtualMachine name.
+func (r *Reconciler) getVMPublishRequests(
+	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (
+	map[string]vmopv1.VirtualMachinePublishRequest,
+	sets.Set[string],
+	sets.Set[string],
+	error) {
+
+	reqs := &vmopv1.VirtualMachinePublishRequestList{}
+	// retrieve from API server directly to avoid duplicate resource creation in case that the newly
+	// created vm publish request has not made it back into the client's cache
+	if err := r.apiReader.List(ctx,
+		reqs,
+		client.InNamespace(ctx.VMGroupPublishRequest.Namespace),
+		client.MatchingLabels{
+			vmopv1.VirtualMachinePublishRequestManagedByLabelKey: ctx.VMGroupPublishRequest.Name,
+		}); err != nil {
+		ctx.Logger.Error(err, "failed to list VirtualMachinePublishRequest")
+		return nil, nil, nil, err
+	}
+
+	existReqsMap := make(map[string]vmopv1.VirtualMachinePublishRequest)
+	pendingReqsSet := sets.Set[string]{}
+	completedReqsSet := sets.Set[string]{}
+	for _, req := range reqs.Items {
+		if metav1.IsControlledBy(&req, ctx.VMGroupPublishRequest) {
+			existReqsMap[req.Spec.Source.Name] = req
+			if req.Status.Ready {
+				completedReqsSet.Insert(req.Spec.Source.Name)
+			} else {
+				pendingReqsSet.Insert(req.Spec.Source.Name)
+			}
+		}
+	}
+
+	return existReqsMap, pendingReqsSet, completedReqsSet, nil
+}
+
+// reconcileStatusImages updates status's image status by taking a map of VMGroupPublishRequests
+// own by the VirtualMachineGroupPublishRequest and converting them into a sorted list of
+// VirtualMachineGroupPublishRequestImageStatus. Sorting prevents reconciling when list is unchanged.
+func (r *Reconciler) reconcileStatusImages(
+	ctx *pkgctx.VirtualMachineGroupPublishRequestContext,
+	existReqsMap map[string]vmopv1.VirtualMachinePublishRequest) {
+
+	imageStatuses := make([]vmopv1.VirtualMachineGroupPublishRequestImageStatus, 0, len(existReqsMap))
+	for _, req := range existReqsMap {
+		imageStatuses = append(imageStatuses, vmopv1.VirtualMachineGroupPublishRequestImageStatus{
+			Source:             req.Spec.Source.Name,
+			PublishRequestName: req.Name,
+			ImageName:          req.Status.ImageName,
+			Conditions:         req.Status.Conditions,
+		})
+	}
+
+	ctx.VMGroupPublishRequest.Status.Images = slices.SortedFunc(slices.Values(imageStatuses),
+		func(a, b vmopv1.VirtualMachineGroupPublishRequestImageStatus) int {
+			return strings.Compare(a.Source, b.Source)
+		})
+}
+
+// reconcileNewPublishRequests compares a set of exist requests and a set of expected requests and
+// creates new requests if they are missing in the exist requests set.
+// It returns nil ONLY when no new publish requests were created, in other words, the set of exist requests
+// and the set of expected requests are the same.
+func (r *Reconciler) reconcileNewPublishRequests(
+	ctx *pkgctx.VirtualMachineGroupPublishRequestContext,
+	existReqsSet, desiredReqsSet sets.Set[string]) error {
+
+	newReqsSet := desiredReqsSet.Difference(existReqsSet)
+	if newReqsSet.Len() == 0 {
+		return nil
+	}
+
+	vmGroupPubReqGvk, err := apiutil.GVKForObject(ctx.VMGroupPublishRequest, r.Scheme())
+	if err != nil {
+		return err
+	}
+
+	vmAPIVersion := vmopv1.GroupVersion.String()
+	clAPIVersion := imgregv1a1.GroupVersion.String()
+	for sourceVM := range newReqsSet {
+		// TTLSecondsAfterFinished is intentionally unset to avoid vm publish request to be
+		// deleted after completion.
+		if err := r.Create(ctx, &vmopv1.VirtualMachinePublishRequest{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:    ctx.VMGroupPublishRequest.Namespace,
+				GenerateName: ctx.VMGroupPublishRequest.Name + "-",
+				OwnerReferences: []metav1.OwnerReference{
+					*metav1.NewControllerRef(ctx.VMGroupPublishRequest, vmGroupPubReqGvk),
+				},
+				Labels: map[string]string{
+					vmopv1.VirtualMachinePublishRequestManagedByLabelKey: ctx.VMGroupPublishRequest.Name,
+				},
+			},
+			Spec: vmopv1.VirtualMachinePublishRequestSpec{
+				Source: vmopv1.VirtualMachinePublishRequestSource{
+					Name:       sourceVM,
+					Kind:       "VirtualMachine",
+					APIVersion: vmAPIVersion,
+				},
+				Target: vmopv1.VirtualMachinePublishRequestTarget{
+					Location: vmopv1.VirtualMachinePublishRequestTargetLocation{
+						Name:       ctx.VMGroupPublishRequest.Spec.Target,
+						Kind:       "ContentLibrary",
+						APIVersion: clAPIVersion,
+					},
+				},
+			},
+		}); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// reconcileStatusCompletedCondition sets vm group publish request's completed condition and completion time
+// once all desired vm publish requests are completed.
+func (r *Reconciler) reconcileStatusCompletedCondition(
+	ctx *pkgctx.VirtualMachineGroupPublishRequestContext,
+	pendingReqsSet, completedReqsSet, desiredReqsSet sets.Set[string]) error {
+
+	if completedReqsSet.Equal(desiredReqsSet) {
+		conditions.MarkTrue(ctx.VMGroupPublishRequest, vmopv1.VirtualMachineGroupPublishRequestConditionComplete)
+		ctx.VMGroupPublishRequest.Status.CompletionTime = metav1.Now()
+		ctx.Logger.Info("VirtualMachineGroupPublishRequest completed successfully")
+		// update to condition will trigger a reconcile to process TTL value
+		return nil
+	}
+
+	conditions.MarkFalse(
+		ctx.VMGroupPublishRequest,
+		vmopv1.VirtualMachinePublishRequestConditionComplete,
+		vmopv1.VirtualMachineGroupPublishRequestConditionReasonPending,
+		"waiting %d more vm publish requests to be completed",
+		pendingReqsSet.Len())
+
+	return nil
+}
+
+// verifySpec ensures spec's source, target, and virtualMachines are set up and validated properly by the webhooks.
+// It sets the completed condition to false with message asking user to re-create the group request.
+func (r *Reconciler) verifySpec(vmGroupPublishReq *vmopv1.VirtualMachineGroupPublishRequest) error {
+	var errs []error
+	if vmGroupPublishReq.Spec.Source == "" {
+		errs = append(errs, fmt.Errorf(undefinedSpecErrorMsg, "source"))
+	}
+	if vmGroupPublishReq.Spec.Target == "" {
+		errs = append(errs, fmt.Errorf(undefinedSpecErrorMsg, "target"))
+	}
+	if len(vmGroupPublishReq.Spec.VirtualMachines) == 0 {
+		errs = append(errs, fmt.Errorf(undefinedSpecErrorMsg, "virtualMachines"))
+	}
+
+	if len(errs) > 0 {
+		// in the rare case that webhooks were down when request was created
+		// set condition to failure
+		conditions.MarkError(vmGroupPublishReq,
+			vmopv1.VirtualMachineGroupPublishRequestConditionComplete,
+			invalidSpecErrorMsg,
+			errors.Join(errs...))
+
+		return pkgerr.NoRequeueError{Message: invalidSpecErrorMsg}
+	}
+
+	return nil
+}
+
+// reconcileSpecTTL deletes the group publish request and its child vm publish requests once the TTL expires.
+func (r *Reconciler) reconcileSpecTTL(ctx *pkgctx.VirtualMachineGroupPublishRequestContext) error {
+	vmGroupPubReq := ctx.VMGroupPublishRequest
+
+	// skip deletions when ttl is nil
+	if vmGroupPubReq.Spec.TTLSecondsAfterFinished == nil {
+		return nil
+	}
+
+	// calculate expiration time based on the completion time and the expected ttl.
+	// when ttl is zero, the expiration completion time is same as completion time which triggers immediate deletion.
+	expirationTime := vmGroupPubReq.Status.CompletionTime.Time.Add(time.Duration(*vmGroupPubReq.Spec.TTLSecondsAfterFinished) * time.Second)
+	if time.Now().Before(expirationTime) {
+		return pkgerr.RequeueError{After: time.Until(expirationTime)}
+	}
+
+	ctx.Logger.Info(fmt.Sprintf(delTTLMsg, "deleting"))
+	if err := r.Delete(ctx, vmGroupPubReq); client.IgnoreNotFound(err) != nil {
+		ctx.Logger.Error(err, fmt.Sprintf(delTTLMsg, "failed to delete"))
+		return err
+	}
+	ctx.Logger.Info(fmt.Sprintf(delTTLMsg, "deleted"))
+	return nil
+}
+
+// deleteVMPublishRequests deletes vm publish requests owned by this group publish request.
+// It is called in ReconcileDelete() but not strictly necessarily since it's owned VirtualMachinePublishRequests have
+// OwnerReference set to it and the k8s GC will clean up them as well if VirtualMachineGroupPublishRequest is deleted.
+// The explicit delete helps speeding up the cleanup process.
+func (r *Reconciler) deleteVMPublishRequests(ctx *pkgctx.VirtualMachineGroupPublishRequestContext) error {
+	reqs, _, _, err := r.getVMPublishRequests(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, req := range reqs {
+		if err := r.Delete(ctx, &req); client.IgnoreNotFound(err) != nil {
+			ctx.Logger.Error(err, "failed to delete vm publish request", "name", req.Name)
+			return err
+		}
+	}
+	return nil
 }
--- a/controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller_unit_test.go
+++ b/controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller_unit_test.go
@@ -1,0 +1,215 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package virtualmachinegrouppublishrequest_test
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinegrouppublishrequest"
+	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
+	"github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	pkgerr "github.com/vmware-tanzu/vm-operator/pkg/errors"
+	"github.com/vmware-tanzu/vm-operator/test/builder"
+)
+
+const (
+	finalizerName = "vmoperator.vmware.com/virtualmachinegrouppublishrequest"
+)
+
+func unitTests() {
+	Describe(
+		"Reconcile",
+		Label(
+			testlabels.Controller,
+			testlabels.API,
+		), unitTestsReconcile,
+	)
+}
+
+func unitTestsReconcile() {
+	var (
+		initObjects       []client.Object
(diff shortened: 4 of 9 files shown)


controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -6,28 +6,41 @@ package virtualmachinegrouppublishrequest
 
 import (
 	"context"
+	goerrors "errors"
```
bryanv: Usually we don't import alias this

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+	}
+
+	createdRequestsMap, pendingRequestsSet, completedRequestsSet, err := r.getVMPublishRequests(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	ctx.VMGroupPublishRequest.Status.Images = r.getSortedImageStatus(createdRequestsMap)
+
+	if err = r.createNewVMPublishRequests(ctx, createdRequestsMap, desiredRequestsSet); err != nil {
```
bryanv: Our usual style for these is to still scope the err var, ie `if err := r..`

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+		retErr = goerrors.Join(retErr, fmt.Errorf(undefinedSpecErrorMsg, "virtualMachines"))
+	}
+
+	return retErr
+}
+
+func (r *Reconciler) reconcileStatus(
+	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (ctrl.Result, error) {
+
+	desiredRequestsSet := sets.Set[string]{}
```
bryanv: This isn't needed until later after something that can fail so move it like L220

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
 	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (_ ctrl.Result, reterr error) {
```
bryanv: While here, there is no defer here so can you remove the named return values?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+		return err
+	}
+
+	var retErr error
+	for toBeCreatedRequest := range toBeCreatedRequestsSet {
+		request := &vmopv1.VirtualMachinePublishRequest{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: ctx.VMGroupPublishRequest.Namespace,
+				Name:      ctx.VMGroupPublishRequest.Spec.Source + "-" + toBeCreatedRequest,
+				OwnerReferences: []metav1.OwnerReference{
```
bryanv, line 292: Use controller-runtime's `SetOwnerReference()` instead (note that you missed setting the UID here)

Also, is there a reason you're not setting this as the controlling owner ref?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+					Kind:       reflect.TypeOf(imgregv1a1.ContentLibrary{}).Name(),
+					APIVersion: vmopv1.GroupVersion.String(),
+				},
+			}
+		}
+		if ctx.VMGroupPublishRequest.Spec.TTLSecondsAfterFinished != nil {
+			request.Spec.TTLSecondsAfterFinished = ctx.VMGroupPublishRequest.Spec.TTLSecondsAfterFinished
+		}
+		if err = r.Create(ctx, request); err != nil {
+			retErr = goerrors.Join(retErr, err)
```
bryanv: For stuff like accumulate the errors into a  `var errs []error` and a final `errors.Join()` for the return value

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+	if err != nil {
+		return err
+	}
+
+	var retErr error
+	for toBeCreatedRequest := range toBeCreatedRequestsSet {
+		request := &vmopv1.VirtualMachinePublishRequest{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: ctx.VMGroupPublishRequest.Namespace,
+				Name:      ctx.VMGroupPublishRequest.Spec.Source + "-" + toBeCreatedRequest,
```
bryanv: Is there an expectation that this name will always be unique?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+	}
+
+	if len(pendingRequestsSet) > 0 {
+		reasons = append(reasons, vmopv1.VirtualMachineGroupPublishRequestConditionReasonPending)
+		messages = append(messages, fmt.Sprintf("waiting vm publish requests for %s to be completed", strings.Join(pendingRequestsSet.UnsortedList(), ", ")))
+	}
+
+	conditions.MarkFalse(vmGroupPublishReq,
+		vmopv1.VirtualMachineGroupPublishRequestConditionComplete,
+		strings.Join(reasons, ", "), "%s", strings.Join(messages, ", "))
```
bryanv: This seems odd to me: I can't think of any other place were the reason is a list

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+			vmopv1.VirtualMachineGroupPublishRequestTTLCondition,
+			fmt.Sprintf("found vm publish requests for %v", slices.Collect(maps.Keys(requests))),
+			ttlErr)
+		return ctrl.Result{}, ttlErr
+	}
+	conditions.Delete(vmGroupPublishReq, vmopv1.VirtualMachineGroupPublishRequestTTLCondition)
+
+	ctx.Logger.Info("deleting vm group publish request", "name: ", vmGroupPublishReq.Name)
+	if err := r.Delete(ctx, vmGroupPublishReq); err != nil {
+		ctx.Logger.Error(err, "failed to delete vm publish request", "name", vmGroupPublishReq.Name)
```
bryanv: Isn't the publish request object name already a part of the logger?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,47 +108,48 @@ func NewReconciler(
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
 
-	obj := &vmopv1.VirtualMachineGroupPublishRequest{}
-	if err := r.apiReader.Get(ctx, req.NamespacedName, obj); err != nil {
+	vmGroupPublishRequest := &vmopv1.VirtualMachineGroupPublishRequest{}
+	if err := r.apiReader.Get(ctx, req.NamespacedName, vmGroupPublishRequest); err != nil {
```
bryanv: Do you know why this is using the API reader client? The publish request controller needed that due to some funky VC API stuff but not clear to me if that applies to this object.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+	}
+
+	if err := r.verifySpec(vmGroupPublishReq); err != nil {
+		// in the rare case that webhooks were down when request was created
+		// set condition to failure
+		conditions.MarkError(vmGroupPublishReq,
+			vmopv1.VirtualMachineGroupPublishRequestConditionComplete,
+			"webhooks failed to mutate/validate spec. please delete and create again.",
+			err)
+		// return nil to stop reconciling
```
bryanv: Please check out our NoRequeueError type, and our ResultFromError() func

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+		conditions.MarkTrue(vmGroupPublishReq, vmopv1.VirtualMachineGroupPublishRequestConditionComplete)
+		vmGroupPublishReq.Status.CompletionTime = metav1.Now()
+		return true
+	}
+
+	missingRequestsSet := desiredRequestsSet.Difference(completedRequestsSet.Union(pendingRequestsSet))
+
+	var reasons, messages []string
+	if len(missingRequestsSet) > 0 {
+		reasons = append(reasons, vmopv1.VirtualMachineGroupPublishRequestConditionReasonMissing)
```
bryanv: I guess this seems odd to be since the very job of this controller is to create these guys, no? In general, create on these guys isn't going to fail, what is a user suppose to do here, why would they care no created vs missing?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -146,12 +160,243 @@ func (r *Reconciler) ReconcileNormal(
...
+
+	if err = r.createNewVMPublishRequests(ctx, createdRequestsMap, desiredRequestsSet); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if completed := r.isCompleted(ctx.VMGroupPublishRequest, pendingRequestsSet, completedRequestsSet, desiredRequestsSet); !completed {
+		return ctrl.Result{RequeueAfter: pkgmgr.DefaultSyncPeriod}, nil
+	}
+
+	return r.removeAfterTTL(ctx)
```
bryanv: Why have this called twice, instead of just letting the next reconcile handle it (since won't isComplete set the condition true?)

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: thanks for the great catch. yes, the name should be unique. I am debating between

- VM group name + VM name + UUID 
- VM group publish name + VM group name + VM name 

Both options should provide uniqueness. I am learning toward the second approach since it is more readable. It is easier to tell the relationship (even though the first option could do the same by looking up the ownerref) with looking up the ownerref. 

What do you think?

zhangheliu: We will also have a webhook to ensure the combined name length meets the requirements and etc

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: It will have a sample out like:
```
Conditions: [{Type:Complete Status:False ObservedGeneration:0 LastTransitionTime:2025-07-23 19:47:09 +0000 UTC Reason:VirtualMachinePublishRequestsMissing, VirtualMachinePublishRequestsPending Message:waiting vm publish requests for dummy-vm1 to be created, waiting vm publish requests for dummy-vm0 to be completed}]
```

It can also be broken down by introducing two new conditions, on which the Completed condition will depend. 
The initial thought process is to have the Completed condition as a simple overview. Users could then look into images[]  condition to get more details about why. 

I could change it to have multiple conditions if this is the preferred pattern we follow.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
bryanv: Use the GenerateName field, probably just with the group request as the prefix. You could include a label for the group request it belongs to and pass that in the later List(). You'd still want to filter by the OwnerRef though on what the List returns.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: good catch, I had two errors pkgs imported earlier and forgot to remove it. thanks!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: The thought was that the LCM of the publish request is managed by the publish request controller, while the group publish request controller only creates and sets TTL in terms of managing the object.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: cool sounds good!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: no and I dont see a reason to use api reader directly. updated.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: thanks for the pointer. updated

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: missingSet contains a set of VMs that have not had a publish request created for it. 
createdSet contains a set of VMs that are created but have not completed yet. 

yeah, like you said, they cared about created v.s. missing is due to the case there is a failure in creating a set of publish request and we need to re-create again. Having the missing set helps provide a quick view for users on the overall status. 

I am not sure about our scale number, but for example,  if we have 500 VMs to do, and the API server is down while we are doing the first 100. We would like to update the group requests condition with the most updated information.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: good catch, fixed. pkgerr comes really handy haha. thanks!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: addressed

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: If we do not need to surface the creation error (ref: https://github.com/vmware-tanzu/vm-operator/pull/1059#discussion_r2226094044) since it is rare, we could also remove it here too.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
bryanv: This controller is creating the underlying publish requests on behalf of the group publish request which is typically when a controlling reference is used.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +110,288 @@ func NewReconciler(
...
+					APIVersion: vmopv1.GroupVersion.String(),
+				},
+			},
+		}
+		if ctx.VMGroupPublishRequest.Spec.Target != "" {
+			request.Spec.Target = vmopv1.VirtualMachinePublishRequestTarget{
+				Location: vmopv1.VirtualMachinePublishRequestTargetLocation{
+					Name:       ctx.VMGroupPublishRequest.Spec.Target,
+					Kind:       reflect.TypeOf(imgregv1a1.ContentLibrary{}).Name(),
+					APIVersion: vmopv1.GroupVersion.String(),
```
bryanv: I don't think this is the correct APIVersion here?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +110,288 @@ func NewReconciler(
...
+	if err := r.List(ctx, requests, client.InNamespace(ctx.VMGroupPublishRequest.Namespace)); err != nil {
+		ctx.Logger.Error(err, "failed to list VirtualMachinePublishRequest")
+		return nil, nil, nil, err
+	}
+
+	createdRequestsMap := make(map[string]vmopv1.VirtualMachinePublishRequest)
+	pendingRequestsSet := sets.Set[string]{}
+	completedRequestsSet := sets.Set[string]{}
+	for _, request := range requests.Items {
+		owned, err := controllerutil.HasOwnerReference(request.OwnerReferences, ctx.VMGroupPublishRequest, r.Scheme())
```
bryanv: Since you switched to a controlling owner ref, can you use `metav1.IsControlledBy()` instead

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +110,288 @@ func NewReconciler(
...
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		if owned {
+			createdRequestsMap[request.Spec.Source.Name] = request
+			if conditions.IsTrue(&request, vmopv1.VirtualMachinePublishRequestConditionComplete) && request.Status.Ready {
```
bryanv: Is there a reason that just checking `request.Status.Ready` isn't good enough?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
bryanv: It isn't that clear to me that info is really that useful to an end user. The missing/pending/completed counts could be reported in the message for the VirtualMachineGroupPublishRequestConditionComplete condition w/ status = false.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +110,288 @@ func NewReconciler(
...
+		vmGroupPublishReq.Status.CompletionTime = metav1.Now()
+		return true
+	}
+
+	missingRequestsSet := desiredRequestsSet.Difference(completedRequestsSet.Union(pendingRequestsSet))
+
+	var reasons, messages []string
+	if len(missingRequestsSet) > 0 {
+		reasons = append(reasons, vmopv1.VirtualMachineGroupPublishRequestConditionReasonMissing)
+		messages = append(messages, fmt.Sprintf("waiting vm publish requests for %s to be created", strings.Join(missingRequestsSet.UnsortedList(), ", ")))
```
bryanv: FYI by using an `UnsortedList()` here and below, most likely the message is going to change on every reconcile regardless if any actual state has changed causing needless churn.

I don't think listing all the missing/pending requests in a condition message is all that useful - the underlying publish request state is better reflected in the group's Status (which I think it is, at least to some extent)

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +110,288 @@ func NewReconciler(
...
+
+	requests, _, _, err := r.getVMPublishRequests(ctx)
+	if err != nil {
+		return err
+	}
+	if len(requests) != 0 {
+		err := fmt.Errorf("failed to delete vm publish requests after TTL")
+		conditions.MarkError(vmGroupPublishReq,
+			vmopv1.VirtualMachineGroupPublishRequestTTLCondition,
+			fmt.Sprintf("found vm publish requests for %v", slices.Collect(maps.Keys(requests))),
```
bryanv: The order of the keys is going to change on every reconcile causing churn. It isn't clear that we really need to include this in the condition anyways. 

Why does this need to wait for the underlying requests to go away anwyays?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +110,288 @@ func NewReconciler(
...
+			request.Spec.Target = vmopv1.VirtualMachinePublishRequestTarget{
+				Location: vmopv1.VirtualMachinePublishRequestTargetLocation{
+					Name:       ctx.VMGroupPublishRequest.Spec.Target,
+					Kind:       reflect.TypeOf(imgregv1a1.ContentLibrary{}).Name(),
+					APIVersion: vmopv1.GroupVersion.String(),
+				},
+			}
+		}
+		if ctx.VMGroupPublishRequest.Spec.TTLSecondsAfterFinished != nil {
+			request.Spec.TTLSecondsAfterFinished = ctx.VMGroupPublishRequest.Spec.TTLSecondsAfterFinished
```
bryanv: This seems odd to me -  isn't the group's TTLSecondsAfterFinished just for the group itself. Each underlying publish is going to complete on its own pace.

Like if a one publish takes a long time, couldn't an already completed request delete itself, I guess causing this controller to create a new one?

test/builder/dummies.go
```diff
@@ -504,6 +513,21 @@ func DummyVirtualMachinePublishRequest(name, namespace, sourceName, itemName, cl
 	}
 }
 
+func VirtualMachineGroupPublishRequest(name, namespace string) *vmopv1.VirtualMachineGroupPublishRequest {
+	return &vmopv1.VirtualMachineGroupPublishRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:       name,
+			Namespace:  namespace,
+			Finalizers: []string{"vmoperator.vmware.com/virtualmachinegrouppublishrequest"},
```
bryanv: I guess it was copied from `DummyVirtualMachinePublishRequest()` above but this is a little odd to include in the finalizer here. Is that actually needed?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: yeah, great catch! I was supposed to set the the TTL only when it is completed (if there is a TTL set). will fix. Thanks!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: yeah, copy&paste error :(

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: no, I just saw they were set at the same time  in the VM publish controller

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: got it. thanks!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: updated

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: updated the logic to have only the counts

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: make sense, updated the logic to only have a high-level count in the condition message.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: fixed

test/builder/dummies.go
zhangheliu: our ReconcileNormal exits right away after adding the finianzer if it is not found. we could add this in the beforeeach function as well.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +108,293 @@ func NewReconciler(
...
+// The key for those objects is the source VirtualMachine name.
+func (r *Reconciler) getVMPublishRequests(
+	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (
+	map[string]vmopv1.VirtualMachinePublishRequest,
+	sets.Set[string],
+	sets.Set[string],
+	error) {
+
+	reqs := &vmopv1.VirtualMachinePublishRequestList{}
+	if err := r.List(ctx, reqs, client.InNamespace(ctx.VMGroupPublishRequest.Namespace)); err != nil {
```
bryanv: Since we're using GenerateName on the underlying publish requests, this will now need to use the API Reader client. Otherwise, there is a chance that a just created publish request hasn't made it back into the client's cache causing this controller to create a duplicate request.

We should probably have a label on the publish requests that we can use it here to filter down the results.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +108,293 @@ func NewReconciler(
...
+	}
+
+	// calculate expiration time based on the completion time and the expected ttl.
+	// when ttl is zero, the expiration completion time is same as completion time which triggers immediate deletion.
+	expirationTime := vmGroupPubReq.Status.CompletionTime.Time.Add(time.Duration(*vmGroupPubReq.Spec.TTLSecondsAfterFinished) * time.Second)
+	if time.Now().Before(expirationTime) {
+		return pkgerr.RequeueError{After: time.Until(expirationTime)}
+	}
+
+	if err := r.deleteVMPublishRequests(ctx); err != nil {
```
bryanv: How about we do this in `ReconcileDelete()` instead? The same cleanup should apply if the group request is deleted before the TTL.

Note though that this clean up isn't strictly necessarily since the publish requests have this group publish request as their owner ref, so the publish requests will get GC.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +108,293 @@ func NewReconciler(
...
+// once all desired vm publish requests are completed.
+func (r *Reconciler) reconcileStatusCompletedCondition(
+	ctx *pkgctx.VirtualMachineGroupPublishRequestContext,
+	pendingReqsSet, completedReqsSet, desiredReqsSet sets.Set[string]) error {
+
+	if completedReqsSet.Equal(desiredReqsSet) {
+		conditions.MarkTrue(ctx.VMGroupPublishRequest, vmopv1.VirtualMachineGroupPublishRequestConditionComplete)
+		ctx.VMGroupPublishRequest.Status.CompletionTime = metav1.Now()
+		// requeue immediately for processing TTL
+		return pkgerr.RequeueError{}
```
bryanv: This should be able to return `nil` here. Setting the condition & completion time will cause an update so an "immediate" reconcile will happen. It might be nice to have a log message here about group completion though.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +108,293 @@ func NewReconciler(
...
+
+	conditions.MarkFalse(
+		ctx.VMGroupPublishRequest,
+		vmopv1.VirtualMachinePublishRequestConditionComplete,
+		vmopv1.VirtualMachineGroupPublishRequestConditionReasonPending,
+		"waiting %d more vm publish requests to be completed",
+		pendingReqsSet.Len())
+
+	// requeue for default sync period while wait for more requests to be completed
+	return pkgerr.RequeueError{After: pkgmgr.DefaultSyncPeriod}
```
bryanv: This is the full resync period so it doesn't make sense to requeue after that period. What this controller should be doing is using EnqueueRequestForOwner - see https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/handler#example-EnqueueRequestForOwner and existing usages in our code.

You should be able to remove the `forceReconcile()` func in the test code after this change

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +108,293 @@ func NewReconciler(
...
+		ctx.Logger.Error(err, "failed to delete vm publish request")
+		return err
+	}
+	return nil
+}
+
+// deleteVMPublishRequests deletes vm publish requests owned by this group publish request.
+func (r *Reconciler) deleteVMPublishRequests(ctx *pkgctx.VirtualMachineGroupPublishRequestContext) error {
+	reqs, _, _, err := r.getVMPublishRequests(ctx)
+	if err != nil || len(reqs) == 0 {
```
bryanv: No need to check for `len(reqs) == 0` since the loop will noop

test/builder/dummies.go
bryanv: OK. These look to be the only dummies that are doing this so it is a little out of place. I'd prefer that we do it in the BeforeEach(), and that's what we're typically doing elsewhere. Also, for like the integration tests it better helps test that the controller really is adding it like we expect.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: good points! yeah, will do

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: good point!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: Totally forgot about that. will do, thanks!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: good catch. should have done it there

zhangheliu: hmm, actually, should we leave it out for now and add the clean up if we actually see any issues in the future?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -56,10 +69,12 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 				controlledType,
 				mgr.GetScheme()),
 		}).
+		Watches(&vmopv1.VirtualMachinePublishRequest{},
+			handler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &vmopv1.VirtualMachineGroupPublishRequest{})).
```
bryanv: I think this should include `handler.OnlyControllerOwner()` since we're using controlling owner refs on these

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +110,308 @@ func NewReconciler(
...
+					*metav1.NewControllerRef(ctx.VMGroupPublishRequest, vmGroupPubReqGvk),
+				},
+				Labels: map[string]string{
+					vmopv1.VirtualMachinePublishRequestManagedByLabelKey: ctx.VMGroupPublishRequest.Name,
+				},
+			},
+			Spec: vmopv1.VirtualMachinePublishRequestSpec{
+				Source: vmopv1.VirtualMachinePublishRequestSource{
+					Name:       sourceVM,
+					Kind:       reflect.TypeOf(vmopv1.VirtualMachine{}).Name(),
```
bryanv: I think our usual convention const/hardcode the Kind string value (we could update the publish request validation webhook to do the same too)

test/builder/dummies.go
zhangheliu: ack, I will create a new story to keep track of fixing existing tests that follow this pattern

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: addressed

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: addressed

api/v1alpha4/virtualmachinegroup_publishrequest_types.go
```diff
@@ -64,13 +68,12 @@ type VirtualMachineGroupPublishRequestSpec struct {
...
+	// completes. After the TTL expires, the resource and its child resources
+	// will be automatically deleted without the user having to take any direct action.
 	//
-	// If this field is unset then the request resource will not be
-	// automatically deleted. If this field is set to zero then the request
-	// resource is eligible for deletion immediately after it finishes.
+	// If this field is unset then the request resource and its child resources
+	// will not be automatically deleted. If this field is set to zero then the request
+	// resource and its child resources are eligible for deletion immediately after it finishes.
 	TTLSecondsAfterFinished *int64 `json:"ttlSecondsAfterFinished,omitempty"`
```
dilyar85: Since you're changing these godocs which will be generated in the CRD file, run `make generate` to get it updated.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -6,60 +6,92 @@ package virtualmachinegrouppublishrequest
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/source"
```
dilyar85, line 26: Nit: extra space here

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
-	if controllerutil.ContainsFinalizer(ctx.Obj, finalizerName) {
-		controllerutil.RemoveFinalizer(ctx.Obj, finalizerName)
+	if controllerutil.ContainsFinalizer(ctx.VMGroupPublishRequest, finalizerName) {
+		controllerutil.RemoveFinalizer(ctx.VMGroupPublishRequest, finalizerName)
 		ctx.Logger.Info("Removed finalizer for VirtualMachineGroupPublishRequest")
+
+		ctx.Logger.Info("Removing owned VirtualMachinePublishRequests")
+		if err := r.deleteVMPublishRequests(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
```
dilyar85, line 154: Is this needed with the owner ref applied to those own resources already? If needed, consider removing the finalizer after all child objects are deleted?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
+	}
+
+	// calculate expiration time based on the completion time and the expected ttl.
+	// when ttl is zero, the expiration completion time is same as completion time which triggers immediate deletion.
+	expirationTime := vmGroupPubReq.Status.CompletionTime.Time.Add(time.Duration(*vmGroupPubReq.Spec.TTLSecondsAfterFinished) * time.Second)
+	if time.Now().Before(expirationTime) {
+		return pkgerr.RequeueError{After: time.Until(expirationTime)}
+	}
+
+	ctx.Logger.Info("deleting vm group publish request")
```
dilyar85: Nit: add "... due to TTL expired" or something similar for better visibility in logs?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
+	if err := r.Delete(ctx, vmGroupPubReq); err != nil {
+		ctx.Logger.Error(err, "failed to delete vm publish request")
```
dilyar85: s/"vm publish request"/"vm group publish request"

Also, ignore not found error if it's already deleted?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
+	for _, req := range existReqsMap {
+		imageStatuses = append(imageStatuses, vmopv1.VirtualMachineGroupPublishRequestImageStatus{
+			Source:             req.Spec.Source.Name,
+			PublishRequestName: req.Name,
+			ImageName:          req.Status.ImageName,
+			Conditions:         req.Status.Conditions,
+		})
+	}
+
+	sort.Slice(imageStatuses, func(i, j int) bool {
```
dilyar85: I think in newer golang version it's recommended to use `slices.Sort/SortFunc()` for better performance.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
+						APIVersion: imgregv1a1.GroupVersion.String(),
+					},
+				},
+			},
+		}); err != nil {
+			return err
+		}
+	}
+
+	return pkgerr.RequeueError{}
```
dilyar85: Is this intentional? Why requeue when the group already watches changes from its owned publish requests?

test/builder/dummies.go
```diff
@@ -504,6 +513,20 @@ func DummyVirtualMachinePublishRequest(name, namespace, sourceName, itemName, cl
...
+func VirtualMachineGroupPublishRequest(name, namespace string) *vmopv1.VirtualMachineGroupPublishRequest {
+	return &vmopv1.VirtualMachineGroupPublishRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: vmopv1.VirtualMachineGroupPublishRequestSpec{
+			Source:          DummyVMGroupName,
+			Target:          DummyContentLibraryName,
+			VirtualMachines: []string{DummyVirtualMachine0Name, DummyVirtualMachine1Name},
```
dilyar85: Consider making this list of VM names as function param for better visibility and flexibility?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
+					*metav1.NewControllerRef(ctx.VMGroupPublishRequest, vmGroupPubReqGvk),
+				},
+				Labels: map[string]string{
+					vmopv1.VirtualMachinePublishRequestManagedByLabelKey: ctx.VMGroupPublishRequest.Name,
+				},
+			},
+			Spec: vmopv1.VirtualMachinePublishRequestSpec{
+				Source: vmopv1.VirtualMachinePublishRequestSource{
+					Name:       sourceVM,
+					Kind:       "VirtualMachinePublishRequest",
```
dilyar85: This should be the `VirtualMachine` kind right?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
+					APIVersion: vmopv1.GroupVersion.String(),
```
dilyar85: Consider initializing this as a var outside loop to avoid calling it for each item.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +127,308 @@ func NewReconciler(
...
 	}
 	ctx.Logger.V(4).Info("Finished Reconciling VirtualMachineGroupPublishRequest Deletion")
 	return ctrl.Result{}, nil
 }
 
-func (r *Reconciler) ReconcileNormal(
-	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (_ ctrl.Result, reterr error) {
-
+func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineGroupPublishRequestContext) error {
 	ctx.Logger.Info("Reconciling VirtualMachinePublishRequest")
```
dilyar85: Not your change but it has the incorrect type in log here.

api/v1alpha4/virtualmachinegroup_publishrequest_types.go
zhangheliu: yeah, I jsut reverted those changes for now. will do those once we work on the webhook changes, since after make generate, it requires to submit the doc chagne first.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: GC should clean them as well. we do it just to make the process faster. Moved it before the finalizer removal.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: yeah, updated. thanks!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: yeah, good catch! thanks!~ it should return nil now

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: great catch. thanks!

test/builder/dummies.go
zhangheliu: yeah sounds good. will do

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller_unit_test.go
```diff
@@ -0,0 +1,214 @@
...
+			vmopv1.VirtualMachinePublishRequestManagedByLabelKey: vmGroupPubReq.Name,
+		})).NotTo(HaveOccurred())
+	Expect(reqs.Items).To(HaveLen(expectedPendingCnt))
+	return reqs.Items
+}
+
+func verifyRequest(req vmopv1.VirtualMachinePublishRequest, vmGroupPub *vmopv1.VirtualMachineGroupPublishRequest) {
+	Expect(req.Name).To(ContainSubstring(vmGroupPub.Name))
+	Expect(metav1.HasLabel(req.ObjectMeta, vmopv1.VirtualMachinePublishRequestManagedByLabelKey)).To(BeTrue())
+	Expect(req.Labels[vmopv1.VirtualMachinePublishRequestManagedByLabelKey]).To(Equal(vmGroupPub.Name))
```
bryanv: Use HaveKeyWithValue matcher - https://onsi.github.io/gomega/#havekeywithvaluekey-any-value-any

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -6,60 +6,93 @@ package virtualmachinegrouppublishrequest
...
+		Reconciler:              r,
+		MaxConcurrentReconciles: ctx.MaxConcurrentReconciles,
+		LogConstructor:          pkgutil.ControllerLogConstructor(controllerNameShort, &vmopv1.VirtualMachineGroupPublishRequest{}, mgr.GetScheme()),
+	})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to VirtualMachineGroupPublishRequest.
+	if err := c.Watch(source.Kind(
```
bryanv: For a controller like this, it is probably best to switch to the builder `ctrl.NewControllerManagedBy()` method and this becomes type becomes just the `For()` argument

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +128,308 @@ func NewReconciler(
...
-func (r *Reconciler) ReconcileNormal(
-	ctx *pkgctx.VirtualMachineGroupPublishRequestContext) (_ ctrl.Result, reterr error) {
-
-	ctx.Logger.Info("Reconciling VirtualMachinePublishRequest")
-	vmGroupPublishReq := ctx.Obj
+func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineGroupPublishRequestContext) error {
+	ctx.Logger.Info("Reconciling VirtualMachineGroupPublishRequest")
+	vmGroupPublishReq := ctx.VMGroupPublishRequest
 
 	if !controllerutil.ContainsFinalizer(vmGroupPublishReq, finalizerName) {
```
bryanv: Can simplify this a bit because `AddFinalizer()` returns a bool if the finalizer was added so the contains can be removed

test/builder/dummies.go
zhangheliu: addressed

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: switched back to NewControllerManagedBy with For() and Owns().

controllers/controllers.go
```diff
@@ -56,6 +57,9 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 	if err := virtualmachinepublishrequest.AddToManager(ctx, mgr); err != nil {
 		return fmt.Errorf("failed to initialize VirtualMachinePublishRequest controller: %w", err)
 	}
+	if err := virtualmachinegrouppublishrequest.AddToManager(ctx, mgr); err != nil {
+		return fmt.Errorf("failed to initialize VirtualMachineGroupPublishRequest controller: %w", err)
+	}
```
dilyar85: Is there a dedicated capability for this feature? If not, we should probably guard it with `pkgcfg.FromContext(ctx).Features.VMGroups` (not enabled yet) to avoid breaking the main.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -95,63 +108,308 @@ func NewReconciler(
...
+	}
+
+	existReqsMap, pendingReqsSet, completedReqsSet, err := r.getVMPublishRequests(ctx)
+	if err != nil {
+		return err
+	}
+
+	r.reconcileStatusImages(ctx, existReqsMap)
+
+	desiredReqsSet := sets.New(ctx.VMGroupPublishRequest.Spec.VirtualMachines...)
```
dilyar85, line 195: The API docs mentioned that "If omitted, this field defaults to the names of all of the VMs currently a member of the group, either directly or indirectly via a nested group." Will there be a follow-up change to address this?

dilyar85 (commented):
Thanks for addressing all previous comments. Left a few more but overall LGTM!

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: Yeah, this alone will spec.target and spec.source will be done in the webhook changes. We have VMSVC-2895 for tracking it.

controllers/controllers.go
zhangheliu: good catch. thanks!

zhangheliu:
> Thanks for addressing all previous comments. Left a few more but overall LGTM!

Thank you and @bryanv  for reviewing!!

PR #1066, opened 2025-07-29
 ✨ Basic Group Placement support

bryanv:

First pass at very basic group placement, and the start of deduping this with the existing VM create workflow but still a lot of work remaining.

Fixes #

```release-note
NONE
```


pkg/providers/vsphere/placement/cluster_placement.go
```diff
@@ -219,39 +219,40 @@ func PlaceVMForCreate(
...
 
-	vmCtx.Logger.V(4).Info("PlaceVmsXCluster request", "placementSpec", vimtypes.ToString(placementSpec))
+	for i, cs := range configSpecs {
+		// Work around PlaceVmsXCluster bug that crashes vpxd when ConfigSpec.Files is nil (still needed?)
+		cs.Files = new(vimtypes.VirtualMachineFileInfo)
+		placementSpec.VmPlacementSpecs[i].ConfigSpec = cs
+	}
 
-	resp, err := object.NewRootFolder(vcClient).PlaceVmsXCluster(vmCtx, placementSpec)
+	logger.V(4).Info("PlaceVmsXCluster request", "spec", vimtypes.ToString(placementSpec))
```
dilyar85: Since this is only once for each VM/VMG deployment, can we log it by default? It would help debug placement related bugs a lot without having to increase the log level and restart VMOP pod every time.

pkg/providers/vsphere/placement/group_placement.go
```diff
@@ -0,0 +1,92 @@
...
+	vcClient *vim25.Client,
+	finder *find.Finder,
+	candidates map[string][]string,
+	configSpecs []vimtypes.VirtualMachineConfigSpec) (map[string][]Recommendation, error) {
+
+	var candidateRPMoRefs []vimtypes.ManagedObjectReference
+
+	for _, rpMoIDs := range candidates {
+		for _, rpMoID := range rpMoIDs {
+			rpMoRef := vimtypes.ManagedObjectReference{Type: "ResourcePool", Value: rpMoID}
```
dilyar85: Nit: s/`"ResourcePool"`/`string(vimtypes.ManagedObjectTypesResourcePool)`

pkg/providers/vsphere/placement/group_placement.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	recommendations, err := getGroupPlacementRecommendations(ctx, vcClient, finder, candidates, configSpecs)
+	if err != nil {
+		return nil, err
+	}
+
+	results := map[string]Result{}
+	for vmName, vmRecommendations := range recommendations {
+		selectedRecommendation := vmRecommendations[rand.Intn(len(vmRecommendations))] // nolint:gosec
```
dilyar85, line 44: Could `vmRecommendations` be empty?

pkg/providers/vsphere/vmprovider_vm_group.go
```diff
@@ -0,0 +1,411 @@
...
+
+			idx := findVMMemberStatus(vm.Name, vmGroup.Status.Members)
+			if idx < 0 {
+				m := vmopv1.VirtualMachineGroupMemberStatus{
+					Name: vm.Name,
+					Kind: "VirtualMachine",
+				}
+				vmGroup.Status.Members = append(vmGroup.Status.Members, m)
+				idx = len(vmGroup.Status.Members) - 1
+			}
```
dilyar85, line 203: I wonder if we should error out if member status cannot be found. This likely means the member was deleted or not linked to the group yet, so probably no need to update the placement status?

pkg/providers/vsphere/vmprovider_vm_group.go
```diff
@@ -0,0 +1,411 @@
...
+	vmName string,
+	result *placement.Result) *vmopv1.VirtualMachinePlacementStatus {
+
+	placementStatus := &vmopv1.VirtualMachinePlacementStatus{}
+	placementStatus.Name = vmName
+	placementStatus.Zone = result.ZoneName
+	placementStatus.Pool = result.PoolMoRef.Value
+
+	if result.HostMoRef != nil {
+		placementStatus.Name = result.HostMoRef.Value
```
dilyar85: Should be `.Node`

pkg/providers/vsphere/placement/group_placement.go
bryanv: No an entry in the map is only added because of a placement result

pkg/providers/vsphere/vmprovider_vm_group.go
bryanv: The VM was a member at the point in time when we got the group in the controller. I'd rather not assume anything else about the status at this point. A later group reconcile can clean up the status (IDK if the code for that already exists)

dilyar85: Sounds good. The VMG controller does clean up the status if the member is removed from a group. I was more concerned when the member existed & wasn't linked anymore, but I guess it's no harm leaving the placement status there.

dilyar85 (approved):

pkg/providers/vsphere/placement/cluster_placement.go
bryanv: I'll log this by default in this MR but it might be too spammy to leave enabled long term.

pkg/providers/vsphere/vmprovider_vm_group.go
```diff
@@ -0,0 +1,251 @@
...
+		return err
+	}
+
+	placementArgs, err := vs.vmGroupGetVMPlacementArgs(ctx, vcClient, groupPlacements)
+	if err != nil {
+		return err
+	}
+
+	if len(placementArgs.configSpecs) == 0 {
+		return nil
```
dilyar85, line 47: Would be nice to leave a log here for better debugging in future (can be a follow-up)

pkg/providers/vsphere/vmprovider_vm_group.go
```diff
@@ -0,0 +1,251 @@
...
+			createArgs, err := vs.vmGroupGetVMCreatePrereqs(vmCtx, vcClient)
+			if err != nil {
+				return nil, err
+			}
+
+			if firstVM {
+				placementArgs.childResourcePoolName = createArgs.ChildResourcePoolName
+				firstVM = false
+			} else if placementArgs.childResourcePoolName != createArgs.ChildResourcePoolName {
+				return nil, fmt.Errorf("all VMs being placed as group must belong to same child ResourcePool")
```
dilyar85, line 92: Include both RP names in the error for improved visibility? (can be a follow-up)

dilyar85 (approved):
LGTM, thanks @bryanv!

PR #1067, opened 2025-07-29
✨Report reserved and used capacity of VMSnapshot in its StoragePolicyUsage

lubronzhan:

### Change in VMSnapshot API and controller
A new type was added in status VirtualMachineSnapshotStorageStatus, and under it, there are two new fields: `requested` and `used`
```
    // +optional

	// Storage describes the observed amount of storage used by a
	// VirtualMachineSnapshot, including the space for FCDs.
	Storage *VirtualMachineSnapshotStorageStatus `json:"storage,omitempty"`



// VirtualMachineSnapshotStorageStatus defines the observed state of a
// VirtualMachineSnapshot's storage.
type VirtualMachineSnapshotStorageStatus struct {
	// +optional

	// Used describes the observed amount of storage used by a
	// VirtualMachineSnapshot, except the space for FCDs.
	Used *resource.Quantity `json:"used,omitempty"`

	// +optional

	// Requested describes the observed amount of storage requested by a
	// VirtualMachineSnapshot. It's a list of requested storage for each
	// storage class.
	// Since a snapshot can have multiple PVCs, it can point to multiple storage
	// classes.
	Requested []VirtualMachineSnapshotStorageStatusRequested `json:"requested,omitempty"`
}

// VirtualMachineSnapshotStorageStatusRequested describes the observed amount of
// storage requested by a VirtualMachineSnapshot for a storage class.
type VirtualMachineSnapshotStorageStatusRequested struct {
	// StorageClass is the name of the storage class.
	StorageClass string `json:"storageClass"`

	// Total describes the total storage space requested by a
	// VirtualMachineSnapshot for the storage class.
	Total *resource.Quantity `json:"total"`
}
```

The algorithm for calculating Used and Requested:
**Requested**:
1. Total requested for each storageclass = VM's Classic Disk size + VM Memory + FCD Disk size.
     1. VM's Classic Disk size is fetched from VM's status volume.
     1. VM Memory is fetched from VMClass.
     1. FCD Disk size is fetched from PVC, or if it's an instance storage volume, directly fetch from VM spec.

**Used**:
1. Calls GetSnapshotSize, it use the same algorithm that vSphere UI used. Extract files for vmSnapshot, excludes the FCD related files

**To notify StoragePolicyUsage controller**
Call `SyncStorageUsageForNamespace` to inject events to the context channel

### Change in StoragePolicyUsage controller
Now it reconciles a StoragePolicyUsage just for VMSnapshot. 
1. When VMSnapshot doesn't have annotation key `csi.vsphere.volume.sync` or it's set to `requested`, assume we need to calculate reserved capacity and include it in SPU. Get the reserved capacity from `VMSnapshot.status.requested`, filtered by the StorageClassName.
1. When VMSnapshot doesn't have annotation `csi.vsphere.volume.sync: completed`, assume the CSI already updated SPU for volumes, we need to calculate vmSnapshot's used capacity and include it in SPU. Get the used capacity for that VM from `VMSnapshot.status.used`.

When a VMSnapshot's **used capacity** is reported in SPU, the logic in SPU controller will automatically ignore its **requested capacity** when calculating SPU's **requested capacity**. The same vice versa. 

### VMProvider change
1. Add GetSnapshotSize func. This use the same algorithm that VC UI used to calculate the VMSnapshot size.
      1. The VMSnapshot.status.used include the size of memory file, the vmdk files, and the vmsn file. Exclude FCD's size. 
      3. **Issue**: the func returns a decimal value, while VC UI gives a xx GB approximate value instead. Should I round the number and return the same as VC UI as well?
1. GetParentSnapshot func
      1. Previous logic is to recursively calling apiserver, now it's refactored to a simple VC call

Part of VM's status
```
    storage:
      requested:
        disks: 10Gi
      total: "19419076776"
      usage:
        disks: "2078279139"
        other: "8681658536"
    uniqueID: vm-1059
    volumes:
    - attached: true
      diskUUID: 6000C292-3c7a-7cbc-c945-0fb35c5a494b
      limit: 10Gi
      name: lubron-vm
      requested: 10Gi
      type: Classic
      used: "2078279139"
    - attached: true
      diskUUID: 6000C299-85a3-b188-e22e-3b6c3c3993d6
      limit: 5Gi
      name: lubron-vol
      requested: 5Gi
      type: Managed
      used: "634390801"
    zone: domain-c54
kind: List
```

Example VMSnapshot
```
k get vmsnapshot -n lubron-test   sn-1 -oyaml
apiVersion: vmoperator.vmware.com/v1alpha4
kind: VirtualMachineSnapshot
metadata:
  annotations:
      {"apiVersion":"vmoperator.vmware.com/v1alpha4","kind":"VirtualMachineSnapshot","metadata":{"annotations":{},"name":"sn-1","namespace":"lubron-test"},"spec":{"vmRef":{"apiVersion":"vmoperator.vmware.com/v1alpha4","kind":"VirtualMachine","name":"lubron-vm"}}}
  creationTimestamp: "2025-08-06T21:14:00Z"
  finalizers:
  - vmoperator.vmware.com/virtualmachinesnapshot
  generation: 1
  name: sn-1
  namespace: lubron-test
  ownerReferences:
  - apiVersion: vmoperator.vmware.com/v1alpha4
    kind: VirtualMachine
    name: lubron-vm
    uid: a086e2f8-b151-422c-b5c0-f4f4a710aaad
  resourceVersion: "6374015"
  uid: 4de85d26-47db-4854-97e7-be5d18c432b8
spec:
  vmRef:
    apiVersion: vmoperator.vmware.com/v1alpha4
    kind: VirtualMachine
    name: lubron-vm
status:
  conditions:
  - lastTransitionTime: "2025-08-06T21:13:59Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineSnapshotReady
  storage:
    requested:
    - storageClass: wcpglobal-storage-profile
      total: 15Gi
    used: "2010145972"
  uniqueID: snapshot-1060
```
The sum of used is different from the UI, since it excludes the FCD's size
image

Snapshot 2:
```
root@423d3bde73db981f7ca7ee7b1f342d8a [ ~ ]# k get vmsnapshot -n lubron-test   sn-2 -oyaml
apiVersion: vmoperator.vmware.com/v1alpha4
kind: VirtualMachineSnapshot
metadata:
  annotations:
      {"apiVersion":"vmoperator.vmware.com/v1alpha4","kind":"VirtualMachineSnapshot","metadata":{"annotations":{},"name":"sn-2","namespace":"lubron-test"},"spec":{"vmRef":{"apiVersion":"vmoperator.vmware.com/v1alpha4","kind":"VirtualMachine","name":"lubron-vm"}}}
  creationTimestamp: "2025-08-06T21:30:29Z"
  finalizers:
  - vmoperator.vmware.com/virtualmachinesnapshot
  generation: 1
  name: sn-2
  namespace: lubron-test
  ownerReferences:
  - apiVersion: vmoperator.vmware.com/v1alpha4
    kind: VirtualMachine
    name: lubron-vm
    uid: a086e2f8-b151-422c-b5c0-f4f4a710aaad
  resourceVersion: "6374019"
  uid: 6a22bb29-d469-4a40-80f0-b1a765ec5eec
spec:
  vmRef:
    apiVersion: vmoperator.vmware.com/v1alpha4
    kind: VirtualMachine
    name: lubron-vm
status:
  conditions:
  - lastTransitionTime: "2025-08-06T21:30:28Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineSnapshotReady
  storage:
    requested:
    - storageClass: wcpglobal-storage-profile
      total: 15Gi
    used: "24142813"
  uniqueID: snapshot-1061
```

Example SPU 
```
root@423d5c6e6d062e746bdc51479b5a8f5a [ ~ ]# k get storagepolicyusage -n lubron-test wcpglobal-storage-profile-vmsnapshot-usage -oyaml
apiVersion: cns.vmware.com/v1alpha2
kind: StoragePolicyUsage
metadata:
  creationTimestamp: "2025-07-31T18:26:50Z"
  generation: 1
  name: wcpglobal-storage-profile-vmsnapshot-usage
  namespace: lubron-test
  ownerReferences:
  - apiVersion: cns.vmware.com/v1alpha2
    blockOwnerDeletion: true
    controller: true
    kind: StoragePolicyQuota
    name: wcpglobal-storage-profile-storagepolicyquota
    uid: 2b5e16f2-5d6e-43c0-8a1c-490c81a1533d
  resourceVersion: "6363809"
  uid: b21dc63f-74b1-4f72-b73f-5e0fbcb03033
spec:
  caBundle: LS0t...Cg==
  resourceApiGroup: vmoperator.vmware.com
  resourceExtensionName: vmware-system-vmop-webhook-service
  resourceExtensionNamespace: vmware-system-vmop
  resourceKind: VirtualMachineSnapshot
  storageClassName: wcpglobal-storage-profile
  storagePolicyId: 7b6f2012-1132-43e9-aef4-b3a223c29818
status:
  quotaUsage:
    reserved: 30Gi
    used: "0"
```

After annotate the VMSnapshot 1 with `csi.vsphere.volume.sync: completed`, the updated SPU:
```
root@423d3bde73db981f7ca7ee7b1f342d8a [ ~ ]# k get storagepolicyusage -n lubron-test wcpglobal-storage-profile-vmsnapshot-usage -oyaml
apiVersion: cns.vmware.com/v1alpha2
kind: StoragePolicyUsage
metadata:
  creationTimestamp: "2025-07-31T18:26:50Z"
  generation: 1
  name: wcpglobal-storage-profile-vmsnapshot-usage
  namespace: lubron-test
  ownerReferences:
  - apiVersion: cns.vmware.com/v1alpha2
    blockOwnerDeletion: true
    controller: true
    kind: StoragePolicyQuota
    name: wcpglobal-storage-profile-storagepolicyquota
    uid: 2b5e16f2-5d6e-43c0-8a1c-490c81a1533d
  resourceVersion: "6411500"
  uid: b21dc63f-74b1-4f72-b73f-5e0fbcb03033
spec:
  caBundle: LS0t...==
  resourceApiGroup: vmoperator.vmware.com
  resourceExtensionName: vmware-system-vmop-webhook-service
  resourceExtensionNamespace: vmware-system-vmop
  resourceKind: VirtualMachineSnapshot
  storageClassName: wcpglobal-storage-profile
  storagePolicyId: 7b6f2012-1132-43e9-aef4-b3a223c29818
status:
  quotaUsage:
    reserved: 15Gi
    used: "2010145972"
```

## Release note
```release-note
Report reserved and used capacity of VMSnapshot in its StoragePolicyUsage
```


lubronzhan:
Found an issue with logic when calculating the used capacity. Let me fix it first

lubronzhan:
> Found an issue with logic when calculating the used capacity. Let me fix it first

Ok the logic was fixed

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -99,6 +100,19 @@ type VirtualMachineSnapshotStatus struct {
...
+	// Used describes the observed amount of storage used by a
+	// VirtualMachineSnapshot.
+	Used *VirtualMachineSnapshotStorageStatusUsed `json:"used,omitempty"`
+}
+
+// VirtualMachineSnapshotStorageStatusUsed describes the observed amount of
+// storage used by a VirtualMachineSnapshot.
+type VirtualMachineSnapshotStorageStatusUsed struct {
+	// Total describes the total storage space used by a VirtualMachineSnapshot.
+	Total *resource.Quantity `json:"total,omitempty"`
```
aruneshpa: Should we clarify here that this does not include the PVC disks?

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
```diff
@@ -107,21 +112,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
 		ctx,
 		req.Namespace,
-		spqutil.StoragePolicyUsageNameForVM(req.Name))
+		req.Name)
 }
 
 func (r *Reconciler) ReconcileNormal(
 	ctx context.Context,
-	namespace, storagePolicyUsageName string) error {
+	scNamespace string,
```
aruneshpa:
```suggestion
	namespace string,
```

Since SCs are cluster scoped and this is just namespace for the resource that is being reconciled.

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
```diff
@@ -107,21 +112,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+	scName string) error {
+
+	errs := []error{}
+	if err := r.ReconcileSPUForVM(ctx, scNamespace, scName); err != nil {
+		errs = append(errs, err)
+	}
+	if err := r.ReconcileSPUForVMSnapshot(ctx, scNamespace, scName); err != nil {
```
aruneshpa: Out of curiosity, why is this in the virtualmachine pacakge?  I would expect this (and all relevant snapshot methods) to be in snapshots package.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
 
 	return caBundle, nil
 }
+
+// CalculateReservedForSnapshot calculates the reserved capacity for a snapshot.
+// 1. Space to be reserved for classic disks (equal to the disk size – to conform to how VolumeSnapshots work).
```
aruneshpa: Nit: Not just VolumeSnapshots, all reservation works like that.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+// 2. Additional space to be reserved if it is memory snapshot.
```
aruneshpa: I would clearly call out that we reserve the space that is equal to the memory of the VM.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+		switch volume.Type {
+		case vmopv1.VirtualMachineStorageDiskTypeClassic:
+			// For classic disk, check if the storage class matches the SPU's storage class.
+			if vm.Spec.StorageClass != storageClass {
+				// Skip the disk if the storage class doesn't match the SPU's storage class
+				logger.V(5).Info("skipping disk because storage class doesn't match the SPU's storage class", "vm", vm.NamespacedName(), "storageClass", vm.Spec.StorageClass, "spuStorageClass", storageClass)
+				continue
+			}
+			// Add the disk size to the total reserved capacity
+			if volume.Limit == nil {
```
aruneshpa: Hmm.  This should be volume.requested no?

```
	// +optional

	// Requested describes the minimum, requested capacity of the volume.
	//
	// Please note, this value is used when calculating a VM's impact to a
	// namespace's storage quota.
	Requested *resource.Quantity `json:"requested,omitempty"`

```
The limit of a volume is what is specified by the user as limit.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+	}
+
+	vm := &vmopv1.VirtualMachine{}
+	if err := k8sClient.Get(ctx, client.ObjectKey{Namespace: vmSnapshot.Namespace, Name: vmSnapshot.Spec.VMRef.Name}, vm); err != nil {
+		return fmt.Errorf("failed to get VM %s: %w", vmSnapshot.Spec.VMRef.Name, err)
+	}
+
+	// Add the VM's memory to the total reserved capacity if it's powered on
+	// TODO (lubron): What if the VM is suspended?
+	if vm.Spec.StorageClass == storageClass && vmSnapshot.Spec.Memory && vm.Status.PowerState == vmopv1.VirtualMachinePowerStateOn {
```
aruneshpa: This shouldn't be the VM's power state.  It is possible that a snapshot was taken as powered off, but then the VM was turned on later -- and the controller is still processing the previous snapshot.

I would recommend just reserving the VM's memory if spec.memory = true for the snapshot.  And file an issue so we can rely on VM snapshot's status.memory for this (since spec.memory may be ignored if the snapshot's power state is off).  The other option would be to look at the snapshot details on VC here -- but that is the last resort since that's time consuming.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+		// Get the vmClass
```
aruneshpa: Nit: Instead of this comment, maybe describe how you are finding out the memory requested by the VM (using the class)?

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+			// Add the disk size to the total reserved capacity
+			if volume.Limit == nil {
+				return fmt.Errorf("limit is not set for classic disk")
+			}
+			total.Add(*volume.Limit)
+		case vmopv1.VirtualMachineStorageDiskTypeManaged:
+			// For FCD, find the pvc of the volume, then check if pvc's storage class
+			// matches the SPU's storage class.
+			// If it doesn't match, skip the disk.
+			// If it matches, add the PVC's limit to the total reserved capacity.
```
aruneshpa: Not the limit, but the size.

And now that I think about it, I don't think we can rely on any "live" state of the VM or any resources because we may be processing a snapshot that was taking quite a while ago.

IOW, we will have to expose all this information in VMSnapshot's status so we can use it here and elsewhere.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+			//
+			// TODO (lubron): the VirtualMachineVolumeSource also contains the InstanceVolumeClaim,
+			// which has a StorageClass.
+			// check in which case we should use it.
```
aruneshpa: Can you file a TODO for this?  This _should_ be fine since underneath, these are just PVCs so the reservation and usage will go through CSI.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+				logger.V(5).Info("skipping disk because corresponding pvc is not found in the vm.spec.volumes", "volume", volume.Name)
+				continue
+			}
+
+			pvc := &corev1.PersistentVolumeClaim{}
+			if err := k8sClient.Get(ctx, client.ObjectKey{Namespace: vmSnapshot.Namespace, Name: pvcName}, pvc); err != nil {
+				return fmt.Errorf("failed to get pvc %s for volume %s: %w", pvcName, volume.Name, err)
+			}
+
+			limit := volume.Limit
```
aruneshpa: Same comment here.  Just use the PVC capacity.

pkg/util/kube/spq/spq.go
```diff
@@ -268,3 +277,103 @@ func GetWebhookCABundle(ctx context.Context, k8sClient client.Client) ([]byte, e
...
+			// If the limit is not set on volume, then use the PVC's limit
+			if limit == nil {
+				if pvc.Spec.Resources.Limits == nil {
+					return fmt.Errorf("limit is not set for managed disk")
+				}
+				limit = pvc.Spec.Resources.Limits.Storage()
+			}
+
+			// Fetch the storage class for the PVC
```
aruneshpa: Hmm.  Why not move this right after you fetch the PVC so we return / skip as early as possible?

pkg/util/kube/spq/spq.go
lubronzhan: I saw your comment here https://chat.google.com/room/AAQAF3ta4NY/XfCPLXh4ShY/XfCPLXh4ShY?cls=10 also found this code https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachine/volume/volume_controller.go#L1026-L1041

Should we calculate the reserved space like above code? Do you mean we should just skip the calculation it in reservation? I thought we want to include everything when calculate reservation

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
lubronzhan: Do you mean getSnapshotSize function?

pkg/util/kube/spq/spq.go
lubronzhan:
```

	// +optional

	// Limit describes the maximum, requested capacity of the volume.
	Limit *resource.Quantity `json:"limit,omitempty"`
```
IC, I thought we want to consider the worst case, like the upper limit of the disk. Let me change this

pkg/util/kube/spq/spq.go
lubronzhan: Do you mean we should calculate reserved capacity **only once at the beginning**? Make sense, it's not like usage that will change overtime

lubronzhan: But then the problem is, a VM could have PVCs with different storageClass, we need to store those requested capacity in VMSnapshot status by each storageClass. Then when calculating the SPU reserved, we know which part of the storage belongs to this SPU

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -172,3 +176,163 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+// Here we use the file name to identify the FCDs because the backingObjectID
+// is not always available. and there is no other uuid available.
+// Example could be found here https://vmw-jira.broadcom.net/browse/VMSVC-2700?focusedId=20244583&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20244583
+func getFCDDevicesFileKeySet(vmCtx pkgctx.VirtualMachineContext, vmDevices []types.BaseVirtualDevice, vmlayoutFileInfos []types.VirtualMachineFileLayoutExFileInfo) sets.Set[int] {
+	fcdDevicesBackingFileNamePrefixSet := sets.Set[string]{}
+	for _, device := range vmDevices {
+		vd, ok := device.(*types.VirtualDisk)
+		if !ok {
+			continue
+		} else if vd.VDiskId != nil && vd.VDiskId.Id != "" { // FCDs have a VDiskId
```
palnabarun: Can save on this indentation. `if !ok || vd.DiskId == nil || vd.DiskId.Id == "" { continue }`

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -191,6 +191,21 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
+
+	ctx.Logger.V(4).Info("Updating snapshot's usage")
+	size, err := r.VMProvider.GetSnapshotSize(ctx, vmSnapshot.Name, vm)
+	if err != nil {
+		return fmt.Errorf("failed to get snapshot size: %w", err)
+	}
+	if vmSnapshot.Status.Used == nil {
+		vmSnapshot.Status.Used = &vmopv1.VirtualMachineSnapshotStorageStatusUsed{}
+	}
+	// TODO(lubron): Now it's default to decimal SI since the size is int64.
```
palnabarun: "decimal SI" -> typo? defaulting to binarySI below.

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
lubronzhan: Got it, just understood how this vm watcher architecture work. I also added the SyncStorageUsageForNamespace in vmsnapshot controller to enqueue events.

pkg/util/kube/spq/spq.go
lubronzhan: Make sense, worst case estimation

pkg/util/kube/spq/spq.go
lubronzhan: Ok adjust the API accordingly

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Actually even though I specify binarySI, it will show as decimal value by default in the CRD. Like the example in the PR description `4943010903`, since the number is very likely not divided by 1024.

Let me make this clearer

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
lubronzhan: Ok I think I wasn't using it correctly
Reading this PR now https://github.com/vmware-tanzu/vm-operator/pull/733

lubronzhan: Let me remove anything related to `SyncStorageUsageForNamespace ` for now in this PR and add it later

lubronzhan: Added VMSVC-2923

lubronzhan:
Need to rebase

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
lubronzhan: NVM, I made it work and verified in e2e

pkg/util/kube/vmsnapshot.go
```diff
@@ -0,0 +1,136 @@
...
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+)
+
+// CalculateReservedForSnapshotPerStorageClass calculates the reserved capacity for a snapshot.
+// And return the requested capacity categorized by each storage class.
+// 1. Space required for classic disks.
+// 2. Space required for VM's memory.
+// 3. Space required for each PVC.
+func CalculateReservedForSnapshotPerStorageClass(
```
lubronzhan, line 25: I'm not putting it in vsphere/virtualmachine/snapshot.go because this is pure k8s api related code, not calling vsphere API

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
```diff
@@ -184,20 +205,95 @@ func (r *Reconciler) ReconcileNormal(
...
+	// Can have disks with different storage classes.
+	vmSnapshotList := &vmopv1.VirtualMachineSnapshotList{}
+	if err := r.Client.List(ctx, vmSnapshotList, client.InNamespace(namespace)); err != nil {
+		return fmt.Errorf(
+			"failed to list VirtualMachineSnapshots in namespace %s: %w", namespace, err)
+	}
+
+	for _, vmSnapshot := range vmSnapshotList.Items {
+		// Update the usage only when the CSI has marked the VMSnapshot with volume sync completed.
+		if vmSnapshot.Annotations != nil && vmSnapshot.Annotations[constants.CSIVSphereVolumeSyncAnnotationKey] == constants.CSIVSphereVolumeSyncAnnotationValueCompleted {
```
bryanv: No need to check for `nil`

controllers/virtualmachine/storagepolicyusage/storagepolicyusage_controller.go
```diff
@@ -272,3 +368,64 @@ func reportReserved(
...
+	vmSnapshot vmopv1.VirtualMachineSnapshot,
+	total *resource.Quantity,
+	storageClass string) error {
+
+	if vmSnapshot.Spec.VMRef == nil {
+		return fmt.Errorf("vmRef is not set")
+	}
+
+	vm := &vmopv1.VirtualMachine{}
+	if err := r.Client.Get(r.Context, client.ObjectKey{Namespace: vmSnapshot.Namespace, Name: vmSnapshot.Spec.VMRef.Name}, vm); err != nil {
```
bryanv: We've already done a List of the VMs. Can we refactor this to use that. Also, this going repeating fetch the same VM if it has multiple snapshots

The funky and unfortunate decision to try to push quota managed in k8s means that this reconcile is called all the time so it needs to be very mindful

lubronzhan: >We've already done a List of the VMs. Can we refactor this to use that. Also, this going repeating fetch the same VM if it has multiple snapshots

Yep it's doable. There is still the case that VM has different storageclass but some of its PVC has the target storageclass, then in that case we still need to fetch the VM.

>The funky and unfortunate decision to try to push quota managed in k8s means that this reconcile is called all the time so it needs to be very mindful

Just curious, how would you do it differently. rely on the external system manager quota and reports failure of creating resource then just reflect that in CR, so that nothing will need to query all resource in a single namespace at a single time?

bryanv: > There is still the case that VM has different storageclass but some of its PVC has the target storageclass, then in that case we still need to fetch the VM.

I'm confused if we've done a list of the VMs why is a get of the VM needed for PVCs?

> Just curious, how would you do it differently. rely on the external system manager quota and reports failure of creating resource then just reflect that in CR, so that nothing will need to query all resource in a single namespace at a single time?

Instead of having us do lazy updating and enforcement via webhooks VC should manage this.

lubronzhan: >I'm confused if we've done a list of the VMs why is a get of the VM needed for PVCs?

Oh wait, PVC's disk usage is only considered when calculating SPU Reservation, not in SPU Usage. Ok then nvm.  

I thought I need to find PVC with same storageClass as SPU, while this PVC is attached to a VM that's pointing to a different StorageClass, since these kind of PVC's usage should also be included in SPU. I've already had those information for reservation in VMSnapshot's status, which was calculated in VMSnapshot controller. So that I don't have to do it inside SPU controller

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -103,6 +104,32 @@ type VirtualMachineSnapshotStatus struct {
 
 	// Conditions describes the observed conditions of the VirtualMachine.
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// +optional
+
+	// Used describes the observed amount of storage used by a
+	// VirtualMachineSnapshot, except the space for FCDs.
+	Used *resource.Quantity `json:"used,omitempty"`
```
aruneshpa, line 122: Like the VM types, can you move it under a `Storage` subtype?
https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha4/virtualmachine_types.go#L1235

`vmsnapshot.stauts.used` doesn't really signify what we are talking about.

api/v1alpha4/virtualmachinesnapshot_types.go
```diff
@@ -103,6 +104,32 @@ type VirtualMachineSnapshotStatus struct {
...
+
+	// +optional
+
+	// Requested describes the observed amount of storage requested by a
+	// VirtualMachineSnapshot. It's a list of requested storage for each
+	// storage class.
+	// Since a snapshot can have multiple PVCs, it can point to multiple storage
+	// classes.
+	Requested []VirtualMachineSnapshotStorageStatusRequested `json:"requested,omitempty"`
```
aruneshpa, line 131: Same comment as above for this field.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -171,6 +187,44 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
 		return ctrl.Result{}, errors.New("VM hasn't been created and has no uniqueID")
 	}
 
+	// Calculate the requested capacity of the snapshot at the beginning only once.
+	if vmSnapshot.Status.Requested == nil {
+		ctx.Logger.V(4).Info("Updating snapshot's status requested capacity")
+		vmSnapshot.Status.Requested = []vmopv1.VirtualMachineSnapshotStorageStatusRequested{}
+		requested, err := kubeutil.CalculateReservedForSnapshotPerStorageClass(ctx, r.Client, r.Logger, *vmSnapshot)
```
aruneshpa, line 198: My understanding was that this was only supposed to happen when the snapshot was not ready.  When it because ready, we would unset the reservation and increase the usage.   Can you confirm this please?

pkg/util/kube/vmsnapshot.go
```diff
@@ -0,0 +1,136 @@
...
+				continue
+			}
+
+			pvc := &corev1.PersistentVolumeClaim{}
+			if err := k8sClient.Get(ctx, client.ObjectKey{Namespace: vmSnapshot.Namespace, Name: pvcSpec.ClaimName}, pvc); err != nil {
+				return nil, fmt.Errorf("failed to get pvc %s for volume %s: %w", pvcSpec.ClaimName, volume.Name, err)
+			}
+
+			volumeRequested := volume.Requested
+			// If the requested is not set on volume, then use the PVC.spec.resources.requests.
```
aruneshpa: Why not rely on the requests directly without the fallback?
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L515

resources:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2586-L2588

Example:
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes

pkg/util/kube/vmsnapshot.go
```diff
@@ -0,0 +1,136 @@
...
+			if volumeRequested == nil {
+				if pvc.Spec.Resources.Requests == nil {
+					return nil, fmt.Errorf("requested is not set for managed disk")
+				}
+				logger.V(5).Info("using PVC.spec.resources.requests for disk", "pvc", pvc.Name, "storageClass", pvc.Spec.StorageClassName)
+				volumeRequested = pvc.Spec.Resources.Requests.Storage()
+			}
+
+			if pvc.Spec.StorageClassName == nil {
```
aruneshpa, line 112: Can we check this as early as possible and return?

pkg/util/kube/vmsnapshot.go
```diff
@@ -0,0 +1,136 @@
...
+				return nil, fmt.Errorf("storage class is not set for pvc %s", pvc.Name)
+			}
+
+			if _, ok := requestedMap[*pvc.Spec.StorageClassName]; !ok {
```
aruneshpa, line 120: This block should be moved up as well.  Basically, you should fetch the PVC spec and make sure that it has SC set and that SC is something that you are interested in.  We are already doing really compute intensive things here (fetching all VMs, all PVCs, classes, and SCs, so every CPU cycle counts).

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -172,3 +176,163 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+	}
+	vmlayout := vmCtx.MoVM.LayoutEx
+	vmDevices := vmCtx.MoVM.Config.Hardware.Device
+
+	var fileKeyList []int
+
+	// Find the VirtualMachineFileLayoutExSnapshotLayout for current snapshot
+	for _, snapshot := range vmlayout.Snapshot {
+		if snapshot.Key.Value == vmSnapshot.Value {
+			// Add the memoryKey if it exists.
```
aruneshpa:
```suggestion
			// Add the file key for the snapshot memory (.vmem) file if present.
```

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -172,3 +176,163 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+			if snapshot.MemoryKey != -1 { // .vmem
+				vmCtx.Logger.V(5).Info("Adding memoryKey", "memoryKey", snapshot.MemoryKey)
+				fileKeyList = append(fileKeyList, int(snapshot.MemoryKey))
+			}
+
+			// .vmsn
+			vmCtx.Logger.V(5).Info("Adding dataKey", "dataKey", snapshot.DataKey)
```
aruneshpa:
```suggestion
			vmCtx.Logger.V(5).Info("Adding the file key for snapshot (.vmsn) file", "dataKey", snapshot.DataKey)
```

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -172,3 +176,163 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+			fileKeyList = append(fileKeyList, int(snapshot.DataKey))
+
+			// Add the last element of fileKeys in the disk.chain.
```
aruneshpa:
```suggestion
			// Add the disk files for the child most delta disk which is the last item in the disk chain.
```

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -172,3 +176,161 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+			// .vmdk files
+			for _, disk := range snapshot.Disk {
+				for _, chain := range disk.Chain {
+					if len(chain.FileKey) > 0 {
+						vmCtx.Logger.V(5).Info("Adding fileKey", "fileKey", chain.FileKey[len(chain.FileKey)-1])
+						fileKeyList = append(fileKeyList, int(chain.FileKey[len(chain.FileKey)-1]))
```
aruneshpa:
```suggestion
						vmCtx.Logger.V(5).Info("Adding the file key for the child most delta disk of the snapshot", "fileKey", chain.FileKey[len(chain.FileKey)-1])
```

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -172,3 +176,161 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+					}
+				}
+			}
+		}
+	}
+
+	// Find the keySets for FCDs from layoutEx.file
+	fcdDeviceFileKeySet := getFCDDevicesFileKeySet(vmCtx, vmDevices, vmlayout.File)
```
aruneshpa:
```suggestion
	// A VM snapshot creates a delta disk for all disks of a VM -- including
	// PVCs that are backed by FCDs.  Since the usage of the delta disks 
	// created on PVCs will already be reported by the VolumeSnapshot 
	// SPU, we need to skip those here to avoid double counting. 
```

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -172,3 +176,161 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+
+// getFCDDevicesFileKeySet extracts fileKeys of layoutEx.Files that are part of the FCDs.
+// Here we use the file name to identify the FCDs because the backingObjectID
+// is not always available. and there is no other uuid available.
+func getFCDDevicesFileKeySet(vmCtx pkgctx.VirtualMachineContext, vmDevices []types.BaseVirtualDevice, vmlayoutFileInfos []types.VirtualMachineFileLayoutExFileInfo) sets.Set[int] {
+	fcdDevicesBackingFileNamePrefixSet := sets.Set[string]{}
+	for _, device := range vmDevices {
+		vd, ok := device.(*types.VirtualDisk)
+		if !ok || vd.VDiskId == nil || vd.VDiskId.Id == "" { // FCD has VDiskId
+			continue
```
aruneshpa, line 263: Consider using this pattern:
```
			device := object.VirtualDeviceList(moVM.Config.Hardware.Device)
			for _, d := range device.SelectByType(&vimtypes.VirtualDisk{}) {
				disk := d.(*vimtypes.VirtualDisk)
				....
			}

``

File: pkg/providers/vsphere/virtualmachine/snapshot.go [hunk 30]
```diff
@@ -172,3 +176,161 @@ func (m snapshotMap) add(parent string, tree []types.VirtualMachineSnapshotTree)
...
+		// "[sharedVmfs-0] fcd/c4f94aecf2cf42ba9bb66fb6c9fc3527-flat.vmdk"
+
+		// In order to find the corresponding fcd related file name in the layoutEx.file,
+		// get the uuid from fileName and check whether the layoutEx.file.Name contains the fcd/<uuid>.
+		// If yes, add the file.Key of that layoutEx.file to the fcdDeviceFileKeySet
+
+		// Use regex matcher to find the string like fcd/[a-zA-Z0-9] from the layoutEx.file.Name.
+		re := regexp.MustCompile(`fcd\/[a-zA-Z0-9]+`)
+		matches := re.FindStringSubmatch(fileName)
+		if len(matches) > 0 {
```
aruneshpa: Can you check and confirm that these snapshotted disks do not have `vDiskId` set to true in the `snapshot.config.hardware.device` that will be the easiest way to tell that these are FCDs.?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
aruneshpa: Nevermind. you are following the same pattern as VM SPU where we either calculate the reserved, or usage -- but update both.   So one of them is always 0.

lubronzhan: Yeah. If the logic is let each reconcile update the SPU's status, then we need some data in Snapshot object to indicate that current vmSnapshot has been removed from reservation and added to usage. But then there is case that updating SPU succeeded, updating that data failed, it would end up in duplicate update. Unless we could have some kind of atomic transaction to make sure updating SPU and VMSnapshot is atomic. Or we put an array data in SPU, something like annotation to show what has been included in reservation and what has been included in usage. That way is also O(n) complexity since we have to parse the whole array and update it.

pkg/util/kube/vmsnapshot.go
lubronzhan: Got it.

`requestedMap[*pvc.Spec.StorageClassName]; !ok {` is not needed to be moved up though, since it's only needed when the storageClass is ran into for the first time and we want to add its value to the map.

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: After discussion we decided to use device keys as identifier to filter files under snapshot disk chain

lubronzhan:
Verified again with 2 snapshot on the VM. The VM has a PVC and it was used for `634390801` byte. The used sum in vmsnapshot doesn't include this number

lubronzhan:
Hi @bryanv could you help take a look again? Thank you!

PR #1068, opened 2025-07-30
 🐛 Disable fast deploy for instance storage VMs

bryanv:

For an instance storage VM during fast deploy, we'll likely end up doing multiple placement calls, because once the IS PVCs are all bound, we need to do a final placement to get the Datastores.

However, since we're not sending in the selected host to that later placement call, we may end up with a different host.

That is fixable - at least with PlaceVM - but IS isn't used very much so defer fixing that until later.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -2194,6 +2194,19 @@ func (vs *vSphereVMProvider) vmCreateGetStoragePrereqs(
 		createArgs.HasInstanceStorage = AddInstanceStorageVolumes(
 			vmCtx,
 			createArgs.VMClass.Spec.Hardware.InstanceStorage)
+
+		if createArgs.HasInstanceStorage && pkgcfg.FromContext(vmCtx).Features.FastDeploy {
+			// Fast deploy breaks instance storage placement since later placement calls
+			// to get the Datastores need to provide the selected host because the PVCs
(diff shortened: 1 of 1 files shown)


dilyar85 (approved):

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2194,6 +2194,19 @@ func (vs *vSphereVMProvider) vmCreateGetStoragePrereqs(
 		createArgs.HasInstanceStorage = AddInstanceStorageVolumes(
 			vmCtx,
 			createArgs.VMClass.Spec.Hardware.InstanceStorage)
+
+		if createArgs.HasInstanceStorage && pkgcfg.FromContext(vmCtx).Features.FastDeploy {
```
aruneshpa, line 2198: I haven't looked at the full trace of function calls, but just double checking that there are no references in code where we have already checked (and possibly acted on) the fast deploy from the context?  Trying to make sure that we don't have a split behavior in a single reconcile where for instance storage, some parts of reconcile behave as if FD is enabled, and from here on out, it is disabled.

bryanv: Yah nothing immediately before this in the VM Provider cares. There is some stuff/checks in the controller that Andrew & I talked about consolidating that but deemed it not worth it.

 I'm working on fixing the placement issues for here that we can address so I hope this is short lived.

PR #1070, opened 2025-07-31
🐛 Do not promote disks twice

akutz:

This patch fixes an issue where disks were being promoted twice due to a stale controller cache on systems with heavy load. The disk promotion reconciler depended upon the promo task reference to be written to the VM status, but upon re-entering reconcile, the status.taskRef field may be empty if the cache is not updated.

This fix removes status.taskRef and instead fetches all of a VM's recent tasks at the beginning of each reconcile.

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/vmware/govmomi/fault"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/ovf"
 	"github.com/vmware/govmomi/pbm"
@@ -752,6 +753,7 @@ var VMUpdatePropertiesSelector = []string{
 	"config",
 	"guest",
 	"layoutEx",
+	"recentTask",
 	"resourcePool",
 	"runtime",
 	"snapshot",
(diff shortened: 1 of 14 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -907,6 +918,42 @@ func (vs *vSphereVMProvider) reconcileStatus(
 		})
 }
 
+func (vs *vSphereVMProvider) getRecentTaskInfo(
+	ctx pkgctx.VirtualMachineContext) (context.Context, error) {
+
+	vcClient, err := vs.getVcClient(ctx)
```
bryanv: The caller already has the vcClient so pass it in?

akutz: Huh, I failed to see that. Sure.

pkg/vmconfig/diskpromo/diskpromo_reconciler.go
```diff
@@ -187,8 +172,11 @@ func (r reconciler) Reconcile(
 	// not participating in snapshots.
 	for i := range allDisks {
 		d := allDisks[i].(*vimtypes.VirtualDisk)
-		if _, ok := snapshotDiskKeys[d.Key]; !ok {
-			if d.VDiskId == nil { // Skip FCDs
+
```
aruneshpa, line 175: If you want, you can reduce the indent a little.
```
for i := range allDisks {
		d := allDisks[i].(*vimtypes.VirtualDisk)
		if _, ok := snapshotDiskKeys[d.Key]; !ok {
			continue
		}
		if d.VDiskId != nil {
			continue
		}
```
(or even merge these two conditions in a single `if` -- but that might be a little hard to read).

aruneshpa (approved):
LGTM.

pkg/vmconfig/diskpromo/diskpromo_reconciler.go
akutz: I used to use `continue`, but I started to move away from it to avoid issues regarding missing coverage. When `continue` is used, you have to be able to verify those negative cases as well. With the way the PR currently is, you only have to cover the one branch.

PR #1075, opened 2025-08-04
✨ Build approximate VM spec when reverting to imported snapshots

palnabarun:

This commit introduces the ability to build an approximate VM spec when
reverting to imported snapshots. The spec is constructed based on the
snapshot's metadata and some fields from the state of the VM in VC,
allowing for a more accurate representation of the VM's state at the
time of the snapshot.

This is done so that the VM spec can reach to a state that likely defines
how the VM was configured when the Snapshot was taken.

NA

## Simulated Test scenario

1. Create a VM using `VirtualMachine` CR.
2. Take a snapshot of the VM from VC.
3. Create a `VirtualMachineSnapshot` CR pointing to the snapshot created in (2)
4. Made some changes to `VirtualMachine` CR.
5. Revert to the snapshot by setting `spec.currentSnapshot` in `VirtualMachine` CR.

The `VirtualMachine` CR should now have spec fields approximated from the workflow introduced in this PR.

```release-note
✨ Build approximate VM spec when reverting to imported snapshots
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -12,6 +12,7 @@ import (
 	"math/rand"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"text/template"
@@ -1337,7 +1338,7 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 	// Restore VM spec and metadata from the snapshot.
 	vmCtx.Logger.Info("Starting VM spec and metadata restoration from snapshot")
 
-	if err := vs.restoreVMSpecFromSnapshot(vmCtx, vcVM, snapObj); err != nil {
+	if err := vs.restoreVMSpecFromSnapshot(vmCtx, vcVM, snapCR, snapObj); err != nil {
 		vmCtx.Logger.Error(err, "Restoring VM spec and metadata from snapshot failed")
 		return true, err
 	}
@@ -1395,9 +1396,10 @@ func (vs *vSphereVMProvider) performSnapshotRevert(
 func (vs *vSphereVMProvider) restoreVMSpecFromSnapshot(
 	vmCtx pkgctx.VirtualMachineContext,
 	vcVM *object.VirtualMachine,
+	snapCR *vmopv1.VirtualMachineSnapshot,
 	snapObj *vimtypes.ManagedObjectReference) error {
 
-	vmYAML, err := vs.getVMYamlFromSnapshot(vmCtx, vcVM, snapObj)
+	vmYAML, err := vs.getVMYamlFromSnapshot(vmCtx, vcVM, snapCR, snapObj)
 	if err != nil {
 		vmCtx.Logger.Error(err, "failed to parse VM spec from snapshot")
 		return err
@@ -1451,6 +1453,7 @@ func (vs *vSphereVMProvider) restoreVMSpecFromSnapshot(
 func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
 	vmCtx pkgctx.VirtualMachineContext,
 	vcVM *object.VirtualMachine,
+	snapCR *vmopv1.VirtualMachineSnapshot,
 	snap *vimtypes.ManagedObjectReference) (string, error) {
 
 	var moSnap mo.VirtualMachineSnapshot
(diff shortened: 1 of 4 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1450,16 +1451,159 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+		chv, err := strconv.ParseInt(moSnap.Config.Version, 10, 32)
+		if err != nil {
+			vmCtx.Logger.V(5).Info("unable to approximate minimum hardware version", "managed object hardware version", moSnap.Config.Version)
+		} else {
+			vm.Spec.MinHardwareVersion = int32(chv)
+		}
+
+		// process BootOptions
+		// TODO(nabarun): Skip BootOptions for now. Only recently added in context of VMGroup.
+		// TODO(AKP): Start a thread with akutz and let him know.
```
palnabarun: @aruneshpa Please confirm this.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1450,16 +1451,159 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+			Advanced: &vmopv1.VirtualMachineAdvancedSpec{
+				ChangeBlockTracking: moSnap.Config.ChangeTrackingEnabled,
+			},
+
+			// TODO(nabarun): How to get this info?
+			// Get using PbmQueryAssociatedProfile and get the Class from that
+			// Alternate, if we assume the below doesn't happen, the StorageClass might as well be
+			// the one on the VM CR since MO will ensure that
+			// TODO(AKP): Ask PM about the corner case whether MO imports a
+			// really old snapshot whose StoragePolicy doesn't exist in NS anymore
```
palnabarun: @aruneshpa Please confirm this.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1450,16 +1452,184 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+			ServerUuid: moSnap.Config.InstanceUuid,
+		})
+
+		if len(profileIds) == 0 {
+			vmCtx.Logger.V(4).Info("VM is not associated with any profile")
+		} else {
+			// Taking only the first one for now
+			// TODO(nabarun): Once consensus is established on what would be the storage class when multiple storage
+			// policies are returned, update this logic if needed
+			profileId := profileIds[0]
```
palnabarun: cc @aruneshpa for a clarification on this

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1450,16 +1452,184 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+
+			if len(storageClasses) == 0 {
+				err := errors.New("error finding storage classes for profile")
+				vmCtx.Logger.Error(err, "", "profileId", profileId)
+				return "", err
+			}
+
+			// in SV, it seems there are two storage classes for the same profileId
+			// TODO(nabarun): Reach consensus on which one to pick up. Currently picking up the first one found.
+			if len(storageClasses) > 1 {
```
palnabarun: @aruneshpa I see there are multiple storage classes in namespace for a profile ID.

```
$ k get sc -o json | jq '.items[] | select(.parameters.storagePolicyID == "1b1dd966-171c-4256-a7aa-887bf4590c8e")'
{
  "allowVolumeExpansion": true,
  "apiVersion": "storage.k8s.io/v1",
  "kind": "StorageClass",
  "metadata": {
    "annotations": {
      "cns.vmware.com/StoragePoolTypeHint": "cns.vmware.com/VMFS"
    },
    "creationTimestamp": "2025-08-11T07:46:29Z",
    "name": "wcpglobal-storage-profile",
    "resourceVersion": "10018",
    "uid": "620fd67f-794c-4f60-a669-d63aa83a8378"
  },
  "parameters": {
    "storagePolicyID": "1b1dd966-171c-4256-a7aa-887bf4590c8e"
  },
  "provisioner": "csi.vsphere.vmware.com",
  "reclaimPolicy": "Delete",
  "volumeBindingMode": "Immediate"
}
{
  "allowVolumeExpansion": true,
  "apiVersion": "storage.k8s.io/v1",
  "kind": "StorageClass",
  "metadata": {
    "annotations": {
      "cns.vmware.com/StoragePoolTypeHint": "cns.vmware.com/VMFS"
    },
    "creationTimestamp": "2025-08-11T07:46:29Z",
    "name": "wcpglobal-storage-profile-latebinding",
    "resourceVersion": "10019",
    "uid": "745d1a5d-7dca-44ec-aafe-06b85dd7d7d1"
  },
  "parameters": {
    "storagePolicyID": "1b1dd966-171c-4256-a7aa-887bf4590c8e"
  },
  "provisioner": "csi.vsphere.vmware.com",
  "reclaimPolicy": "Delete",
  "volumeBindingMode": "WaitForFirstConsumer"
}
```

Currently picking up whichever is the first one. WDYT?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
 	ecList := object.OptionValueList(moSnap.Config.ExtraConfig)
 	raw, exists := ecList.GetString(backupapi.VMResourceYAMLExtraConfigKey)
 	if !exists {
-		vmCtx.Logger.V(4).Info("VM does not have the necessary ExtraConfig in the Snapshot",
+		vmCtx.Logger.V(4).Info("VM does not have the necessary ExtraConfig in the Snapshot. Approximating the spec from the VirtualMachineConfigInfo",
```
aruneshpa: This will happen _once_ per revert -- which is an infrequent operation in itself.  And this is an indication that this VM is being reverted from an imported snapshot.  I think we should make this an info level log.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
-		// again.
-		return "", nil
+
+		// For VMs that are imported, we do not have the
+		// ExtraConfig set in the snapshot. In this case, we
+		// will approximate the spec from the VirtualMachineConfigInfo.
+		// This is a best-effort attempt to restore the VM spec
+		// and metadata from the snapshot.
+
+		// TODO(future): Check if the VM is actually imported.
```
aruneshpa: We absolutely should do this only for imported VM snapshots that are created by mobility service to prevent reverts from greenfield snapshots that were taken externally.  Mobility service uses `vmoperator.vmware.com/imported-snapshot` annotation to differentiate such snapshots from external snapshots.  Can we ensure that we only get into this entire code path if the snapshot was imported?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+				Name:      vmCtx.VM.Name,
+				Namespace: vmCtx.VM.Namespace,
+			},
+		}
+
+		// Labels and Annotations are a TODO item. Discussions are going on about
+		// storing Label and Annotations in VC.
+		//
+		//	Much of the above is governed by the conclusions in
+		//	https://vmw-confluence.broadcom.net/display/WCP/Reverting+a+VM+to+an+Imported+VM+Snapshot
```
aruneshpa: This is an internal link so doesn't really add any value in putting it on GH.  Please remove it.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+			// We only support VM snapshots and affinity policies are not part of VMs.
+			Affinity: nil,
+
+			// It's not possible to construct the crypto spec looking at the VM.
+			Crypto: nil,
+
+			// It doesn't matter since in VCD, VMs can't be customized upon subsequent boots.  Except using Force Customization which is an admin operation.
+			Bootstrap: nil,
+
+			// These are set to the field defaults
```
aruneshpa: Consider omitting all of these so they can take their default value.  It is possible that we change the default value at some point and we don't want to have to come back here and fix it again.  Same comment for all other fields that we are defaulting.

aruneshpa: If something has complicated explanation for why we are defaulting, then I am okay keeping that (e.g., reserved profiles).

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+
+		pbmClient, err := pbm.NewClient(vmCtx, vcVM.Client())
+		if err != nil {
+			vmCtx.Logger.Error(err, "error creating pbm client")
+			return "", err
+		}
+		profileIDs, err := pbmClient.QueryAssociatedProfile(vmCtx, pbmtypes.PbmServerObjectRef{
+			ObjectType: "virtualMachine",
+			Key:        vmCtx.VM.Status.UniqueID,
+			ServerUuid: moSnap.Config.InstanceUuid,
```
aruneshpa: Hmm.  This isn't quite right. You are passing in the VM's instance UUID.  `ServerUuid` here refers to vCenter's UUID.
One way to get that would be to get the service content (you can do that using any of the clients e.g., vim)

Here's a reference to how you can get this: https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmprovider_vm.go#L178

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+		}
+
+		if len(profileIDs) == 0 {
+			vmCtx.Logger.V(4).Info("VM is not associated with any profile")
+		} else {
+			// Taking only the first one for now
+			// TODO(nabarun): Once consensus is established on what would be the storage class when multiple storage
+			// policies are returned, update this logic if needed
+			profileID := profileIDs[0]
+			storageClasses, err := spq.GetStorageClassesForPolicy(vmCtx, vs.k8sClient, vmCtx.VM.Namespace, profileID.UniqueId)
```
aruneshpa: It turns out that when you take a snapshot of a VM using storage policy `A`, and then update / modify the VM to use policy `B`, and then revert to the snapshot, the VM continues to have the policy `A`.  So, let's just simplify all of this and set the VM's spec.storageClass to the one from the current VM (like we do for groupName).

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+					Name: fmt.Sprintf("eth-%d", i),
+				}
+			}
+			vm.Spec.Network.Interfaces = interfaces
+		}
+
+		// MinHardwareVersion is set to the Current Hardware Version of the VM
+		chv, err := strconv.ParseInt(moSnap.Config.Version, 10, 32)
+		if err != nil {
+			vmCtx.Logger.V(5).Info("unable to approximate minimum hardware version", "managed object hardware version", moSnap.Config.Version)
```
aruneshpa: We follow a convention of making the log markers camel cased so it's easier to search through.  So, consider calling it `hwVersionFromSnapshot` instead of `managed object hardware version`

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1488,188 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+		} else {
+			vm.Spec.MinHardwareVersion = int32(chv)
+		}
+
+		// TODO(future): Labels and Annotations are skipped for now as design discussions are
```
aruneshpa, line 1596: This is also called on L1509.  Please remove the reference from there.

pkg/providers/vsphere/vmprovider_vm.go
palnabarun: Makes sense. The defaults can themselves change in the future.

pkg/providers/vsphere/vmprovider_vm.go
palnabarun: Good catch.

pkg/providers/vsphere/vmprovider_vm.go
palnabarun: Good catch. Thanks!

pkg/providers/vsphere/vmprovider_vm.go
palnabarun: Pruned the list of variables being set in the block. Only the ones needing explanation and the ones that are actually needed to be set are kept.

aruneshpa:
I think you may have accidentally nuked the PR template during the update?  Can we use the template so everything is consistent (and add the release note)?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1487,149 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
 	ecList := object.OptionValueList(moSnap.Config.ExtraConfig)
 	raw, exists := ecList.GetString(backupapi.VMResourceYAMLExtraConfigKey)
 	if !exists {
-		vmCtx.Logger.V(4).Info("VM does not have the necessary ExtraConfig in the Snapshot",
+		// fetch te snapshot CR
+		snapCR := &vmopv1.VirtualMachineSnapshot{}
+		if err := vs.k8sClient.Get(
```
aruneshpa: We are fetching the snapshot in the calling method already.  Can you pass that in this method?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1487,149 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+			ctrlclient.ObjectKey{Name: vmCtx.VM.Spec.CurrentSnapshot.Name, Namespace: vmCtx.VM.Namespace},
+			snapCR); err != nil {
+			vmCtx.Logger.Error(err, "Error getting snapshot CR", "snapshotName", vmCtx.VM.Spec.CurrentSnapshot.Name)
+			return "", err
+		}
+
+		// if the Snapshot has ImportedSnapshotAnnotation, then only perform an approximation of the VM spec
+		_, found := snapCR.GetAnnotations()[vmopv1.ImportedSnapshotAnnotation]
+		if !found {
+			return "", nil
```
aruneshpa, line 1476: Log something here on Info so we know that a revert was attempted on an invalid (external) snapshot - which is an unsupported workflow.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1486,16 +1487,149 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+		}
+
+		vmCtx.Logger.Info("VM does not have the necessary ExtraConfig in the Snapshot. Approximating the spec from the VirtualMachineConfigInfo",
```
aruneshpa: Call out more explicitly that VM is being reverted to an imported snapshot?

palnabarun:
Added back the template. 👍🏽

pkg/providers/vsphere/vmprovider_vm.go
palnabarun: Done. I was avoiding changing method signatures, but with only two methods to be changed, I am fine.

palnabarun:
There seems to be something wrong with the GH Pipelines

```
pkg/providers/vsphere/vmprovider_resourcepolicy_test.go:1: : # github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere_test [github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere.test]
Error: pkg/providers/vsphere/vmprovider_vm_test.go:3417:38: undefined: vmlifecycle.FindSnapshotNameInTree (typecheck)
```


```
~/s/v/vm-operator on 🌱 imported-snapshots/create-vm-spec (0799166) [📦🤷‍] on ⛵ kind-exp:kind-exp () [💣 1]
❯ rg FindSnapshotNameInTree
```

palnabarun:
To whoever is looking at the PR in future, the conversation moved to #1111.

PR #1081, opened 2025-08-05
🌱 Disable FastDeploy+Direct / NVRAM file support

akutz:

This patch temporarily disables FastDeploy+Direct mode as well as support for FastDeploy with NVRAM files. They may both be re-enabled by setting a single flag to true.

This patch is required to mitigate a bug in vpxd on main related to vSAN datastores. Currently they do not support creating VMs using an explicit path.

Fixes `NA`

```release-note
Disable FastDeploy+Direct / FastDeploy support for NVrAM files.
```


akutz:
Replaced by #1084.

PR #1083, opened 2025-08-06
🌱 Exclude snapshot related disks/files when calculating VM status storage usage

lubronzhan:
Consume new `GetVirtualDiskInfoByUUID` in govmomi introduced by https://github.com/vmware/govmomi/pull/3832 

Now when calculating the usage for VM, it exclude those snapshot related delta disks. Since we will calculate snapshot related usage in VMSnapshot CR instead of VirtualMachine CR.

And for VMI's usage, we still include all chains. Related detail can be found here https://github.com/vmware-tanzu/vm-operator/pull/1083#discussion_r2261235151

**More Info**
Updated VM status when the VM has two snapshots
```
    volumes:
    - attached: true
      diskUUID: 6000C292-3c7a-7cbc-c945-0fb35c5a494b
      limit: 10Gi
      name: lubron-vm
      requested: 10Gi
      type: Classic
      used: "24117592"
    - attached: true
      diskUUID: 6000C299-85a3-b188-e22e-3b6c3c3993d6
      limit: 5Gi
      name: lubron-vol
      requested: 5Gi
      type: Managed
      used: "1049058"
```

VM's disk chain
image
1. The disk 2000 is the Classic disk
2. The disk 2001 is the Managed disk

Disk 2000's size correspond to size of (file15 + file16)
image
24117248 + 344 == 24117592

Disk 2001's size corresponds to size of (file17 + file18)
image
482 + 1048576 == 1049058

```release-note
Exclude snapshot related disks/files when calculating VM status storage usage
```

### DIFF
--- a/api/v1alpha4/virtualmachine_storage_types.go
+++ b/api/v1alpha4/virtualmachine_storage_types.go
@@ -202,14 +202,16 @@ type VirtualMachineStorageStatusUsed struct {
 	// +optional
 
 	// Disks describes the total storage space used by a VirtualMachine's
-	// non-PVC disks.
+	// non-PVC disks. This does not include the any child / delta disks
+	// created for snapshots.
 	Disks *resource.Quantity `json:"disks,omitempty"`
 
 	// +optional
 
 	// Other describes the total storage space used by the VirtualMachine's
-	// non disk files, ex. the configuration file, swap space, logs, snapshots,
-	// etc.
+	// non-disk files, ex. the configuration file, swap space, logs, etc.
+	// This does not include the non-disk files created for snapshots,
+	// ex. snapshot data, list, and memory files.
 	Other *resource.Quantity `json:"other,omitempty"`
 }
 
--- a/pkg/providers/vsphere/contentlibrary/content_library_utils.go
+++ b/pkg/providers/vsphere/contentlibrary/content_library_utils.go
@@ -185,7 +185,12 @@ func UpdateVmiWithVirtualMachine(
 
 			if uuid != "" {
 				di, _ := vmdk.GetVirtualDiskInfoByUUID(
-					ctx, nil, vm, false, uuid)
+					ctx,
+					nil,   /* the client is not needed since props aren't refetched */
+					vm,    /* use props from this object */
+					false, /* do not refetch props */
+					false, /* include disks related to snapshots */
+					uuid)
 				capacity = kubeutil.BytesToResource(di.CapacityInBytes)
 				size = kubeutil.BytesToResource(di.Size)
 			}
--- a/pkg/providers/vsphere/vmlifecycle/update_status.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status.go
@@ -849,7 +849,11 @@ func updateStorageUsage(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 				vimtypes.VirtualMachineFileLayoutExFileTypeDigestExtent:
 
 				// Skip disks
+			case vimtypes.VirtualMachineFileLayoutExFileTypeSnapshotList,
+				vimtypes.VirtualMachineFileLayoutExFileTypeSnapshotMemory,
+				vimtypes.VirtualMachineFileLayoutExFileTypeSnapshotData:
 
+				// Skip snapshot related non-disk files
 			default:
 				other += f.UniqueSize
 			}
@@ -968,7 +972,13 @@ func updateVolumeStatus(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 		if diskIndex, ok := existingDisksInStatus[diskUUID]; ok {
 			// The disk is already in the list of volume statuses, so update the
 			// existing status with the usage information.
-			di, _ := vmdk.GetVirtualDiskInfoByUUID(ctx, nil, moVM, false, diskUUID)
+			di, _ := vmdk.GetVirtualDiskInfoByUUID(
+				ctx,
+				nil,   /* the client is not needed since props aren't refetched */
+				moVM,  /* use props from this object */
+				false, /* do not refetch props */
+				true,  /* exclude disks related to snapshots */
+				diskUUID)
 			vm.Status.Volumes[diskIndex].Used = kubeutil.BytesToResource(di.UniqueSize)
 			if di.CryptoKey.ProviderID != "" || di.CryptoKey.KeyID != "" {
 				vm.Status.Volumes[diskIndex].Crypto = &vmopv1.VirtualMachineVolumeCryptoStatus{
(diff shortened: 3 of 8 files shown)


api/v1alpha4/virtualmachine_storage_types.go
```diff
@@ -202,14 +202,14 @@ type VirtualMachineStorageStatusUsed struct {
 	// +optional
 
 	// Disks describes the total storage space used by a VirtualMachine's
-	// non-PVC disks.
+	// non-PVC disks. This does not include the snapshot related disks.
```
aruneshpa:
```suggestion
	// non-PVC disks. This does not include the any child / delta disks created for snapshots.
```

api/v1alpha4/virtualmachine_storage_types.go
```diff
@@ -202,14 +202,14 @@ type VirtualMachineStorageStatusUsed struct {
...
 	Disks *resource.Quantity `json:"disks,omitempty"`
 
 	// +optional
 
 	// Other describes the total storage space used by the VirtualMachine's
-	// non disk files, ex. the configuration file, swap space, logs, snapshots,
-	// etc.
+	// non disk files, ex. the configuration file, swap space, logs, etc.
+	// This does not include the snapshot related non disk files.
```
aruneshpa:
```suggestion
	// This does not include the any child / delta disks created for snapshots.
    ```

Same in other places maybe?

File: pkg/providers/vsphere/contentlibrary/content_library_utils.go [hunk 3]
```diff
@@ -184,8 +184,8 @@ func UpdateVmiWithVirtualMachine(
 			}
 
 			if uuid != "" {
-				di, _ := vmdk.GetVirtualDiskInfoByUUID(
-					ctx, nil, vm, false, uuid)
+				di, _ := virtualmachine.GetVirtualDiskInfoByUUID(
```
aruneshpa: VMIs can have snapshots (VMIs shouldn't even have delta disks), so I think here we need to calculate the entire chain.  Now, it doesn't matter because I only expect one disk in the chain for the VMIs, but still worth checking with @akutz to confirm that we are not missing anything.

aruneshpa (approved):
LGTM.  Thanks for this @lubronzhan !

api/v1alpha4/virtualmachine_storage_types.go
lubronzhan: mm The `other` just contains files, not disks. I change this to same syntax

pkg/providers/vsphere/virtualmachine/devices.go
```diff
@@ -99,3 +106,164 @@ func CreateInstanceStorageDiskDevices(isVolumes []vmopv1.VirtualMachineVolume) [
...
+type VirtualDiskInfo struct {
+	CapacityInBytes int64
+	DeviceKey       int32
+	FileName        string
+	Size            int64
+	UniqueSize      int64
+	CryptoKey       VirtualDiskCryptoKey
+}
+
+// GetVirtualDiskInfoByUUID returns information about a virtual disk identified
```
akutz: Why not update GoVmomi instead of copying this into VM Op?

akutz (changes requested):

pkg/providers/vsphere/contentlibrary/content_library_utils.go
akutz: VMIs as OVFs will not have chains, but project-scoped libraries (backed by VMs) could, so yes, we will need the entire chain. Or at least the final node in it with a size that represents the totals of the chain.

pkg/providers/vsphere/virtualmachine/devices.go
lubronzhan: I discussed with @aruneshpa that this way it would be faster since we don't have to bump govmomi to a new version.

lubronzhan:
Created PR in govmomi to be able to allow excluding snapshot when calculating size.

Now this PR consumes the new govmomi sha. 

When calculating VM usage, exclude snapshot, when calculating VMI usage, include all disks files

api/v1alpha4/virtualmachine_storage_types.go
```diff
@@ -202,14 +202,15 @@ type VirtualMachineStorageStatusUsed struct {
...
+	// created for snapshots.
 	Disks *resource.Quantity `json:"disks,omitempty"`
 
 	// +optional
 
 	// Other describes the total storage space used by the VirtualMachine's
-	// non disk files, ex. the configuration file, swap space, logs, snapshots,
-	// etc.
+	// non disk files, ex. the configuration file, swap space, logs, etc.
+	// This does not include the non disk files created for snapshots.
```
akutz: nit: `s/non disk/non-disk/g`. 

Also, suggest `non-disk files created for snapshots, ex. snapshot data, list, and memory files.`

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -968,7 +972,8 @@ func updateVolumeStatus(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 		if diskIndex, ok := existingDisksInStatus[diskUUID]; ok {
 			// The disk is already in the list of volume statuses, so update the
 			// existing status with the usage information.
-			di, _ := vmdk.GetVirtualDiskInfoByUUID(ctx, nil, moVM, false, diskUUID)
+			// Exclude files in disk chains for snapshots.
+			di, _ := vmdk.GetVirtualDiskInfoByUUID(ctx, nil, moVM, false, true, diskUUID)
```
akutz: suggest:

```golang
di, _ := vmdk.GetVirtualDiskInfoByUUID(
	ctx,
	nil,   /* the client is not needed since props aren't refetched */
	moVM,  /* use props from this object */
	false, /* do not refetch props */
	true,  /* exclude disks related to snapshots */ 
	diskUUID)
```

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -979,7 +984,8 @@ func updateVolumeStatus(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 		} else if !isFCD {
 			// The disk is a classic, non-FCD that must be added to the list of
 			// volume statuses.
-			di, _ := vmdk.GetVirtualDiskInfoByUUID(ctx, nil, moVM, false, diskUUID)
+			// Exclude files in disk chains for snapshots.
+			di, _ := vmdk.GetVirtualDiskInfoByUUID(ctx, nil, moVM, false, true, diskUUID)
```
akutz: suggest:

```golang
di, _ := vmdk.GetVirtualDiskInfoByUUID(
	ctx,
	nil,   /* the client is not needed since props aren't refetched */
	moVM,  /* use props from this object */
	false, /* do not refetch props */
	true,  /* exclude disks related to snapshots */ 
	diskUUID)
```

PR #1086, opened 2025-08-08
🌱 Only create SPU for snapshot when its feature gate is enabled

lubronzhan:
Any resource related to VMSnapshot should be created only if its feature gate is enabled.

Verified in real env, by default
```
I0808 05:26:24.886316       1 storagepolicyquota_controller.go:160] "Skip creating storage policy usage for resource kind" logger="controllers.StoragePolicyQuota" kind="VirtualMachineSnapshot"
```
Once the feature gate is enabled, SPU is created
```
I0808 05:49:30.072534       1 capabilties.go:172] "Updated features from capabilities" logger="setup" dryRun=false diff="TKGMultipleCL=true,VMSnapshots=true,WorkloadDomainIsolation=true"

 k get storagepolicyusage -n lubron-test wcpglobal-storage-profile-vmsnapshot-usage
NAME                                         AGE
wcpglobal-storage-profile-vmsnapshot-usage   25m
```

```release-note
None
```

### DIFF
--- a/controllers/storagepolicyquota/storagepolicyquota_controller.go
+++ b/controllers/storagepolicyquota/storagepolicyquota_controller.go
@@ -107,8 +107,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
-	if err := r.ReconcileNormal(ctx, &obj); err != nil {
-		pkgutil.FromContextOrDefault(ctx).Error(err, "Failed to ReconcileNormal StoragePolicyQuota")
+	logger := pkgutil.FromContextOrDefault(ctx)
+
+	if err := r.ReconcileNormal(ctx, logger, &obj); err != nil {
+		logger.Error(err, "Failed to ReconcileNormal StoragePolicyQuota")
 		return ctrl.Result{}, err
 	}
 
@@ -117,6 +119,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 
 func (r *Reconciler) ReconcileNormal(
 	ctx context.Context,
+	logger logr.Logger,
 	spq *spqv1.StoragePolicyQuota) error {
 
 	caBundle, err := spqutil.GetWebhookCABundle(ctx, r.Client)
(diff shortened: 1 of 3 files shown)


controllers/storagepolicyquota/storagepolicyquota_controller.go
```diff
@@ -137,26 +137,34 @@ func (r *Reconciler) ReconcileNormal(
...
+			nameFunc: spqutil.StoragePolicyUsageNameForVMSnapshot,
+			enabled:  pkgcfg.FromContext(ctx).Features.VMSnapshots,
 		},
 	}
 
 	var errs []error
 	for i := range objs {
 		for _, resourceKind := range resourceKinds {
+			if !resourceKind.enabled {
+				r.Logger.V(5).Info("Skip creating storage policy usage for resource kind", "kind", resourceKind.kind)
```
bryanv: It doesn't matter too much for this situation but if you're going to log, prefer using the logger that controller-runtime places in the context with `pkgutil.FromContextOrDefault(ctx)`

lubronzhan: Oh thanks for the reminding. Let me update this in a follow up

aruneshpa: By followup, do you mean in a new revision, or a followup MR?  I think this can be fixed in this change since it's low risk and low effort.

akutz: Please update this PR to use what @bryanv provided. Also, please use log-level 4. Log level 2 is standard, and 4 tends to be our verbose mode. Having additional levels beyond that is complicated because it also pulls in a bunch of CR logging.

Suggestion:

```golang
logger := pkgutil.FromContextOrDefault(ctx)

for i := range objs {
		for _, resourceKind := range resourceKinds {
			if !resourceKind.enabled {
				logger.V(4).Info(
					"Skip creating storage policy usage for resource",
					"kind", resourceKind.kind)
```

lubronzhan: Got it. I found some other place that needs to be updated, was gonna update them all in a follow up. Now all updated in this PR

lubronzhan: MM since I need to cherry pick this PR, let me remove those change and add them in follow up instead

PR #1088, opened 2025-08-08
🌱Notify CSI for volume snapshot sync by adding the annotation

lubronzhan:

After the VMSnapshot is ready, we need to notify the CSI to sync volumeSnapshot SPU usage by adding the annotation on VMSnapshot.  This PR adds that annotation when necessary

**Other note**
The annotation was added when the vmsnapshot is marked as ready
```
root@423d35f0e284a74b4fdab08a554ca925 [ ~ ]# k get vmsnapshot -A -oyaml
apiVersion: v1
items:
- apiVersion: vmoperator.vmware.com/v1alpha4
  kind: VirtualMachineSnapshot
  metadata:
    annotations:
      csi.vsphere.volume.sync: requested
        {"apiVersion":"vmoperator.vmware.com/v1alpha4","kind":"VirtualMachineSnapshot","metadata":{"annotations":{},"name":"sn-1","namespace":"lubron-test"},"spec":{"vmRef":{"apiVersion":"vmoperator.vmware.com/v1alpha4","kind":"VirtualMachine","name":"lubron-vm"}}}
    creationTimestamp: "2025-08-08T16:54:34Z"
    finalizers:
    - vmoperator.vmware.com/virtualmachinesnapshot
...  
status:
    conditions:
    - lastTransitionTime: "2025-08-08T16:54:35Z"
      message: ""
      reason: "True"
      status: "True"
      type: VirtualMachineSnapshotReady
    storage:
      requested:
      - storageClass: wcpglobal-storage-profile
        total: 25Gi
      used: "4976565820"
    uniqueID: snapshot-1067
kind: List
metadata:
  resourceVersion: ""
```

```release-note
None
```

### DIFF
--- a/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
+++ b/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
@@ -20,6 +20,7 @@ import (
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
 	vmopv1common "github.com/vmware-tanzu/vm-operator/api/v1alpha4/common"
+	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	"github.com/vmware-tanzu/vm-operator/pkg/constants"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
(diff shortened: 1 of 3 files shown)


lubronzhan:
Fixed the test failure

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -169,6 +170,8 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
 		return ctrl.Result{}, errVMRefNil
 	}
 
+	ensureCSIVolumeSyncAnnotation(vmSnapshot)
```
aruneshpa: Can you move the conditions.IsTrue check here instead of the called method?  That would make it clear to the reader that we only evaluate the annotation if the snapshot is ready.

aruneshpa (approved):

PR #1091, opened 2025-08-09
✨ Introduce VirtualMachineGroupPublishRequest Webhooks

zhangheliu:
It introduces the validation and mutation webhooks for VirtualMachineGroupPublishRequest

Fixes # N/A

N/A

```release-note
Introduce VirtualMachineGroupPublishRequest Webhooks
```


webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,152 @@
...
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+	"github.com/vmware-tanzu/vm-operator/pkg/builder"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/webhooks/common"
+)
+
+// +kubebuilder:webhook:path=/default-mutate-vmoperator-vmware-com-v1alpha4-virtualmachinegrouppublishrequest,mutating=true,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachinegrouppublishrequests,verbs=create,versions=v1alpha4,name=default.mutating.virtualmachinegrouppublishrequest.v1alpha4.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinegrouppublishrequests,verbs=get;list;update
+// +kubebuilder:rbac:groups=imageregistry.vmware.com,resources=contentlibraries,verbs=get;list
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinegroups,verbs=get;list
```
dilyar85: Nit: this is typically put right before the `AddToManager` function in all controller/webhook files.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,152 @@
...
+		return admission.Allowed("")
+	}
+
+	return m.mutateCrate(ctx)
+}
+
+// mutateCrate sets spec.source to the VM group publish request's name if it is omitted,
+// spec.target to a default content library if one exists, and spec.virtualMachines to
+// all the VMs under source VM group.
+func (m mutator) mutateCrate(ctx *pkgctx.WebhookRequestContext) admission.Response {
```
dilyar85: Typo `s/mutateCrate/mutateCreate`

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,152 @@
...
+
+	vmGroup := &vmopv1.VirtualMachineGroup{}
+	if err := m.client.Get(ctx, vmGroupKey, vmGroup); err != nil {
+		return nil, err
+	}
+
+	for _, member := range vmGroup.Status.Members {
+		switch member.Kind {
+		case "VirtualMachine":
+			vms = append(vms, member.Name)
```
dilyar85: I wonder if we should add the VM only if it's currently linked to the group (by checking its condition)?

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
dilyar85: I don't think the list of VM Group is needed here?

webhooks/virtualmachinepublishrequest/validation/virtualmachinepublishrequest_validator.go
```diff
@@ -165,6 +173,38 @@ func (v validator) validateTargetLocation(ctx *pkgctx.WebhookRequestContext, vmp
...
+	return fmt.Errorf("%q label is reserved", vmopv1.VirtualMachinePublishRequestManagedByLabelKey)
+}
+
+// validateUpdateVMGroupPublishRequestOwnership ensures no updates are applied to a VirtualMachineGroupPublishRequest
+// managed VirtualMachinePublishRequest and the managed by VirtualMachineGroupPublishRequest label can only be added
+// during a VirtualMachinePublishRequest creation.
+func (v validator) validateUpdateVMGroupPublishRequestOwnership(oldVMPub, vmPub *vmopv1.VirtualMachinePublishRequest) error {
+	if metav1.HasLabel(oldVMPub.ObjectMeta, vmopv1.VirtualMachinePublishRequestManagedByLabelKey) {
+		return fmt.Errorf("cannot update VirtualMachineGroupPublishRequest owned VirtualMachinePublishRequest")
+	}
```
dilyar85, line 207: This should be checked only for non-admin accounts to not block VMOP updating the object?

webhooks/common/utils_test.go
```diff
@@ -0,0 +1,119 @@
...
+	),
+	func() {
+		var (
+			ctx *builder.UnitTestContext
+		)
+		BeforeEach(func() {
+			ctx = builder.NewUnitTestContext()
+		})
+
+		When("there is an issue with client", func() {
```
bryanv: What's the issue with the fake client this test is covering?

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,152 @@
...
+		case "VirtualMachineGroup":
+			groupMembers, err := m.retrieveVMGroupMembers(ctx, ctrlclient.ObjectKey{Namespace: vmGroupKey.Namespace, Name: member.Name})
```
bryanv: Can you create an issue to handle the possible loop / recursive depth? (The same issue exists in the Group controller)

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,152 @@
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return admission.Errored(http.StatusNotFound, err)
+			}
+			return admission.Errored(http.StatusInternalServerError, err)
+		}
+		groupPubReq.Spec.VirtualMachines = vms
+	}
+	// ensure vms are unique and sorted
+	groupPubReq.Spec.VirtualMachines = sets.List(sets.New(groupPubReq.Spec.VirtualMachines...))
```
bryanv: I'm not sure if we want to change the whatever order this was created with. Like a user made a copy-pasta error that is basically swallowed here

The unique part is better asserted in the validation webhook, and should fail for dupes

webhooks/virtualmachinegrouppublishrequest/validation/virtualmachinegrouppublishrequest_validator.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+// +kubebuilder:webhook:verbs=create;update,path=/default-validate-vmoperator-vmware-com-v1alpha4-virtualmachinegrouppublishrequest,mutating=false,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachinegrouppublishrequests,versions=v1alpha4,name=default.validating.virtualmachinegrouppublishrequest.v1alpha4.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinegrouppublishrequests,verbs=get;list
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachinegroups,verbs=get
+// +kubebuilder:rbac:groups=imageregistry.vmware.com,resources=contentlibraries,verbs=get
+
+const (
+	webHookName                = "default"
+	mutatingWebhookDownWarning = "The mutating webhook should have filled it if empty."
```
bryanv: This is a weird error message. The failure policy is to fail. I think if the Group didn't contain any VMs the user would get this error message, no?

webhooks/virtualmachinegrouppublishrequest/validation/virtualmachinegrouppublishrequest_validator.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return fieldErrs
+}
+
+// validateSpecVMs ensures spec.VirtualMachines is defined either via the user or the mutation webhook.
+// It does not check the existence of those VMs since it would require multiple calls to the API server.
+// Rather, each VM publish request's status is reflected on VM group publish request's status.
+func (validator) validateSpecVMs(specPath *field.Path, groupPubReq *vmopv1.VirtualMachineGroupPublishRequest) *field.Error {
+	if len(groupPubReq.Spec.VirtualMachines) == 0 {
+		return field.Required(specPath.Child("virtualMachines"), mutatingWebhookDownWarning)
```
bryanv: Maybe the mutation webhook should fail instead with a more specific error message.

And then this validation webhook fails if this is empty.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,152 @@
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return admission.Errored(http.StatusNotFound, err)
+			}
+			return admission.Errored(http.StatusInternalServerError, err)
+		}
+		groupPubReq.Spec.Target = defaultCL.Name
+	}
+
+	if len(groupPubReq.Spec.VirtualMachines) == 0 {
```
bryanv, line 102: If this _isn't_ empty at create, what's the expected behavior? The Source is ignored (but then it is odd to me that we're updating the Source if it was empty)

webhooks/virtualmachinepublishrequest/validation/virtualmachinepublishrequest_validator_intg_test.go
```diff
@@ -82,6 +96,37 @@ func intgTestsValidateCreate() {
...
+				builder.DummyVMGroupPublishRequestName, ctx.vmPub.Namespace, []string{})
+			vmGroupPubReq.SetUID("dummy-uid")
+			ctx.vmPub.Labels = map[string]string{vmopv1.VirtualMachinePublishRequestManagedByLabelKey: vmGroupPubReq.Name}
+		})
+		AfterEach(func() {
+			vmGroupPubReq = nil
+		})
+
+		It("should deny the request if owner reference is not set", func() {
+			Eventually(func(g Gomega) {
```
bryanv: Does this really need to be in Eventually() ?

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: looking at vmgroup controller, looks like it only sets the members' status after the linked condition is true.

webhooks/common/utils_test.go
zhangheliu: it is purely for the code line coverage report. i can remove this since there is not much value added

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: ack, linked to vm group publish epic.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: yeah, was thinking that the API could be a little more tolerance but it makes sense to keep user input without modification.

webhooks/virtualmachinegrouppublishrequest/validation/virtualmachinegrouppublishrequest_validator.go
zhangheliu: Was thinking about the extreme case when the mutating webhook is down while the validating webhook is still up (an operator turn off the webhook by following certain runbooks and forgets to put it back).

webhooks/virtualmachinegrouppublishrequest/validation/virtualmachinegrouppublishrequest_validator.go
zhangheliu: make sense, was trying to point users to check the mutating webhook configuration when those optional fields are empty.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: Per our current API definition: 

The virtualMachines only get populated when it is empty. We are not declaring any relationships between the group and virtualMachines fields yet. 

My understanding is that virtualMachines is always the source of truth for what VM publish requests to create. The source group feild is only there to help if virtualMachine fields are empty. 

There could be a case when a user defines virtualMachines and sets the source group to some random groups. I am not sure if we should still want to check if the vm group exists or not. 

perhaps: 
we could clearly define in our API stating:

either 

- if there is a relationship between virtualMachines and source group, vaildating webhook ensures all vms defined are in the group 
- if there is NO relations between  virtualMachines and source group, mutating webhook will set spec.source to empty, ignore user's input since it does not matter

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
dilyar85: This would also help you fail early if the VM CR doesn't exist yet (the group link condition is true only when the VM member exists and points to the correct group).

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: Just had this discussion during the demo meeting. Looks like we are going with: if virtualMachines is omitted, it will be all the members in the source group. If it is defined, we need to validate to ensure those VMs are a subset of the group. 

will update the API and webhooks accordingly.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: sounds good, thanks~

webhooks/common/utils_test.go
```diff
@@ -0,0 +1,313 @@
...
+			})
+		})
+
+		When("there is a default CL", func() {
+			BeforeEach(func() {
+				Expect(ctx.Client.Create(ctx, builder.DummyDefaultContentLibrary(
+					builder.DummyContentLibraryName+"-0", "", ""))).To(Succeed())
+			})
+
+			It("should return NotFound err", func() {
```
lubronzhan: nit: should succeeds

webhooks/common/utils.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
+)
+
+const (
+	DefaultIImagePublishContentLibraryLabelKey = "imageregistry.vmware.com/default"
```
dilyar85: Extra `I`?

webhooks/common/utils.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if len(clList.Items) == 0 {
+		return nil, apierrors.NewNotFound(schema.GroupResource{
+			Group:    imgregv1a1.GroupVersion.Group,
+			Resource: "ContentLibrary",
+		}, "")
+	}
+
+	if len(clList.Items) > 1 {
+		return nil, apierrors.NewBadRequest("more than one default ContentLibrary found")
```
dilyar85: Worth adding the namespace and/or `clList.Items` for better visibility in error logs?

webhooks/common/utils.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return validationErrs
+}
+
+// RetrieveVMGroupMembers retrieves all the group linked VMs under a VM group recursively.
+// An error is return if a loop is detected among nested groups or a group has duplicated members.
+func RetrieveVMGroupMembers(ctx context.Context, c ctrlclient.Client,
+	vmGroupKey ctrlclient.ObjectKey, visitedGroups *sets.Set[string]) (sets.Set[string], error) {
+	if visitedGroups.Has(vmGroupKey.Name) {
+		return nil, fmt.Errorf("a loop is detected among groups")
```
dilyar85: Consider adding the duplicate group name in the error message?

webhooks/common/utils.go
```diff
@@ -0,0 +1,105 @@
...
+		if fieldErr != nil {
+			validationErrs = append(validationErrs, fieldErr.Error())
+		}
+	}
+	return validationErrs
+}
+
+// RetrieveVMGroupMembers retrieves all the group linked VMs under a VM group recursively.
+// An error is return if a loop is detected among nested groups or a group has duplicated members.
+func RetrieveVMGroupMembers(ctx context.Context, c ctrlclient.Client,
```
dilyar85: I feel like `pkg/util/vmopv1` might be a better place to put this function since we need to do similar check in group placement or other controller logic. Not a blocker for this PR but just a heads-up this might get moved later.

webhooks/virtualmachinegrouppublishrequest/validation/virtualmachinegrouppublishrequest_validator.go
```diff
@@ -0,0 +1,179 @@
...
+	fieldErrs = append(fieldErrs, validation.ValidateImmutableField(groupPubReq.Spec.Target, oldGroupPubReq.Spec.Target, specPath.Child("target"))...)
+	fieldErrs = append(fieldErrs, validation.ValidateImmutableField(groupPubReq.Spec.VirtualMachines, oldGroupPubReq.Spec.VirtualMachines, specPath.Child("virtualMachines"))...)
+
+	return fieldErrs
+}
+
+// validateSpecVMs ensures spec.VirtualMachines is defined either via the user or the mutation webhook.
+// It does not check the existence of those VMs since it would require multiple calls to the API server.
+// Rather, each VM publish request's status is reflected on VM group publish request's status.
+func (v validator) validateSpecVMs(specPath *field.Path,
```
dilyar85: I would rather highlight what this function actually does - it verifies the `spec.virtualMachines` is a subset of all members in the group defined in `spec.source`.

webhooks/virtualmachinegrouppublishrequest/validation/virtualmachinegrouppublishrequest_validator.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	visitedGroups := &sets.Set[string]{}
+	vmGroupSet, err := common.RetrieveVMGroupMembers(ctx, v.client,
+		ctrlclient.ObjectKey{Namespace: groupPubReq.Namespace, Name: groupPubReq.Spec.Source}, visitedGroups)
+	if err != nil {
+		return field.InternalError(specPath.Child("virtualMachines"), err)
+	}
+	if !vmGroupSet.IsSuperset(sets.New(groupPubReq.Spec.VirtualMachines...)) {
+		return field.Invalid(specPath.Child("virtualMachines"), "", "virtual machines must in the source group")
```
dilyar85: Nit: `virtual machines must be a direct or indirect member of the source group` or something like that for more clarity.

webhooks/common/utils.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+
+	imgregv1a1 "github.com/vmware-tanzu/image-registry-operator-api/api/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
```
dilyar85: This needs to be `v1alpha5` now since the new API has been introduced (same as other places importing the `vmopv1` package in your change).

config/webhook/manifests.yaml
```diff
@@ -157,6 +177,27 @@ webhooks:
...
+      name: webhook-service
+      namespace: system
+      path: /default-validate-vmoperator-vmware-com-v1alpha4-virtualmachinegrouppublishrequest
+  failurePolicy: Fail
+  name: default.validating.virtualmachinegrouppublishrequest.v1alpha4.vmoperator.vmware.com
+  rules:
+  - apiGroups:
+    - vmoperator.vmware.com
+    apiVersions:
+    - v1alpha4
```
lubronzhan: Version here also needs update

config/webhook/manifests.yaml
```diff
@@ -46,6 +46,26 @@ webhooks:
...
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /default-mutate-vmoperator-vmware-com-v1alpha4-virtualmachinegrouppublishrequest
+  failurePolicy: Fail
+  name: default.mutating.virtualmachinegrouppublishrequest.v1alpha4.vmoperator.vmware.com
```
lubronzhan: Version here needs update

config/webhook/manifests.yaml
```diff
@@ -46,6 +46,26 @@ webhooks:
...
+  rules:
+  - apiGroups:
+    - vmoperator.vmware.com
+    apiVersions:
+    - v1alpha4
```
lubronzhan: Version here

config/webhook/manifests.yaml
```diff
@@ -157,6 +177,27 @@ webhooks:
...
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /default-validate-vmoperator-vmware-com-v1alpha4-virtualmachinegrouppublishrequest
+  failurePolicy: Fail
+  name: default.validating.virtualmachinegrouppublishrequest.v1alpha4.vmoperator.vmware.com
```
lubronzhan: Version here

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/vmware-tanzu/vm-operator/pkg/builder"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/webhooks/common"
+)
+
+const (
+	webHookName = "default"
+)
+
+// +kubebuilder:webhook:path=/default-mutate-vmoperator-vmware-com-v1alpha4-virtualmachinegrouppublishrequest,mutating=true,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachinegrouppublishrequests,verbs=create,versions=v1alpha4,name=default.mutating.virtualmachinegrouppublishrequest.v1alpha4.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
```
lubronzhan: version here

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,130 @@
...
+			if apierrors.IsNotFound(err) {
+				return admission.Errored(http.StatusNotFound, err)
+			}
+			return admission.Errored(http.StatusInternalServerError, err)
+		}
+		if vmSet.Len() == 0 {
+			return admission.Errored(http.StatusBadRequest, fmt.Errorf("%q group does not have any memebers", groupPubReq.Spec.Source))
+		}
+
+		// ensure vms are unique and sorted when it is set by the mutator
```
lubronzhan, line 116: Now it's not sorted

webhooks/virtualmachinegrouppublishrequest/validation/virtualmachinegrouppublishrequest_validator.go
```diff
@@ -0,0 +1,179 @@
...
+	oldGroupPubReq, err := v.vmGroupPublishRequestFromUnstructured(ctx.OldObj)
+	if err != nil {
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
+
+	return common.BuildValidationResponse(ctx, nil,
+		common.ConvertFieldErrorsToStrings(v.validateUpdateSpec(groupPubReq, oldGroupPubReq)), nil)
+}
+
+// validateUpdateSpec ensures no updates to spce immutable fields.
```
lubronzhan:
```suggestion
// validateUpdateSpec ensures no updates to spec's immutable fields.
```

webhooks/virtualmachinepublishrequest/validation/virtualmachinepublishrequest_validator.go
```diff
@@ -165,6 +173,43 @@ func (v validator) validateTargetLocation(ctx *pkgctx.WebhookRequestContext, vmp
...
+
+	for _, ownerRef := range vmPub.OwnerReferences {
+		if ownerRef.Kind == "VirtualMachineGroupPublishRequest" &&
+			ownerRef.Name == vmPub.Labels[vmopv1.VirtualMachinePublishRequestManagedByLabelKey] {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("%q label is reserved", vmopv1.VirtualMachinePublishRequestManagedByLabelKey)
```
lubronzhan: It would be helpful to show what's described in the function comment. There is no `VirtualMachineGroupPublishRequest` in owner ref

webhooks/common/utils.go
lubronzhan: Yeah there is no conflict detected if the change is in the new file. @zhangheliu  better do a search on `v1alpha4` in the github file change page. Since some change is in yaml or comment, might be missed

webhooks/common/utils.go
zhangheliu: addressed, it was in webhook util folder since it was only used by webhooks.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: list returns a sorted slice,  we keep user's input order if it is not empty but sort the VMs if we are getting all members.

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -308,6 +308,9 @@ func (r *Reconciler) reconcileNewPublishRequests(
 						Kind:       "ContentLibrary",
 						APIVersion: clAPIVersion,
 					},
+					Item: vmopv1.VirtualMachinePublishRequestTargetItem{
+						Name: ctx.VMGroupPublishRequest.Name + "-" + sourceVM,
```
dilyar85, line 312: I recalled  we discussed using the namespace name to ensure uniqueness. Was that changed?

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,131 @@
...
+		groupPubReq.Spec.Source = groupPubReq.Name
+	}
+
+	if groupPubReq.Spec.Target == "" {
+		defaultCL, err := common.RetrieveDefaultImagePublishContentLibrary(ctx, m.client, groupPubReq.Namespace)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return admission.Errored(http.StatusNotFound, fmt.Errorf("cannot find a default content library with the %q label", common.DefaultImagePublishContentLibraryLabelKey))
+			}
+			return admission.Errored(http.StatusInternalServerError, err)
```
dilyar85, line 97: I don't think we should return the 500 if there're more than one default CLs found.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if len(groupPubReq.Spec.VirtualMachines) == 0 {
+		visitedGroups := &sets.Set[string]{}
+		vmSet, err := vmopv1util.RetrieveVMGroupMembers(ctx, m.client,
+			ctrlclient.ObjectKey{Namespace: groupPubReq.Namespace, Name: groupPubReq.Spec.Source}, visitedGroups)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return admission.Errored(http.StatusNotFound, err)
+			}
+			return admission.Errored(http.StatusInternalServerError, err)
```
dilyar85, line 110: Ditto. This should be something like bad request since the error could be returned by duplicate or loop members in the VM group?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
zhangheliu: Yeah, VCFA folks wanted to have this format since they could set the VMGroupPublishRequest name to be unique if needed.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: yeah, good catch.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: hmm, this depends on what the VM group API is intended to do. Is the VM group API ensuring that a duplicated member/looped group should never have the status group-linked set to true? If so, I think this should return an internal error.

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
zhangheliu: addressed. thanks~

webhooks/virtualmachinegrouppublishrequest/mutation/virtualmachinegrouppublishrequest_mutator.go
dilyar85: Yep, a VM Group should never have duplicate members in its status (it's not enforced yet but will be soon). What you said makes sense. Resolving the comment.

PR #1095, opened 2025-08-14
:sparkles: Support zonal affinity policy between VMs

aruneshpa:

This change introduces support for specifying zonal affinity policies between VMs using the VM's spec.affinity fields.

Two VMs can specify zonal affinity with each other by specifying a labelSelector.matchLabels in the affinity spec.  VM operator then constructs the placement spec for the VMs that includes these affinity policies before requesting DRS for placement recommendations.  DRS now includes these affinity constraints when evaluating placement for VMs.

This has been validated in a stretched environment by ensuring that two VMs that specify zonal affinity are placed in the same zone.

```release-note
Support zonal affinity policy between VMs
```


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2515,6 +2522,57 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpec(
 	return nil
 }
 
+func (vs *vSphereVMProvider) vmCreateGenConfigSpecAffinityPolicy(
```
bryanv: I don't think this needs to be a provider method. I was kind of expecting something in the `virtualmachine` package along side `CreateConfigSpec()` (or just do it as a part of that function) so we can more easily test that the expected ConfigSpec is being created.

aruneshpa: I started with putting it in the virtualmachine package, but it seems we have a pattern of taking the generated ConfigSpec and modifying it to add other things. 
Like, these methods:
https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmprovider_vm.go#L2503-L2513

pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -148,9 +151,69 @@ func CreateConfigSpec(
...
+							PolicyStrictness:  string(vimtypes.VmPlacementPolicyVmPlacementPolicyStrictnessPreferredDuringPlacementIgnoredDuringExecution),
+							PolicyTopology:    string(vimtypes.VmPlacementPolicyVmPlacementPolicyTopologyVSphereZone),
+						})
+					}
+				}
+			}
+		}
+	}
+
+	configSpec.VmPlacementPolicies = placementPols
```
dilyar85: Should we set this only if `placementPols` is not empty?

pkg/providers/vsphere/virtualmachine/configspec_test.go
```diff
@@ -192,6 +194,46 @@ var _ = Describe("CreateConfigSpec", func() {
...
+			)
+			BeforeEach(func() {
+				pkgcfg.SetContext(vmCtx, func(config *pkgcfg.Config) {
+					config.Features.VMPlacementPolicies = true
+				})
+
+				labelKey = "node-pool"
+				labelValue = "node-pool-1"
+
+				vmCtx.VM.Spec.Affinity = &vmopv1.VirtualMachineAffinitySpec{
```
dilyar85, line 217: Add a test for `VMAffinity.PreferredDuringSchedulingIgnoredDuringExecution` (or just include it in the current VM.Spec.Affinity) to verify that `PolicyStrictness` is set properly as well?

pkg/providers/vsphere/virtualmachine/configspec.go
aruneshpa: I wasn't checking for that because this is an option field.  vCenter will just drop it during Reconfigure.  But since it is confusing to readers, I am going to add the length check.

pkg/providers/vsphere/virtualmachine/configspec_test.go
aruneshpa: Added a test.

pkg/providers/vsphere/virtualmachine/configspec_test.go
```diff
@@ -5,10 +5,11 @@
 package virtualmachine_test
 
 import (
-	"context"
+	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
bryanv, line 14: Sort down below

pkg/providers/vsphere/virtualmachine/configspec_test.go
```diff
@@ -192,6 +194,82 @@ var _ = Describe("CreateConfigSpec", func() {
...
+				pkgcfg.SetContext(vmCtx, func(config *pkgcfg.Config) {
+					config.Features.VMPlacementPolicies = true
+				})
+
+				reqLabelKey = "node-pool"
+				reqLabelValue = "node-pool-1"
+				prefLabelKey = "node-pool"
+				prefLabelValue = "node-pool-2"
+
+				vmCtx.VM.Spec.Affinity = &vmopv1.VirtualMachineAffinitySpec{
```
bryanv, line 217: In one of the existing basic tests assert that this is `nil`

PR #1097, opened 2025-08-14
:seedling: Prevent snapshot of VKS nodes

aruneshpa:

This change modifies the VM snapshot webhook to reject snapshot requests on VKS nodes.  This is done by checking for CAPI added labels on the VKS nodes which is already a contract / pattern that is followed in VM operator for other workflows (e.g., backup).

Since webhook validation based on labels is always race-y, this change also modifies the snapshot reconciliation workflow in the provider to skip taking snapshot of VKS nodes.

```release-note
Prevent snapshot of VKS nodes
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -1262,6 +1262,13 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 		vmCtx.Logger.V(5).Info("No current snapshot specified, skipping revert")
 		return false, nil
 	}
+
+	// Skip snapshot revert processing for VKS/TKG nodes
+	if kubeutil.HasCAPILabels(vmCtx.VM.Labels) {
+		vmCtx.Logger.V(4).Info("Skipping snapshot revert for VKS/TKG node",
+			"vmName", vmCtx.VM.Name, "snapshotName", vmCtx.VM.Spec.CurrentSnapshot.Name)
+		return false, nil
+	}
 	desiredSnapshotName := vmCtx.VM.Spec.CurrentSnapshot.Name
 
 	// Check if the snapshot CR exists.
@@ -1524,6 +1531,13 @@ func (vs *vSphereVMProvider) reconcileSnapshot(
 
 	// If no snapshots found, nothing to do
 	if len(vmSnapshots) == 0 {
+		return nil
+	}
+
+	// Skip snapshot processing for VKS/TKG nodes
+	if kubeutil.HasCAPILabels(vmCtx.VM.Labels) {
+		vmCtx.Logger.V(4).Info("Skipping snapshot processing for VKS/TKG node",
+			"vmName", vmCtx.VM.Name)
 		return nil
 	}
 
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -1927,6 +1927,11 @@ func (v validator) validateSnapshot(
 		allErrs = append(allErrs, field.Required(snapshotPath.Child("name"), ""))
 	}
 
+	// Check if the VM is a VKS node and prevent snapshot revert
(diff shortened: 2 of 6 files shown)


webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -19,8 +19,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+	vmopv1common "github.com/vmware-tanzu/vm-operator/api/v1alpha4/common"
```
lubronzhan: v1alpha5

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1262,6 +1262,13 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 		vmCtx.Logger.V(5).Info("No current snapshot specified, skipping revert")
 		return false, nil
 	}
+
+	// Skip snapshot revert processing for VKS/TKG nodes
+	if kubeutil.HasCAPILabels(vmCtx.VM.Labels) {
+		vmCtx.Logger.V(4).Info("Skipping snapshot revert for VKS/TKG node",
```
bryanv, line 1268: This logger should already have the VM name. Is there like a condition we could report this on instead?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1927,6 +1927,11 @@ func (v validator) validateSnapshot(
 		allErrs = append(allErrs, field.Required(snapshotPath.Child("name"), ""))
 	}
 
+	// Check if the VM is a VKS/TKG node and prevent snapshot revert
+	if kubeutil.HasCAPILabels(vm.Labels) {
+		allErrs = append(allErrs, field.Forbidden(snapshotPath, "snapshot revert is not allowed for VKS/TKG nodes"))
```
bryanv: AFAIK this is just VKS nowadays

lubronzhan (approved):
Just a small issue. Everything else looks good!

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -136,6 +143,30 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
 	return common.BuildValidationResponse(ctx, nil, validationErrs, nil)
 }
 
+// validateVMNotVKSNode checks if the referenced VM is a VKS/TKG node and returns an error if it is.
+func (v validator) validateVMNotVKSNode(ctx *pkgctx.WebhookRequestContext, vmRef *vmopv1common.LocalObjectRef, namespace string) error {
```
bryanv, line 147: How about just in the snapshot controller?

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Hmm.  If we report a condition like that, when would we clear it?  It will be a one time condition that will stick, which isn't nice?
And since the VKS VMs are always created with these labels, we would never get to this code anyway since the webhook would prevent the snapshot.  I added this as an additional safety check.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
aruneshpa: I think there's value in keeping this in the webhooks since this is going to be pretty deterministic.  Let me know if you disagree.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Merging this after our conversation.  We can revisit this later if it turns out to be an issue.

PR #1099, opened 2025-08-14
🌱 Enqueue reconcile between groups and linked members

dilyar85:

This PR has the following two changes:

1. Add a new mapper func to trigger reconciliation of VMs when their group changes
   - ensuring VM status is updated with the group's linked condition as soon as it is adopted by a group
   - allowing immediate placement of the VM when the group has placement result ready

2. Updated the existing mapper func to trigger reconciliation of VM Groups when a linked member is updated
   - ensuring VM Group's member status reflects the latest state of its members

Fixes N/A.

None.

```release-note
Enqueue reconcile between group to linked VM members.
```

### DIFF
--- a/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
+++ b/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
@@ -63,9 +63,9 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 	return ctrl.NewControllerManagedBy(mgr).
 		For(controlledType).
 		Watches(&vmopv1.VirtualMachineGroup{},
-			handler.EnqueueRequestsFromMapFunc(vmGroupToParentGroupMapperFn())).
+			handler.EnqueueRequestsFromMapFunc(MemberToGroupMapperFn(ctx))).
 		Watches(&vmopv1.VirtualMachine{},
-			handler.EnqueueRequestsFromMapFunc(vmToParentGroupMapperFn())).
+			handler.EnqueueRequestsFromMapFunc(MemberToGroupMapperFn(ctx))).
 		WithOptions(controller.Options{
 			MaxConcurrentReconciles: ctx.MaxConcurrentReconciles,
 			LogConstructor:          pkgutil.ControllerLogConstructor(controllerNameShort, controlledType, mgr.GetScheme()),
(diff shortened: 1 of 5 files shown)


controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -300,7 +290,7 @@ func (r *Reconciler) reconcileMembers(
 		for _, member := range bootOrder.Members {
 			key := member.Kind + "/" + member.Name
 
-			// Check if we have an existing status to update, or create a new one.
+			// Check if we have an existing status to update, or create new one.
```
dilyar85, line 293: Just keeping it under the 80-char limit ;)

pkg/util/vmopv1/vm.go
```diff
@@ -390,6 +390,58 @@ func EncryptionClassToVirtualMachineMapper(
...
+			group     = o.(*vmopv1.VirtualMachineGroup)
+			namespace = group.Namespace
+			requests  = make([]reconcile.Request, 0, len(group.Status.Members))
+		)
+
+		for _, m := range group.Status.Members {
+			if m.Kind == "VirtualMachine" &&
+				conditions.IsTrue(&m, vmopv1.VirtualMachineGroupMemberConditionGroupLinked) {
+				// Only trigger a reconcile if the VM condition doesn't have
+				// the group linked condition true, or if the VM is not placed.
```
bryanv: Should we also check if the VirtualMachineGroupMemberConditionPlacementReady is set for the member VM?

pkg/util/vmopv1/vm.go
```diff
@@ -390,6 +390,58 @@ func EncryptionClassToVirtualMachineMapper(
...
+				vmName := m.Name
+				vm := &vmopv1.VirtualMachine{}
+				key := client.ObjectKey{Namespace: namespace, Name: vmName}
+				if err := k8sClient.Get(ctx, key, vm); err != nil {
+					continue
+				}
+
+				if !conditions.IsTrue(vm, vmopv1.VirtualMachineGroupMemberConditionGroupLinked) ||
```
aruneshpa, line 426: Should we also skip VMs that are being deleted, or do we not care?  I am split because we are listing those  VMs anyway so it doesn't save us much except a reconcile request.

pkg/util/vmopv1/vm.go
dilyar85: I didn't include that condition check because a VM Group can update to adopt an existing (already placed) VM member which will not have that condition in the VM Group member status. I think we should still trigger a reconciliation in that case to update the group linked condition in the VM status immediately when the group adopts the VM.

pkg/util/vmopv1/vm.go
dilyar85: Yeah probably not worth it as the triggered reconcile request simply adds deletion timestamp and returns.

pkg/util/vmopv1/vm.go
bryanv: OK. For adoption cases should we check if the VM's Spec.GroupName is for this group?

dilyar85: > OK. For adoption cases should we check if the VM's Spec.GroupName is for this group?

That's what `conditions.IsTrue(&m, vmopv1.VirtualMachineGroupMemberConditionGroupLinked)` is for :)
The condition will be true only if the VM.Spec.GroupName has the same group name set.

dilyar85: Spoke with @bryanv offline. There could be an edge case where the VM.Spec.GroupName is updated to a new group, while the old group hasn't been reconciled to update its member status group linked condition to false. Updated the function to also check for VM.Spec.GroupName to address this.

bryanv (approved):

PR #1102, opened 2025-08-14
🌱 Patch vm status with snapshot info

lubronzhan:
Path VM's status with currentSnapshot and rootSnapshots info.

currentSnapshot has been taken care of before, mostly add logic about root Snapshot here

```release-note
None
```

### DIFF
--- a/pkg/providers/vsphere/vmlifecycle/update_status.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status.go
@@ -125,6 +125,9 @@ func ReconcileStatus(
 		if err := updateCurrentSnapshotStatus(vmCtx, k8sClient); err != nil {
 			errs = append(errs, err)
 		}
+		if err := updateRootSnapshots(vmCtx, k8sClient); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
 	zoneName := vm.Labels[topology.KubernetesTopologyZoneLabelKey]
@@ -1277,4 +1280,64 @@ func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, tar
 		}
 	}
 	return ""
+}
+
+// updateRootSnapshots updates the VM status to reflect the
+// root snapshots on the VM.
+func updateRootSnapshots(vmCtx pkgctx.VirtualMachineContext, k8sClient ctrlclient.Client) error {
+	vmCtx.Logger.V(4).Info("Updating root snapshots in VM status")
+
+	vm := vmCtx.VM
+	// If VM has no snapshots, clear the root snapshots from the status.
+	if vmCtx.MoVM.Snapshot == nil {
+		if len(vm.Status.RootSnapshots) > 0 {
+			vmCtx.Logger.V(4).Info("VM has no snapshots, clearing status.rootSnapshots")
+			vm.Status.RootSnapshots = nil
+		}
+		return nil
+	}
+
+	// If VM has no root snapshots, clear the root snapshots from the status.
+	if len(vmCtx.MoVM.Snapshot.RootSnapshotList) == 0 {
+		if len(vm.Status.RootSnapshots) > 0 {
+			vmCtx.Logger.V(4).Info("VM has no root snapshots, clearing status.rootSnapshots")
+			vm.Status.RootSnapshots = nil
+		}
+		return nil
+	}
+
+	// Refresh the root snapshots from the VM mo
+	var newRootSnapshots []common.LocalObjectRef //nolint:prealloc
+	for _, rootSnapshot := range vmCtx.MoVM.Snapshot.RootSnapshotList {
+		objKey := ctrlclient.ObjectKey{
+			Name:      rootSnapshot.Name,
+			Namespace: vm.Namespace,
+		}
+
+		rootSnapshotCR := &vmopv1.VirtualMachineSnapshot{}
+		if err := k8sClient.Get(vmCtx, objKey, rootSnapshotCR); err != nil {
+			if !apierrors.IsNotFound(err) {
+				vmCtx.Logger.Error(err, "Failed to get VirtualMachineSnapshot custom resource",
+					"snapshotName", rootSnapshot.Name)
+				return err
+			}
+			// TODO (lubron) Snapshot custom resource doesn't exist, but the snapshot
+			// could be created by admin on the vSphere. We will reference these kind
+			// of snapshots in other way, like externally created snapshots, so the snapshot
+			// graph reflects the actual state of the VM in vSphere.
+			vmCtx.Logger.V(4).Info("VirtualMachineSnapshot custom resource not found, skipping",
+				"snapshotName", rootSnapshot.Name)
+			continue
+		}
+
+		newRootSnapshots = append(newRootSnapshots, common.LocalObjectRef{
+			APIVersion: rootSnapshotCR.APIVersion,
+			Kind:       rootSnapshotCR.Kind,
+			Name:       rootSnapshotCR.Name,
+		})
+	}
+
+	vm.Status.RootSnapshots = newRootSnapshots
+
+	return nil
 }
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -1300,11 +1300,8 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 
 	vmCtx.Logger.V(4).Info("Found desired snapshot for the VM", "snapshotRef", snapObj.Reference().Value)
 
-	// Check if we need to revert (current snapshot != desired snapshot)
-	if needsRevert := vs.needsSnapshotRevert(vmCtx, snapObj); !needsRevert {
-		vmCtx.Logger.V(4).Info("VM already points to the desired snapshot",
-			"snapshotName", desiredSnapshotName)
-		return false, nil
+	if vmCtx.MoVM.Snapshot.CurrentSnapshot.Value == snapObj.Reference().Value {
+		vmCtx.Logger.V(4).Info("Reverting to the current snapshot")
 	}
 
 	vmCtx.Logger.Info("Starting snapshot revert operation",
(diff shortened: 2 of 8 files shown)


pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -125,6 +125,9 @@ func ReconcileStatus(
 		if err := updateCurrentSnapshotStatus(vmCtx, k8sClient); err != nil {
 			errs = append(errs, err)
 		}
+		if err := updateRootSnapshots(vmCtx, k8sClient); err != nil {
```
aruneshpa, line 128: Why not handle current snapshot here in the same change as well?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1278,3 +1281,69 @@ func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, tar
...
+	// If VM has no snapshots, clear the status.
+	if vmCtx.MoVM.Snapshot == nil {
+		if len(vm.Status.RootSnapshots) > 0 {
+			vmCtx.Logger.V(4).Info("VM has no snapshots, clearing status.rootSnapshots")
+			vm.Status.RootSnapshots = nil
+		}
+		return nil
+	}
+
+	// If VM has no root snapshots, clear the status.
```
aruneshpa: Here and above:
```suggestion
	// If VM has no root snapshots, clear the root snapshots from the status.
```

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1278,3 +1281,69 @@ func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, tar
...
+// root snapshots on the VM.
+func updateRootSnapshots(vmCtx pkgctx.VirtualMachineContext, k8sClient ctrlclient.Client) error {
+	vmCtx.Logger.V(4).Info("Updating root snapshots in VM status")
+
+	vm := vmCtx.VM
+	// If VM has no snapshots, clear the status.
+	if vmCtx.MoVM.Snapshot == nil {
+		if len(vm.Status.RootSnapshots) > 0 {
+			vmCtx.Logger.V(4).Info("VM has no snapshots, clearing status.rootSnapshots")
+			vm.Status.RootSnapshots = nil
```
aruneshpa, line 1295: Why not set the current snapshots to nil / empty here as well?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1278,3 +1281,69 @@ func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, tar
...
+		}
+		return nil
+	}
+
+	// If VM has no root snapshots, clear the status.
+	if len(vmCtx.MoVM.Snapshot.RootSnapshotList) == 0 {
+		if len(vm.Status.RootSnapshots) > 0 {
+			vmCtx.Logger.V(4).Info("VM has no root snapshots, clearing status.rootSnapshots")
+			vm.Status.RootSnapshots = nil
```
aruneshpa, line 1304: If the VM doesn't have any root snapshots, then it can't have a current snapshot.  You can set the current snapshot to empty.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1278,3 +1281,69 @@ func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, tar
...
+		}
+
+		rootSnapshotCR := &vmopv1.VirtualMachineSnapshot{}
+		if err := k8sClient.Get(vmCtx, objKey, rootSnapshotCR); err != nil {
+			if !apierrors.IsNotFound(err) {
+				vmCtx.Logger.Error(err, "Failed to get VirtualMachineSnapshot custom resource",
+					"snapshotName", rootSnapshot.Name)
+				return err
+			}
+			// Snapshot custom resource doesn't exist, skip it.
```
aruneshpa: Can you add a TODO here?  We talked about this the other day where this can be an _externally_ created snapshot that we need to handle.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1278,3 +1281,69 @@ func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, tar
...
+			vmCtx.Logger.V(4).Info("VirtualMachineSnapshot custom resource not found, skipping",
+				"snapshotName", rootSnapshot.Name)
+			continue
+		}
+
+		// If the snapshot is being deleted, don't update the status.
+		if !rootSnapshotCR.DeletionTimestamp.IsZero() {
```
aruneshpa: Why do we care about the deletion timestamp here?  When the object is deleted, the deletion will automatically update the tree anyway?
As a principal, the vCenter and Supervisor snapshot trees for a VM must match as much as possible.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1278,3 +1281,69 @@ func FindSnapshotNameInTree(snapshots []vimtypes.VirtualMachineSnapshotTree, tar
...
+		}
+
+		newRootSnapshots = append(newRootSnapshots, common.LocalObjectRef{
+			APIVersion: rootSnapshotCR.APIVersion,
+			Kind:       rootSnapshotCR.Kind,
+			Name:       rootSnapshotCR.Name,
+		})
+	}
+
+	if len(newRootSnapshots) > 0 {
```
aruneshpa: Does the length check matter?  Since this is an optional field, it will be skipped / omitted from the patch automatically IIRC.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1302,8 +1302,9 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 
 	// Check if we need to revert (current snapshot != desired snapshot)
 	if needsRevert := vs.needsSnapshotRevert(vmCtx, snapObj); !needsRevert {
-		vmCtx.Logger.V(4).Info("VM already points to the desired snapshot",
+		vmCtx.Logger.V(4).Info("VM already points to the desired snapshot, clearing the spec.currentSnapshot",
 			"snapshotName", desiredSnapshotName)
+		vmCtx.VM.Spec.CurrentSnapshot = nil
```
aruneshpa: Hmm.  We shouldn't be clearing this field.  Instead, we need to remove this entire method / check and let revert go through if the observed current snapshot matches requested current snapshot.  This is the "revert a VM to the current snapshot workflow".

I can handle this in a followup.

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: There is another function that handles currentSnapshot.
https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmlifecycle/update_status.go#L1177-L1202

We could merge these two funcs. But I keep them separate for clarity

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: The currentSnapshot is taken care of here.
There is another function that handles currentSnapshot.
[main/pkg/providers/vsphere/vmlifecycle/update_status.go#L1177-L1202](https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmlifecycle/update_status.go#L1177-L1202)

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: We have this logic check in https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmlifecycle/update_status.go#L1177-L1202 as well.
Do you want me to change that as well?

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: If not checked, `vm.Status.RootSnapshots` will be assigned with an empty array `[]common.LocalObjectRef{}`, instead of nil. Just want to make the format of empty `vm.Status.RootSnapshots` consistent against above code `vm.Status.RootSnapshots = nil`.
So our test code could just check whether vm.Status.RootSnapshots is nil, instead of also need to check if it's empty `[]common.LocalObjectRef{}`.

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: Oh yeah you are right. Let me just modify it here, should be a simple change

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: It's right above the func I added `updateCurrentSnapshotStatus `

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: Let me remove it and add `//nolint:prealloc` so lint won't complain

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1300,11 +1300,12 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 
 	vmCtx.Logger.V(4).Info("Found desired snapshot for the VM", "snapshotRef", snapObj.Reference().Value)
 
-	// Check if we need to revert (current snapshot != desired snapshot)
-	if needsRevert := vs.needsSnapshotRevert(vmCtx, snapObj); !needsRevert {
-		vmCtx.Logger.V(4).Info("VM already points to the desired snapshot",
-			"snapshotName", desiredSnapshotName)
-		return false, nil
+	// Log a message if the VM already points to the current snapshot, but new change
+	// might be added to the current VM, so we want to continue with the revert.
```
aruneshpa: This is a very common workflow.  To the point where VC has an API for this.  Just say something like:
"Reverting to the current snapshot"

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: Will follow up in a different PR

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1300,11 +1300,9 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
...
 	vmCtx.Logger.V(4).Info("Found desired snapshot for the VM", "snapshotRef", snapObj.Reference().Value)
 
-	// Check if we need to revert (current snapshot != desired snapshot)
-	if needsRevert := vs.needsSnapshotRevert(vmCtx, snapObj); !needsRevert {
-		vmCtx.Logger.V(4).Info("VM already points to the desired snapshot",
-			"snapshotName", desiredSnapshotName)
-		return false, nil
+	if vmCtx.MoVM.Snapshot != nil && vmCtx.MoVM.Snapshot.CurrentSnapshot != nil &&
+		vmCtx.MoVM.Snapshot.CurrentSnapshot.Value == snapObj.Reference().Value {
+		vmCtx.Logger.V(4).Info("Reverting to the current snapshot")
```
bryanv, line 1304: Did you really intent to get rid of the return? If so, you're checking things here for nil that are checked later on

lubronzhan: Yes, I intend to get rid of it. Since ideally we should support revert to current snapshot. Then in that case the previous `needsSnapshotRevert` func would always return true, thus removing the return here. Since it's not necessary any more.

I should remove the check for `vmCtx.MoVM.Snapshot != nil && vmCtx.MoVM.Snapshot.CurrentSnapshot != nil` since these are impossible after running `virtualmachine.FindSnapshot` and find the target snapshot, the code shouldn't be able to reach here if `vmCtx.MoVM.Snapshot == nil || vmCtx.MoVM.Snapshot.CurrentSnapshot == nil`

PR #1107, opened 2025-08-18
🐛 Propagate the config context into the prober

bryanv:

If after a VC creds rotation - which will close the existing VC client - the prober is the first to try to create a new VC client, it will panic without the Config in the context so pass down that context into the probe context.

Fixes #

```release-note
NONE
```


dilyar85 (approved):
LGTM!

lubronzhan (approved):

PR #1109, opened 2025-08-18
🌱 Patch remaining VMSnapshot status and some refactor

lubronzhan:
1. This update parent snapshot's children list after a new snapshot is taken.
2. Update power state status of snapshot
1. Refactor:
     1. Add a new func `SyncVMSnapshotTreeStatus` in vmlifecycle that updates currentSnapshot and rootSnapshots of a VM according to the state of moVM. This func is called when snapshot is `create/revert/delete`, and called in reconcileStatus
     1. Export `SyncVMSnapshotTreeStatus ` in vmprovider API so that vmsnapshot controller could use it
             1. Cleaned up some tests in vmsnapshot controller since it doesn't have the logic to update  currentSnapshot and rootSnapshots according to different tree topology.
     1. Move snapshot related vmprovider funcs to a new file vmprovider_vm_snapshot.go, as well as their tests
     1. Refactor `GetParentSnapshot` so that it make govmomi call inside 

This PR also update vm's status.currentSnapshot when the snapshot is deleted. It shouldn't touch spec.currentSnapshot.

Current VM snapshot graph
image
What vm status and snapshot status shows

**VM**
```
    currentSnapshot:
      kind: VirtualMachineSnapshot
      name: sn-4
    rootSnapshots:
    - apiVersion: vmoperator.vmware.com/v1alpha5
      kind: VirtualMachineSnapshot
      name: sn-1
```
**sn-1**
```
status:
  children:
  - apiVersion: vmoperator.vmware.com/v1alpha5
    kind: VirtualMachineSnapshot
    name: sn-2
  - apiVersion: vmoperator.vmware.com/v1alpha5
    kind: VirtualMachineSnapshot
    name: sn-3
  conditions:
```
**sn-2**
```
status:
  children:
  - apiVersion: vmoperator.vmware.com/v1alpha5
    kind: VirtualMachineSnapshot
    name: sn-4
```

Now deleting **sn-1**
image
**VM** becomes
```
    currentSnapshot:
      kind: VirtualMachineSnapshot
      name: sn-4
    rootSnapshots:
    - apiVersion: vmoperator.vmware.com/v1alpha5
      kind: VirtualMachineSnapshot
      name: sn-2
    - apiVersion: vmoperator.vmware.com/v1alpha5
      kind: VirtualMachineSnapshot
      name: sn-3
```
**sn-3**
```
status:
  children:
  - apiVersion: vmoperator.vmware.com/v1alpha5
    kind: VirtualMachineSnapshot
    name: sn-4
```

```release-note
None
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -11,7 +11,6 @@ import (
 	"maps"
 	"math/rand"
 	"path"
-	"slices"
 	"sort"
 	"strings"
 	"sync"
@@ -1380,7 +1379,7 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevert(
 		"afterRestoreLabels", vmCtx.VM.Labels,
 		"afterRestoreSpecCurrentSnapshot", vmCtx.VM.Spec.CurrentSnapshot)
 
-	if err := vmlifecycle.UpdateCurrentSnapshotStatus(vmCtx, vs.k8sClient); err != nil {
+	if err := vmlifecycle.SyncVMSnapshotTreeStatus(vmCtx, vs.k8sClient); err != nil {
 		return true, err
 	}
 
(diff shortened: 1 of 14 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1627,6 +1628,12 @@ func (vs *vSphereVMProvider) reconcileSnapshot(
 		return err
 	}
 
+	// Update the children snapshot status of parent snapshot
+	if err = vs.patchChildrenSnapshotStatus(vmCtx, vs.k8sClient, snapshotToProcess); err != nil {
```
bryanv: You don't need to pass in the `vs.k8sClient` here to this method

For a later cleanup I don't think a lot of this really needs to be provider methods in the first place but moved into the vm utils or a specific snapshot utils files in this directory so that can be tested separately.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -3041,3 +3048,53 @@ func (vs *vSphereVMProvider) markSnapshotFailed(vmCtx pkgctx.VirtualMachineConte
...
 }
+
+// patchChildrenSnapshotStatus find the parent snapshot of current snapshot
+// and patches the children snapshot status of the parent snapshot by adding
+// the current snapshot to the children list of the parent snapshot.
+func (vs *vSphereVMProvider) patchChildrenSnapshotStatus(vmCtx pkgctx.VirtualMachineContext, k8sClient ctrlclient.Client,
+	vmSnapshot *vmopv1.VirtualMachineSnapshot) error {
+	// Use vsphereprovider's GetParentSnapshot since it will fetch latest
+	// version of moVM from vsphere, after a new snapshot was taken above.
+	parentSnapshot, err := vs.GetParentSnapshot(vmCtx, vmSnapshot.Name, vmCtx.VM)
```
bryanv: Don't do this. This will have to lookup the VM again, and the caller already has the `vcVM`. Just get the properties you need. `virtualmachine.GetParentSnapshot()` probably shouldn't be taking the vmCtx just to get the prop field that it needs out of that.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: This works.
But can you think through the implementation if we yank this block (and the patchSnapshotStatus) block from here and put it in update_status.go for a VM where we walk over the entire snapshot tree and patch these statuses?

When a snapshot, or revert happens, we will immediately requeue the request.  And the first thing we will do in the next reconcile is to patch the status, so should be fine?

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: Thanks. Forget `Properties` already makes the govmomi call

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: That should work. Let me try modify that

lubronzhan:
Did a big refactor by moving funcs of snapshot to a different file

PR #1111, opened 2025-08-19
✨ Build approximate VM spec when reverting to imported snapshots

palnabarun:

This commit introduces the ability to build an approximate VM spec when
reverting to imported snapshots. The spec is constructed based on the
snapshot's metadata and some fields from the state of the VM in VC,
allowing for a more accurate representation of the VM's state at the
time of the snapshot.

This is done so that the VM spec can reach to a state that likely defines
how the VM was configured when the Snapshot was taken.

NA

## Simulated Test scenario

1. Create a VM using `VirtualMachine` CR.
2. Take a snapshot of the VM from VC.
3. Create a `VirtualMachineSnapshot` CR pointing to the snapshot created in (2)
4. Made some changes to `VirtualMachine` CR.
5. Revert to the snapshot by setting `spec.currentSnapshot` in `VirtualMachine` CR.

The `VirtualMachine` CR should now have spec fields approximated from the workflow introduced in this PR.

```release-note
✨ Build approximate VM spec when reverting to imported snapshots
```


palnabarun:
To whoever is looking at the PR in future, the conversation moved here from #1075.

aruneshpa (approved):

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1466,16 +1469,141 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+			// currentSnapshot is only used to trigger a revert, so this must be empty.
+			CurrentSnapshot: nil,
+
+			// Group name is overridden to the VM's current group.
+			// We can't know (or guess) what the group the VM was when the snapshot was taken.
+			GroupName: vmCtx.VM.Spec.GroupName,
+
+			// When Snapshot is taken with Policy A and then VM is updated to use Policy B,
+			// after a revert to Snapshot with Policy A is performed, VM continues to have Policy A.
+			StorageClass: vmCtx.VM.Spec.StorageClass,
```
lubronzhan, line 1540: The `vmCtx.VM.Spec.StorageClass` seems to be the current VM's storageClass? Not like what the description says.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1466,16 +1469,141 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
...
+			interfaces := make([]vmopv1.VirtualMachineNetworkInterfaceSpec, len(vNICs))
+			for i := range vNICs {
+				interfaces[i] = vmopv1.VirtualMachineNetworkInterfaceSpec{
+					Name: fmt.Sprintf("eth%d", i),
+				}
+			}
+			vm.Spec.Network.Interfaces = interfaces
+		}
+
+		// MinHardwareVersion is set to the Current Hardware Version of the VM
```
lubronzhan, line 1589: Question, I assume moSnap stores hardware version of the snapshot when the snapshot is taken, other than the current hardware version of the VM?

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3210,8 +3210,103 @@ func vmTests() {
...
+							config.Features.VMImportNewNet = true
+						})
+					})
+
+					It("should successfully revert to desired snapshot and approximate the VM Spec", func() {
+						if vm.Labels == nil {
+							vm.Labels = make(map[string]string)
+						}
+
+						vm.Labels[kubeutil.CAPVClusterRoleLabelKey] = ""
```
lubronzhan, line 3242: What's the goal of this label? To skip reconcileSnapshot?

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3210,8 +3210,103 @@ func vmTests() {
...
+						Expect(ecMap).ToNot(HaveKey(backupapi.VMResourceYAMLExtraConfigKey))
+
+						// Create first snapshot in vCenter
+						task, err := vcVM.CreateSnapshot(ctx, vmSnapshot.Name, "first snapshot", false, false)
+						Expect(err).ToNot(HaveOccurred())
+						Expect(task.Wait(ctx)).To(Succeed())
+
+						// Create first snapshot CR
+						// Mark the snapshot as ready so that the snapshot workflow doesn't try to create it.
+						conditions.MarkTrue(vmSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition)
```
lubronzhan, line 3262: Are you trying to use `CAPVClusterRoleLabelKey ` to skip reconciling snapshot?

Here it doesn't work probably because when you create, it doesn't update the status, you need to call the status API separately to update resource's condition

e.g.:
```
cur := &vmopv1.VirtualMachineSnapshot{}
Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(vmSnapshot), cur)).To(Succeed())

conditions.MarkTrue(cur, vmopv1.VirtualMachineSnapshotReadyCondition)
Expect(ctx.Client.Status().Update(ctx, cur)).To(Succeed())
```

lubronzhan (commented):
I don't have enough VC related context to decide whether VMSpec config is expected, tried to review as much as I can. 

Overall looks good to me. Just a few questions.

pkg/providers/vsphere/vmprovider_vm.go
palnabarun: More context for this: https://github.com/vmware-tanzu/vm-operator/pull/1075#discussion_r2279921642

I have added another small note to the comment to clarify.

pkg/providers/vsphere/vmprovider_vm.go
palnabarun: This is coming from moSnap's VMConfigInfo.

pkg/providers/vsphere/vmprovider_vm_test.go
palnabarun: No, to skip backup.

pkg/providers/vsphere/vmprovider_vm_test.go
palnabarun: Right, the snapshot reconcillation is also paused when the label is set.

pkg/providers/vsphere/vmprovider_vm.go
lubronzhan: So it's the previous hardware version than to the current hardware version?

lubronzhan (approved):

PR #1113, opened 2025-08-20
:seedling: Skip reconcile if a VM snapshot revert is in progress

aruneshpa:

This change modifies the VM validation webhook to:
- Reject requests to modify spec.currentSnapshot if a revert operation is already underway.
- Skip and return from the reconcile if a snapshot revert request is in progress.  This is done by relying on the SnapshotInProgress annotation introduced in a previous commit.

If the snapshot revert fails due to some terminal error, user intervention may be required to remove the annotation to re-trigger the reconcile.

Testing Done:
- Other than unit and integration tests, validated the webhook change in a live env:
```
k apply --dry-run=server -f vm-invalid.yaml
secret/parunesh-vm-bootstrap-data configured (server dry run)
Error from server (spec.currentSnapshot: Forbidden: a snapshot revert
is already in progress):
```

```release-note
Skip reconcile if a VM snapshot revert is in progress
```


pkg/context/virtualmachine_context.go
```diff
@@ -76,6 +83,28 @@ func HasVMRunningTask(ctx context.Context, checkSharedLock bool) bool {
...
+		if t.State == vimtypes.TaskInfoStateRunning &&
+			slices.Contains(snapshotTaskDescriptionIDs, t.DescriptionId) {
+			return true
+		}
+	}
+	return false
+}
+
+// HasVMRunningSnapshotRevertTask returns true if the VM has a running snapshot revert task.
+func HasVMRunningSnapshotRevertTask(ctx context.Context) bool {
```
lubronzhan: This func is not used anywhere? Could be cleaned up.

pkg/context/virtualmachine_context_test.go
```diff
@@ -0,0 +1,254 @@
...
+				},
+			}
+			ctx = pkgctx.WithVMRecentTasks(ctx, tasks)
+
+			result := pkgctx.HasVMRunningSnapshotTask(ctx)
+			Expect(result).To(BeTrue())
+		})
+	})
+
+	Describe("HasVMRunningSnapshotRevertTask", func() {
```
lubronzhan: This could be cleaned up as well

lubronzhan (approved):
Overall looks good. Just a small comment.

Are the definitions for task names (e.g., VirtualMachine.revertToSnapshot) located in the Bora repository?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -857,6 +835,47 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+
+	//
+	// Reconcile a snapshot restore operation.
+	//
+	// Reverting a VM to a snapshot results in the VM's spec getting
+	// overwritten from the snapshot. For now, we do not allow any
+	// other changes to the VM's spec if a revert is being requested
+	// to avoid having to merge the spec post revert.
+	//
+	if pkgcfg.FromContext(vmCtx).Features.VMSnapshots {
```
lubronzhan, line 865: It's intentional to move the revert snapshot after backupState because of the test failure?

aruneshpa: Yes.  I want to ensure that the status update, and backup always happen.  When backup is done, that will update the ConfigSpec and a new reconcile will be queued by the async watcher.
The next reconcile will perform the revert.

bryanv (approved):

PR #1115, opened 2025-08-20
🌱 Move API conversion tests into sub-module

akutz:

This patch relocates the API conversion tests into a distinct sub-module under the existing API module. This reduces the API module's dependency graph.

Fixes `NA`

```release-note
Move API tests to the distinct sub-module ./api/test.
```


api/go.mod
```diff
@@ -4,41 +4,27 @@ go 1.24.0
...
-require github.com/onsi/ginkgo/v2 v2.23.4
-
 require (
-	github.com/google/go-cmp v0.7.0
-	github.com/google/uuid v1.6.0
-	github.com/onsi/gomega v1.36.3
-	k8s.io/api v0.33.0
-	k8s.io/apimachinery v0.33.0
+	k8s.io/api v0.31.0
+	k8s.io/apimachinery v0.31.0
```
dilyar85, line 9: Can we keep the new versions here (which would also be consistent with the root go.mod)?

akutz: Nope. I downgraded to make the API module more compatible. This was fully intentional.

PR #1119, opened 2025-08-21
:sparkles: Support VM-VM anti-affinity at Zonal level

aruneshpa:

This change introduces the support for specifying anti-affinity among VMs with respect to zones.  This allows a VM to specify an effective policy that can say "do not place this VM in the same zone with other VMs that have these labels".

One change to call out is that now we are sending in a list of tags that should be attached to the VMs even if a VM doesn't specify any affinity policy at all.  This is so that when we stat supporting dynamic policy evaluations, the tags specified by day-2 policies are already present on the VMs.

### Testing Done:
- Other than the unit tests here, I have evaluated this in a real environment where the anti-affinity policy is being respected.

```release-note
Support VM-VM anti-affinity at Zonal level
```


dilyar85 (approved):
LGTM!

pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -348,3 +387,75 @@ func initResourceAllocation(ap **vimtypes.ResourceAllocationInfo) {
...
+	for key, value := range labels {
+		// TODO: We should skip labels that are added by VM operator.
+		label := fmt.Sprintf("%s:%s", key, value)
+		tagsToAttach = append(tagsToAttach, label)
+	}
+
+	// Sort the labels to maintain consistent ordering.
+	slices.Sort(tagsToAttach)
+
+	// This is a little weird that we are sending all tags to the "first"
```
varunsl-brcm, line 457: Agreed that this is weird. We are working towards lifting `tagsToAttach` one level up in the config spec. Still in design phase. Will keep you updated if that changes.

aruneshpa: Thanks Varun.  I have filed a story to follow up once you have modified the API.

PR #1116, opened 2025-08-21
🌱 Add hardware/guest info to VM status

akutz:
**What does this PR do, and why is it needed?**

This patch adds new information to the VM status:

* `crypto.hasVTPM`
* `hardware.numCPUs`
* `hardware.cpuReservation`
* `hardware.memory`
* `hardware.memoryReservation`
* `hardware.nvidiaVGPUs`
* `guest.guestID`
* `guest.guestFullName`

The property `hardware.nvidiaVGPUs` is a slice of data structures that include the vGPU profile and the type of migration supported by the device:

* `None`
* `Normal` (standard vMotion)
* `Enhanced` (sub-second vMotion)

**Which issue(s) is/are addressed by this PR?** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes `NA`

**Are there any special notes for your reviewer**:

I still need to add unit tests around this.

**Please add a release note if necessary**:

```release-note
Add info about VM CPU/memory config, guest OS, vTPM, and Nvidia vGPUs to VM status.
```

PR #1122, opened 2025-08-22
✨Sync snapshot tree status

lubronzhan:

This is to sync the status of whole snapshot tree of a VM.
This func is called when a snapshot is create/revert/delete, as well as reconcileStatus.

Other than that, this also removes `GetParentSnapshot` from vmprovider interface since it's no longer used

Look for TODO, that's for handling the external snapshot

```release-note

```


pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1420,23 +1420,150 @@ func updateProbeStatusGuestInfo(
...
 	vmCtx pkgctx.VirtualMachineContext,
 	k8sClient ctrlclient.Client) error {
+	vmCtx.Logger.V(4).Info("Syncing snapshot tree status")
 
-	if err := updateCurrentSnapshotStatus(vmCtx, k8sClient); err != nil {
+	// Process the whole snapshot tree in updateSnapshotTreeChildrenStatus, and store
+	// the snapshot custom resources in this map.
+	// This map can be used in other functions later.
+	vmSnapshotMap := map[string]*vmopv1.VirtualMachineSnapshot{}
+	if err := updateSnapshotTreeChildrenStatus(vmCtx, k8sClient, vmSnapshotMap); err != nil {
```
palnabarun: We can remove the map since client.Get calls are cheap and served from cache. If we want to optimize VMOP <--> API Server calls, then it's okay to keep.

Otherwise, we can save ourselves some code complexity by fetching the VM Snapshot objects whenever we require them.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1432,6 +1439,114 @@ func SyncVMSnapshotTreeStatus(
...
+	vmCtx pkgctx.VirtualMachineContext,
+	k8sClient ctrlclient.Client) error {
+	mo := vmCtx.MoVM
+	if mo.Snapshot == nil || len(mo.Snapshot.RootSnapshotList) == 0 {
+		return nil
+	}
+
+	if mo.Snapshot == nil || len(mo.Snapshot.RootSnapshotList) == 0 {
+		return nil
+	}
```
palnabarun: Duplicate checks.

palnabarun (approved):
Other than the nit above, LGTM.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1432,6 +1439,114 @@ func SyncVMSnapshotTreeStatus(
...
+
+	vmSnapshot := &vmopv1.VirtualMachineSnapshot{}
+	objKey := ctrlclient.ObjectKey{
+		Name:      snapshotName,
+		Namespace: vmCtx.VM.Namespace,
+	}
+
+	if err := k8sClient.Get(vmCtx, objKey, vmSnapshot); err != nil {
+		if !apierrors.IsNotFound(err) {
+			vmCtx.Logger.Error(err, "Failed to get VirtualMachineSnapshot custom resource",
```
bryanv: Nit: I don't think including `custom resource` is really needed

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1480,19 +1595,14 @@ func updateCurrentSnapshotStatus(
 		"snapshotRef", currentSnapMoref.Value)
 
 	// Check if there's a VirtualMachineSnapshot custom resource with this name.
-	vmSnapshot := &vmopv1.VirtualMachineSnapshot{}
-	objKey := ctrlclient.ObjectKey{
-		Name:      snapshotName,
-		Namespace: vm.Namespace,
+	vmSnapshot, err := getSnapshotCR(vmCtx, k8sClient, snapshotName)
+	if err != nil {
+		vmCtx.Logger.Error(err, "Failed to get snapshot custom resource", "snapshotName", snapshotName)
```
bryanv: Log either here for in `getSnapshotCR()` but not both

pkg/util/kube/vmsnapshot.go
```diff
@@ -43,7 +46,7 @@ func CalculateReservedForSnapshotPerStorageClass(
 	if vmSnapshot.Spec.Memory {
 		// Fetch the requested memory size from the VMClass.
 		vmClass := &vmopv1.VirtualMachineClass{}
-		if err := k8sClient.Get(ctx, client.ObjectKey{Namespace: vmSnapshot.Namespace, Name: vm.Spec.ClassName}, vmClass); err != nil {
+		if err := k8sClient.Get(ctx, ctrlclient.ObjectKey{Namespace: vmSnapshot.Namespace, Name: vm.Spec.ClassName}, vmClass); err != nil {
```
bryanv, line 49: We things like resize this may not be what the VM is actually currently requested (nor its current amount of memory)

pkg/util/kube/vmsnapshot.go
```diff
@@ -140,3 +143,22 @@ func CalculateReservedForSnapshotPerStorageClass(
...
+	snapPatch := ctrlclient.MergeFrom(vmSnapshot.DeepCopy())
+	vmSnapshot.Status.UniqueID = snapMoRef.Reference().Value
+	vmSnapshot.Status.Quiesced = vmSnapshot.Spec.Quiesce != nil
+	vmSnapshot.Status.PowerState = vmCtx.VM.Status.PowerState
+
+	pkgcnd.MarkTrue(vmSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition)
+
+	if err := k8sClient.Status().Patch(vmCtx, vmSnapshot, snapPatch); err != nil {
+		return fmt.Errorf(
+			"failed to patch snapshot status resource %s/%s: err: %s",
```
bryanv: Use `%w`

pkg/util/kube/vmsnapshot_test.go
```diff
@@ -263,6 +267,94 @@ var _ = Describe("CalculateReservedForSnapshot", func() {
...
+		initObjects []ctrlclient.Object
+
+		vmCtx pkgctx.VirtualMachineContext
+	)
+
+	BeforeEach(func() {
+		vm := builder.DummyBasicVirtualMachine("test-vm", "dummy-ns")
+
+		vmCtx = pkgctx.VirtualMachineContext{
+			Context: pkgcfg.WithConfig(pkgcfg.Config{MaxDeployThreadsOnProvider: 16}),
```
bryanv: Why is `MaxDeployThreadsOnProvider` here?

pkg/util/kube/vmsnapshot_test.go
```diff
@@ -263,6 +267,94 @@ var _ = Describe("CalculateReservedForSnapshot", func() {
...
+				err := kubeutil.PatchSnapshotSuccessStatus(vmCtx, k8sClient, vmSnapshot, snapMoRef)
+				Expect(err).ToNot(HaveOccurred())
+
+				snapObj := &vmopv1.VirtualMachineSnapshot{}
+				Expect(k8sClient.Get(vmCtx, ctrlclient.ObjectKey{Name: vmSnapshot.Name, Namespace: vmSnapshot.Namespace}, snapObj)).To(Succeed())
+				Expect(snapObj.Status.UniqueID).To(Equal(snapMoRef.Value))
+				Expect(snapObj.Status.Quiesced).To(BeTrue())
+				Expect(snapObj.Status.PowerState).To(Equal(vmCtx.VM.Status.PowerState))
+				Expect(snapObj.Status.Conditions).To(HaveLen(1))
+				Expect(snapObj.Status.Conditions[0].Type).To(Equal(vmopv1.VirtualMachineSnapshotReadyCondition))
```
bryanv: Use either like the conditions `IsTrue()` or we have have condition Matcher.

pkg/providers/vsphere/vmprovider_vm_snapshot_test.go
```diff
@@ -31,324 +25,179 @@ import (
...
-
-		AfterEach(func() {
-			k8sClient = nil
-			initObjects = nil
+		When("VM is not found", func() {
+			BeforeEach(func() {
+				vm.Status.UniqueID = ""
+			})
+			It("should return error", func() {
+				Expect(vmProvider.SyncVMSnapshotTreeStatus(ctx, vm)).To(HaveOccurred())
```
bryanv: This reads weird - how about `ToNot(Succeed())`

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_intg_test.go
```diff
@@ -469,18 +437,8 @@ func intgTestsReconcile() {
...
-						Eventually(func(g Gomega) {
-							tmpParent := &vmopv1.VirtualMachineSnapshot{}
-							g.Expect(vcSimCtx.Client.Get(ctx, parentSSObjKey, tmpParent)).To(Succeed())
-							g.Expect(tmpParent).To(Not(BeNil()))
-							g.Expect(tmpParent.Status.Children).To(HaveLen(2))
-							g.Expect(tmpParent.Status.Children).To(ContainElement(*newLocalObjectRefWithSnapshotName(vmSnapshotL3Node1Name)))
-							g.Expect(tmpParent.Status.Children).To(ContainElement(*newLocalObjectRefWithSnapshotName(vmSnapshotL3Node2Name)))
-							g.Expect(tmpParent.Status.Children).ToNot(ContainElement(*newLocalObjectRefWithSnapshotName(vmSnapshotL2Name)))
-						}).Should(Succeed(), "waiting for parent snapshot's children to be updated")
+						// Skip checking parent snapshot's children because it is synced through fakeVMProvider.SyncVMSnapshotTreeStatusFn
```
bryanv: What do you mean? You can set the Fn to whatever capture variables, no?

pkg/util/kube/vmsnapshot_test.go
lubronzhan: Good catch, I just moved this test to here. Let me remove this and similar change.

pkg/util/kube/vmsnapshot.go
lubronzhan: Good point! 
I guess I can just use `VirtualMachineMemoryAllocationStatus` introduced by Andrew now? No need to check VMClass anymore
https://github.com/vmware-tanzu/vm-operator/pull/1116/files#diff-ffa3aff55a61266b1c1b6a3198947fea3d91ccd632acfefd52651be3cf8305c2R1168-R1178

lubronzhan: To unblock @palnabarun , I can update  this func in a follow up PR

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller_intg_test.go
lubronzhan: Yeah, but it also expect Fn to call k8sclient.Update( ) to update snapshotCR's status. But SyncVMSnapshotTreeStatusFn only accepts `ctx context.Context, vm *vmopv1.VirtualMachine`, it can't use the provider's builtin k8sClient. Later the checks just fetch the newest vmSnapshot CR using the k8sclient but the change cant't be detected because of that.

I haven't found other funcs in vmprovider interface that calls k8sClient that really test k8sClient related feature. Do you know other way I could test this?

lubronzhan: I guess I could try add k8sClient as a func to fakeVMProvider, and see if that works. Let me try

lubronzhan: NVM, I can call the vcSimCtx inside a Fn .. Didn't know I can do that

pkg/util/kube/vmsnapshot.go
lubronzhan: Created a Jira. Mark this one as done

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1432,6 +1439,110 @@ func SyncVMSnapshotTreeStatus(
...
+		}
+		// TODO (Lubron): If ref is nil, it means the snapshot custom resource is not found
+		// and it's an external snapshot.
+		if ref != nil {
+			children = append(children, *ref)
+		}
+	}
+
+	// Update Status.Children if changed.
+	if !reflect.DeepEqual(curSnapshot.Status.Children, children) {
```
aruneshpa: children is a slice, so it may not be ordered and thereforce, the deepequal may fail?

aruneshpa (approved):
Looks okay.

aruneshpa:
Make sure you also verify that other workflows like delete snapshot are modifying the children appropriately.

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: I checked the UI looks like the order is sorted on snapshot.value. In case the VC source code didn't do that, let me do this differently then

lubronzhan:
Let me verify with delete

lubronzhan:
Verified again

PR #1123, opened 2025-08-22
✨ vSphere Policy APIs and PolicyEvaluation controller

akutz:

This patch defines the vSphere Policy APIs and the PolicyEvaluation controller.

Fixes `NA`

```release-note
Define the vSphere Policy APIs.
```


external/vsphere-policy/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// +optional
+
+	// GuestFamily matches the workload's guest family.
+	// Valid values are: Darwin, Linux, Netware, Other, Solaris, and Windows.
+	GuestFamily *GuestFamilyMatcherSpec `json:"guestFamily,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Pod;VirtualMachine
+type WorkloadKind string
```
fstrudel, line 177: You already have WorkloadType (L40)?

external/vsphere-policy/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,235 @@
...
+	ValueSelectorOpEqual    ValueSelectorOperator = "Equal"
+	ValueSelectorOpNotEqual ValueSelectorOperator = "NotEqual"
+)
+
+type StringMatcherSpec struct {
+	// +optional
+	// +kubebuilder:default=Equal
+
+	// Op describes the operation performed against the specified value.
+	Op ValueSelectorOperator `json:"op,omitempty"`
```
fstrudel, line 139: Do we expect only Equal/NotEqual and no StartsWith for string in the first version?

external/vsphere-policy/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,235 @@
...
+	// Value describes the subject of the evaluation.
+	Value GuestFamilyType `json:"value,omitempty"`
+}
+
+type MatchGuestSpec struct {
+	// +optional
+
+	// GuestID matches the workload's guest ID.
+	//
+	// Please the following location for the value values:
```
fstrudel: Do you mean the  `Value` in the `GuestFamilyMatcherSpec`? (since there's no `Value` directly in that spec, not sure what "value values" refers to . Here and other places too)

external/vsphere-policy/api/v1alpha1/common_types.go
fstrudel: (and should that be a pointer if optional? Here and other places too)

external/vsphere-policy/api/v1alpha1/computepolicy_types.go
```diff
@@ -0,0 +1,99 @@
+// © Broadcom. All Rights Reserved.
```
fstrudel, line 1: Are you not using the standard copyright (with year) ? -> https://github-vcf.devops.broadcom.net/vcf/vmware-git_tools/blob/eac5adec3918d2b3e369ce29b11f72f1e49590ac/src/git_tools/update_copyrights.py#L71
```
// Copyright (c) 2025 Broadcom. All Rights Reserved.
// Broadcom Confidential. The term "Broadcom" refers to Broadcom Inc.
// and/or its subsidiaries.
```

external/vsphere-policy/api/v1alpha1/common_types.go
akutz: I do not use pointers for primitives when it is not necessary. I do not believe in this approach as it can lead to NPEs. In this case an empty string is perfectly fine as a zero value.

external/vsphere-policy/api/v1alpha1/common_types.go
akutz: It's a bad comment, will fix. Thanks @fstrudel !

external/vsphere-policy/api/v1alpha1/computepolicy_types.go
akutz: Because not all projects do (see GoVmomi) and this way we do not have to routinely update the year.

external/vsphere-policy/api/v1alpha1/common_types.go
akutz: > Do we expect only Equal/NotEqual and no StartsWith for string in the first version?

I can add `StartsWith`. Any other suggestions?

external/vsphere-policy/api/v1alpha1/common_types.go
akutz: Dammit. Thanks!

external/vsphere-policy/api/v1alpha1/common_types.go
akutz: Ah, it should read `valid values`.

external/vsphere-policy/api/v1alpha1/common_types.go
akutz: @fstrudel Voila!

```go
// +kubebuilder:validation:Enum=Equal;NotEqual;Contains;NotContains;HasPrefix;NotHasPrefix;HasSuffix;NotHasSuffix;Match;NotMatch
type ValueSelectorOperator string

const (
	// ValueSelectorOpEqual executes a == b.
	ValueSelectorOpEqual ValueSelectorOperator = "Equal"

	// ValueSelectorOpNotEqual executes a != b.
	ValueSelectorOpNotEqual ValueSelectorOperator = "NotEqual"

	// ValueSelectorOpHasPrefix executes strings.HasPrefix(a, b).
	ValueSelectorOpHasPrefix ValueSelectorOperator = "HasPrefix"

	// ValueSelectorOpNotHasPrefix executes !strings.HasPrefix(a, b).
	ValueSelectorOpNotHasPrefix ValueSelectorOperator = "NotHasPrefix"

	// ValueSelectorOpHasSuffix executes strings.HasSuffix(a, b).
	ValueSelectorOpHasSuffix ValueSelectorOperator = "HasSuffix"

	// ValueSelectorOpNotHasSuffix executes !strings.HasSuffix(a, b).
	ValueSelectorOpNotHasSuffix ValueSelectorOperator = "NotHasSuffix"

	// ValueSelectorOpMatch executes regexp.MatchString(a, b).
	ValueSelectorOpMatch ValueSelectorOperator = "Match"

	// ValueSelectorOpNotMatch executes !regexp.MatchString(a, b).
	ValueSelectorOpNotMatch ValueSelectorOperator = "NotMatch"

	// ValueSelectorOpContains executes strings.Contains(a, b).
	ValueSelectorOpContains ValueSelectorOperator = "Contains"

	// ValueSelectorOpNotContains executes !strings.Contains(a, b).
	ValueSelectorOpNotContains ValueSelectorOperator = "NotContains"
)
```

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -734,49 +737,37 @@ func (vs *vSphereVMProvider) createVirtualMachineAsync(
 		chanErr <- ErrCreate
 	}
 
-	_, k8sErr := controllerutil.CreateOrPatch(
-		ctx,
-		vs.k8sClient,
-		ctx.VM,
-		func() error {
+	objPatch := ctrlclient.MergeFrom(ctx.VM.DeepCopy())
```
akutz, line 739: @bryanv I went ahead and got rid of this anachronistic use of `CreateOrPatch` about which you warned me in the past. Now it just does a pure status patch to ensure we do not hit weirdness.

controllers/vspherepolicy/policyevaluation/policyevaluation_controller.go
```diff
@@ -0,0 +1,583 @@
...
+
+	return nil
+}
+
+func (r *Reconciler) reconcileMatchingComputePolicies(
+	ctx context.Context,
+	obj *vspherepolv1.PolicyEvaluation) error {
+
+	var list vspherepolv1.ComputePolicyList
+	if err := r.Client.List(ctx, &list); err != nil {
```
lubronzhan: Looks like `ComputePolicy` is namespaceScoped. Is it expected that policy in other namespace can also match the  policyEvaludation?

lubronzhan: I saw below you choose policyEvaluation's namespace to find the computePolicy, I believe here it's better to specify the namespace as well? `client.InNamespace(obj.Namespace)`
```
	var (
		pol vspherepolv1.ComputePolicy
		key = client.ObjectKey{
			Namespace: obj.Namespace,
			Name:      ref.Name,
		}
	)

	if err := r.Client.Get(ctx, key, &pol); err != nil {
		return fmt.Errorf("failed to get compute policy: %w", err)
	}
```

controllers/vspherepolicy/policyevaluation/policyevaluation_controller.go
```diff
@@ -0,0 +1,583 @@
...
+
+	var (
+		pol vspherepolv1.ComputePolicy
+		key = client.ObjectKey{
+			Namespace: obj.Namespace,
+			Name:      ref.Name,
+		}
+	)
+
+	if err := r.Client.Get(ctx, key, &pol); err != nil {
```
lubronzhan, line 539: All ComputePolicys are fetched  fetched in `reconcileMatchingComputePolicies`. Maybe you can extract the List on the top level and store in a map then pass to addComputePolicyRef so that you don't need to fetch it again?

controllers/vspherepolicy/policyevaluation/policyevaluation_controller.go
```diff
@@ -0,0 +1,583 @@
...
+			}
+		)
+		if err := r.Client.Get(ctx, tpk, &tp); err != nil {
+			return fmt.Errorf("failed to get tag policy %q: %w", tpn, err)
+		}
+		tags = append(tags, tp.Spec.Tags...)
+	}
+
+	// Check if this policy is already in the results to avoid duplicates.
+	for _, p := range obj.Status.Policies {
```
lubronzhan: nit: 
```suggestion
  if slices.ContainsFunc(obj.Status.Policies, p vspherepolv1.PolicyEvaluationResult)bool{
    return p.Name == pol.Name && p.Kind == computePolicyKind
  }
  return nil
  }
```

controllers/vspherepolicy/policyevaluation/policyevaluation_controller_test.go
```diff
@@ -0,0 +1,2427 @@
...
+									},
+								},
+							},
+						},
+					}
+					withObjs = append(withObjs, computePolicy)
+
+					obj.Spec.Policies = []vspherepolv1.LocalObjectRef{
+						{
+							APIVersion: "vsphere.policy.vmware.com/v1alpha1",
```
lubronzhan: Nit: could be replaced with `vspherepolv1.GroupVersion.String()` so you don't have to change it in the future

external/vsphere-policy/api/v1alpha1/computepolicy_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// the namespace.
+	Match *MatchSpec `json:"match,omitempty"`
+
+	// +optional
+	// +listType=set
+
+	// Tags specifies the names of the TagPolicy objects in the same namespace
+	// that contain the information about the vSphere tags used to activate this
+	// policy.
+	Tags []string `json:"tags,omitempty"`
```
lubronzhan, line 46: I got confused when looking at the tests that has both tagPolicy.spec.tags vs computePolicy.spec.tags . Is it better to rename this to TagPolicies?

pkg/providers/vsphere/contentlibrary/content_library_utils.go
```diff
@@ -101,14 +86,17 @@ func UpdateVmiWithVirtualMachine(
 	status.Firmware = vm.Config.Firmware
 
 	// Populate the guest info.
-	status.OSInfo.ID = vm.Config.GuestId
-	if vm.Guest != nil {
-		status.OSInfo.Type = vm.Guest.GuestFamily
-	}
+	status.OSInfo.Type = vm.Config.GuestId
```
lubronzhan, line 89: naive question, is this expected?

pkg/vmconfig/policy/policy_reconciler_util.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// getTagsFromPolicyEvaluationResults returns the list of vSphere tag UUIDs
+// from one of more PolicyEvaluationResult objects.
+func getTagsFromPolicyEvaluationResults(
+	results ...vspherepolv1.PolicyEvaluationResult) []string {
+
+	var (
+		active    []string
+		activeMap = map[string]struct{}{}
```
lubronzhan: nit: `sets.Set[string]`, it has builtin func `UnsortedList` to turn into a slice

test/builder/testdata/images/ttylinux-pc_i486-16.1.ova
lubronzhan, line 1: Interesting, I didn't know we have ova committed

lubronzhan (commented):
I tried my best to review. Thanks for making the vspherepolicy controller clean and easy to understand!

controllers/vspherepolicy/policyevaluation/policyevaluation_controller.go
akutz: Yes, policy is intentionally namespace scoped. You are correct, I should be using a namespace filter. Good catch! I will add a test with policies in other namespaces to account for this. Seriously, good eye sir!

controllers/vspherepolicy/policyevaluation/policyevaluation_controller.go
akutz: No, this is correct. It's better to utilize the cache and keep the two areas isolated. Plus, the list isn't fetched in the same call stack. It would be necessary to over-optimize this to avoid doing a `Get` against the cache in another function.

external/vsphere-policy/api/v1alpha1/computepolicy_types.go
akutz: They are the same on purpose. That way both have `spec.tags`. Different objects often have similar field names.

pkg/providers/vsphere/contentlibrary/content_library_utils.go
akutz: Yes! Interestingly the OVF's `vmw:guestType` is actually the VIM guest ID. The OVF guest ID is based on CIM's guest ID. See https://github.com/vmware/govmomi/pull/3845 for more info.

test/builder/testdata/images/ttylinux-pc_i486-16.1.ova
akutz: Yes, for testing. It is super, super tiny. The built-in test suite automatically uploads this to vC Sim as a CLS item.

controllers/vspherepolicy/policyevaluation/policyevaluation_controller.go
lubronzhan: Yeah I forgot about the client-go cache again 😂

external/vsphere-policy/api/v1alpha1/computepolicy_types.go
lubronzhan: Ok got it. Reviewed the policyEvaluation.status.policy.tags definition again

lubronzhan (approved):
LGTM!

PR #1124, opened 2025-08-22
 🌱 VM Affinity and Anti-Affinity webhook validations

bryanv:

Fixes #

```release-note
NONE
```

### DIFF
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -167,6 +168,7 @@ func (v validator) ValidateCreate(ctx *pkgctx.WebhookRequestContext) admission.R
 	fieldErrs = append(fieldErrs, v.validateBootOptions(ctx, vm)...)
 	fieldErrs = append(fieldErrs, v.validateSnapshot(ctx, vm, nil)...)
 	fieldErrs = append(fieldErrs, v.validateGroupName(ctx, vm)...)
+	fieldErrs = append(fieldErrs, v.validateVMAffinity(ctx, vm)...)
 
 	validationErrs := make([]string, 0, len(fieldErrs))
 	for _, fieldErr := range fieldErrs {
@@ -1392,6 +1394,7 @@ func (v validator) validateImmutableFields(
 	}
 	allErrs = append(allErrs, v.validateImmutableReserved(ctx, vm, oldVM)...)
 	allErrs = append(allErrs, v.validateImmutableNetwork(ctx, vm, oldVM)...)
+	allErrs = append(allErrs, v.validateImmutableVMAffinity(ctx, vm, oldVM)...)
 
 	return allErrs
 }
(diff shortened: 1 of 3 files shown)


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1951,3 +1953,121 @@ func (v validator) validateGroupName(
...
+
+	path := field.NewPath("spec", "affinity")
+
+	if affinity.ZoneAffinity != nil {
+		allErrs = append(allErrs, field.Forbidden(path.Child("zoneAffinity"), "zone affinity is not allowed"))
+	}
+
+	if affinity.ZoneAntiAffinity != nil {
+		allErrs = append(allErrs, field.Forbidden(path.Child("zoneAntiAffinity"), "zone anti-affinity is not allowed"))
+	}
```
dilyar85, line 1983: Are these fields forbidden temporarily or always? Curious why we introduced them in spec if they cannot be set...

webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
```diff
@@ -3053,6 +3053,193 @@ func unitTestsValidateCreate() {
...
+										},
+									},
+									TopologyKey: "topology.kubernetes.io/zone",
+								},
+							},
+						}
+					},
+					validate: doValidateWithMsg(
+						`spec.affinity.vmAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].labelSelector.matchExpressions[0].operator: Unsupported value: "NotIn": supported values: "In"`,
+						`spec.affinity.vmAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].topologyKey: Unsupported value: "topology.kubernetes.io/host": supported values: "topology.kubernetes.io/zone`,
```
dilyar85: Nit: missing quote at the end

webhooks/virtualmachine/validation/virtualmachine_validator.go
bryanv: So I think this is the least bad `field.Foo()` to use here - the others require the value which for a whole struct is not going to be very pretty in the output - those are more geared towards standalone fields. It is what I used for the bootstrap stuff. But yah `field.NotSupported` would be nicer.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -269,6 +270,7 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
 	fieldErrs = append(fieldErrs, v.validateBootOptions(ctx, vm)...)
 	fieldErrs = append(fieldErrs, v.validateSnapshot(ctx, vm, oldVM)...)
 	fieldErrs = append(fieldErrs, v.validateGroupName(ctx, vm)...)
+	fieldErrs = append(fieldErrs, v.validateVMAffinity(ctx, vm)...)
```
aruneshpa: Affinity cannot be changed on a VM once specified (at least for the upcoming release).  We should enforce immutability.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1951,3 +1953,121 @@ func (v validator) validateGroupName(
...
+	if a := affinity.VMAffinity; a != nil {
+		p := path.Child("vmAffinity")
+
+		if rs := a.RequiredDuringSchedulingIgnoredDuringExecution; len(rs) > 0 {
+			p := p.Child("requiredDuringSchedulingIgnoredDuringExecution")
+
+			for idx, rs := range rs {
+				p := p.Index(idx)
+
+				if rs.LabelSelector != nil {
```
aruneshpa, line 1995: Hmm.  Trying to parse this...

I just realized that LabelSelector and MatchExpressions both can be specified.  Ultimately, we are saying place the VM with other VMs with some tag.

I will file a story to handle that in the provider code.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1951,3 +1953,121 @@ func (v validator) validateGroupName(
...
+				}
+
+				if rs.TopologyKey != topology.KubernetesTopologyZoneLabelKey {
+					allErrs = append(allErrs, field.NotSupported(
+						p.Child("topologyKey"), rs.TopologyKey, []string{topology.KubernetesTopologyZoneLabelKey}))
+				}
+			}
+		}
+
+		if ps := a.PreferredDuringSchedulingIgnoredDuringExecution; len(ps) > 0 {
```
aruneshpa: VM-VM affinity at zone level supports required _and_ preferred during scheduling.  So we should allow this.
https://github.com/vmware-tanzu/vm-operator/blob/6bbba44a5b69ff00cd92e35c2595cbc873edaa2c/pkg/providers/vsphere/virtualmachine/configspec.go#L194

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1951,3 +1953,121 @@ func (v validator) validateGroupName(
...
+		p := path.Child("vmAffinity")
+
+		if rs := a.RequiredDuringSchedulingIgnoredDuringExecution; len(rs) > 0 {
+			p := p.Child("requiredDuringSchedulingIgnoredDuringExecution")
+
+			for idx, rs := range rs {
+				p := p.Index(idx)
+
+				if rs.LabelSelector != nil {
+					if len(rs.LabelSelector.MatchExpressions) != 0 {
```
aruneshpa: I know we said we won't match labels, but that was more for AAF usecase where we don't enforce labels.  Here the VM must have this label because we will send this tag to DRS.  It will be confusing to the user, so we might as well be explicit.

We have the entire VM object here anyway, so should be easy to compare that the values specified here are present in metadata.Labels.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1951,3 +1953,169 @@ func (v validator) validateGroupName(
...
+				}
+
+				if rs.TopologyKey != topology.KubernetesTopologyZoneLabelKey {
+					allErrs = append(allErrs, field.NotSupported(
+						p.Child("topologyKey"), rs.TopologyKey, []string{topology.KubernetesTopologyZoneLabelKey}))
+				}
+			}
+		}
+
+		if len(a.RequiredDuringSchedulingPreferredDuringExecution) > 0 {
```
aruneshpa, line 2096: I think this is already handled at L2047.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1951,3 +1953,169 @@ func (v validator) validateGroupName(
...
+				"VM anti-affinity with RequiredDuringSchedulingPreferredDuringExecution is not allowed"))
+		}
+
+		if len(a.PreferredDuringSchedulingPreferredDuringExecution) > 0 {
+			p := p.Child("preferredDuringSchedulingPreferredDuringExecution")
+
+			for idx, rs := range a.PreferredDuringSchedulingPreferredDuringExecution {
+				p := p.Index(idx)
+
+				if rs.LabelSelector != nil {
```
aruneshpa, line 2107: This also needs to be updated to allow MatchExpressions, but only `In` operator.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1951,3 +1953,169 @@ func (v validator) validateGroupName(
...
+	return allErrs
+}
+
+func (v validator) validateImmutableVMAffinity(
+	_ *pkgctx.WebhookRequestContext,
+	vm, oldVM *vmopv1.VirtualMachine) field.ErrorList {
+
+	var allErrs field.ErrorList
+
+	if !equality.Semantic.DeepEqual(vm.Spec.Affinity, oldVM.Spec.Affinity) {
```
aruneshpa, line 2139: IDK how far do we want to take this, but there can potentially be a bunch of lists in the affinity rules (MatchExpressions) and changing their order should not reject the request ideally :(

webhooks/virtualmachine/validation/virtualmachine_validator.go
bryanv: L2047 is `RequiredDuringSchedulingIgnoredDuringExecution` where here is `...PreferredDuringExecution`

webhooks/virtualmachine/validation/virtualmachine_validator.go
bryanv: OK

webhooks/virtualmachine/validation/virtualmachine_validator.go
bryanv: I don't think allowing those non changes changes is worth the effort.

PR #1125, opened 2025-08-22
:seedling: Enqueue reconcile on snapshot and root snapshot updates

aruneshpa:

This change modifies the watcher to enqueue an event when the snapshot (the current snapshot) and the root snapshot (list of root snapshots) of a VM change.  This allows us to not worry about updating / patching the snapshot tree once we are done with the snapshot or revert operation.  Once the snapshot is done, it will queue a reconcile and refresh the snapshot tree.

### Testing Done:
- Unit tests
- Verified on a testbed that the watcher gets called when a snapshot is created on vCenter directly.  The list of properties returned by vCenter also include the (current) snapshot, and root snapshots.

```release-note
Enqueue reconcile on snapshot and root snapshot updates
```


pkg/util/vsphere/watcher/watcher.go
```diff
@@ -27,9 +27,11 @@ func DefaultWatchedPropertyPaths() []string {
 		"config.extraConfig",
 		"config.hardware.device",
 		"config.keyId",
+		"snapshot",
 		"guest.ipStack",
 		"guest.net",
 		"guestHeartbeatStatus",
+		"rootSnapshot",
```
dilyar85: Looks like they're sorted in alphabetical order?

pkg/util/vsphere/watcher/watcher_test.go
```diff
@@ -419,6 +419,33 @@ var _ = Describe("Start", func() {
...
+				// Assert that a result is signaled due to the VM entering the
+				// scope of the watcher.
+				assertResult(cluster1vm1, "my-namespace-1", "my-name-1")
+
+				// Assert no more results are signaled.
+				assertNoResult()
+
+				// Invoke a snapshot operation that should update the
+				// root snapshots, and current snapshot of the VM.
+				t, err := cluster1vm1.CreateSnapshot(ctx, "snap-1", "description", false, false)
```
lubronzhan, line 433: It would be best if you could also add a case of revertSnapshot (only update current snapshot)

pkg/util/vsphere/watcher/watcher.go
aruneshpa: I had this as `currentSnapshot` (because that's what govmomi calls this field) in alphabetical order; had to change it to `snapshot`, but forgot to fix the ordering.  😬 
Thanks for spotting this.

PR #1127, opened 2025-08-25
 🌱 Various GetSnapshotSize changes

bryanv:

 - Build up the file disk/size map first so we don't have to keep a slice of all the keys to do the addition after the loop
 - Exit the loop once we find the snapshot by MoID

Fixes #

```release-note
NONE
```


pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -31,31 +31,28 @@ type SnapshotArgs struct {
...
 )
 
 func SnapshotVirtualMachine(args SnapshotArgs) (*vimtypes.ManagedObjectReference, error) {
-	obj := args.VMSnapshot
-	vm := args.VcVM
-	// Find snapshot by name
-	snapMoRef, _ := vm.FindSnapshot(args.VMCtx, obj.Name)
+	snapshotName := args.VMSnapshot.Name
+
+	snapMoRef, _ := args.VcVM.FindSnapshot(args.VMCtx, snapshotName) // TODO: Use our FindSnapshot impl and check error?
```
lubronzhan, line 42: Do you mean `FindSnapshotInTree` func?

lubronzhan: Or you mean we should check the error

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -174,69 +167,63 @@ func GetSnapshotSize(vmCtx pkgctx.VirtualMachineContext, vmSnapshot *vimtypes.Ma
...
 	}
 
-	var total int64
-	for _, fileKey := range fileKeyList {
-		if fileSize, ok := fileKeyMap[fileKey]; ok {
-			total += fileSize
-		}
-		// Assume the size is 0 if it's not found in the fileKeyMap
-	}
+	// TODO: Error if snapshot not found?
```
lubronzhan: IC, return early if `vmlayout.Snapshot` is empty. Ok will add it later

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -95,50 +94,17 @@ func (vs *vSphereVMProvider) GetSnapshotSize(
 		return 0, fmt.Errorf("failed to get VirtualMachine %q: %w", vmCtx.VM.Name, err)
 	}
 
-	var moVM mo.VirtualMachine
-
-	if err = vcVM.Properties(ctx, vcVM.Reference(), []string{"snapshot", "layoutEx", "config.hardware.device"}, &moVM); err != nil {
+	err = vcVM.Properties(ctx, vcVM.Reference(), []string{"snapshot", "layoutEx", "config.hardware.device"}, &vmCtx.MoVM)
```
lubronzhan, line 97: nit: These two lines could be combined

lubronzhan (approved):
LGTM. Thanks for all the refactoring!

pkg/providers/vsphere/virtualmachine/snapshot.go
bryanv: Yah this is using govmoni's `FindSnapshot` which doesn't return very friendly errors, so we also have our own impl that does. I think that's why this spot is ignoring the error which it probably shouldn't be.

pkg/providers/vsphere/virtualmachine/snapshot.go
bryanv: Sure.

pkg/providers/vsphere/vmprovider_vm_snapshot.go
bryanv: I split it because just the Properties call line is getting kind of long already

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -95,50 +94,17 @@ func (vs *vSphereVMProvider) GetSnapshotSize(
...
 		return 0, fmt.Errorf("failed to find snapshot %q: %w", vmSnapshotName, err)
 	}
 
-	size := virtualmachine.GetSnapshotSize(
-		vmCtx, vmNapshot)
-
-	return size, nil
-}
-
-func (vs *vSphereVMProvider) GetParentSnapshot(
```
bryanv, line 117: This was removed from the interface in a prior commit but this impl wasn't removed.

pkg/providers/vsphere/virtualmachine/snapshot.go
aruneshpa, line 42: Yes, we should use our impl.  The reason for duplicating and using our custom impl was to avoid having to create a property collector.  I think that still aplies.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -31,31 +31,28 @@ type SnapshotArgs struct {
...
 func SnapshotVirtualMachine(args SnapshotArgs) (*vimtypes.ManagedObjectReference, error) {
-	obj := args.VMSnapshot
-	vm := args.VcVM
-	// Find snapshot by name
-	snapMoRef, _ := vm.FindSnapshot(args.VMCtx, obj.Name)
+	snapshotName := args.VMSnapshot.Name
+
+	snapMoRef, _ := args.VcVM.FindSnapshot(args.VMCtx, snapshotName) // TODO: Use our FindSnapshot impl and check error?
 	if snapMoRef != nil {
 		// TODO: Handle revert to snapshot. Need a way to compare currentSnapshot's moID
```
aruneshpa: Not your change, but this is an outdated TODO.  Can you remove it?  Thanks!

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -174,70 +167,64 @@ func GetSnapshotSize(vmCtx pkgctx.VirtualMachineContext, vmSnapshot *vimtypes.Ma
...
+	for _, snapshot := range vmLayout.Snapshot {
+		if snapshot.Key.Value != vmSnapshot.Value {
+			continue
+		}
 
-				if len(disk.Chain) == 0 {
-					vmCtx.Logger.V(5).Info("skipping the disk since its chain is empty", "diskKey", disk.Key)
-					continue
-				}
+		fcdDeviceKeySet := getFCDDeviceKeySet(vmCtx)
```
aruneshpa: Even though this is only running once when we are at the snapshot that we are interested in, it reads as if we are doing this a bunch of times :(
What was the reason for moving this in the loop?  Definition closer to where it is used?
If so, maybe we extract the snapshot layout by iterating over all the snapshots and all other processing happens outside the loop at top level indent?

bryanv: Sure. I add some find snapshot by MoID func or refactor this inline to make it clearer

aruneshpa (approved):

PR #1133, opened 2025-08-26
✨ Add support for Unmanaged snapshots in VM snapshot references

palnabarun:

This makes references to VM snapshots use a new struct `VirtualMachineSnapshotReference` which can refer to both Managed and Unmanaged snapshots. The existing fields `CurrentSnapshot` and `RootSnapshots` in `VirtualMachineStatus` and `Children` in `VirtualMachineSnapshotStatus` have been updated to use this new struct.

This also adds conversion logic to convert between v1alpha4 and v1alpha5 APIs. The conversion logic assumes that all snapshot references in v1alpha4 are to Unmanaged snapshots. The conversion logic has been tested in the integration tests.

This also updates the `SyncVMSnapshotTreeStatus` function to
appropriately set the new fields, even in case of Unmanaged snapshots.

NA

Snippets from manual tests:

```
# CurrentSnapshot is pointing to a snapshot created through vCenter
# RootSnapshots shows a root snapshot created through Supervisor
$ k get vm vm -o yaml
...
  currentSnapshot:
    type: Unmanaged
...
  rootSnapshots:
  - reference:
      kind: VirtualMachineSnapshot
      name: snap-1
    type: Managed
...
```

Conversion:
```
$ k get virtualmachines vm -o yaml
...
  currentSnapshot:
    reference:
      kind: VirtualMachineSnapshot
      name: snap-3
    type: Managed
...
  rootSnapshots:
  - reference:
      kind: VirtualMachineSnapshot
      name: snap-1
    type: Managed
...

$ k get virtualmachines.v1alpha4.vmoperator.vmware.com vm -o yaml
...
  currentSnapshot:
    kind: VirtualMachineSnapshot
    name: snap-3
...
  rootSnapshots:
  - apiVersion: vmoperator.vmware.com/v1alpha5
    kind: VirtualMachineSnapshot
    name: snap-1
...
```

```release-note
✨ Add support for Unmanaged snapshots in VM snapshot references by adding a new type to store reference types
```


pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1466,31 +1466,38 @@ func updateSnapshotChildrenStatus(
...
 	if err != nil {
 		return nil, err
 	}
 
 	if curSnapshot == nil {
-		// TODO (Lubron): This is for external snapshots.
+		// If the current snapshot custom resource is not found,
+		// it means it's an Unmanaged snapshot.
+		// Nothing to update here.
 		vmCtx.Logger.V(4).Info("Snapshot custom resource not found, skipping",
```
lubronzhan:
```suggestion
		vmCtx.Logger.V(4).Info("VirtualMachineSnapshot not found, skipping",
```

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1466,31 +1466,38 @@ func updateSnapshotChildrenStatus(
...
 		if err != nil {
 			return nil, err
 		}
-		// TODO (Lubron): If ref is nil, it means the snapshot custom resource is not found
-		// and it's an external snapshot.
+
 		if ref != nil {
 			children = append(children, *ref)
+		} else {
+			vmCtx.Logger.V(4).Info("Child snapshot custom resource not found, adding a reference to an Unmanaged snapshot",
```
lubronzhan:
```suggestion
			vmCtx.Logger.V(4).Info("Child VirtualMachineSnapshot not found, adding an Unmanaged reference",
```

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1671,24 +1684,29 @@ func updateRootSnapshots(
...
 	for _, rootSnapshot := range vmCtx.MoVM.Snapshot.RootSnapshotList {
 		rootSnapshotCR, err := getSnapshotCR(vmCtx, k8sClient, rootSnapshot.Name)
 		if err != nil {
 			return err
 		}
 
 		if rootSnapshotCR == nil {
-			// TODO (Lubron): This is for external snapshots.
-			vmCtx.Logger.V(4).Info("Snapshot custom resource not found, skipping",
+			vmCtx.Logger.V(4).Info("Snapshot custom resource not found, adding an Unmnaged reference",
```
lubronzhan:
```suggestion
			vmCtx.Logger.V(4).Info("VirtualMachineSnapshot not found, adding an Unmanaged reference",
```

pkg/providers/vsphere/vmlifecycle/update_status_test.go
```diff
@@ -3242,9 +3242,11 @@ var _ = Describe("Snapshot status", func() {
...
-					Expect(vmCtx.VM.Status.CurrentSnapshot.APIVersion).To(Equal(vmSnapshot.APIVersion))
-					Expect(vmCtx.VM.Status.CurrentSnapshot.Kind).To(Equal(vmSnapshot.Kind))
-					Expect(vmCtx.VM.Status.CurrentSnapshot.Name).To(Equal(vmSnapshot.Name))
+					Expect(vmCtx.VM.Status.CurrentSnapshot.Type).To(Equal(vmopv1.VirtualMachineSnapshotReferenceTypeManaged))
+					Expect(vmCtx.VM.Status.CurrentSnapshot.SnapshotReference).ToNot(BeNil())
+					Expect(vmCtx.VM.Status.CurrentSnapshot.SnapshotReference.APIVersion).To(Equal(vmSnapshot.APIVersion))
+					Expect(vmCtx.VM.Status.CurrentSnapshot.SnapshotReference.Kind).To(Equal(vmSnapshot.Kind))
+					Expect(vmCtx.VM.Status.CurrentSnapshot.SnapshotReference.Name).To(Equal(vmSnapshot.Name))
 				})
 			})
```
lubronzhan, line 3424: The test below here `When("the snapshot custom resource doesn't exist", func() {`, it should expect a non empty `CurrentSnapshot ` after this change I assume?


I think the test is incorrect because it doesn't actually removes the vmSnapshot since it has finalizer
```
// Set a finalizer
vmSnapshot.Finalizers = append(vmSnapshot.Finalizers, "random-finalizer")
Expect(ctx.Client.Update(ctx, vmSnapshot)).To(Succeed())
// Mark the snapshot for deletion. The finalizer will
// make sure the object is not garbage collected.
Expect(ctx.Client.Delete(ctx, vmSnapshot)).To(Succeed())`
```

lubronzhan (approved):
Overall looks good! Just some small comments

lubronzhan:
Just to confirm, if it's unmanaged type, we just set unmanaged type, no need to set the name in the localObjectRef? If yes, could you put it in the spec definition? Thanks

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1598,10 +1607,11 @@ func updateCurrentSnapshotStatus(
 	}
 
 	if vmSnapshot == nil {
-		// TODO (Lubron): This is for external snapshots.
 		vmCtx.Logger.V(4).Info("VirtualMachineSnapshot custom resource not found, clearing status",
```
bryanv: This is not being cleared anymore

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1466,31 +1466,38 @@ func updateSnapshotChildrenStatus(
...
 		vmCtx.Logger.V(4).Info("Snapshot custom resource not found, skipping",
 			"snapshotName", node.Name)
 		return nil, nil
 	}
 
 	// Recurse into children and collect their refs (only if their CRs exist).
-	children := make([]common.LocalObjectRef, 0, len(node.ChildSnapshotList))
+	children := make([]vmopv1.VirtualMachineSnapshotReference, 0, len(node.ChildSnapshotList))
 	for i := range node.ChildSnapshotList {
 		child := node.ChildSnapshotList[i]
```
bryanv: Nit: save a line an use `i, child := range ...` ?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1671,24 +1684,29 @@ func updateRootSnapshots(
...
 		if err != nil {
 			return err
 		}
 
 		if rootSnapshotCR == nil {
-			// TODO (Lubron): This is for external snapshots.
-			vmCtx.Logger.V(4).Info("Snapshot custom resource not found, skipping",
+			vmCtx.Logger.V(4).Info("Snapshot custom resource not found, adding an Unmnaged reference",
 				"snapshotName", rootSnapshot.Name)
+			newRootSnapshots = append(newRootSnapshots, vmopv1.VirtualMachineSnapshotReference{
```
bryanv, line 1638: Care there be multiple of these? Is there something later set to help identify which is which?

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -139,6 +139,31 @@ type VirtualMachineSnapshotStorageStatus struct {
...
 }
 
+type VirtualMachineSnapshotReferenceType string
+
+const (
+	// VirtualMachineSnapshotReferenceTypeManaged represents a snapshot that is managed by the Supervisor Cluster.
+	VirtualMachineSnapshotReferenceTypeManaged VirtualMachineSnapshotReferenceType = "Managed"
+
+	// VirtualMachineSnapshotReferenceTypeUnmanaged represents a snapshot that is managed by the backing infrastructure (e.g., vSphere).
+	VirtualMachineSnapshotReferenceTypeUnmanaged VirtualMachineSnapshotReferenceType = "Unmanaged"
```
bryanv, line 190: Is this not possible to test via vcsim? I don't see this being used in any of the tests

lubronzhan: It's not tested with vcsim, but tested with mocked moVM in update_status_test.go
https://github.com/vmware-tanzu/vm-operator/pull/1133/files#diff-fab70a3f39d63d3c216b52ec02f18f299ef5767ffcb982d22d998963a73b4af3R3966

bryanv: Oh my bad. That diff was collapsed

pkg/providers/vsphere/vmlifecycle/update_status_test.go
```diff
@@ -3924,7 +3962,9 @@ var _ = Describe("Snapshot status", func() {
 
 				It("should clear the root snapshots status", func() {
 					Expect(vmlifecycle.ReconcileStatus(vmCtx, ctx.Client, vcVM, data)).To(Succeed())
-					Expect(vmCtx.VM.Status.RootSnapshots).To(BeNil())
+					Expect(vmCtx.VM.Status.RootSnapshots).To(ContainElement(vmopv1.VirtualMachineSnapshotReference{
```
bryanv, line 4167: Here and elsewhere `ContainElement` is being used how about `HaveExactElements` or assert the length

pkg/providers/vsphere/vmlifecycle/update_status_test.go
palnabarun: Corrected the test logic. In "the snapshot custom resource doesn't exist", the object was not deleted because of the vmsnapshot finalizer in the dummy object.

pkg/providers/vsphere/vmlifecycle/update_status.go
palnabarun: Yes, it's possible that multiple of these are there. There isn't an info to identify them.

We discussed about having the snap moID as a field in `VirtualMachineSnapshotReference` when the snapshot is not managed, but we didn't find an utility of having it at the moment.

When someone has snapshots in VC which are not managed by VMOP, we don't support operations like reverting to those snapshots.

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -139,6 +139,31 @@ type VirtualMachineSnapshotStorageStatus struct {
 	Requested []VirtualMachineSnapshotStorageStatusRequested `json:"requested,omitempty"`
 }
 
+type VirtualMachineSnapshotReferenceType string
+
+const (
+	// VirtualMachineSnapshotReferenceTypeManaged represents a snapshot that is managed by the Supervisor Cluster.
```
aruneshpa: vvvvvvvvNit: Just Supervisor is sufficient.  Drop the "Cluster" (anywhere you are using it).

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -139,6 +139,31 @@ type VirtualMachineSnapshotStorageStatus struct {
...
+	VirtualMachineSnapshotReferenceTypeManaged VirtualMachineSnapshotReferenceType = "Managed"
+
+	// VirtualMachineSnapshotReferenceTypeUnmanaged represents a snapshot that is managed by the backing infrastructure (e.g., vSphere).
```
aruneshpa: Maybe just call out that a snapshot that is not managed by Supervisor.  Then give an example of how such a snapshot could be created.  For example, a snapshot created by a backup application, or a snapshot created by a provider or VI admin by directly interacting with vCenter.

lubronzhan (approved):
LGTM!

pkg/providers/vsphere/vmprovider_vm_snapshot_test.go
```diff
@@ -178,9 +178,13 @@ func vmSnapshotTests() {
 		It("should sync the VM's current and root snapshots status", func() {
 			Expect(vmProvider.SyncVMSnapshotTreeStatus(ctx, vm)).To(Succeed())
 			Expect(vm.Status.CurrentSnapshot).ToNot(BeNil())
-			Expect(vm.Status.CurrentSnapshot.Name).To(Equal(vmSnapshot.Name))
+			Expect(vm.Status.CurrentSnapshot.Type).To(Equal(vmopv1.VirtualMachineSnapshotReferenceTypeManaged))
+			Expect(vm.Status.CurrentSnapshot.SnapshotReference).To(Not(BeNil()))
```
bryanv: We typically use `ToNot(...` instead

api/v1alpha4/virtualmachine_conversion.go
```diff
@@ -128,6 +130,37 @@ func restore_v1alpha5_VirtualMachineHardware(dst, src *vmopv1.VirtualMachine) {
...
 }
 
+func Convert_common_LocalObjectRef_To_v1alpha5_VirtualMachineSnapshotReference(
+	in *vmopv1a4common.LocalObjectRef, out *vmopv1.VirtualMachineSnapshotReference, s apiconversion.Scope) error {
+	if in == nil {
+		return nil
+	}
+
+	// Convert LocalObjectRef to the nested SnapshotReference field
+	out.SnapshotReference = &vmopv1common.LocalObjectRef{
```
aruneshpa: Could `out` not be nil here?

api/v1alpha4/virtualmachine_conversion.go
```diff
@@ -128,6 +130,37 @@ func restore_v1alpha5_VirtualMachineHardware(dst, src *vmopv1.VirtualMachine) {
...
+		Name:       in.Name,
+	}
+
+	return nil
+}
+
+func Convert_v1alpha5_VirtualMachineSnapshotReference_To_common_LocalObjectRef(
+	in *vmopv1.VirtualMachineSnapshotReference, out *vmopv1a4common.LocalObjectRef, s apiconversion.Scope) error {
+	if in == nil || in.SnapshotReference == nil {
+		// return if the SnapshotReference is nil
```
aruneshpa: Pretty self explanatory, I think you can remove the comment.

api/v1alpha4/virtualmachine_conversion.go
```diff
@@ -128,6 +130,37 @@ func restore_v1alpha5_VirtualMachineHardware(dst, src *vmopv1.VirtualMachine) {
 	}
 }
 
+func Convert_common_LocalObjectRef_To_v1alpha5_VirtualMachineSnapshotReference(
```
aruneshpa, line 139: Have you verified that the conversion works by creating an object on one version and then fetching them on the other version?

TBH, this may not be necessary at all once we rip out the snapshots from v1a4 since they were never meant to be there in the first place.

api/v1alpha5/virtualmachinepublishrequest_types.go
```diff
@@ -112,6 +112,10 @@ const (
 	// VirtualMachinePublishRequestManagedByLabelKey is the label key to identify VirtualMachinePublishRequest that
 	// is created as a part of the VirtualMachineGroupPublishRequest.
 	VirtualMachinePublishRequestManagedByLabelKey = GroupName + "/virtualmachinepublishrequest-managed-by"
+
+	// VirtualMachinePublishRequestUUIDLabelKey is the label key to store UUID of the VirtualMachinePublishRequest that
+	// clones a source virtual machine, as a virtual machine template.
+	VirtualMachinePublishRequestUUIDLabelKey = GroupName + "/virtualmachinepublishrequest-uuid"
```
aruneshpa: This should not be part of your change?

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -139,6 +139,31 @@ type VirtualMachineSnapshotStorageStatus struct {
...
+	VirtualMachineSnapshotReferenceTypeUnmanaged VirtualMachineSnapshotReferenceType = "Unmanaged"
+)
+
+type VirtualMachineSnapshotReference struct {
+	// Type is the type of the snapshot reference.
+	//
+	// The value can be one of the following:
+	// - Managed -> the snapshot is managed by the Supervisor Cluster
+	// - Unmanaged -> the snapshot is managed by the backing infrastructure (e.g., vSphere).
+	// +kubebuilder:validation:Enum=Managed;Unmanaged
```
aruneshpa: Do you see a point in defaulting to "managed"?

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -139,6 +139,31 @@ type VirtualMachineSnapshotStorageStatus struct {
...
+	Type VirtualMachineSnapshotReferenceType `json:"type"`
+
+	// +optional
+
+	// SnapshotReference is the reference to the snapshot in the Supervisor Cluster.
+	SnapshotReference *vmopv1common.LocalObjectRef `json:"snapshotReference,omitempty"`
```
aruneshpa: How about calling it just "Reference", otherwise, we introduce stuttering (`VirtualMachineSnapshotReference.SnapshotReference`)

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
aruneshpa, line 1: Same for this?  I think this got pulled in as part of an unclean rebase?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1445,12 +1451,16 @@ func updateSnapshotChildrenStatus(
 	}
 
 	// Return this node's LocalObjectRef to the parent.
```
aruneshpa:
```suggestion
	// Return this node's reference to the parent.
```

api/v1alpha5/virtualmachinepublishrequest_types.go
palnabarun: Yeah. Not sure what happened with the rebase. Commit message also had changed.

api/v1alpha5/virtualmachinesnapshot_types.go
palnabarun: Makes sense. Updated the field.

api/v1alpha4/virtualmachine_conversion.go
palnabarun: I don't think so. The calling function sends a reference to an existing object. The object itself might have unset fields, which we are filling in here.

api/v1alpha5/virtualmachinesnapshot_types.go
palnabarun: Makes sense to set a reasonable default to "Managed".

api/v1alpha4/virtualmachine_conversion.go
palnabarun: Yup. Conversions work fine. Added a sample in the PR comment.

aruneshpa (approved):

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -145,6 +145,38 @@ type VirtualMachineSnapshotStorageStatus struct {
 	Requested []VirtualMachineSnapshotStorageStatusRequested `json:"requested,omitempty"`
 }
 
+type VirtualMachineSnapshotReferenceType string
```
akutz, line 178: This should use a kubebuilder enum to be correct. Plus, exported symbols should be commented correctly. See VirtualMachinePowerState as an example:

```golang
// +kubebuilder:validation:Enum=PoweredOff;PoweredOn;Suspended

// VirtualMachinePowerState defines a VM's desired and observed power states.
type VirtualMachinePowerState string
```

So, it would be:

```golang
// +kubebuilder:validation:Enum=Managed;Unmanaged

// VirtualMachineSnapshotReferenceTypeManaged defines type of snapshot.
type VirtualMachineSnapshotReferenceTypeManaged string
```

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -145,6 +145,38 @@ type VirtualMachineSnapshotStorageStatus struct {
...
+	// VirtualMachineSnapshotReferenceTypeUnmanaged represents a snapshot
+	// that is not managed by the Supervisor. These snapshots might have been
+	// created by a provider or a VI Admin directly on vCenter, or a snapshot
+	// created by a backup service.
+	// The Supervisor does not manage the lifecycle of such snapshots.
+	VirtualMachineSnapshotReferenceTypeUnmanaged VirtualMachineSnapshotReferenceType = "Unmanaged"
+)
+
+type VirtualMachineSnapshotReference struct {
+	// +kubebuilder:validation:Enum=Managed;Unmanaged
```
akutz, line 194: Remove this and put it on the type alias per my earlier comment.

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -145,6 +145,38 @@ type VirtualMachineSnapshotStorageStatus struct {
...
+	// +kubebuilder:default=Managed
+
+	// Type is the type of the snapshot reference.
+	//
+	// The value can be one of the following:
+	// - Managed
+	// - Unmanaged
+	Type VirtualMachineSnapshotReferenceType `json:"type"`
```
akutz, line 202: Please update the comment to be:

```golang
	// Type describes the type of snapshot reference.
	//
	// Possible values are: Managed, Unmanaged.
```

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -145,6 +145,38 @@ type VirtualMachineSnapshotStorageStatus struct {
...
+
+	// +optional
+
+	// Reference is the reference to the snapshot in the Supervisor.
```
akutz, line 206: Should the comment indicate this is only set for managed snapshots?

api/v1alpha5/virtualmachinesnapshot_types.go
palnabarun: Assuming you meant this:
```
// +kubebuilder:validation:Enum=Managed;Unmanaged

// VirtualMachineSnapshotReferenceType defines the type of the snapshot reference.
type VirtualMachineSnapshotReferenceType string
 ```

PR #1135, opened 2025-08-27
:seedling: Remove the zone annotation from VM backup

aruneshpa:

Today, a VM being backed up includes the zone annotation.  As a result, if a VM is being restored from backup in an environment where the zone doesn't exist, the VM registration will fail.

This problem is much more severe for the fail-over scenarios since the primary and secondary sites are different and can have distinct zones.

### Testing Done
- Modified the existing unit tests to validate that the label is unset
- Validated in a real env that the label is dropped from the backup

```release-note
Remove the zone annotation from VM backup
```


pkg/providers/vsphere/virtualmachine/backup.go
```diff
@@ -301,9 +302,19 @@ func getDesiredVMResourceYAMLForBackup(
 func trimBackupFields(obj client.Object) {
 	if annotations := obj.GetAnnotations(); len(annotations) > 0 {
 		delete(annotations, corev1.LastAppliedConfigAnnotation)
+
 		obj.SetAnnotations(annotations)
```
bryanv, line 305: I think there are additional values that we'd want to remove too - https://github.com/vmware-tanzu/vm-operator/blob/4ac019e5dcfe0fbb27a5ebfe372378355e3133d5/pkg/constants/constants.go#L127-L147. Thoughts?

aruneshpa: Absolutely agree.  However, I want to club those in a different PR because I want one PR with all the labels and annotation and for that, I want to get an ACK from VCFA.

Is it okay if I keep this one scoped to the DR operator ask?

bryanv: Sure thing

dilyar85 (approved):

PR #1137, opened 2025-08-28
:seedling: Apply VM name label on the snapshot custom resource

aruneshpa:

This change introduces a new label on a `VirtualMachineSnapshot` resource that points to the VM custom resource for which the snapshot is taken.  This label is added in the mutation webhook when the resource is created.  The value of the label is extracted from snapshot resource's `spec.vmRef.name`.

This allows clients (such as VCFA) to query the snapshots resources corresponding to a given virtual machine.

The mutation webhook _only_ intercepts the Create operations since we do not support changing the `vmRef` of an existing snapshot resource.

To enforce this label never changing, introduce a validation in the validation webhook.

```release-note
Apply VM name label on the snapshot custom resource
```


webhooks/virtualmachinesnapshot/mutation/virtualmachinesnapshot_mutator.go
```diff
@@ -0,0 +1,111 @@
...
+		vmSnapshot.Labels = make(map[string]string)
+	}
+
+	// Check if the label already exists with the correct value
+	vmName := vmSnapshot.Spec.VMRef.Name
+	if currentValue, exists := vmSnapshot.Labels[vmopv1.VMNameForSnapshotLabel]; exists && currentValue == vmName {
+		return false
+	}
+
+	vmSnapshot.Labels[vmopv1.VMNameForSnapshotLabel] = vmName
```
bryanv: Shouldn't this only set if not set, and otherwise let the validation webhook fail it?

webhooks/virtualmachinesnapshot/mutation/virtualmachinesnapshot_mutator.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil, err
+	}
+	return vmSnapshot, nil
+}
+
+// SetVMNameLabel sets the VM name label on the snapshot if it has a vmRef.
+// Returns true if the snapshot was mutated, false otherwise.
+func SetVMNameLabel(vmSnapshot *vmopv1.VirtualMachineSnapshot) bool {
+	// Only set the label if there's a vmRef with a name.
+	if vmSnapshot.Spec.VMRef == nil || vmSnapshot.Spec.VMRef.Name == "" {
```
bryanv, line 96: JW: why is this effectively required field a pointer in the first place? We've got nil checks for it all over :(

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -143,6 +144,32 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
+// of the snapshot resource. This validation is only applicable during
+// an update operation.
+func (v validator) validateImmutableVMNameLabel(
+	ctx *pkgctx.WebhookRequestContext,
+	vmSnapshot, oldVMSnapshot *vmopv1.VirtualMachineSnapshot) field.ErrorList {
+
+	var allErrs field.ErrorList
+
+	vmNameLabelPath := field.NewPath("metadata", "labels").Key(vmopv1.VMNameForSnapshotLabel)
+	if oldVMSnapshot != nil && oldVMSnapshot.Labels != nil {
```
bryanv: Can we simplify this - `oldVMSnapshot` will never be `nil` since is only call for updates

webhooks/virtualmachinesnapshot/mutation/virtualmachinesnapshot_mutator.go
aruneshpa: Yeah, I am going to file a story to handle this.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -143,6 +144,32 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
+
+	var allErrs field.ErrorList
+	vmNameLabelPath := field.NewPath("metadata", "labels").Key(vmopv1.VMNameForSnapshotLabel)
+
+	if oldVMSnapshot != nil && oldVMSnapshot.Labels[vmopv1.VMNameForSnapshotLabel] != "" {
+		// Since we are only validating this during
+		// Update, the new VM is always guaranteed to have
+		// a non nil label with the VM name set.
+		newVMNameLabel := vmSnapshot.Labels[vmopv1.VMNameForSnapshotLabel]
+		oldVMNameLabel := oldVMSnapshot.Labels[vmopv1.VMNameForSnapshotLabel]
```
bryanv: Maybe for readability, store the earlier lookup so don't do it again here

bryanv (approved):

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
lubronzhan: Or no need to check `oldVMSnapshot.Labels[vmopv1.VMNameForSnapshotLabel] == ""` since it's impossible to be ""

webhooks/virtualmachinesnapshot/mutation/virtualmachinesnapshot_mutator_intg_test.go
```diff
@@ -0,0 +1,79 @@
...
+	BeforeEach(func() {
+		ctx = newIntgMutatingWebhookContext()
+		vmSnapshot = ctx.vmSnapshot.DeepCopy()
+	})
+	AfterEach(func() {
+		ctx = nil
+	})
+
+	Describe("mutate", func() {
+		Context("basic webhook functionality", func() {
```
palnabarun, line 64: Can we have an integration test as well which checks if the object was correctly mutated and stamped with the label?

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -143,6 +144,32 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
+// of the snapshot resource. This validation is only applicable during
+// an update operation.
+func (v validator) validateImmutableVMNameLabel(
+	ctx *pkgctx.WebhookRequestContext,
+	vmSnapshot, oldVMSnapshot *vmopv1.VirtualMachineSnapshot) field.ErrorList {
+
+	var allErrs field.ErrorList
+	vmNameLabelPath := field.NewPath("metadata", "labels").Key(vmopv1.VMNameForSnapshotLabel)
+
+	if oldVMSnapshot != nil && oldVMSnapshot.Labels[vmopv1.VMNameForSnapshotLabel] != "" {
```
palnabarun: Is this check required? This only gets called on Update operations. AFAIU, oldVMSnapshot will not be nil on update operations.

Same for the label check, if the object was mutated and validated on create correctly, the label will always be set.

aruneshpa: Just trying to be defensive when coding.  Did not want to assume the condition of the object and since nil check isn't expensive, I think it is okay to keep it?

aruneshpa: Three people commented the same thing, so I updated the code.

PR #1139, opened 2025-08-29
✨Add snapshot condition report

lubronzhan:
Add snapshot condition report.
```
// indicate whether snapshot is ready
- VirtualMachineSnapshotReadyCondition
     // indicate whether snapshot is created
    - VirtualMachineSnapshotCreatedCondition
     // indicate whether csi volume sync has been done for this snapshot
    - VirtualMachineSnapshotCSIVolumeSyncedCondition
```

For `VirtualMachineSnapshotCreatedCondition`
1. Not setting it by default
2. When reconcileSnapshot sees the snapshot has `VirtualMachineSnapshotCreatedCondition` set to true, then skip
3. When snapshot is being created, set reason to `in progress`
4. If creation failed, set reason to `failed`
5. Set to true once snapshot is created.
6. ReconcileSnapshotRevert will only process the snapshot if it has `VirtualMachineSnapshotReadyCondition` set to true, means it needs to have `VirtualMachineSnapshotCSIVolumeSyncedCondition` to true as well.

For `VirtualMachineSnapshotCSIVolumeSyncedCondition`
1. Not setting it by default
2. Set to `requested` once VMSnapshot controller sets the csi sync annotation, only after `VirtualMachineSnapshotCreatedCondition` is true
3. Set to true once the annotation is set to completed

For `VirtualMachineSnapshotReadyCondition`
Only set to true when both `VirtualMachineSnapshotCreatedCondition` and `VirtualMachineSnapshotCSIVolumeSyncedCondition` are ready

Besides above:
1. Refactored some tests as well.
2. Also remove some of the vmCtx.Logger.Error

```release-note
Add snapshot condition report
```

### DIFF
--- a/api/v1alpha5/virtualmachinesnapshot_types.go
+++ b/api/v1alpha5/virtualmachinesnapshot_types.go
@@ -79,11 +79,41 @@ type QuiesceSpec struct {
 const (
 	// VirtualMachineSnapshotReadyCondition represents the condition
 	// that the virtual machine snapshot is ready.
+	// It's ready only when both VirtualMachineSnapshotCreatedCondition
+	// and VirtualMachineSnapshotCSIVolumeSyncedCondition are true.
 	VirtualMachineSnapshotReadyCondition = "VirtualMachineSnapshotReady"
 
-	// VirtualMachineSnapshotInProgressReason represents that a
-	// snapshot is in progress.
-	VirtualMachineSnapshotInProgressReason = "VirtualMachineSnapshotInProgress"
+	// VirtualMachineSnapshotWaitingForCreationReason documents that the virtual machine
+	// snapshot is waiting for creation.
+	VirtualMachineSnapshotWaitingForCreationReason = "VirtualMachineSnapshotWaitingForCreation"
+
+	// VirtualMachineSnapshotWaitingForCSISyncReason documents that the virtual machine
+	// snapshot is waiting for CSI volume sync.
+	VirtualMachineSnapshotWaitingForCSISyncReason = "VirtualMachineSnapshotWaitingForCSISync"
+)
+
+const (
+	// VirtualMachineSnapshotCreatedCondition exposes the status of
+	// the virtual machine snapshot creation.
+	VirtualMachineSnapshotCreatedCondition = "VirtualMachineSnapshotCreated"
+
+	// VirtualMachineSnapshotCreationInProgressReason documents that the
+	// virtual machine snapshot creation is in progress.
+	VirtualMachineSnapshotCreationInProgressReason = "VirtualMachineSnapshotCreationInProgress"
+
+	// VirtualMachineSnapshotCreationFailedReason documents that the virtual machine
+	// snapshot creation has failed.
+	VirtualMachineSnapshotCreationFailedReason = "VirtualMachineSnapshotCreationFailed"
+)
+
+const (
+	// VirtualMachineSnapshotCSIVolumeSyncedCondition expose the status of
+	// syncing CSI volume for this virtual machine snapshot.
+	VirtualMachineSnapshotCSIVolumeSyncedCondition = "VirtualMachineSnapshotCSISynced"
+
+	// VirtualMachineSnapshotCSIVolumeSyncInProgressReason documents that the
+	// CSI volume sync is in progress for this virtual machine snapshot.
+	VirtualMachineSnapshotCSIVolumeSyncInProgressReason = "VirtualMachineSnapshotCSISyncInProgress"
 )
 
 // VirtualMachineSnapshotStatus defines the observed state of VirtualMachineSnapshot.
@@ -170,11 +200,11 @@ type VirtualMachineSnapshot struct {
 	Status VirtualMachineSnapshotStatus `json:"status,omitempty"`
 }
 
-func (vmSnapshot *VirtualMachineSnapshot) NamespacedName() string {
+func (vmSnapshot VirtualMachineSnapshot) NamespacedName() string {
 	return vmSnapshot.Namespace + "/" + vmSnapshot.Name
 }
 
-func (vmSnapshot *VirtualMachineSnapshot) GetConditions() []metav1.Condition {
+func (vmSnapshot VirtualMachineSnapshot) GetConditions() []metav1.Condition {
 	return vmSnapshot.Status.Conditions
 }
 
--- a/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
+++ b/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
@@ -18,15 +18,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
-	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
+	pkgcnd "github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	"github.com/vmware-tanzu/vm-operator/pkg/constants"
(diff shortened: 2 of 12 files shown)


api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -73,11 +73,33 @@ type QuiesceSpec struct {
...
+	// virtual machine snapshot creation is in progress.
+	VirtualMachineSnapshotInProgressReason = "VirtualMachineSnapshotCreationInProgress"
+
+	// VirtualMachineSnapshotFailedReason documents that the virtual machine
+	// snapshot creation has failed.
+	VirtualMachineSnapshotFailedReason = "VirtualMachineSnapshotFailed"
+
+	// VirtualMachineSnapshotCSIVolumeSyncCondition expose the status of
+	// syncing CSI volume for this virtual machine snapshot.
+	VirtualMachineSnapshotCSIVolumeSyncCondition = "VirtualMachineSnapshotCSISync"
```
palnabarun:
```suggestion
	// VirtualMachineSnapshotCSIVolumeSyncedCondition expose the status of
	// syncing CSI volume for this virtual machine snapshot.
	VirtualMachineSnapshotCSIVolumeSyncedCondition = "VirtualMachineSnapshotCSISynced"
```

Conditions typically are named in a way which denotes a final steady state.

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -73,11 +73,33 @@ type QuiesceSpec struct {
...
+
+	// VirtualMachineSnapshotCSIVolumeSyncRequestedReason documents that the
+	// CSI volume sync has been requested for this virtual machine snapshot.
+	VirtualMachineSnapshotCSIVolumeSyncRequestedReason = "VirtualMachineSnapshotCSISyncRequested"
```
palnabarun: How does `VirtualMachineSnapshotCSISyncInProgress` sound as the reason?

lubronzhan: I named it requested, since the annotation also use the `requested` as the value. But yeah, inProgress should be more readable for the user

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -319,6 +297,105 @@ func ensureCSIVolumeSyncAnnotation(vmSnapshot *vmopv1.VirtualMachineSnapshot) {
 	}
 	// As long as the value is not set to completed by CSI driver, we need to mark it as requested.
 	if vmSnapshot.Annotations[constants.CSIVSphereVolumeSyncAnnotationKey] != constants.CSIVSphereVolumeSyncAnnotationValueCompleted {
-		vmSnapshot.Annotations[constants.CSIVSphereVolumeSyncAnnotationKey] = constants.CSIVSphereVolumeSyncAnnotationValueRequest
+		vmSnapshot.Annotations[constants.CSIVSphereVolumeSyncAnnotationKey] = constants.CSIVSphereVolumeSyncAnnotationValueRequested
+	}
+}
+
+func setSnapshotReadyCondition(vmSnapshot *vmopv1.VirtualMachineSnapshot) {
```
aruneshpa: maybe call it reconcileSnapshotConditions since you are essentially manipulating three conditions here, Ready being one of them.

api/v1alpha5/virtualmachinesnapshot_types.go
```diff
@@ -73,11 +73,45 @@ type QuiesceSpec struct {
...
-	// VirtualMachineSnapshotInProgressReason represents that a
-	// snapshot is in progress.
-	VirtualMachineSnapshotInProgressReason = "VirtualMachineSnapshotInProgress"
+	// VirtualMachineSnapshotWaitingForCreationReason documents that the virtual machine
+	// snapshot is waiting for creation.
+	VirtualMachineSnapshotWaitingForCreationReason = "VirtualMachineSnapshotWaitingForCreation"
+
+	// VirtualMachineSnapshotWaitingForCSISyncReason documents that the virtual machine
+	// snapshot is waiting for CSI volume sync.
+	VirtualMachineSnapshotWaitingForCSISyncReason = "VirtualMachineSnapshotWaitingForCSISync"
```
aruneshpa, line 92: Can you please group these together so all the creation condition and reasons are together in a block.  Ditto for the CSI condition and reasons.

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -190,49 +191,26 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
...
 	}
-	// TODO(lubron): Now the format will very likely to fall back to decimalSI
-	// since the size is int64 and it's not a multiple of 1024.
-	// We can think of refactoring to show a more humain readble value without loosing
-	// the precision in the future.
-	total := kubeutil.BytesToResource(size)
 
-	if vmSnapshot.Status.Storage.Used == nil {
-		vmSnapshot.Status.Storage.Used = resource.NewQuantity(0, resource.BinarySI)
+	if err := r.calculateUsedCapacity(ctx); err != nil {
```
aruneshpa, line 216: Can you remind me when we are decrementing the requested capacity?  We should do that when the CSI sync is done (creation is not sufficient).

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -319,6 +297,105 @@ func ensureCSIVolumeSyncAnnotation(vmSnapshot *vmopv1.VirtualMachineSnapshot) {
...
+	}
+}
+
+func setSnapshotReadyCondition(vmSnapshot *vmopv1.VirtualMachineSnapshot) {
+	// If the snapshot is being deleted, skip setting the condition.
+	if !vmSnapshot.DeletionTimestamp.IsZero() {
+		return
 	}
+
+	if pkgcnd.Get(vmSnapshot, vmopv1.VirtualMachineSnapshotCreatedCondition) == nil {
```
aruneshpa: Hmm.  Since we are manipulating the conditions on the snapshot resource from the VM provider code anyway, would the following workflow be simpler:
- Set the "SnapshotCreationInProgress" from the VM provider when the snapshot is being processed.  We don't set any conditions before then
- If creation fails, the VM provider sets the created to fails with failed reason
- If creation succeeds, then the snapshot controller takes over by applying the CSI annotation and setting the CSI sync condition.  We can also use the created condition to update the storage (like you are already doing).
Thoughts?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: In SPU controller, it will not include Snapshot as part of requested capacity if it has CSIVolumeSync annotation set to `Completed`.

On VMSnapshot, we record both reserved and usage, so it would be easier for SPU controller to calculate the total number

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
lubronzhan: Ok, I read from somewhere it's a good practice to always have a condition. Let me change that

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
aruneshpa: Got it.  So, we clear out (decrement) the reserved capacity in in the SPU if the snapshot is completed.  Thanks!

aruneshpa (approved):

lubronzhan:
Found a bug in CSI, if merge this, then can't test revert in new VC build, since Snapshot won't be `ready` state. It might impact how VCFA test this feature

lubronzhan:
Ok gonna merge this, since above bug only impacts if the user have PVC attached to their VM

PR #1140, opened 2025-08-29
🌱  Emit a condition for a failed snapshot revert operation

palnabarun:

When a snapshot revert operation fails, we now emit a condition on the VirtualMachine object to indicate the failure. This condition provides details about the failure, allowing users to understand the reason behind the unsuccessful revert attempt.

Once the revert operation fails, users will see a condition named `VirtualMachineSnapshotRevertSucceeded` set to `False` on the VirtualMachine object. The condition's message will contain information about the error encountered during the revert process.

Also, enhances documentations for the `VirtualMachineSnapshotRevertInProgressAnnotationKey` to clarify its behavior in failure scenarios.

None

### Manual testing is done on a VC
1. A VM is created with a PVC, and then two VM snapshots were created in succession `sn-1 -> sn-2`. Such were created to have a few revert points to revert to.
2. With that running state, a CSI `VolumeSnapshot` was taken of the PVC.
3. Then another VM snapshot was taken `sn-3` to create another revert point for testing.

Now with the VM running with a state after `sn-3`, we attempt a revert to `sn-1`. This fails with a resulting state of VM.

```
root@420ff86af566197915ddbd611ec9ccab [ ~ ]# k get vm vm -o yaml
kind: VirtualMachine
metadata:
  annotations:
    ...
    vmoperator.vmware.com/snapshot-revert-in-progress: ""
    ...
  creationTimestamp: "2025-08-28T10:57:24Z"
  finalizers:
  - vmoperator.vmware.com/virtualmachine
  generation: 6
  labels:
    topology.kubernetes.io/zone: domain-c36
  name: vm
  namespace: nabarun
  resourceVersion: "3243679"
  uid: a92fa694-efc5-431b-aeb9-ad046b5fbda9
spec:
  ...
  currentSnapshot:
    kind: VirtualMachineSnapshot
    name: sn-1
  ...
  volumes:
  - name: data-disk
    persistentVolumeClaim:
      claimName: vm-data-disk
status:
  ...
  conditions:
  ...
  - lastTransitionTime: "2025-08-29T06:12:59Z"
    message: 'An error occurred while reverting to a snapshot: The specified snapshot
      is not revertable. The virtual machine cannot be reverted when crossing a CSI
      VolumeSnapshot. Please delete the VolumeSnapshots and retry the revert.'
    reason: VirtualMachineSnapshotRevertInvalid
    status: "False"
    type: VirtualMachineSnapshotRevertSucceeded
  ...
  currentSnapshot:
    kind: VirtualMachineSnapshot
    name: sn-3
  ...
```

And the following events were in the event log.

```
root@420ff86af566197915ddbd611ec9ccab [ ~ ]# k get events
LAST SEEN   TYPE      REASON          OBJECT              MESSAGE
16m         Normal    UpdateSuccess   virtualmachine/vm   Update success
85s         Normal    UpdateSuccess   virtualmachine/vm   Update success
85s         Warning   UpdateFailure   virtualmachine/vm   failed to reconcile snapshot revert: The operation is not allowed in the current state.
```

### vcsim tests

Future TODO: To add a way in govmomi to inject faults and check for them.

```release-note
🌱  Emit a condition for a failed snapshot revert operation
```

### DIFF
--- a/api/v1alpha5/virtualmachine_types.go
+++ b/api/v1alpha5/virtualmachine_types.go
@@ -59,6 +59,32 @@ const (
 )
 
 const (
+	// VirtualMachineSnapshotRevertSucceeded indicates that the VM
+	// has been reverted to a snapshot.
+	VirtualMachineSnapshotRevertSucceeded = "VirtualMachineSnapshotRevertSucceeded"
+
+	// VirtualMachineSnapshotRevertInProgressReason indicates that the
+	// revert operation is in progress.
+	VirtualMachineSnapshotRevertInProgressReason = "VirtualMachineSnapshotRevertInProgress"
+
+	// VirtualMachineSnapshotRevertTaskFailedReason indicates that the
+	// revert operation is invalid.
+	VirtualMachineSnapshotRevertTaskFailedReason = "VirtualMachineSnapshotRevertTaskFailed"
+
+	// VirtualMachineSnapshotRevertFailedInvalidVMManifestReason indicates
+	// that the revert operation has failed due to invalid VM spec to revert to.
+	VirtualMachineSnapshotRevertFailedInvalidVMManifestReason = "VirtualMachineSnapshotRevertFailedInvalidVMManifest"
+
+	// VirtualMachineSnapshotRevertSkippedReason indicates that the
+	// revert operation was skipped.
+	VirtualMachineSnapshotRevertSkippedReason = "VirtualMachineSnapshotRevertSkipped"
+
+	// VirtualMachineSnapshotRevertFailedReason indicates that the
+	// revert operation failed for some reason.
+	VirtualMachineSnapshotRevertFailedReason = "VirtualMachineSnapshotRevertFailed"
+)
+
+const (
 	// GuestBootstrapCondition exposes the status of guest bootstrap from within
 	// the guest OS, when available.
 	GuestBootstrapCondition = "GuestBootstrap"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -162,6 +162,14 @@ const (
 	// VirtualMachineSnapshotRevertInProgressAnnotationKey is the
 	// annotation key to indicate that a VM snapshot revert is in
 	// progress.
+	//
+	// This annotation is set when a snapshot revert operation is initiated
+	// and is removed once the operation completes successfully.
+	//
+	// If the revert operation fails, this annotation may remain on the VM,
+	// indicating that the revert is still in progress. In such cases, manual
+	// intervention will be required to remove the annotation and retry the
+	// revert operation.
 	VirtualMachineSnapshotRevertInProgressAnnotationKey = "vmoperator.vmware.com/snapshot-revert-in-progress"
 
 	// VMProvKeepDisksExtraConfigKey is set in ExtraConfig with the value a
--- a/pkg/providers/vsphere/vmprovider_vm_snapshot.go
+++ b/pkg/providers/vsphere/vmprovider_vm_snapshot.go
@@ -276,7 +276,16 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevertDoTask(
 
 	// Skip snapshot revert processing for VKS/TKG nodes.
 	if kubeutil.HasCAPILabels(vmCtx.VM.Labels) {
-		logger.V(4).Info("Skipping revert for VKS node")
+		message := "Skipping snapshot revert for VKS node"
+
+		pkgcnd.MarkFalse(vmCtx.VM,
+			vmopv1.VirtualMachineSnapshotRevertSucceeded,
+			vmopv1.VirtualMachineSnapshotRevertSkippedReason,
+			"%s",
+			message,
+		)
+
+		logger.V(4).Info(message)
 		return nil
 	}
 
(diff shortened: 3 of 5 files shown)


pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -48,6 +48,20 @@ func errToConditionReasonAndMessage(err error) (string, string) {
 	}
 }
 
+func errorMessageFromTaskInfoWithKey(taskInfo *vimtypes.TaskInfo, key string) string {
+	if taskInfo == nil || taskInfo.Error == nil || taskInfo.Error.Fault == nil {
+		return ""
+	}
+
+	for _, fm := range taskInfo.Error.Fault.GetMethodFault().FaultMessage {
```
lubronzhan: GetMethodFault() returns an interface, do you want to check if it's nil as well

pkg/constants/constants.go
```diff
@@ -162,6 +162,14 @@ const (
...
 	// progress.
+	//
+	// This annotation is set when a snapshot revert operation is initiated
+	// and is removed once the operation completes successfully.
+	//
+	// If the revert operation fails, this annotation may remain on the VM,
+	// indicating that the revert is still in progress. In such cases, manual
+	// intervention will be required to remove the annotation and retry the
+	// revert operation.
 	VirtualMachineSnapshotRevertInProgressAnnotationKey = "vmoperator.vmware.com/snapshot-revert-in-progress"
```
lubronzhan, line 173: IC, so this annotation is needed if the user wants to retry the revert. Still it would be best if you could also use your new condition to show the revert progress if it's in progress or it's completed.

Because with current logic, once a VM revert fail, then he retry, there is no logic to remove the condition `VirtualMachineSnapshotRevertSucceeded: false`. You should ensure it's set to true if revert succeed at least.

palnabarun: Added a state to denote the in-progress revert.

> Because with current logic, once a VM revert fail, then he retry, there is no logic to remove the condition VirtualMachineSnapshotRevertSucceeded: false. You should ensure it's set to true if revert succeed at least.

When the revert succeeds, the status is wiped. So, the condition would be wiped.

api/v1alpha5/virtualmachine_types.go
```diff
@@ -56,6 +56,15 @@ const (
 	// VirtualMachineClassConfigurationSynced indicates that the VM's current configuration is synced to the
 	// current version of its VirtualMachineClass.
 	VirtualMachineClassConfigurationSynced = "VirtualMachineClassConfigurationSynced"
+
+	// VirtualMachineSnapshotRevertSucceeded indicates that the VM has been reverted to a snapshot.
+	VirtualMachineSnapshotRevertSucceeded = "VirtualMachineSnapshotRevertSucceeded"
+
+	// VirtualMachineSnapshotRevertInvalidReason indicates that the revert operation is invalid.
+	VirtualMachineSnapshotRevertInvalidReason = "VirtualMachineSnapshotRevertInvalid"
```
aruneshpa: Invalid could mean a lot of things (e.g., invalid YAML payload in the snapshot, invalid / corrupted snapshot etc).

What do you think about calling it `SnapshotRevertTaskFailedReason`?

pkg/providers/vsphere/vmprovider_vm_utils.go
```diff
@@ -48,6 +48,25 @@ func errToConditionReasonAndMessage(err error) (string, string) {
 	}
 }
 
+func errorMessageFromTaskInfoWithKey(taskInfo *vimtypes.TaskInfo, key string) string {
```
aruneshpa, line 51: Just double checking, are there any helper messages for this in govmomi?

pkg/constants/constants.go
lubronzhan: IC forgot this https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmprovider_vm_snapshot.go#L440

pkg/providers/vsphere/vmprovider_vm_utils.go
lubronzhan: Saw old code doing similar things here https://github.com/vmware-tanzu/vm-operator/blob/4578f25e5bd528fdcc6c1d1f6aa32166944e748e/pkg/providers/vsphere/resources/vm.go#L241-L250

Maybe could be combined.

lubronzhan (approved):
LGTM!

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -390,6 +400,53 @@ func (vs *vSphereVMProvider) performSnapshotRevert(
...
+			conditionMessage += " The virtual machine cannot be reverted when crossing a CSI VolumeSnapshot. " +
+				"VolumeSnapshots should be deleted before retrying the revert."
+		}
+
+		// The condition reason and message are based on the error encountered.
+		// The condition would be reset when a revert is successful, since the Status is wiped out after a revert.
+		pkgcnd.MarkError(vmCtx.VM,
+			vmopv1.VirtualMachineSnapshotRevertSucceeded,
+			vmopv1.VirtualMachineSnapshotRevertInvalidReason,
+			fmt.Errorf("%s", conditionMessage),
```
bryanv: `MarkError` is really intended to be used when you already have an error to save a little boilerplate - use `MarkFalse` here instead

api/v1alpha5/virtualmachine_types.go
palnabarun: Done. And added more reasons to the revert lifecycle.

pkg/providers/vsphere/vmprovider_vm_snapshot.go
palnabarun: Done.

pkg/providers/vsphere/vmprovider_vm_utils.go
palnabarun: The above code in [vm-operator/pkg/providers/vsphere/resources/vm.go](https://github.com/vmware-tanzu/vm-operator/blob/4578f25e5bd528fdcc6c1d1f6aa32166944e748e/pkg/providers/vsphere/resources/vm.go#L241-L250) doesn't go beyond the FaultMessage itself. In snapshot's case, we need to go a little deeper to see if it actually failed due to crossing a CSI Snapshot while attempting the revert, by checking the message key.

lubronzhan: Yeah make sense, these are two different cases.

api/v1alpha5/virtualmachine_types.go
```diff
@@ -58,6 +58,32 @@ const (
...
+	// revert operation is in progress.
+	VirtualMachineSnapshotRevertInProgressReason = "VirtualMachineSnapshotRevertInProgress"
+
+	// VirtualMachineSnapshotRevertTaskFailedReason indicates that the
+	// revert operation is invalid.
+	VirtualMachineSnapshotRevertTaskFailedReason = "VirtualMachineSnapshotRevertTaskFailed"
+
+	// VirtualMachineSnapshotRevertFailedInvalidYAMLReason indicates
+	// that the revert operation has failed due to invalid VM spec to revert to.
+	VirtualMachineSnapshotRevertFailedInvalidYAMLReason = "VirtualMachineSnapshotRevertFailedInvalidYAML"
```
aruneshpa: Can you change this to somthing generic?  Like "InvalidVMManifest"?  We are talking about changing the stored YAML to JSON and I don't want the Condition to be affected by that change.

aruneshpa (approved):
LGTM with a suggestion.

PR #1142, opened 2025-08-30
🌱 Fix up snapshot-related vm provider code

akutz:

This patch fixes several issues in the snapshot-related code in the VM provider:

* Adhere to the file's preferred line-length, 80.
* Use pkgerr.RequeueErr instead of returning a tuple, like all the other reconcile functions do in the updateVirtualMachine function.
* Moves the business logic of snapshots into the snapshot helper functions instead of updateVirtualMachine.
* Moves the snapshot functions into vmprovider_vm_snapshot.go which was seemingly created for this purpose.
* Returns wrapped errors instead of logging errors in-place and then returning the inner errors.
* Updates the PatchSnapshotSuccessStatus function in kubeutil to not use a VirtualMachineContext as no other file in the entire kubeutil package uses anything but context.Context, by design.
* Uses logr.Logger's Info key/value pattern instead of using Sprintf.
* Uses the contextual logger and its WithValues function to ensure contextual data is always logged on the call stack, ex. adding the desired snapshot name to the logger via WithValues.
* Moved all logging to V(4) or less since anything above this incurs K8s logging and is too noisy.

Fixes `NA`

```release-note
NONE
```


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -869,48 +883,28 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+		if err := vs.reconcileSnapshotRevert(vmCtx, vcVM); err != nil {
 			if pkgerr.IsNoRequeueError(err) {
 				return errOrReconcileErr(reconcileErr, err)
 			}
 			reconcileErr = getReconcileErr("snapshot revert", reconcileErr, err)
-		} else if requeue {
-			// Snapshot revert was successful and requires requeue for next reconcile
-			// to operate on the updated VM spec. Return any accumulated reconcile errors.
-			vmCtx.Logger.V(4).Info("Snapshot revert completed successfully, requeuing reconcile")
-			return reconcileErr
```
dilyar85: Nit: combine these together since they're both related to snapshots (it also avoids duplicate checks on the same feature)? Something like:
```suggestion
	//
	// 6. Reconcile snapshot task and revert
	//
	if pkgcfg.FromContext(vmCtx).Features.VMSnapshots {
		if err := vs.reconcileSnapshotTask(vmCtx, vcVM); err != nil {
			return errOrReconcileErr(reconcileErr, err)
		}

		if err := vs.reconcileSnapshotRevert(vmCtx, vcVM); err != nil {
			if pkgerr.IsNoRequeueError(err) {
				return errOrReconcileErr(reconcileErr, err)
			}
			reconcileErr = getReconcileErr("snapshot revert", reconcileErr, err)
		}
	}
```

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -936,13 +930,15 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
-	// Reconcile the current snapshot of this VM.
+	//
+	// 10. Reconcile current snapshot
+	//
 	if pkgcfg.FromContext(vmCtx).Features.VMSnapshots {
-		if err := vs.reconcileSnapshot(vmCtx, vcVM); err != nil {
+		if err := vs.reconcileCurrentSnapshot(vmCtx, vcVM); err != nil {
 			if pkgerr.IsNoRequeueError(err) {
 				return err
 			}
```
dilyar85: Not your change but looks like all other places have `return errOrReconcileErr(reconcileErr, err)` in this case. Should this be updated too for consistency?

pkg/providers/vsphere/vmprovider_vm.go
akutz: Let me think about that. Thanks!

pkg/providers/vsphere/vmprovider_vm.go
akutz: Yes, it should be. Thank you!

PR #1143, opened 2025-09-02
 🌱 Verify  if a VM Template is Associated with a VM Publish Request

zhangheliu:
The PR adds the VM publish request's UUID to the cloned VM template's extra configurations, so it could be used to verify if the VM template is associated with a VM publish request. The UUID extra configuration is removed when the VM template is used for creating a new VM. 

Fixes # N/A

More integration tests will be added once the final end-to-end flow for publishing an inventory VM is complete.  

```release-note
N/A
```

### DIFF
--- a/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
+++ b/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vapi/library"
+	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 
 	imgregv1a1 "github.com/vmware-tanzu/image-registry-operator-api/api/v1alpha1"
@@ -459,7 +460,7 @@ func (r *Reconciler) checkIsTargetValid(ctx *pkgctx.VirtualMachinePublishRequest
 		}
 		contentLibraryType = contentLibraryV1A2.Spec.Type
 	}
-	ctx.Logger.Info("target location has library type: %q", contentLibraryType)
+	ctx.Logger.Info("target location type", "LibraryType", contentLibraryType)
 
 	item, err := r.getItemByName(ctx, contentLibraryType, string(ctx.ContentLibrary.Spec.UUID), targetItemName)
 
@@ -754,8 +755,10 @@ func (r *Reconciler) getUploadedItemID(ctx *pkgctx.VirtualMachinePublishRequestC
 	return parseItemIDFromTaskResult(task.Result)
 }
 
-// isItemCorrelatedWithVMPub checks if the item is correlated with this VM Publish request
-// by checking the item description. We add the vmPub uuid to the target item description when publishing.
+// isItemCorrelatedWithVMPub checks if the item is correlated with this VM
+// Publish request. It checks item's description for vmPub UUID if the item
+// is a CLS supported library item. It checks item's extraConfig for vmPub UUID
+// if the item is a VM template. The vmPub UUID is added during publishing.
 //
 // Currently, there is no easy way to link published item to the VM publish request
 // other than activation ID.
@@ -769,8 +772,22 @@ func (r *Reconciler) isItemCorrelatedWithVMPub(
 
 	switch tItem := item.(type) {
 	case *object.VirtualMachine:
-		// TODO(abaruni) Verify if VM is associated with this publish request
-		return tItem.Reference().Value, true, nil
+		var o mo.VirtualMachine
+		if err := tItem.Properties(ctx,
+			tItem.Reference(),
+			[]string{"config.extraConfig"},
+			&o); err != nil {
+			return "", false, err
+		}
+		for _, opt := range o.Config.ExtraConfig {
+			if optKV := opt.GetOptionValue(); optKV != nil &&
+				optKV.Key == vmopv1.VirtualMachinePublishRequestUUIDLabelKey {
+				if optKV.Value.(string) == string(ctx.VMPublishRequest.UID) {
+					return tItem.Reference().Value, true, nil
+				}
+				return "", false, nil
+			}
+		}
 	case *library.Item:
 		if tItem.Description != nil {
 			descriptions := itemDescriptionReg.FindStringSubmatch(*tItem.Description)
--- a/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
+++ b/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
@@ -732,34 +732,6 @@ func unitTestsReconcile() {
 						vmopv1.VirtualMachinePublishRequestConditionUploaded)).To(BeTrue())
 				})
 			})
-
-			When("Publishing to inventory", func() {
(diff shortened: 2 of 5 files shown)


controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -769,8 +772,24 @@ func (r *Reconciler) isItemCorrelatedWithVMPub(
...
+		if err := tItem.Properties(ctx,
+			tItem.Reference(),
+			[]string{"config.extraConfig"},
+			&o); err != nil {
+			return "", false, err
+		}
+		for _, opt := range o.Config.ExtraConfig {
+			if optKV := opt.GetOptionValue(); optKV != nil &&
+				optKV.Key == vmopv1.VirtualMachinePublishRequestUUIDLabelKey &&
+				optKV.Value.(string) == string(ctx.VMPublishRequest.UID) {
```
bryanv: There shouldn't be duplicate keys, so we can return true/false as soon as we find the key

zhangheliu: good catch, addressed. thanks~

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/vmware/govmomi/vim25/mo"
```
abaruni: can move this import down with the other govmomi packages

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -459,7 +460,7 @@ func (r *Reconciler) checkIsTargetValid(ctx *pkgctx.VirtualMachinePublishRequest
 		}
 		contentLibraryType = contentLibraryV1A2.Spec.Type
 	}
-	ctx.Logger.Info("target location has library type: %q", contentLibraryType)
+	ctx.Logger.Info("target location type", "LibraryTypeContentLibrary", contentLibraryType)
```
abaruni: LibraryTypeContentLibrary is a type of ContentLibrary. see https://github.com/vmware-tanzu/image-registry-operator-api/blob/main/api/v1alpha2/contentlibrary_types.go#L116-L124

maybe we can change that to just `LibraryType`. good catch on the incorrect usage here. thanks

pkg/providers/vsphere/virtualmachine/publish.go
```diff
@@ -95,7 +93,13 @@ func CloneVM(
 		},
 		Template: true,
 		Config: &vimtypes.VirtualMachineConfigSpec{
-			Annotation: descriptionPrefix + vmPubReq.Status.TargetRef.Item.Description,
+			Annotation: vmPubReq.Status.TargetRef.Item.Description,
```
abaruni, line 96: If we are going to use the ExtraConfig to associate the template with the VM Publish Request, then we can remove the annotation here. It's kind of an either/or. Either we do the association with the annotation, or we do it with the ExtraConfig

pkg/providers/vsphere/vmlifecycle/create_clone.go
```diff
@@ -78,6 +80,13 @@ func createCloneSpec(
 		}
 	}
 
+	// If a VM template is cloned via a VM publish request, the VM publish
+	// request UUID is added to the extraConfig. The UUID is removed when the
+	// template is used to clone a new VM.
+	cloneSpec.Config.ExtraConfig = pkgutil.OptionValues(
+		cloneSpec.Config.ExtraConfig).Delete(
+		vmopv1.VirtualMachinePublishRequestUUIDLabelKey)
+
```
abaruni: There isn't any harm in doing this, but the calling function https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmlifecycle/create_clone.go#L22-L54 is not actually invoked for the inventory-backed content library flow. For vm-backed images we use https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmlifecycle/create.go#L59-L63

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
```diff
@@ -748,15 +748,11 @@ func unitTestsReconcile() {
...
-						Expect(err).NotTo(HaveOccurred())
-
-						// Update TargetValid condition.
-						Expect(conditions.IsTrue(vmpub,
-							vmopv1.VirtualMachinePublishRequestConditionTargetValid)).To(BeTrue())
-						Expect(conditions.IsTrue(vmpub,
-							vmopv1.VirtualMachinePublishRequestConditionUploaded)).To(BeTrue())
+						Expect(err).To(HaveOccurred())
+						Expect(err.Error()).To(ContainSubstring(
+							"VM template with nil vim25 client"))
```
abaruni: i don't think this is necessary since the nil check for the client isn't necessay, but if it were, then we should add a test case in addition to keeping the current one.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -769,8 +772,26 @@ func (r *Reconciler) isItemCorrelatedWithVMPub(
 
 	switch tItem := item.(type) {
 	case *object.VirtualMachine:
-		// TODO(abaruni) Verify if VM is associated with this publish request
-		return tItem.Reference().Value, true, nil
+		if pkgutil.IsNil(tItem.Client()) {
+			return "", false, fmt.Errorf(
+				"found a same name VM template with nil vim25 client")
+		}
```
abaruni: This check isn't necessary. This shouldn't ever be the case unless we are going out of our way to nil out the client.

Since the item itself comes from https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmprovider.go#L453, the Reference that is constructed ultimately sets the client to be the same one used in the call to FindChild

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
zhangheliu: yeah, agree. we could remove those and have a story to follow up with end-to-end integ test once we hammer out the details needed to set the Complete condition to True for VCD CL. I will remove this and also the cient check

pkg/providers/vsphere/vmlifecycle/create_clone.go
zhangheliu: good catch. thanks

pkg/providers/vsphere/virtualmachine/publish.go
zhangheliu: dont we still need to store user-provided item description, e.gl "foo"? 


	// User-provided description of the virtual machine.
	//
	// Because this property
	// is optional in the virtual machine configuration, it is necessary
	// to pass an explicit empty string in a ConfigSpec object to remove an annotation
	// that is already present in the `VirtualMachineConfigInfo`
	// for a virtual machine.
	//
	// Reconfigure privilege: VirtualMachine.Config.Rename
	Annotation string `xml:"annotation,omitempty" json:"annotation,omitempty"`

lubronzhan (approved):
LGTM!

PR #1144, opened 2025-09-02
🌱 Fix filename typo

uucloud:

Fix filename typo: `rename pkg/config/capabilities/{capabilties.go => capabilities.go}`


dilyar85 (approved):
Thanks @uucloud for your contribution!

PR #1145, opened 2025-09-02
:sparkles: Add ability to validate quota on VirtualMachinePublishRequests for inventory-backed ContentLibraries

abaruni:

This PR adds validation logic for VirtualMachinePublishRequest when the target ContentLibrary is subject to quota enforcement.

Fixes #

```release-note
Add target validation for VirtualMachinePublishRequests when target ContentLibrary is subject to quota enforcement.
```


controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -458,6 +459,16 @@ func (r *Reconciler) checkIsTargetValid(ctx *pkgctx.VirtualMachinePublishRequest
 			return err
 		}
 		contentLibraryType = contentLibraryV1A2.Spec.Type
+
+		if contentLibraryType == imgregv1.LibraryTypeInventory && contentLibraryV1A2.Labels != nil {
```
zhangheliu: would metav1.HasLabel() be helpful here?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -511,6 +522,53 @@ func (r *Reconciler) getItemByName(ctx *pkgctx.VirtualMachinePublishRequestConte
...
+	// If this annotation is not present on the VM Publish Request, then we have not begun quota
+	// validation. We need to apply this annotation along with the requested capacity annotation
+	// and await processing.
+	if _, ok := vmPubReq.Annotations[pkgconst.AsyncQuotaPerformCheckAnnotationKey]; !ok {
+		vm := ctx.VM
+		if vm.Status.Storage != nil && vm.Status.Storage.Total != nil {
+			if vmPubReq.Annotations == nil {
+				vmPubReq.Annotations = make(map[string]string)
+			}
+			vmPubReq.Annotations[pkgconst.AsyncQuotaPerformCheckAnnotationKey] = "true"
```
zhangheliu, line 535: Should we have done this one-time operation in the mutating webhook?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -511,6 +522,53 @@ func (r *Reconciler) getItemByName(ctx *pkgctx.VirtualMachinePublishRequestConte
...
+			vmPubReq.Annotations[pkgconst.AsyncQuotaCheckRequestedCapacityAnnotationKey] = vm.Status.Storage.Total.String()
+
+			return fmt.Errorf("quota validation is needed for this request")
```
zhangheliu: Should we just log the message only if we intend to add the annotations in the controller?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -511,6 +522,53 @@ func (r *Reconciler) getItemByName(ctx *pkgctx.VirtualMachinePublishRequestConte
...
+		}
+		return fmt.Errorf("unable get storage total for VM %q for quota validation", vm.Name)
+	}
+	// Check for the existence of the requested-quota annotation. If this annotation is present, then we are
+	// currently awaiting quota verification, and we should halt any further processing.
+	if _, ok := vmPubReq.Annotations[pkgconst.AsyncQuotaCheckRequestedCapacityAnnotationKey]; ok {
+
+		return fmt.Errorf("quota validation has not completed for this request")
```
zhangheliu: same as above. not sure if we need to return an error for re-queues since annotation change will trigger reconciliation anyway

pkg/constants/constants.go
```diff
@@ -183,4 +183,28 @@ const (
 	// has completed the sync of the volume. This annotation is set by the CSI driver
 	// when the sync is completed.
 	CSIVSphereVolumeSyncAnnotationValueCompleted = "completed"
+
+	// AsyncQuotaPerformCheckAnnotationKey is the annotation key indicating that the ContentLibrary
+	// associated with a VirtualMachinePublishRequest is subject to VCFA quota approval.
+	AsyncQuotaPerformCheckAnnotationKey = "vmware-system-vcfa-storage-quota"
```
zhangheliu, line 192: Isn't this the CL label key? Do we need to insert this in the publish request?

pkg/constants/constants.go
```diff
@@ -183,4 +183,28 @@ const (
...
+
+	// AsyncQuotaCheckRequestedCapacityAnnotationKey is the annotation key added
+	// by the VirtualMachinePublishRequest mutation webhook to initiate quota validation. Adding
+	// this annotation will indicate that the controller must provide the requested capacity
+	// and block request fulfillment until further notice.
+	AsyncQuotaCheckRequestedCapacityAnnotationKey = "vcfa.storage.quota.check.vmware.com/requested-capacity"
```
zhangheliu, line 198: should we add a validating webhook to ensure it is not writable by unprivileged users?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: we want to return the error to halt processing at that point. if we don't, then we'll continue to attempting to publish, which we don't want

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: we want to return the error to halt processing at that point. if we don't, then we'll continue to attempting to publish, which we don't want. we want to ensure quota validation succeeds before we continue processing

pkg/constants/constants.go
abaruni: yes, this is a label key. good catch. we use this to indicate that the content library is subject to quota restrictions first and foremost. we also reuse it to indicate that the controller has marked this publish request as having initiated quota validation.

pkg/constants/constants.go
abaruni: this needed to be updated. the original intent was to have a mutation webhook handle adding this annotation. this has since been changed to remove the mutation webhook from the flow and have this handled solely by the controller. will update the comment to be accurate. good catch here

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: the mutation webhook has been removed from the flow. this now exists to signal the controller that we have initiated quota validation so that we don't get stuck in a loop.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: sure. it won't provide any functional benefit, but would prettify things a tad.

zhangheliu: I think metav1.HasLabel() checks for nil as well.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
zhangheliu: when returning error, the same publish request is re-queued. Shouldn't we just return nil to halt the processing? when there are annotation/label changes, the publish request will get queued again.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: it doesn't

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -511,6 +521,57 @@ func (r *Reconciler) getItemByName(ctx *pkgctx.VirtualMachinePublishRequestConte
...
+	// If this annotation is not present on the VM Publish Request, then we have not begun quota
+	// validation. We need to apply this annotation along with the requested capacity annotation
+	// and await processing.
+	if !metav1.HasAnnotation(vmPubReq.ObjectMeta, pkgconst.AsyncQuotaPerformCheckAnnotationKey) {
+		vm := ctx.VM
+		if vm.Status.Storage != nil && vm.Status.Storage.Total != nil {
+			if vmPubReq.Annotations == nil {
+				vmPubReq.Annotations = make(map[string]string)
+			}
+			vmPubReq.Annotations[pkgconst.AsyncQuotaPerformCheckAnnotationKey] = "true"
```
zhangheliu, line 535: qq: is it agreed that AsyncQuotaPerformCheckAnnotationKey is immutable?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -511,6 +521,57 @@ func (r *Reconciler) getItemByName(ctx *pkgctx.VirtualMachinePublishRequestConte
...
+			message := vmPubReq.Annotations[pkgconst.AsyncQuotaCheckMessageAnnotationKey]
+			err := pkgerr.NoRequeueError{
+				Message: fmt.Sprintf("quota validation has failed with message %q", message),
+			}
+			conditions.MarkError(vmPubReq,
+				vmopv1.VirtualMachinePublishRequestConditionTargetValid,
+				vmopv1.TargetContentLibraryFailedQuotaCheckReason,
+				err)
+
+			return err
```
zhangheliu, line 564: pkgerr.NoRequeueError?

zhangheliu (approved):

lubronzhan (approved):
LGTM!

PR #1146, opened 2025-09-02
🌱 Reverts a VM to the same power state as that of the snapshot

palnabarun:

For non-imported snapshots, so far the power state was being taken from the backed up VM YAML in ExtraConfig for the snapshot. This might not be the correct power state because, for example, in the case when one takes a snapshot of a Powered On VM without it's memory state, the snapshot power state would be off. The VM's resultant power state post revert should also be off.

None

```release-note
Reverts a VM to the same power state as that of the snapshot
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm_snapshot.go
+++ b/pkg/providers/vsphere/vmprovider_vm_snapshot.go
@@ -444,6 +444,11 @@ func (vs *vSphereVMProvider) restoreVMSpecFromSnapshot(
 	// Empty out the status.
 	vmCtx.VM.Status = vmopv1.VirtualMachineStatus{}
 
+	// reconcile the power state of the VM post revert
+	if err := vs.reconcilePostRevertPowerState(vmCtx, snapObj); err != nil {
+		return fmt.Errorf("failed to reconcile power state post revert: %w", err)
+	}
+
 	// Note that since we swapped out the annotations from the
 	// snapshot, it will implicitly remove the "restore-in-progress"
 	// annotation. But go ahead and clean it up again in case the
@@ -482,7 +487,7 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
 		// If the Snapshot has ImportedSnapshotAnnotation, then approximate the
 		// VM spec.
 		return getVMYamlFromSnapshotFromSynthesizedSpec(
-			vmCtx, snap, moSnap, snapRef)
+			vmCtx, snap, moSnap)
 	}
 
 	logger.V(4).Info("Found backup data in snapshot config",
@@ -501,8 +506,7 @@ func (vs *vSphereVMProvider) getVMYamlFromSnapshot(
 func getVMYamlFromSnapshotFromSynthesizedSpec(
 	vmCtx pkgctx.VirtualMachineContext,
 	snapCR *vmopv1.VirtualMachineSnapshot,
-	moSnap mo.VirtualMachineSnapshot,
-	snap *vimtypes.ManagedObjectReference) (string, error) {
+	moSnap mo.VirtualMachineSnapshot) (string, error) {
 
 	logger := pkglog.FromContextOrDefault(vmCtx)
 
@@ -535,28 +539,6 @@ func getVMYamlFromSnapshotFromSynthesizedSpec(
 			APIVersion: vmopv1.GroupVersion.String(),
 		},
 		Spec: synthesizeVMSpecForSnapshot(*vmCtx.VM, vmCtx.MoVM, moSnap),
-	}
-
-	// Inferred from the power state of the snapshot
-	// It doesn't matter what the power state of the VM is at the time of
-	// revert.
-	// The VM is always reverted in the same power state as the snapshot.
-	snapshot := vmlifecycle.FindSnapshotInTree(
-		vmCtx.MoVM.Snapshot.RootSnapshotList, snap.Value)
-	if snapshot == nil {
-		// weird situation here
-		// the snapshot being processed doesn't exist in the snapshot tree
-		// this shouldn't be happening
-		return "", fmt.Errorf(
-			"failed to find snapshot %q in tree", snap.Value)
-	}
-
-	// Convert from vimtypes.VirtualMachinePowerState to
-	// vmopv1.VirtualMachinePowerState
-	if snapshot.State == vimtypes.VirtualMachinePowerStatePoweredOn {
-		vm.Spec.PowerState = vmopv1.VirtualMachinePowerStateOn
-	} else {
-		vm.Spec.PowerState = vmopv1.VirtualMachinePowerStateOff
 	}
 
 	// Number of elements in the spec.network.interfaces list must be same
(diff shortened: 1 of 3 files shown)


pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -405,6 +405,14 @@ func (vs *vSphereVMProvider) restoreVMSpecFromSnapshot(
 	snapCR *vmopv1.VirtualMachineSnapshot,
 	snapObj *vimtypes.ManagedObjectReference) error {
 
+	snapshot := vmlifecycle.FindSnapshotInTree(vmCtx.MoVM.Snapshot.RootSnapshotList, snapObj.Value)
```
lubronzhan: Did this really happened? If not you can just move it inside reconcilePostRevertPowerState since it's only used there?

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3209,6 +3209,92 @@ func vmTests() {
...
+
+							task, err = vcVM.CreateSnapshot(ctx, secondSnapshot.Name, "second snapshot", false, false)
+							Expect(err).ToNot(HaveOccurred())
+							Expect(task.Wait(ctx)).To(Succeed())
+
+							// Create second snapshot CR
+							// Mark the snapshot as ready so that the snapshot workflow doesn't try to create it.
+							conditions.MarkTrue(secondSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition)
+							// Snapshot should be owned by the VM resource.
+							Expect(controllerutil.SetOwnerReference(&o, vmSnapshot, ctx.Scheme)).To(Succeed())
```
lubronzhan: This was done one line 3241.

pkg/providers/vsphere/vmprovider_vm_snapshot.go
palnabarun: I was refactoring this earlier as this call was not needed inside the synthesize method.

Now, all the power state revert happens at one single place.

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -819,3 +824,35 @@ func (vs *vSphereVMProvider) reconcileCurrentSnapshot(
...
+	// handle that case here.
+	switch snapshot.State {
+	case vimtypes.VirtualMachinePowerStatePoweredOn:
+		vmCtx.VM.Spec.PowerState = vmopv1.VirtualMachinePowerStateOn
+	case vimtypes.VirtualMachinePowerStatePoweredOff:
+		vmCtx.VM.Spec.PowerState = vmopv1.VirtualMachinePowerStateOff
+	case vimtypes.VirtualMachinePowerStateSuspended:
+		vmCtx.VM.Spec.PowerState = vmopv1.VirtualMachinePowerStateSuspended
+	default:
+		// Unknown state, log and default to off
```
aruneshpa, line 872: You are not just logging, but returning an error.  I agree that defaulting to powered off makes sense.

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3209,6 +3207,90 @@ func vmTests() {
 						Expect(currentSnap).ToNot(BeNil())
 						Expect(currentSnap.Name).To(Equal(vmSnapshot.Name))
 					})
+
+					Context("and the snapshot was taken when VM was powered on and is now powered off", func() {
+						It("should successfully power on the VM after reverting to a Snapshot in PoweredOn state", func() {
```
aruneshpa, line 3212: Hmnm.  I don't quite grok this.
You are creating two snapshots, both _without_ memory, so the power state of the snapshot should be powered off.  And therefore, when the VM is reverted, it should be in powered off state?

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -819,3 +824,35 @@ func (vs *vSphereVMProvider) reconcileCurrentSnapshot(
 
 	return nil
 }
+
+func (vs *vSphereVMProvider) reconcilePostRevertPowerState(
+	vmCtx pkgctx.VirtualMachineContext,
+	snapObj *vimtypes.ManagedObjectReference) error {
+
+	snapshot := vmlifecycle.FindSnapshotInTree(vmCtx.MoVM.Snapshot.RootSnapshotList, snapObj.Value)
```
bryanv: I think a later useful refactor would be for our `FindSnapshot` to return the `VirtualMachineSnapshotTree` instead of just its moRef so we don't need a later lookup here

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -440,6 +440,11 @@ func (vs *vSphereVMProvider) restoreVMSpecFromSnapshot(
 	// Empty out the status.
 	vmCtx.VM.Status = vmopv1.VirtualMachineStatus{}
 
+	// reconcile the power state of the VM post revert
+	if err := vs.reconcilePostRevertPowerState(vmCtx, snapObj); err != nil {
```
bryanv, line 448: Is the similar code in `getVMYamlFromSnapshotFromSynthesizedSpec` still needed?

palnabarun: Meant to remove that one. Thanks for catching it.

pkg/providers/vsphere/vmprovider_vm_snapshot.go
palnabarun: Good catch. I meant to log and return.

pkg/providers/vsphere/vmprovider_vm_test.go
palnabarun: Fixing this behaviour in vcsim. vcsim wasn't correctly setting snap power state memory=false.

palnabarun: Fixed with https://github.com/vmware/govmomi/pull/3853. Bringing in the govmomi update.

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -819,3 +801,35 @@ func (vs *vSphereVMProvider) reconcileCurrentSnapshot(
...
+	case vimtypes.VirtualMachinePowerStatePoweredOn:
+		vmCtx.VM.Spec.PowerState = vmopv1.VirtualMachinePowerStateOn
+	case vimtypes.VirtualMachinePowerStatePoweredOff:
+		vmCtx.VM.Spec.PowerState = vmopv1.VirtualMachinePowerStateOff
+	case vimtypes.VirtualMachinePowerStateSuspended:
+		vmCtx.VM.Spec.PowerState = vmopv1.VirtualMachinePowerStateSuspended
+	default:
+		// Unknown state, log and default to off
+		// this is very very unlikely to happen
+		vmCtx.Logger.Info("Unknown snapshot power state, defaulting to Off")
```
bryanv: Might as well log the unexpected power state just in case

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3209,6 +3207,90 @@ func vmTests() {
...
+							Expect(vcVM.Properties(ctx, vcVM.Reference(), []string{"snapshot"}, &moVM)).To(Succeed())
+							Expect(moVM.Snapshot).ToNot(BeNil())
+
+							snapshot := vmlifecycle.FindSnapshotInTree(moVM.Snapshot.RootSnapshotList, moVM.Snapshot.CurrentSnapshot.Value)
+							Expect(snapshot).ToNot(BeNil())
+							Expect(snapshot.State).To(Equal(vimtypes.VirtualMachinePowerStatePoweredOn))
+
+							// Snapshot should be owned by the VM resource.
+							o := vmopv1.VirtualMachine{}
+							Expect(ctx.Client.Get(ctx, client.ObjectKeyFromObject(vm), &o)).To(Succeed())
```
bryanv: Is this Get() needed since you already have the VM?

pkg/providers/vsphere/vmprovider_vm_test.go
palnabarun: This is fixed now.

pkg/providers/vsphere/vmprovider_vm_snapshot.go
palnabarun: Makes sense. Currently, FindSnapshot replicates the equivalent method from govmomi.

palnabarun:
Unrelated tests are failing due to the govmomi bump. Investigating.

palnabarun:
Fix for the issue is in https://github.com/vmware-tanzu/vm-operator/pull/1155

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3209,6 +3207,83 @@ func vmTests() {
...
+
+							// Create second snapshot
+							secondSnapshot = builder.DummyVirtualMachineSnapshotWithMemory("", "test-second-snap", vm.Name)
+							secondSnapshot.Namespace = nsInfo.Namespace
+
+							task, err = vcVM.CreateSnapshot(ctx, secondSnapshot.Name, "second snapshot", true, false)
+							Expect(err).ToNot(HaveOccurred())
+							Expect(task.Wait(ctx)).To(Succeed())
+
+							// Create second snapshot CR
```
lubronzhan, line 3249: nit: this line could be removed

pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3209,6 +3207,83 @@ func vmTests() {
...
+							// Mark the snapshot as ready so that the snapshot workflow doesn't try to create it.
+							conditions.MarkTrue(secondSnapshot, vmopv1.VirtualMachineSnapshotReadyCondition)
+							// Snapshot should be owned by the VM resource.
+							Expect(controllerutil.SetOwnerReference(vm, secondSnapshot, ctx.Scheme)).To(Succeed())
+							Expect(ctx.Client.Create(ctx, secondSnapshot)).To(Succeed())
+
+							// Verify the VM is powered on
+							Expect(vm.Status.PowerState).To(Equal(vmopv1.VirtualMachinePowerStateOn))
```
lubronzhan, line 3257: This check doesn't seem useful unless you revert to second snapshot then check the power is turned on?

lubronzhan (approved):
LGTM!

pkg/providers/vsphere/vmprovider_vm_test.go
palnabarun: This is just to check whether the VM is actually Powered On.

pkg/providers/vsphere/vmprovider_vm_test.go
palnabarun: Below this is the snapshot CR creation.

PR #1147, opened 2025-09-02
🌱 Relocate VirtualMachineOrGroup interface and fix VM/VMG get functions receivers

dilyar85:

This PR moves the `VirtualMachineOrGroup` interface declaration from `api/` to `pkg/util/vmopv1` package since it's not user facing and duplicate across all v1a2 to v1a5 versions.

This PR also fixes the receiver of `Get` functions in VM and VMG types to not use pointer and return kind in hard-coded string (as `object.Kind` could be empty, see https://github.com/kubernetes/client-go/issues/541).

Finally, this PR deletes the `pkg/util/vmopv1/owner_refs.go` and its test file by moving its single, group-specific function into the `pkg/util/vmopv1/vmgroup.go` file.

Fixes N/A.

None.

```release-note
Relocate VirtualMachineOrGroup interface and fix VM/VMG get functions receivers.
```


pkg/util/vmopv1/vmgroup_test.go
```diff
@@ -355,3 +356,180 @@ var _ = Describe("UpdateGroupLinkedCondition",
 		})
 	},
 )
+
+var _ = Describe("RemoveStaleGroupOwnerRef", func() {
+	const (
+		apiVersion = "vmoperator.vmware.com/v1alpha5"
```
lubronzhan: nit: could be replaced with `vmopv1.GroupVersion.String()` so no need to update this when we bump to newer version

dilyar85: Good call! Updated as suggested.

PR #1149, opened 2025-09-03
:seedling: Do not include already existing VMs in placement call

aruneshpa:

This change modifies the group placement workflow such that we no longer try to ask DRS for recommendation to place VMs that already exist on vSphere.  For now, this change is done by checking that the VM's status.uniqueID is set.  This is safe for the most part since status update is one of the first things we do during a reconcile and uniqueID is one of the first fields to be updated.

A new reason is introduced for the
VirtualMachineGroupMemberPlacementReady condition called "AlreadyPlaced" which represents a VM that is already placed.

Additionally, this change also handled steady state update of a Group's member statuses.  If a Group's member status is wiped, we populate those with PlacementReady condition with "AlreadyPlaced" reason.

Some other minor cleanups in the tests.

Testing Done:
1. Import / adopting a VM:
- Created a VM
- Created a group with the VM as the member
- Modified the VM to point to the group (spec.groupName)
- Verified that the condition was set:
```
  members:
  - conditions:
    - lastTransitionTime: "2025-09-06T01:15:52Z"
      message: ""
      reason: "True"
      status: "True"
      type: GroupLinked
    - lastTransitionTime: "2025-09-06T01:15:52Z"
      message: VM already placed
      reason: AlreadyPlaced
      status: "True"
      type: PlacementReady
```

2. Steady state status reconstruction
- After the above test, patched the object to remove the status.
```
 kubectl patch vmg -n parunesh-ns vmg --type=json --subresource=status   -p='[{"op": "replace", "path": "/status", "value": {}}]'
virtualmachinegroup.vmoperator.vmware.com/vmg patched
```
- Verified that the Status was re-populated:
```
  members:
  - conditions:
    - lastTransitionTime: "2025-09-06T01:18:15Z"
      message: ""
      reason: "True"
      status: "True"
      type: GroupLinked
    - lastTransitionTime: "2025-09-06T01:18:15Z"
      message: VM already placed
      reason: AlreadyPlaced
      status: "True"
      type: PlacementReady
```

```release-note
Do not include already existing VMs in placement call
```

### DIFF
--- a/api/v1alpha5/virtualmachinegroup_types.go
+++ b/api/v1alpha5/virtualmachinegroup_types.go
@@ -20,6 +20,11 @@ const (
 	// VirtualMachineGroupMemberConditionPlacementReady indicates that the
 	// member has a placement decision ready.
 	VirtualMachineGroupMemberConditionPlacementReady = "PlacementReady"
+
+	// VirtualMachineGroupMemberAlreadyPlacedReason indicates that the
+	// group member already exists on the back-end infrastructure
+	// (e.g., vSphere) and therefore does not require placement.
+	VirtualMachineGroupMemberAlreadyPlacedReason = "AlreadyPlaced"
 )
 
 // GroupMember describes a member of a VirtualMachineGroup.
--- a/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
+++ b/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
@@ -443,6 +443,20 @@ func (r *Reconciler) reconcileMember(
 	}
 
 	if member.Kind == vmKind {
+		vm := obj.(*vmopv1.VirtualMachine)
+
+		// If the VM already exists, set the member condition appropriately.
+		if vm.Status.UniqueID != "" {
+			if !conditions.IsTrue(ms, vmopv1.VirtualMachineGroupMemberConditionPlacementReady) {
+				conditions.Set(ms, &metav1.Condition{
+					Type:    vmopv1.VirtualMachineGroupMemberConditionPlacementReady,
+					Status:  metav1.ConditionTrue,
+					Reason:  vmopv1.VirtualMachineGroupMemberAlreadyPlacedReason,
+					Message: "VM already placed",
+				})
+			}
+		}
+
 		vmStatusPowerState := obj.GetPowerState()
 
 		if vmStatusPowerState == "" {
@@ -467,7 +481,6 @@ func (r *Reconciler) reconcileMember(
 			goto patchMember
 		}
 
-		vm := obj.(*vmopv1.VirtualMachine)
 		vmSpecMatchGroup := vm.Spec.PowerState == ctx.VMGroup.Spec.PowerState
 		if updatePowerState || vmSpecMatchGroup {
 			// VM power state is syncing with group or pending status update.
@@ -526,6 +539,14 @@ patchMember:
 	return nil
 }
 
+// reconcilePlacement reconciles and updates the placement status of
+// the members of the group and its children groups.
+//
+// It does so by collecting all the members across all groups that
+// need to be placed and by calling into the VM provider to get
+// placement recommendations from DRS. Group member statuses are then
+// updated with the placement recommendations. VMs members that
+// already exist are skipped from placement.
 func (r *Reconciler) reconcilePlacement(
 	ctx *pkgctx.VirtualMachineGroupContext) error {
 
@@ -633,7 +654,17 @@ func (r *Reconciler) getVMForPlacement(
 		return nil, fmt.Errorf("VM %q is assigned to group %q instead of expected %q", vmName, gn, vmGroup.Name)
 	}
 
-	// TODO: Probably need to check more stuff on the VM?
+	// If the VM has uniqueID set, then we don't need to do placement
+	// for it.
+	//
+	// Preexisting VMs would already be reconciled and thus will have
+	// MemberLinked condition set. Therefore, will not be part of
+	// placement. But on the off chance that the condition doesn't
+	// exist, we still check for existing VMs explicitly.
+	if vm.Status.UniqueID != "" {
+		return nil, nil
+	}
+
 	return vm, nil
 }
 
--- a/controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
+++ b/controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -45,28 +46,54 @@ var _ = Describe(
 		var (
 			ctx *builder.IntegrationTestContext
 
-			vm1Key, vm2Key           types.NamespacedName
+			vm1Key, vm2Key, vm3Key   types.NamespacedName
+			vm1, vm2, vm3            *vmopv1.VirtualMachine
 			vmGroup1Key, vmGroup2Key types.NamespacedName
 		)
 
+		setupGroupWithMembers := func(groupKey types.NamespacedName, members []vmopv1.VirtualMachineGroupBootOrderGroup, parentGroupName ...string) {
+			GinkgoHelper()
+			group := &vmopv1.VirtualMachineGroup{}
+			Expect(ctx.Client.Get(ctx, groupKey, group)).To(Succeed())
+			groupCopy := group.DeepCopy()
+			groupCopy.Spec.BootOrder = members
+			if len(parentGroupName) > 0 {
+				groupCopy.Spec.GroupName = parentGroupName[0]
+			}
+			Expect(ctx.Client.Patch(ctx, groupCopy, client.MergeFrom(group))).To(Succeed())
+		}
+
+		assignVMToGroup := func(vmKey types.NamespacedName, groupName string) {
+			GinkgoHelper()
+			vm := &vmopv1.VirtualMachine{}
+			Expect(ctx.Client.Get(ctx, vmKey, vm)).To(Succeed())
+			vmCopy := vm.DeepCopy()
+			vmCopy.Spec.GroupName = groupName
+			Expect(ctx.Client.Patch(ctx, vmCopy, client.MergeFrom(vm))).To(Succeed())
+		}
+
 		BeforeEach(func() {
 			ctx = suite.NewIntegrationTestContext()
-
 			vmGroup1Key = types.NamespacedName{
-				Name:      "vmgroup-1",
+				Name:      "vmg-1-" + uuid.NewString(),
 				Namespace: ctx.Namespace,
 			}
 			vmGroup2Key = types.NamespacedName{
-				Name:      "vmgroup-2",
+				Name:      "vmg-2-" + uuid.NewString(),
 				Namespace: ctx.Namespace,
 			}
 
 			vm1Key = types.NamespacedName{
-				Name:      "vm-1",
+				Name:      "vm-1-" + uuid.NewString(),
 				Namespace: ctx.Namespace,
 			}
 			vm2Key = types.NamespacedName{
-				Name:      "vm-2",
+				Name:      "vm-2-" + uuid.NewString(),
+				Namespace: ctx.Namespace,
+			}
+
+			vm3Key = types.NamespacedName{
+				Name:      "vm-3-" + uuid.NewString(),
 				Namespace: ctx.Namespace,
 			}
 
@@ -79,7 +106,7 @@ var _ = Describe(
 			}
 			Expect(ctx.Client.Create(ctx, vmGroup1)).To(Succeed())
 
-			vm1 := &vmopv1.VirtualMachine{
+			vm1 = &vmopv1.VirtualMachine{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace: vm1Key.Namespace,
 					Name:      vm1Key.Name,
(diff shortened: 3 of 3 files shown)


controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -514,6 +514,7 @@ patchMember:
 			member.Kind, member.Name, err)
 	}
 
+	ctx.Logger.Info("ARUNESH: marking member as linked", "member", ms)
```
lubronzhan:
```suggestion
	ctx.Logger.V(4).Info("Marking member as linked", "member", ms)
```

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -527,6 +528,16 @@ patchMember:
 	return nil
 }
 
+// reconcilePlacement reconciles and updates the placement status of
+// the members of the group and it's children groups.
```
lubronzhan:
```suggestion
// the members of the group and its children groups.
```

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -554,8 +563,46 @@ func (r *Reconciler) reconcilePlacement(
...
+		if err := r.VMProvider.PlaceVirtualMachineGroup(ctx, ctx.VMGroup, trimmedPlacements); err != nil {
+			return err
+		}
+	}
+
+	// The placement recommendation returned from the provider already has the
+	// member status populated. For VMs that did not require placement fix up
+	// the status if needed.
+	for i, placement := range groupPlacements {
+		for j := range groupPlacements[i].VMMembers {
```
lubronzhan:
```suggestion
		for _, member := range groupPlacements[i].VMMembers {
```

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -554,8 +563,46 @@ func (r *Reconciler) reconcilePlacement(
...
+			member := groupPlacements[i].VMMembers[j]
+			if _, ok := existingVMs[member.Name]; ok {
+				idx := findVMMemberStatus(member.Name, placement.VMGroup.Status.Members)
```
lubronzhan: nit: Can be replaced with `slices.IndexFunc(placement.VMGroup.Status.Members, func(member VirtualMachineGroupMemberStatus) bool{
   return members.Name == vmName && members.Kind == "VirtualMachine" 
})`

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -554,8 +563,46 @@ func (r *Reconciler) reconcilePlacement(
...
+	if len(trimmedPlacements) > 0 {
+		if err := r.VMProvider.PlaceVirtualMachineGroup(ctx, ctx.VMGroup, trimmedPlacements); err != nil {
+			return err
+		}
+	}
+
+	// The placement recommendation returned from the provider already has the
+	// member status populated. For VMs that did not require placement fix up
+	// the status if needed.
+	for i, placement := range groupPlacements {
```
lubronzhan:
```
	// Even if no VMs need placement, we go through with the patching since
	// there may be existing VMs on the provider that don't need to be placed,
	// but their status need to be patched.
```


But here if `groupPlacements ` is empty, it won't enter the loop to update the status？ So we can still exit early in previous code?

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -634,12 +727,47 @@ func (r *Reconciler) getVMForPlacement(
 		return nil, fmt.Errorf("VM %q is assigned to group %q instead of expected %q", vmName, gn, vmGroup.Name)
 	}
 
-	// TODO: Probably need to check more stuff on the VM?
+	// // Check if VM already exists in vCenter and skip placement if it does
+	// if isExistingVM(ctx, vm) {
```
lubronzhan: Better to do the check early here?

api/v1alpha5/virtualmachinegroup_types.go
```diff
@@ -21,6 +21,11 @@ const (
 	// VirtualMachineGroupMemberConditionPlacementReady indicates that the
 	// member has a placement decision ready.
 	VirtualMachineGroupMemberConditionPlacementReady = "PlacementReady"
+
+	// VirtualMachineGroupMemberAlreadyPlacedReason indicates that the
+	// group member already exists on the back-end infrastructure
+	// (e.g., vSphere) and therefore does not require placement.
+	VirtualMachineGroupMemberAlreadyPlacedReason = "AlreadyPlaced"
```
dilyar85, line 27: Should this be added to the v1a2-v1a4 packages as well to keep consistent with other constants above?

dilyar85: Or we can just drop this since we don't have other condition reasons (e.g. NotFound/NotMember for group linked condition and Pending/NotSynced for power state synced condition) declared in the api package.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -444,6 +444,20 @@ func (r *Reconciler) reconcileMember(
...
+		vm := obj.(*vmopv1.VirtualMachine)
+
+		// If the VM already exists, set the member condition appropriately.
+		if vm.Status.UniqueID != "" {
+			if !conditions.IsTrue(ms, vmopv1.VirtualMachineGroupMemberConditionPlacementReady) {
+				conditions.Set(ms, &metav1.Condition{
+					Type:    vmopv1.VirtualMachineGroupMemberConditionPlacementReady,
+					Status:  metav1.ConditionTrue,
+					Reason:  vmopv1.VirtualMachineGroupMemberAlreadyPlacedReason,
+					Message: "VM already placed",
```
dilyar85, line 455: Nit: clarify the VM is placed **out of the group placement** in the Message field?

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -527,6 +540,14 @@ patchMember:
...
 }
 
+// reconcilePlacement reconciles and updates the placement status of
+// the members of the group and its children groups.
+//
+// It does so by collecting all the members across all groups that
+// need to be placed and by calling into the VM provider to get
+// placement recommendations from DRS. Group member statuses are then
+// updated with the placement recommendations. VMs members that
+// already exist are skipped from placement.
```
dilyar85, line 549: Thanks for adding these comments!

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
```diff
@@ -45,41 +46,66 @@ var _ = Describe(
...
 			ctx *builder.IntegrationTestContext
 
-			vm1Key, vm2Key           types.NamespacedName
+			vm1Key, vm2Key, vm3Key   types.NamespacedName
+			vm1, vm2, vm3            *vmopv1.VirtualMachine
+			vmGroup1, vmGroup2       *vmopv1.VirtualMachineGroup
 			vmGroup1Key, vmGroup2Key types.NamespacedName
 		)
 
+		setupGroupWithMembers := func(g Gomega, groupKey types.NamespacedName, members []vmopv1.VirtualMachineGroupBootOrderGroup, parentGroupName ...string) {
```
dilyar85: Nit: no need to add the `g Gomega` param as this function is not called in any Eventually block.

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
```diff
@@ -45,41 +46,66 @@ var _ = Describe(
...
+			vmCopy.Spec.GroupName = groupName
+			g.Expect(ctx.Client.Patch(ctx, vmCopy, client.MergeFrom(vm))).To(Succeed())
+		}
+
 		BeforeEach(func() {
 			ctx = suite.NewIntegrationTestContext()
-
 			vmGroup1Key = types.NamespacedName{
-				Name:      "vmgroup-1",
+				Name:      "grp-1-" + uuid.NewString(),
```
dilyar85: Nit: `vmg` is the short name for the group CRD

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
dilyar85, line 1: Why change the test name since there is no unit_test file to distinguish? The current name is also consistent with other controllers that have single test file (e.g. VMIC).

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
```diff
@@ -45,41 +46,66 @@ var _ = Describe(
 		var (
 			ctx *builder.IntegrationTestContext
 
-			vm1Key, vm2Key           types.NamespacedName
+			vm1Key, vm2Key, vm3Key   types.NamespacedName
+			vm1, vm2, vm3            *vmopv1.VirtualMachine
+			vmGroup1, vmGroup2       *vmopv1.VirtualMachineGroup
```
dilyar85: These vars are not needed since they're not referenced anywhere in the test case. Every time when we need to get/patch VM/VMG, we always fetch it from the cluster. Declaring these vars here might end up with stale objects if someone uses `=` instead of `:=`.

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
```diff
@@ -449,86 +479,73 @@ var _ = Describe(
...
+					g.Expect(conditions.IsTrue(member, vmopv1.VirtualMachineGroupMemberConditionPlacementReady)).To(BeTrue())
+					if expectedReason != "" {
+						c := conditions.Get(member, vmopv1.VirtualMachineGroupMemberConditionPlacementReady)
+						g.Expect(c.Reason).To(Equal(expectedReason))
+					}
+				}
 
+				Eventually(func(g Gomega) {
+					expectPlacementCondition(g, vmGroup2Key, vm2Key.Name, vmopv1.VirtualMachineGroupMemberAlreadyPlacedReason)
+					expectPlacementCondition(g, vmGroup1Key, vm1Key.Name, "") // VM was placed through provider.  So, no already placed reason.
```
dilyar85: By `provider` did you mean `group`?

dilyar85 (approved):
Left some minor comments, overall LGTM!

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
lubronzhan: Or set them to nil in `AfterEach`

lubronzhan (approved):
LGTM! I remembered there were tons of change when I first reviewed. Did you just get rid of all of them

aruneshpa:
> LGTM! I remembered there were tons of change when I first reviewed. Did you just get rid of all of them

Yes.  I refactored the change after the first version.

api/v1alpha5/virtualmachinegroup_types.go
aruneshpa: @bryanv I don't think we need to backport this to older versions since we don't convert status Conditions between versions.  Objections?

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
aruneshpa: Not necessarily, right?  Even if the VM was placed via the group, but the Group status was wiped afterwards, we want the steady state status reconciliation to populate the status.

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
aruneshpa: I mixed EventuallyWithOffset with passing the matcher.  Fixed.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
dilyar85: > Not necessarily, right? Even if the VM was placed via the group, but the Group status was wiped afterwards, we want the steady state status reconciliation to populate the status.

Ah, right! In that case maybe drop the message since it's not adding any much info besides the `AlreadyPlaced` from reason?

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
aruneshpa: I _was_ adding unit tests in the first version.  So had to rename the file.  Now in the latest version of the diff, the UTs are removed.  Will rename the file back.

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
dilyar85: I would kinda prefer to drop them if they're not referenced at all during the test case.

controllers/virtualmachinegroup/virtualmachinegroup_controller_intg_test.go
aruneshpa: I meant VM provider.  Clarified to call this out.

api/v1alpha5/virtualmachinegroup_types.go
bryanv: I wouldn't expect the down version VMGroup user - just VKS right? - to need to programmatically care about this reason so I think we could forgo this

Conversion just acts on whatever string is there so these const don't matter (but v1a2+ there is no actual conversion of the conditions since they're all metav1.Conditions now

aruneshpa: Yeah, there are no other consumers other than VKS.  And AFAICT, they don't care about this reason specifically.  They would still continue to wait for the node to be ready via the providerID.

PR #1151, opened 2025-09-04
🌱 Update WFFC PVCs earlier in volume reconciliation

bryanv:

For WFFC PVCs set the selected-node annotation before the check for single stepping the attachments, so ideally we can update all the PVCs together to reduce the total time needed to bind all of the PVCs.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/virtualmachine/volume/volume_controller.go
+++ b/controllers/virtualmachine/volume/volume_controller.go
@@ -35,13 +35,13 @@ import (
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	pkgconst "github.com/vmware-tanzu/vm-operator/pkg/constants"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
 	pkgmgr "github.com/vmware-tanzu/vm-operator/pkg/manager"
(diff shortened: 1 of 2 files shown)


controllers/virtualmachine/volume/volume_controller.go
```diff
@@ -690,23 +693,14 @@ func (r *Reconciler) processAttachments(
...
-
 		// If VM hardware version doesn't meet minimal requirement, don't create CNS attachment.
 		// We only fetch the hardware version when this is the first time to create the CNS attachment.
 		//
 		// This HW version check block is along with removing VMI hardware version check from VM validation webhook.
 		// We used to deny requests if a PVC is specified in the VM spec while VMI hardware version is not supported.
 		// In this case, no attachments can be created if VM hardware version doesn't meet the requirement.
 		// So if a VM has volume attached, we can safely assume that it has passed the hardware version check.
-		if len(volumeStatuses) == 0 && len(attachments) == 0 {
+		if checkHWVersion {
```
aruneshpa, line 703: JW, are you initializing the `checkHwVersion` variable for readability so that it is clear why we are doing this?  Cause I would imagine the comment block explaining this with the length check.

bryanv: This only needs to be done once. With things being shuffled around, the `len(volumeStatuses) == 0` may no longer be true despite already doing this check

PR #1153, opened 2025-09-04
🐛 Do not unset NetOP network name with default network

bryanv:

NetOP will update the NetworkInterface Spec with the default network name so we don't want to then clear it.

Noticed when tracking down the cause of NetworkInterface with a high generation count.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/network/network.go
+++ b/pkg/providers/vsphere/network/network.go
@@ -390,7 +390,11 @@ func createNetOPNetworkInterface(
 		netIf.Labels[VMNameLabel] = vmCtx.VM.Name
 		netIf.Labels[VMInterfaceNameLabel] = interfaceSpec.Name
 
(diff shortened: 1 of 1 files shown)


pkg/providers/vsphere/network/network.go
```diff
@@ -390,7 +390,11 @@ func createNetOPNetworkInterface(
 		netIf.Labels[VMNameLabel] = vmCtx.VM.Name
 		netIf.Labels[VMInterfaceNameLabel] = interfaceSpec.Name
 
-		netIf.Spec.NetworkName = networkRefName
+		// NetOp will update the Spec with the default network name so we don't clear that
+		// here if using the default network.
```
dilyar85, line 394: Thanks for adding this comment for context!

dilyar85 (approved):
LGTM!

PR #1158, opened 2025-09-05
:seedling: Remove all unused fields and types from v1alpha3 and v1alpha4

aruneshpa:

We shipped some types and fields in older versions of the API (v1alpha3 and v1alpha4) without the backing implementation.  These were never meant to be released in these API versions.

This change removes these types and fields.  Technically, this is a breaking change, but since no one was using these types (because they were never implemented) on these versions, it is safe to remove.

Note that this change does not solve the issue of CRDs being installed in v1alpha5 version.  That is a known issue and is being worked on separately. 

```release-note
Remove all unused fields and types from v1alpha3 and v1alpha4
```


config/crd/bases/vmoperator.vmware.com_virtualmachinesnapshots.yaml
```diff
@@ -16,261 +16,6 @@ spec:
     singular: virtualmachinesnapshot
   scope: Namespaced
   versions:
-  - name: v1alpha4
```
lubronzhan, line 19: Do we want to remove this completely? Right now CSI, VCFA(not sure) are still fetching snapshot using v1alpha4, we should let them know to update their code

lubronzhan: CSI code https://github.com/nikhilbarge/vsphere-csi-driver/blob/use_volumeclaim_for_vm_snapshot/pkg/syncer/cnsoperator/controller/virtualmachinesnapshot/virtualmachinesnapshot_controller.go#L170-L172

aruneshpa: Thanks for pointing this out.  Waiting for CSI to move to v1alpha5 before I merge this.

aruneshpa:
This is still blocked on https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/3567 since CSI is referencing snapshot on v1a4.

palnabarun (approved):

PR #1161, opened 2025-09-09
🌱Set disableAutoRegistration in EC key when revert snapshot

lubronzhan:

Revert snapshot will restore VM Spec from Extra Config. This will trigger WCP auto registration, which adds `paused` label to the VM. So we want to skip the registration so reconciliation could continue.

Fixes #

```release-note

```


pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -420,6 +422,13 @@ func (vs *vSphereVMProvider) reconcileSnapshotRevertDoTask(
 		return err
 	}
 
+	// TODO: (Lubron) Do we really need this step since the key was removed
+	// after the revert?
+	// Enable auto registration after revert.
+	if err := enableAutoRegistration(vmCtx, vcVM); err != nil {
```
lubronzhan, line 428: Need to figure out whether snapshot's EC contains this, if yes, apply it still

lubronzhan:
Close this since it doesn't work

PR #1163, opened 2025-09-09
:bug: Fix issue where publish requests to inventory-backed CL was not marked complete

abaruni:

There was an issue where a VM Publish Request to an inventory-backed Content Library was not being marked completed. This was due to a call to CLS failing during the check for the vmi. This check needed to be updated to account for non-CLS backed libraries in order to allow for this check to succeed.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
+++ b/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
@@ -27,7 +27,6 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vapi/library"
-	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 
 	imgregv1a1 "github.com/vmware-tanzu/image-registry-operator-api/api/v1alpha1"
@@ -39,18 +38,19 @@ import (
 	pkgconst "github.com/vmware-tanzu/vm-operator/pkg/constants"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	pkgerr "github.com/vmware-tanzu/vm-operator/pkg/errors"
+	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
 	"github.com/vmware-tanzu/vm-operator/pkg/metrics"
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
-	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
 	pkgutil "github.com/vmware-tanzu/vm-operator/pkg/util"
 )
 
 const (
-	finalizerName           = "vmoperator.vmware.com/virtualmachinepublishrequest"
-	deprecatedFinalizerName = "virtualmachinepublishrequest.vmoperator.vmware.com"
-	TaskDescriptionID       = "com.vmware.ovfs.LibraryItem.capture"
+	finalizerName               = "vmoperator.vmware.com/virtualmachinepublishrequest"
+	deprecatedFinalizerName     = "virtualmachinepublishrequest.vmoperator.vmware.com"
+	OVFCaptureTaskDescriptionID = "com.vmware.ovfs.LibraryItem.capture"
+	CloneTaskDescriptionID      = "VirtualMachine.clone"
 
 	// waitForTaskTimeout represents the timeout to wait for task existence in task manager.
 	// When calling `CreateOVF` API, there is no guarantee that we can get its task info due to
@@ -370,23 +370,24 @@ func (r *Reconciler) removeVMPubResourceFromCluster(ctx *pkgctx.VirtualMachinePu
 // doesn't exist, or is not in Created phase.
 func (r *Reconciler) checkIsSourceValid(ctx *pkgctx.VirtualMachinePublishRequestContext) error {
 	vmPubReq := ctx.VMPublishRequest
-	vm := &vmopv1.VirtualMachine{}
-	objKey := client.ObjectKey{Name: vmPubReq.Status.SourceRef.Name, Namespace: vmPubReq.Namespace}
-	err := r.Get(ctx, objKey, vm)
-	if err != nil {
-		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
-		if apierrors.IsNotFound(err) {
-			conditions.MarkError(vmPubReq,
-				vmopv1.VirtualMachinePublishRequestConditionSourceValid,
-				vmopv1.SourceVirtualMachineNotExistReason,
-				err)
+	if ctx.VM == nil {
+		vm := &vmopv1.VirtualMachine{}
+		objKey := client.ObjectKey{Name: vmPubReq.Status.SourceRef.Name, Namespace: vmPubReq.Namespace}
+		err := r.Get(ctx, objKey, vm)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				conditions.MarkError(vmPubReq,
+					vmopv1.VirtualMachinePublishRequestConditionSourceValid,
+					vmopv1.SourceVirtualMachineNotExistReason,
+					err)
+			}
+			return fmt.Errorf("failed to get VirtualMachine %v: %w", objKey, err)
 		}
-		return err
+		ctx.VM = vm
 	}
-	ctx.VM = vm
 
-	if vm.Status.UniqueID == "" {
-		err = errors.New("VM hasn't been created and has no uniqueID")
+	if ctx.VM.Status.UniqueID == "" {
+		err := errors.New("VM hasn't been created and has no uniqueID")
 		conditions.MarkError(vmPubReq,
 			vmopv1.VirtualMachinePublishRequestConditionSourceValid,
 			vmopv1.SourceVirtualMachineNotCreatedReason,
@@ -402,20 +403,14 @@ func (r *Reconciler) checkIsSourceValid(ctx *pkgctx.VirtualMachinePublishRequest
 // It is invalid if the content library doesn't exist, an item with the same name in the CL exists.
 func (r *Reconciler) checkIsTargetValid(ctx *pkgctx.VirtualMachinePublishRequestContext) error {
 	vmPubReq := ctx.VMPublishRequest
-	contentLibrary := &imgregv1a1.ContentLibrary{}
-	targetLocationName := vmPubReq.Spec.Target.Location.Name
-	targetItemName := vmPubReq.Status.TargetRef.Item.Name
-	objKey := client.ObjectKey{Name: targetLocationName, Namespace: vmPubReq.Namespace}
-	if err := r.Get(ctx, objKey, contentLibrary); err != nil {
-		ctx.Logger.Error(err, "failed to get ContentLibrary", "cl", objKey)
-		if apierrors.IsNotFound(err) {
-			conditions.MarkError(vmPubReq,
-				vmopv1.VirtualMachinePublishRequestConditionTargetValid,
-				vmopv1.TargetContentLibraryNotExistReason,
-				err)
+
+	if ctx.ContentLibrary == nil {
+		if err := r.getTargetLibrary(ctx); err != nil {
+			return err
 		}
-		return err
 	}
+	contentLibrary := ctx.ContentLibrary
+	targetItemName := vmPubReq.Status.TargetRef.Item.Name
 
 	if !contentLibrary.Spec.Writable {
 		err := fmt.Errorf("target location %s is not writable", contentLibrary.Status.Name)
@@ -444,42 +439,23 @@ func (r *Reconciler) checkIsTargetValid(ctx *pkgctx.VirtualMachinePublishRequest
 		return err
 	}
 
-	ctx.ContentLibrary = contentLibrary
-
-	contentLibraryType := imgregv1.LibraryTypeContentLibrary
-	if pkgcfg.FromContext(ctx).Features.InventoryContentLibrary {
-		contentLibraryV1A2 := &imgregv1.ContentLibrary{}
-		if err := r.Get(ctx, objKey, contentLibraryV1A2); err != nil {
-			ctx.Logger.Error(err, "failed to get v1alpha2 ContentLibrary", "cl", objKey)
-			if apierrors.IsNotFound(err) {
-				conditions.MarkError(vmPubReq,
-					vmopv1.VirtualMachinePublishRequestConditionTargetValid,
-					vmopv1.TargetContentLibraryNotExistReason,
-					err)
-			}
-			return err
-		}
-		contentLibraryType = contentLibraryV1A2.Spec.Type
-
-		if contentLibraryType == imgregv1.LibraryTypeInventory {
-			// Check if quota validation is needed
-			if metav1.HasLabel(contentLibraryV1A2.ObjectMeta, pkgconst.AsyncQuotaPerformCheckAnnotationKey) {
-
-				if err := r.checkContentLibraryQuota(ctx); err != nil {
-					return err
-				}
+	if ctx.ContentLibraryType == imgregv1.LibraryTypeInventory {
+		// Check if quota validation is needed
+		if metav1.HasLabel(ctx.ContentLibraryV1A2.ObjectMeta, pkgconst.AsyncQuotaPerformCheckAnnotationKey) {
+			if err := r.checkContentLibraryQuota(ctx); err != nil {
+				return err
 			}
 		}
 	}
-	ctx.Logger.Info("target location type", "LibraryType", contentLibraryType)
 
-	item, err := r.getItemByName(ctx, contentLibraryType, string(ctx.ContentLibrary.Spec.UUID), targetItemName)
+	item, err := r.getItemByName(ctx)
 
 	if err != nil {
 		return fmt.Errorf("failed to get item %q from library: %w", targetItemName, err)
 	}
 
 	if !pkgutil.IsNil(item) {
+		objKey := client.ObjectKey{Name: vmPubReq.Spec.Target.Location.Name, Namespace: vmPubReq.Namespace}
 		ctx.Logger.Info("target item already exists in the content library",
 			"library", objKey, "itemName", targetItemName)
 		// If item already exists in the content library and attempt is not zero,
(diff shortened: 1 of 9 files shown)


controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -574,6 +563,47 @@ func (r *Reconciler) checkContentLibraryQuota(ctx *pkgctx.VirtualMachinePublishR
 	return nil
 }
 
+// checkIsStorageProfileValid checks if the encryption status of the source VM's disks is compatible with
+// the target Content Library's storage profile. A source VM with encrypted disks cannot be published to
+// a Content Library whose storage profile does not support encryption. This check is only performed if
+// the ContentLibrary LibraryType is "Inventory".
+func (r *Reconciler) checkIsStorageProfileValid(
```
lubronzhan: Maybe the check of `ctx.VM.Status.Crypto` could also be moved into this func to comply with the description?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -663,23 +701,61 @@ func (r *Reconciler) getPublishRequestTask(ctx *pkgctx.VirtualMachinePublishRequ
...
+	if publishTask.DescriptionId != OVFCaptureTaskDescriptionID {
+		err = fmt.Errorf("failed to find expected task %s, found %s instead", OVFCaptureTaskDescriptionID, publishTask.DescriptionId)
 		logger.Error(err, "task doesn't exist")
 		return nil, err
 	}
 
 	return publishTask, nil
 }
 
+// getCreateOVFTask gets task with description id VirtualMachine.clone and whose result.Value is the MOID of the target
```
lubronzhan: nit, 
```suggestion
// getCloneVMTask gets task with description id VirtualMachine.clone and whose result.Value is the MOID of the target
```

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -663,23 +701,61 @@ func (r *Reconciler) getPublishRequestTask(ctx *pkgctx.VirtualMachinePublishRequ
...
 
 	// Return the first task.
 	// We would never send multiple CreateOvf requests with the same actID,
 	// so that we should never get multiple tasks.
 	publishTask := &tasks[0]
-	if publishTask.DescriptionId != TaskDescriptionID {
-		err = fmt.Errorf("failed to find expected task %s, found %s instead", TaskDescriptionID, publishTask.DescriptionId)
+	if publishTask.DescriptionId != OVFCaptureTaskDescriptionID {
+		err = fmt.Errorf("failed to find expected task %s, found %s instead", OVFCaptureTaskDescriptionID, publishTask.DescriptionId)
 		logger.Error(err, "task doesn't exist")
```
lubronzhan: It's related to your change, but this could be removed and wrapped in return?

pkg/providers/vsphere/vmprovider.go
```diff
@@ -459,6 +461,82 @@ func (vs *vSphereVMProvider) GetItemFromInventoryByName(
 	return vm, nil
 }
 
+func (vs *vSphereVMProvider) GetCloneTasksForVM(
```
lubronzhan, line 464: Looks like this function is generic enough that not only works for clone task? Could it be renamed to something like `GetTasksForVMByDescriptionID`?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
```diff
@@ -718,14 +774,159 @@ func unitTestsReconcile() {
...
+							clv1a2 := builder.DummyContentLibraryV1A2("dummy-cl", vm.Namespace, "dummy-id")
+
+							initObjects = append(initObjects, clv1a2)
+						})
+						JustBeforeEach(func() {
+							pkgcfg.SetContext(ctx, func(config *pkgcfg.Config) {
+								config.Features.InventoryContentLibrary = true
+							})
+
+							fakeVMProvider.GetCloneTasksForVMFn = func(
```
lubronzhan: This could be removed

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
```diff
@@ -718,14 +774,159 @@ func unitTestsReconcile() {
...
+								ctx context.Context,
+								vm *vmopv1.VirtualMachine,
+								moID, descriptionID string) ([]vimtypes.TaskInfo, error) {
+
+								return nil, nil
+							}
+							fakeVMProvider.GetItemFromInventoryByNameFn = func(
```
lubronzhan, line 750: This could be removed as well

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
```diff
@@ -718,14 +774,159 @@ func unitTestsReconcile() {
...
+								ctx context.Context,
+								contentLibrary, itemName string) (object.Reference, error) {
+
+								return &object.VirtualMachine{}, nil
+							}
+						})
+						When("target already exists in the inventory", func() {
+							JustBeforeEach(func() {
+								fakeVMProvider.ContainsExtraConfigEntryFn = func(
```
lubronzhan: This could be removed

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
abaruni: No, this function is used in more than one place. The path that is being exercised here requires that the returned value is not nil

lubronzhan: IC. Thanks

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -944,19 +998,45 @@ func getPublishRequestActID(vmPub *vmopv1.VirtualMachinePublishRequest) string {
...
+	default:
+		return "", fmt.Errorf("expect task result type to be ContentLibraryItem or VirtualMachine, got %s instead", item.Type)
+	}
+}
+
+func (r *Reconciler) getTargetLibrary(ctx *pkgctx.VirtualMachinePublishRequestContext) error {
+	targetLocationName := ctx.VMPublishRequest.Spec.Target.Location.Name
+	objKey := client.ObjectKey{Name: targetLocationName, Namespace: ctx.VMPublishRequest.Namespace}
+
+	contentLibrary := &imgregv1a1.ContentLibrary{}
```
zhangheliu, line 979: It makes sense to call get r.Get(ctx, objKey, contentLibraryV1A2) every reconcile loop since we need ContentLibraryType throughout different stages of the process. JW if we should keep r.Get(ctx, objKey, contentLibraryV1A1) as it is on where we are reading it.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -651,8 +681,16 @@ func (r *Reconciler) checkIsComplete(ctx *pkgctx.VirtualMachinePublishRequestCon
...
+		return r.getCloneVMTask(ctx)
+	}
+
+	return r.getCreateOVFTask(ctx)
+}
+
+// getCreateOVFTask gets task with description id com.vmware.ovfs.LibraryItem.capture and specified actid in taskManager.
+func (r *Reconciler) getCreateOVFTask(ctx *pkgctx.VirtualMachinePublishRequestContext) (*vimtypes.TaskInfo, error) {
 	actID := getPublishRequestActID(ctx.VMPublishRequest)
 	logger := ctx.Logger.WithValues("activationID", actID)
```
zhangheliu, line 643: why do we have to separate getPublishRequestTask into getCreateOVFTask() and getCreateOVFTask()? It looks like GetTasksByActID could handle both types of tasks already.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -320,10 +322,12 @@ func (r *Reconciler) publishVirtualMachine(ctx *pkgctx.VirtualMachinePublishRequ
 			ctx.Logger.Error(err, "update VirtualMachinePublishRequest status failed")
 			return err
 		}
+		vm := ctx.VM
+		contentLibrary := ctx.ContentLibrary
 
 		go func() {
 			actID := getPublishRequestActID(vmPublishReq)
-			itemID, pubErr := r.VMProvider.PublishVirtualMachine(ctx, ctx.VM, vmPublishReq, ctx.ContentLibrary, actID)
+			itemID, pubErr := r.VMProvider.PublishVirtualMachine(ctx, vm, vmPublishReq, contentLibrary, actID)
```
dilyar85: I don't see these vars used anywhere else besides this function call. Was this change intentional?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -320,10 +322,12 @@ func (r *Reconciler) publishVirtualMachine(ctx *pkgctx.VirtualMachinePublishRequ
 			ctx.Logger.Error(err, "update VirtualMachinePublishRequest status failed")
 			return err
 		}
+		vm := ctx.VM
```
bryanv: Why are these needed?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -663,23 +701,61 @@ func (r *Reconciler) getPublishRequestTask(ctx *pkgctx.VirtualMachinePublishRequ
...
+	targetItemName := vmPubReq.Status.TargetRef.Item.Name
+
+	vm := &vmopv1.VirtualMachine{}
+	objKey := client.ObjectKey{Name: vmPubReq.Status.SourceRef.Name, Namespace: vmPubReq.Namespace}
+	err := r.Get(ctx, objKey, vm)
+	if err != nil {
+		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
+		return nil, err
+	}
+	ctx.VM = vm
```
bryanv: Can we improve this flow so we don't then refetch the VM if we get to `publishVirtualMachine()`?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -574,6 +563,47 @@ func (r *Reconciler) checkContentLibraryQuota(ctx *pkgctx.VirtualMachinePublishR
...
+	if storageClassName == "" {
+		return nil
+	}
+
+	sc := storagev1.StorageClass{}
+	if err := r.Get(ctx, client.ObjectKey{Name: storageClassName}, &sc); err != nil {
+		return fmt.Errorf("failed to get storageClass %q: %w", storageClassName, err)
+	}
+	storagePolicyID, err := kubeutil.GetStoragePolicyID(sc)
+
```
bryanv: Please remove these empty lines

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -663,23 +701,61 @@ func (r *Reconciler) getPublishRequestTask(ctx *pkgctx.VirtualMachinePublishRequ
...
+	vmPubReq := ctx.VMPublishRequest
+	targetItemName := vmPubReq.Status.TargetRef.Item.Name
+
+	vm := &vmopv1.VirtualMachine{}
+	objKey := client.ObjectKey{Name: vmPubReq.Status.SourceRef.Name, Namespace: vmPubReq.Namespace}
+	err := r.Get(ctx, objKey, vm)
+	if err != nil {
+		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
+		return nil, err
+	}
```
dilyar85: No need to log the error if it's wrapped into the returned error:
```suggestion
	if err := r.Get(ctx, objKey, vm); err != nil {
		return nil, fmt.Errorf("failed to get VirtualMachine %q: %w", objKey, err)
	}
```

pkg/providers/vsphere/vmprovider.go
```diff
@@ -459,6 +461,82 @@ func (vs *vSphereVMProvider) GetItemFromInventoryByName(
...
+		Logger:  logger,
+		VM:      vm,
+	}
+
+	client, err := vs.getVcClient(vmCtx)
+	if err != nil {
+		return nil, err
+	}
+
+	vcVM, err := vs.getVM(vmCtx, client, false)
```
bryanv: Use `true` here so it returns an error if the VM isn't found?

pkg/providers/vsphere/vmprovider.go
```diff
@@ -459,6 +461,82 @@ func (vs *vSphereVMProvider) GetItemFromInventoryByName(
...
 
+func (vs *vSphereVMProvider) GetCloneTasksForVM(
+	ctx context.Context,
+	vm *vmopv1.VirtualMachine,
+	moID, descriptionID string) ([]vimtypes.TaskInfo, error) {
+
+	logger := pkglog.FromContextOrDefault(ctx)
+
+	vmCtx := pkgctx.VirtualMachineContext{
+		Context: ctx,
```
bryanv: Use ` vs.getOpID()` to set additional context on this to VC

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -663,23 +701,61 @@ func (r *Reconciler) getPublishRequestTask(ctx *pkgctx.VirtualMachinePublishRequ
...
+
+	if targetObjRef == nil {
+		return nil, nil
+	}
+	targetMOID := targetObjRef.Reference().Value
+
+	tasks, err := r.VMProvider.GetCloneTasksForVM(ctx, vm, targetMOID, CloneTaskDescriptionID)
+	if err != nil {
+		return nil, fmt.Errorf("error getting recent clone tasks for VM %q: %w", objKey.Name, err)
```
dilyar85: Nit: `objKey` might be better as it also includes the namespace info.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: we don't have the ability to pass in an activation ID for clone operation. so while the clone task has an activation ID, we don't have any idea what that activation ID is until we fetch the tasks for the VM, sort through tasks that have the correct DescriptionID, i.e., for clones, and have a result value that has a MOID that matches the VM template we created.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -663,23 +701,61 @@ func (r *Reconciler) getPublishRequestTask(ctx *pkgctx.VirtualMachinePublishRequ
...
+	}
+	ctx.VM = vm
+
+	targetObjRef, err := r.VMProvider.GetItemFromInventoryByName(ctx, ctx.ContentLibraryV1A2.Spec.ID, targetItemName)
+	if err != nil {
+		return nil, fmt.Errorf("error getting item %q from inventory: %w", targetItemName, err)
+	}
+
+	if targetObjRef == nil {
+		return nil, nil
```
dilyar85, line 697: Worth adding a V4 log here?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -944,19 +998,58 @@ func getPublishRequestActID(vmPub *vmopv1.VirtualMachinePublishRequest) string {
...
+
+	contentLibrary := &imgregv1a1.ContentLibrary{}
+	if err := r.Get(ctx, objKey, contentLibrary); err != nil {
+		ctx.Logger.Error(err, "failed to get ContentLibrary", "cl", objKey)
+		if apierrors.IsNotFound(err) {
+			conditions.MarkError(vmPubReq,
+				vmopv1.VirtualMachinePublishRequestConditionTargetValid,
+				vmopv1.TargetContentLibraryNotExistReason,
+				err)
+		}
```
dilyar85: Should we return the error here? If so, no need to log the error with wrapping.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -944,19 +998,58 @@ func getPublishRequestActID(vmPub *vmopv1.VirtualMachinePublishRequest) string {
...
+		contentLibraryV1A2 := &imgregv1.ContentLibrary{}
+		if err := r.Get(ctx, objKey, contentLibraryV1A2); err != nil {
+			ctx.Logger.Error(err, "failed to get v1alpha2 ContentLibrary", "cl", objKey)
+			if apierrors.IsNotFound(err) {
+				conditions.MarkError(vmPubReq,
+					vmopv1.VirtualMachinePublishRequestConditionTargetValid,
+					vmopv1.TargetContentLibraryNotExistReason,
+					err)
+			}
+		}
```
dilyar85: Ditto.

pkg/providers/vm_provider_interface.go
```diff
@@ -59,6 +59,9 @@ type VirtualMachineProviderInterface interface {
 
 	GetItemFromLibraryByName(ctx context.Context, contentLibrary, itemName string) (*library.Item, error)
 	GetItemFromInventoryByName(ctx context.Context, contentLibrary, itemName string) (object.Reference, error)
+	GetCloneTasksForVM(ctx context.Context, vm *vmopv1.VirtualMachine, moID, descriptionID string) ([]vimtypes.TaskInfo, error)
+	ContainsExtraConfigEntry(ctx context.Context, objVM *object.VirtualMachine, key, value string) (bool, error)
```
bryanv, line 63: Can the existing `GetVirtualMachineProperties()` be used instead?

pkg/providers/vsphere/virtualmachine/publish.go
```diff
@@ -88,9 +88,7 @@ func CloneVM(
 			Folder: &folderRef,
 			// Causes a linked clone to be unlinked and consolidated
 			DiskMoveType: string(vimtypes.VirtualMachineRelocateDiskMoveOptionsMoveAllDiskBackingsAndDisallowSharing),
-			Profile: []vimtypes.BaseVirtualMachineProfileSpec{
-				&vimtypes.VirtualMachineDefinedProfileSpec{ProfileId: storagePolicyID},
-			},
+			Profile:      nil,
```
bryanv: Why nil here?

pkg/providers/vm_provider_interface.go
abaruni: i will end up using this function, but i will refactor in an upcoming PR to remove those bespoke functions and use this on in place

PR #1164, opened 2025-09-10
🌱 add allowList and denyList for extraConfig

zhangheliu:

Fixes # N/A

```release-note

```


pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,79 @@
...
+func FilteredExtraConfig(
+	in pkgutil.OptionValues,
+	reset bool) (pkgutil.OptionValues, error) {
+	allowKeys := sets.New([]string{
+		"vmservice.pvc",
+		"vmservice.vmi.labels",
+		"vmservice.vmpub.uuid"}...)
+	denyKeyPrefixes := sets.New([]string{
+		"guestinfo",
+		"vmservice"}...)
```
dilyar85: These should be defined as constants in a separate place for better clarity.

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return o.Config.ExtraConfig, nil
+}
+
+// FilteredExtraConfig removes or resets those key entries which have prefix
+// that is in the deny key prefix list, but not in the allow key list.
+// It returns a new filtered OptionValues object if the operation is successful.
+func FilteredExtraConfig(
+	in pkgutil.OptionValues,
+	reset bool) (pkgutil.OptionValues, error) {
```
dilyar85, line 69: I wonder if this function should take `allowKeys` and `denyKeys` as params and just apply those filters, so that it's more flexible with different filters called.

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,79 @@
...
+		"vmservice.pvc",
+		"vmservice.vmi.labels",
+		"vmservice.vmpub.uuid"}...)
+	denyKeyPrefixes := sets.New([]string{
+		"guestinfo",
+		"vmservice"}...)
+
+	filtered := in.Map()
+	for k, v := range filtered {
+		if allowKeys.Has(k) || denyKeyPrefixes.Has(k) {
```
dilyar85: `denyKeyPrefixes.Has(k)` is not needed here?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2306,9 +2306,13 @@ func (vs *vSphereVMProvider) vmCreateGenConfigSpecImage(
...
 	// Inherit the image's vAppConfig.
 	createArgs.ConfigSpec.VAppConfig = imgConfigSpec.VAppConfig
-
+	
 	// Merge the image's extra config.
-	if srcEC := imgConfigSpec.ExtraConfig; len(srcEC) > 0 {
+	if srcEC, err := virtualmachine.FilteredExtraConfig(
+		imgConfigSpec.ExtraConfig,
+		false); err != nil {
+		return err
```
dilyar85, line 2315: Is this needed? I thought we're only filtering the EC keys during publish not deployment?

pkg/providers/vsphere/virtualmachine/extraconfig.go
zhangheliu: this is to handle the case where essentially the key itself matches the prefix while we only search for "prefix.*" wildcard pattern.

pkg/providers/vsphere/vmprovider_vm.go
zhangheliu: i think we are doing both ways

pkg/providers/vsphere/virtualmachine/extraconfig.go
zhangheliu: hmm, we dont have a use case for it yet. maybe we could update this method once we find a use case?

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,90 @@
...
+var (
+	extraConfigDenyKeyPrefixes = [...]string{
+		"guestinfo.",
+		"vmservice."}
+	extraConfigAllowKeys = [...]string{
+		"vmservice.pvc",
+		"vmservice.vmi.labels"}
+)
+
+func GetExtraConfigFromObject(
```
lubronzhan, line 36: Nice refactor! Looks like https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/virtualmachine/guestinfo.go#L20 can be refactored to use this as well

pkg/providers/vsphere/virtualmachine/extraconfig.go
lubronzhan: Then the func name and description should reflect the actual behavior of the func I think. so it's easier  to just follow what @dilyar85 suggested?

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,90 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
```
lubronzhan:
```suggestion
// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
```

pkg/providers/vsphere/virtualmachine/extraconfig_test.go
```diff
@@ -0,0 +1,156 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
```
lubronzhan:
```suggestion
// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
```

pkg/providers/vsphere/virtualmachine/extraconfig.go
zhangheliu: hmm, I am not a big fan of additional logic when we do not see a clear use for the change, even if it is a small one, such as looping and inserting into the set.

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,90 @@
...
+	unexpectedTypeErrMsg = "expected value's datatype string (got %T)"
+)
+
+var (
+	extraConfigDenyKeyPrefixes = [...]string{
+		"guestinfo.",
+		"vmservice."}
+	extraConfigAllowKeys = [...]string{
+		"vmservice.pvc",
+		"vmservice.vmi.labels"}
```
dilyar85: Super nit: can we move the closing bracket to a new line so it's clearer? It'll also generate one less diff line when adding new items later. 
```suggestion
	extraConfigDenyKeyPrefixes = [...]string{
		"guestinfo.",
		"vmservice.",
	}
	extraConfigAllowKeys = [...]string{
		"vmservice.pvc",
		"vmservice.vmi.labels",
	}
```

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+func GetExtraConfigFromObject(
+	ctx context.Context,
+	vm *object.VirtualMachine) (pkgutil.OptionValues, error) {
+	var o mo.VirtualMachine
+	if err := vm.Properties(
```
dilyar85, line 40: Add a sanity check on `vm` not being nil?

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,90 @@
...
+					if _, ok := v.(string); !ok {
+						return nil, fmt.Errorf("failed to filter extraConfig: "+
+							unexpectedTypeErrMsg, v)
+					}
+					filtered[k] = ""
+				} else {
+					delete(filtered, k)
+				}
+			}
+		}
```
dilyar85: Add a `break` at the end of inside `if strings.HasPrefix(lowerK, denyKeyPrefix)` statement (no need to check others if it's already matching one denied key).

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,90 @@
...
+// It returns a new filtered OptionValues object if the operation is successful.
+func FilteredExtraConfig(
+	in pkgutil.OptionValues,
+	reset bool) (pkgutil.OptionValues, error) {
+	allowKeys := sets.New(extraConfigAllowKeys[:]...)
+	filtered := in.Map()
+
+	for k, v := range filtered {
+		lowerK := strings.ToLower(k)
+		if allowKeys.Has(lowerK) {
```
dilyar85, line 75: How does it work with the allowed **prefix** like `vmservice.pvc`?

pkg/providers/vsphere/virtualmachine/extraconfig.go
zhangheliu: if we are doing nil checks, we need to do vm.Reference() and vm.Client() nil checks as well. This method is meant to be called by the internal stack with vm setup properly already. I will update the description to make it more clear that it is expected that VM object is valid.

pkg/providers/vsphere/virtualmachine/extraconfig.go
zhangheliu: due to the small denykeyprefixes size (2), currently, it is not a big deal but yeah, I will add this improvement. thanks for pointing it out.

pkg/providers/vsphere/virtualmachine/extraconfig.go
zhangheliu: hm, I am under the impression that vmservice.pvc is a key and we do not have an allowPrefix list. we could add a allow prefix keys list if there is a requirement for it

dilyar85: The correct key for PVC data should be `vmservice.virtualmachine.pvc.disk.data` as defined in https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/backup/api/backup_types.go#L90

pkg/providers/vsphere/virtualmachine/extraconfig.go
dilyar85: I just thought it'd be clearer if we had the filter list defined as a param (also it would make this utility function more flexible). I'm also fine doing it the current way and it shouldn't be a blocker for this PR.

pkg/providers/vsphere/virtualmachine/extraconfig.go
```diff
@@ -0,0 +1,95 @@
...
+
+var (
+	extraConfigDenyKeyPrefixes = [...]string{
+		"guestinfo.",
+		"vmservice.",
+	}
+	extraConfigAllowKeys = [...]string{
+		"vmservice.virtualmachine.pvc.disk.data",
+		"vmservice.vmi.labels",
+	}
```
dilyar85, line 31: Can we include `vmservice.vmpub.uuid` in the allow keys as well? Since the deny list has the `vmservice.` prefix, it could remove that vmpub key if it's added before the filter call.

dilyar85 (commented):
LGTM except one comment on vmpub key.

pkg/providers/vsphere/virtualmachine/extraconfig.go
zhangheliu: we would still like to remove the vmservice.vmpub.uuid key during VM creation.

dilyar85: Ah, I thought it's supposed to be _always_ allowed. Thanks for the clarification!

PR #1165, opened 2025-09-10
:seedling: Allow regular users to unset zone label on VM resource

aruneshpa:

This change allows unprivileged users to unset / remove the Zone label on the VM resource if the VM has a failed over or restored annotation. Administrators users are responsible to invoke RegisterVM API -- so we should allow that.

```release-note
 Allow regular users to unset zone label on VM resource
```

### DIFF
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -1593,16 +1593,28 @@ func (v validator) validateAvailabilityZone(
 		if oldVal := oldVM.Labels[topology.KubernetesTopologyZoneLabelKey]; oldVal != "" {
 			newVal := vm.Labels[topology.KubernetesTopologyZoneLabelKey]
 
-			// Privileged accounts are allowed to update the
-			// availability zone label on the VM. This is used during
+			// Zone label is being unset on the VM.  This is used during
 			// restore, or a fail-over where the restored environment
 			// may not have access to the zone from backup.
-			//
-			// All other modifications are rejected.
-			if ctx.IsPrivilegedAccount {
-				return allErrs
+			if newVal == "" {
+				// Privileged accounts are allowed to unset the
+				// availability zone label on the VM.
+				//
+				// The annotation checks here are not ideal since
+				// these are applied on the restored / registered VMs
+				// and may stay there forever potentially allowing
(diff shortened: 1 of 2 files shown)


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1593,16 +1593,28 @@ func (v validator) validateAvailabilityZone(
...
+				//
+				// The annotation checks here are not ideal since
+				// these are applied on the restored / registered VMs
+				// and may stay there forever potentially allowing
+				// constant modifications to a VM's zone by any user.
+				// But since this is just unsetting the zone, there's
+				// no potential harm.
+				if ctx.IsPrivilegedAccount ||
+					metav1.HasAnnotation(vm.ObjectMeta, vmopv1.RestoredVMAnnotation) ||
+					metav1.HasAnnotation(vm.ObjectMeta, vmopv1.FailedOverVMAnnotation) {
```
dilyar85, line 1611: Can you also add a comment explaining why we're allowing zone label unset for restored/failed over VMs for future reference please?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1593,16 +1593,28 @@ func (v validator) validateAvailabilityZone(
...
+			if newVal == "" {
+				// Privileged accounts are allowed to unset the
+				// availability zone label on the VM.
+				//
+				// The annotation checks here are not ideal since
+				// these are applied on the restored / registered VMs
+				// and may stay there forever potentially allowing
+				// constant modifications to a VM's zone by any user.
+				// But since this is just unsetting the zone, there's
+				// no potential harm.
```
bryanv, line 1608: Once it has been cleared, does the existing code set to whatever, like the VM's not current zone? I'm worried that could bork things

akutz: We only set it during placement / create. But yeah, it could be race-y since we use it for WffC.

aruneshpa:
This is no longer needed because we ended up modifying the RegisterVM logic to not reset the zone label for in-place restores.

aruneshpa:
This is no longer needed because we ended up modifying the RegisterVM logic to not reset the zone label for in-place restores.

PR #1169, opened 2025-09-11
🌱Refactor some snapshot related code

lubronzhan:
Refactoring snapshot code
1. Change `FindSnapshot` to return `VirtualMachineSnapshotTree` instead of `ManagedObjectReference`.
2. Change `SnapshotVirtualMachine` to return `VirtualMachineSnapshotTree` so following tasks could use it.
3. Not update `vmCtx.MoVM` while doing Property() call to fetch "snapshot" property, since `vmCtx.MoVM` will be shared by all the reconcile functions in updateVirtualMachine
4. Remove `FindSnapshotInTree` and `GetParentSnapshot` since they are not used anymore
5. `PatchSnapshotSuccessStatus` uses snapshot.state as source of truth to get powerstate
6. Add `ConvertPowerState` func to convert  `vimtypes.VirtualMachinePowerState` to `vmopv1.VirtualMachinePowerState`
7. Some logging clean up

```release-note
None
```

### DIFF
--- a/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
+++ b/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
@@ -139,14 +139,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 			if reterr == nil {
 				reterr = err
 			}
-			vmSnapshotCtx.Logger.Error(err, "patch failed")
 		}
 	}()
 
 	if !vmSnapshot.DeletionTimestamp.IsZero() {
 		if err := r.ReconcileDelete(vmSnapshotCtx); err != nil {
-			vmSnapshotCtx.Logger.Error(err, "Failed to delete VirtualMachineSnapshot")
-			return ctrl.Result{}, err
+			return ctrl.Result{},
+				fmt.Errorf("failed to delete VirtualMachineSnapshot: %w", err)
 		}
 		return ctrl.Result{}, nil
 	}
@@ -172,8 +171,7 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) 
 	vm := &vmopv1.VirtualMachine{}
 	objKey := client.ObjectKey{Name: vmSnapshot.Spec.VMRef.Name, Namespace: vmSnapshot.Namespace}
 	if err := r.Get(ctx, objKey, vm); err != nil {
-		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
-		return ctrl.Result{}, err
+		return ctrl.Result{}, fmt.Errorf("failed to get VirtualMachine %q: %w", objKey, err)
 	}
 
 	ctx.VM = vm
@@ -243,8 +241,7 @@ func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineSnapshotContext) 
 			// We don't sync SPU since we don't know which StorageClass of the VM.
 			return nil
 		}
-		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
-		return err
+		return fmt.Errorf("failed to get VirtualMachine %q: %w", objKey, err)
 	}
 
 	// Enqueue the storage class to sync corresponding SPU.
--- a/pkg/providers/vsphere/virtualmachine/snapshot.go
+++ b/pkg/providers/vsphere/virtualmachine/snapshot.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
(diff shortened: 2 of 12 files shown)


controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -243,8 +241,7 @@ func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VirtualMachineSnapshotContext)
 			// We don't sync SPU since we don't know which StorageClass of the VM.
 			return nil
 		}
-		ctx.Logger.Error(err, "failed to get VirtualMachine", "vm", objKey)
-		return err
+		return fmt.Errorf("failed to get VirtualMachine: %w, objKey: %s", err, objKey)
```
dilyar85: Nit suggestion:
```suggestion
		return fmt.Errorf("failed to get VirtualMachine %q: %w", objKey, err)
```

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -158,17 +188,18 @@ func (m snapshotMap) add(parent string, tree []vimtypes.VirtualMachineSnapshotTr
...
-func GetSnapshotSize(vmCtx pkgctx.VirtualMachineContext, vmSnapshot *vimtypes.ManagedObjectReference) int64 {
+// The only difference is that this function does not include the size of FCDs,
+// and returns an error if the snapshot is not found.
+func GetSnapshotSize(logger logr.Logger,
+	moVM mo.VirtualMachine,
+	vmSnapshot *vimtypes.ManagedObjectReference) (int64, error) {
 	if vmSnapshot == nil || vmSnapshot.Value == "" {
-		vmCtx.Logger.V(5).Info("vmSnapshot is nil or empty")
-		return 0
+		return 0, fmt.Errorf("vmSnapshot is nil or empty")
```
dilyar85: Mention the type (VC's ManagedObject) to distinguish from K8s type?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -180,11 +211,10 @@ func GetSnapshotSize(vmCtx pkgctx.VirtualMachineContext, vmSnapshot *vimtypes.Ma
 	}
 
 	if snapshot == nil {
-		// TODO: Error if snapshot not found?
-		return 0
+		return 0, fmt.Errorf("snapshot not found")
```
dilyar85: Include snapshot key/value for better identification in logs?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -207,33 +237,33 @@ func GetSnapshotSize(vmCtx pkgctx.VirtualMachineContext, vmSnapshot *vimtypes.Ma
...
-	return total
+	return total, nil
 }
 
 // getFCDDeviceKeySet returns a set of device keys of disk devices that are FCDs.
-func getFCDDeviceKeySet(vmCtx pkgctx.VirtualMachineContext) sets.Set[int32] {
+func getFCDDeviceKeySet(moVM mo.VirtualMachine) sets.Set[int32] {
 	deviceKeysSet := sets.Set[int32]{}
-	device := object.VirtualDeviceList(vmCtx.MoVM.Config.Hardware.Device)
+	device := object.VirtualDeviceList(moVM.Config.Hardware.Device)
```
dilyar85, line 274: Can `moVM.Config` be nil?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -66,27 +74,48 @@ func CreateSnapshot(args SnapshotArgs) (*vimtypes.ManagedObjectReference, error)
...
+		quiesceSpec)
 	if err != nil {
 		return nil, err
 	}
 
 	taskInfo, err := t.WaitForResult(args.VMCtx)
 	if err != nil {
-		args.VMCtx.Logger.V(5).Error(err, "create snapshot task failed", "taskInfo", taskInfo)
+		return nil, fmt.Errorf("failed to create snapshot: %w. taskInfo: %+v",
+			err, taskInfo)
```
dilyar85: Do we need to include every field from `taskInfo` in the returned error? If not, consider capturing the helpful fields only, or have a verbose error log just for taskInfo? Something like:

```suggestion
if taskInfo != nil {
        args.VMCtx.Logger.V(5).Error(err, "create snapshot task failed", "taskInfo", taskInfo)
}
return nil, fmt.Errorf("failed to create snapshot: %w", err)
```

pkg/util/kube/vmsnapshot.go
```diff
@@ -170,3 +169,19 @@ func PatchSnapshotSuccessStatus(
...
+	case vimtypes.VirtualMachinePowerStatePoweredOff:
+		return vmopv1.VirtualMachinePowerStateOff
+	case vimtypes.VirtualMachinePowerStateSuspended:
+		return vmopv1.VirtualMachinePowerStateSuspended
+	default:
+		logger.Info("Unknown snapshot power state, defaulting to Off",
+			"powerState", powerState)
+		return vmopv1.VirtualMachinePowerStateOff
+	}
+}
```
dilyar85: Is `pkg/util/vmopv1/vm.go` a better place for this function since it's not specific to current vm snapshot file?

dilyar85 (approved):
Left some comments, overall LGTM! Thanks @lubronzhan for the cleanup.

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: Gonna use `moSnapshot` as the keyword to align with `moVM`

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -194,11 +232,11 @@ func GetSnapshotSize(vmCtx pkgctx.VirtualMachineContext, vmSnapshot *vimtypes.Ma
 
 	// Add the file key for the snapshot memory (.vmem) file if present.
 	if snapshot.MemoryKey != -1 {
-		vmCtx.Logger.V(5).Info("Adding memoryKey", "memoryKey", snapshot.MemoryKey)
+		logger.V(5).Info("Adding memoryKey", "memoryKey", snapshot.MemoryKey)
```
aruneshpa, line 235: Here and elsewhere in the change: I thought we said we will not have any log level below 4?

aruneshpa (approved):
LGTM.

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: mm these are very verbose log. Maybe just keep it as V(5) for now? I could still see many V(5) logging in the repo. Bryan also think it's better to keep V(5)

lubronzhan: Let me merge this and check with Andrew.

PR #1171, opened 2025-09-12
:seedling: Support explicit zone specification from group members

aruneshpa:

Placement of a VM is driven via the VirtualMachineGroup if the VM is part of a group (specified by spec.groupName).  This means we don't send in the list of placement candidate resource pools to DRS when asking for placement recommendations.  DRS returns recommendations based on the hardware and the placement policies specified with the VM.

However, if a group member wants to explicitly be placed in a particular zone, they should be able to specify the zone label on the member.  This change adds support for that.  This will be used by VKS to support scale-outs of node pools.

```release-note
Support explicit zone specification from group members
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -1396,12 +1396,33 @@ func (vs *vSphereVMProvider) vmCreateDoPlacement(
 
 	if pkgcfg.FromContext(vmCtx).Features.VMGroups &&
 		vmCtx.VM.Spec.GroupName != "" {
-		vmCtx.Logger.Info(
-			"Getting VM placement result from its group",
-			"groupName", vmCtx.VM.Spec.GroupName,
-		)
 
-		return vs.vmCreateDoPlacementByGroup(vmCtx, vcClient, createArgs)
+		// First update and check if the VM is linked to its group.
+		if err := vmopv1util.UpdateGroupLinkedCondition(
+			vmCtx,
+			vmCtx.VM,
+			vs.k8sClient,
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/vmprovider_vm_test.go
```diff
@@ -3978,6 +3978,89 @@ func vmTests() {
 			})
 		})
 
+		Context("Zone Label Override for VM Groups", func() {
```
dilyar85: Consider moving this under the existing VirtualMachineGroup -> VM Creation Context here: https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmprovider_vm_test.go#L1295-L1313

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1396,12 +1396,33 @@ func (vs *vSphereVMProvider) vmCreateDoPlacement(
...
+		if !pkgcnd.IsTrue(
+			vmCtx.VM,
+			vmopv1.VirtualMachineGroupMemberConditionGroupLinked,
+		) {
+			return fmt.Errorf("VM is not linked to its group")
+		}
+
+		// If the VM has an explicit zone label, skip group placement
+		// and use the regular placement flow to respect the zone override.
+		if zoneName := vmCtx.VM.Labels[topology.KubernetesTopologyZoneLabelKey]; zoneName == "" {
```
lubronzhan: I think you want to replace this with `LabelTopologyZone`?

lubronzhan (approved):
lgtm

PR #1174, opened 2025-09-13
:seedling: Use upstream consts for Zone and Host name label keys

aruneshpa:

While browsing through some code upstream, I found that these are defined upstream.  So, there's no point in maintaining our own variables.  This has a few added benefits of reducing imports in packages -- including reducing dependence in WCP service since we need to reference the zone label during restore.

```release-note
NONE
```


dilyar85 (approved):
LGTM, thanks @aruneshpa!

PR #1176, opened 2025-09-15
🐛 Fix NPE on async create VM

akutz:

This patch fixes an NPE that can occur on async create VM after commit a9e1dfbc0045d9abf50cc281a2a8843e9a1e1314.

Fixes #

I _really_ wanted to write a test to validate this, but I cannot with GoVmomi at the moment. The issue is the way one injects faults into the `CreateVm_Task` API on the `Folder` object. It happens like so:

1. Define a fake folder that returns an error:

```golang

type fakeVimFolder struct {
	*simulator.Folder
}

func (f *fakeVimFolder) CreateVMTask(ctx *simulator.Context, c *vimtypes.CreateVM_Task) soap.HasFault {
	return &methods.CreateVM_TaskBody{
		Fault_: simulator.Fault("fake", &vimtypes.InvalidRequest{}),
	}
}
```

2. Register fake folder in the simulator:

```golang
						folderWithErr := &fakeVimFolder{
							Folder: &simulator.Folder{
								Folder: mo.Folder{
									ManagedEntity: mo.ManagedEntity{
										ExtensibleManagedObject: mo.ExtensibleManagedObject{
											Self: nsInfo.Folder.Reference(),
										},
									},
								},
							},
						}
						ctx.SimulatorContext().Map.Put(folderWithErr)
```

The problem is that there is code in vC Sim that does the following:

```golang
func (p *ResourcePool) ImportVApp(ctx *Context, req *types.ImportVApp) soap.HasFault {
	body := new(methods.ImportVAppBody)

	spec, ok := req.Spec.(*types.VirtualMachineImportSpec)
	if !ok {
		body.Fault_ = Fault(fmt.Sprintf("%T: type not supported", spec), &types.InvalidArgument{InvalidProperty: "spec"})
		return body
	}

	dc := ctx.Map.getEntityDatacenter(p)
	folder := ctx.Map.Get(dc.VmFolder).(*Folder)
```

Well, that last line there is a problem since that folder is no longer a `*simulator.Folder`, but a `*fakeVimFolder` in order to return the injected fault. A panic occurs.

Until we update GoVmomi to allow injecting a createVm fault, we cannot validate this with a test in VM Op. To expedite this fix, let's get the PR merged and we can work on the follow-up in parallel. Thanks!

```release-note
Fix NPE in async create VM workflow.
```


abaruni (approved):
Thank you!!!

PR #1179, opened 2025-09-16
🌱Add mutating webhook for rac and wsfc workload

lubronzhan:

Auto fill PVC's config based on the application type.

Referenced from the API definition [here](https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha5/virtualmachine_storage_types.go#L106-L124)
```
	//   - OracleRAC      -- The volume is configured with
	//                       diskMode=IndependentPersistent and
	//                       sharingMode=MultiWriter
	//   - MicrosoftWSFC  -- The volume is configured with
	//                       diskMode=IndependentPersistent
```

```release-note
Add mutating webhook for rac and wsfc workload
```

### DIFF
--- a/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
+++ b/webhooks/virtualmachine/mutation/virtualmachine_mutator.go
@@ -245,6 +245,11 @@ func (m mutator) Mutate(ctx *pkgctx.WebhookRequestContext) admission.Response {
 				return admission.Denied(err.Error())
 			}
 		}
+		if pkgcfg.FromContext(ctx).Features.VMSharedDisks {
+			if _, err := SetPVCVolumeDefaultsOnCreate(ctx, m.client, modified); err != nil {
+				return admission.Denied(err.Error())
+			}
+		}
 	case admissionv1.Update:
 		oldVM, err := m.vmFromUnstructured(ctx.OldObj)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -864,3 +869,37 @@ func CleanupApplyPowerStateChangeTimeAnno(
...
+			v.PersistentVolumeClaim.SharingMode = vmopv1.VolumeSharingModeMultiWriter
+			wasMutated = true
+		case vmopv1.VolumeApplicationTypeMicrosoftWSFC:
+			v.PersistentVolumeClaim.DiskMode = vmopv1.VolumeDiskModeIndependentPersistent
+			wasMutated = true
+		case "":
+		// skip
+		default:
+			return false, field.Invalid(
+				field.NewPath("spec", fmt.Sprintf("volumes[%d]", i),
```
bryanv: Use`Index()` instead of `fmt.Sprintf`

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -864,3 +869,37 @@ func CleanupApplyPowerStateChangeTimeAnno(
...
+			v.PersistentVolumeClaim.DiskMode = vmopv1.VolumeDiskModeIndependentPersistent
+			v.PersistentVolumeClaim.SharingMode = vmopv1.VolumeSharingModeMultiWriter
+			wasMutated = true
+		case vmopv1.VolumeApplicationTypeMicrosoftWSFC:
+			v.PersistentVolumeClaim.DiskMode = vmopv1.VolumeDiskModeIndependentPersistent
+			wasMutated = true
+		case "":
+		// skip
+		default:
+			return false, field.Invalid(
```
bryanv: NotSupported?

webhooks/virtualmachine/mutation/virtualmachine_mutator_intg_test.go
```diff
@@ -786,4 +789,112 @@ func intgTestsMutating() {
 			})
 		})
 	})
+
+	Context("SetPVCVolumeDefaultsOnCreate", func() {
```
bryanv, line 793: It looks like these integration tests are dups of the unit tests - I don't think it is useful to cover the same things here, let the unit tests be more exhaustive

zhangheliu (approved):

aruneshpa (approved):
LGTM.

One question.  I think we said that we were going to make the application type immutable for now.  If so, maybe add those changes in the validation webhook as well?

lubronzhan:
> LGTM.
> 
> One question. I think we said that we were going to make the application type immutable for now. If so, maybe add those changes in the validation webhook as well?

Yep, let me add those in follow up PR

PR #1182, opened 2025-09-16
🌱 Remove VM Operator managed labels in VM placement 

dilyar85:

This PR updates the `attachVMLabelsToPolicy` function to remove any VM-Operator managed labels (identified by `vmoperator.vmware.com` in label key), so that we're not sending unrelated labels to DRS for VM placement.

This PR also adds a validating check to disallow setting any VM-Operator managed labels in VM's affinity/anti-affinity polices. 

Fixes N/A.

Deployed this change to an internal testbed and verified the placement policy in VM's creation ConfigSpec:

- Created a VM using following YAML spec

```yaml
kind: VirtualMachine
metadata:
  name: test-vm-af-1
  namespace: sdiliyaer-test
  labels:
    app: nginx
    tier: web
    vmoperator.vmware.com/test: test-prefix-label
    test.vmoperator.vmware.com/test: test-domain-label
spec:
  groupName: vmg-root
  className: best-effort-small
  imageName: photon-5.0
  affinity:
    vmAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            tier: web
        topologyKey: topology.kubernetes.io/zone
``` 

- Verified the ConfigSpec used for VM creation didn't include VM-Operator managed labels

```yaml
    "vmPlacementPolicies": [
        {
            "_typeName": "VmPlacementPolicy",
            "tagsToAttach": [
                "app:nginx",
                "tier:web"
            ]
        }
    ]
```

```release-note
Remove VM Operator managed labels in VM placement  and disallow them in VM's AF/AAF policy spec.
```


webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
```diff
@@ -3348,6 +3348,122 @@ func unitTestsValidateCreate() {
...
+										},
+										MatchExpressions: []metav1.LabelSelectorRequirement{
+											{
+												Key:      "vmoperator.vmware.com/paused",
+												Operator: metav1.LabelSelectorOpIn,
+												Values:   []string{"true"},
+											},
+										},
+									},
+									TopologyKey: "topology.kubernetes.io/zone",
```
lubronzhan: nit:
could be replaced with `corev1.LabelTopologyZone`

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -2155,6 +2156,10 @@ func (v validator) validateVMAffinity(
 						allErrs = append(allErrs, field.Forbidden(p, "label selector must match VM"))
 					}
 
+					if kubeutil.HasVMOperatorLabels(rs.LabelSelector.MatchLabels) {
```
lubronzhan, line 2159: Is it better to do it before `selector.Matches(vmLabelSet)`? Same for all below logic.

dilyar85: I don't think it's really matter since we're reporting all errors together?

webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
dilyar85: Thanks. Updated here and other test cases too.

webhooks/virtualmachine/validation/virtualmachine_validator.go
lubronzhan: mm yeah make sense

pkg/util/kube/label.go
```diff
@@ -0,0 +1,58 @@
...
+	// created by CAPV.
+	CAPVClusterRoleLabelKey = "capv.vmware.com/cluster.role"
+
+	// VMOperatorLabelDomain is the domain for VM Operator managed labels.
+	VMOperatorLabelDomain = "vmoperator.vmware.com"
+)
+
+// HasCAPILabels returns true if the VM has a label indicating it was created by
+// Cluster API such as CAPW or CAPV.
+func HasCAPILabels(labels map[string]string) bool {
```
zhangheliu, line 26: Do we still need to support CAPW?

pkg/util/kube/label.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+
+	return false
+}
+
+// RemoveVMOperatorLabels returns a copy of the labels map with all VM Operator
+// managed labels removed. VM Operator managed labels are identified with the
+// "vmoperator.vmware.com" domain in their key.
+func RemoveVMOperatorLabels(labels map[string]string) map[string]string {
```
zhangheliu, line 48: nit: wondering the ratio between the size of the deletion v.s. the whole map. Would in place deletion help with performance if the ratio is small?

dilyar85: This function is called from [attachVMLabelsToPolicy](https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/virtualmachine/configspec.go#L285) by passing the `vmCtx.VM.Labels` param. Since Go maps are referenced types, if we do in place deletion it would remove the VMOP managed labels from the VM itself as well.

pkg/util/kube/label.go
dilyar85: Not sure. This was just copy pasted from the original location. If it's not needed anymore, we can drop it in a follow-up PR.

webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
```diff
@@ -3348,6 +3348,122 @@ func unitTestsValidateCreate() {
...
+											"vmoperator.vmware.com/paused": "true",
+										},
+									},
+									TopologyKey: corev1.LabelTopologyZone,
+								},
+							},
+						}
+					},
+					validate: doValidateWithMsg(
+						`spec.affinity.vmAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].labelSelector: Forbidden: label selector can not contain VM Operator managed labels (vmoperator.vmware.com)`),
```
bryanv: Shouldn't the path include `matchLabels `?

dilyar85: Good catch! Updated all occurrences in webhook code and tests. Thanks for pointing this out.

PR #1186, opened 2025-09-17
✨ Random / parallel testing to speed up testing

akutz:

This patch updates VM Op to run its unit/integration tests in a random order, in parallel (using as many processes as there are CPUs).

This change reduces the time to test one of the larger packages, ./pkg/providers/vsphere, from 502s to only 88s.

The randomization also ensures both the production code and test code is reslient and more decomposed.

Finally, if there is a need to maintain ordering, because VM Op uses Ginkgo v2, the Ordered decorator may be utilized. There are examples of this in the code-base already, such as in ./pkg/exit/exit_test.go. Just search the project for `Ordered,` to find such examples, ex.: `grep -r 'Ordered,' * `.

Fixes `NA`

```release-note
Speed up and stabilize testing by randomizing and parallelizing all VM Op unit and integration tests.
```


lubronzhan (approved):
Thanks for refactoring! Especially those in snapshot

PR #1188, opened 2025-09-18
:sparkles: Add skeleton controller to reconcile volumes through batch attach

aruneshpa:

This change introduces the skeleton controller which will eventually replace the existing volume controller to reconcile the volumes using a new CnsNodeVmBatchAttachment API.

A lot of this code is placeholder methods to enable parallel execution of this effort.

```release-note
None
```

### DIFF
--- a/controllers/virtualmachine/volumebatch/volumebatch_controller.go
+++ b/controllers/virtualmachine/volumebatch/volumebatch_controller.go
@@ -1,0 +1,678 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package volumebatch
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	cnsv1alpha1 "github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/api/v1alpha1"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
+	"github.com/vmware-tanzu/vm-operator/pkg/patch"
+	"github.com/vmware-tanzu/vm-operator/pkg/record"
+	pkgutil "github.com/vmware-tanzu/vm-operator/pkg/util"
+)
+
+const (
+	controllerName = "volumebatch"
+)
+
+// AddToManager adds this package's controller to the provided manager.
+func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) error {
+	var (
+		controllerNameShort = fmt.Sprintf("%s-controller", strings.ToLower(controllerName))
+		controllerNameLong  = fmt.Sprintf("%s/%s/%s", ctx.Namespace, ctx.Name, controllerNameShort)
+	)
+
+	r := NewReconciler(
+		ctx,
+		mgr.GetClient(),
+		ctrl.Log.WithName("controllers").WithName("volumebatch"),
+		record.New(mgr.GetEventRecorderFor(controllerNameLong)),
+	)
+
+	c, err := controller.New(controllerName, mgr, controller.Options{
+		Reconciler:              r,
+		MaxConcurrentReconciles: ctx.MaxConcurrentReconciles,
+		LogConstructor:          pkglog.ControllerLogConstructor(controllerNameShort, &vmopv1.VirtualMachine{}, mgr.GetScheme()),
+	})
+	if err != nil {
+		return err
+	}
+
+	if err := c.Watch(source.Kind(
+		mgr.GetCache(),
+		&vmopv1.VirtualMachine{},
+		&handler.TypedEnqueueRequestForObject[*vmopv1.VirtualMachine]{},
+	)); err != nil {
+		return fmt.Errorf("failed to start VirtualMachine watch: %w", err)
+	}
+
+	// Watch for changes to CnsNodeVmBatchAttachment, and enqueue a
+	// request to the owner VirtualMachine.
+	if err := c.Watch(source.Kind(
+		mgr.GetCache(),
+		&cnsv1alpha1.CnsNodeVmBatchAttachment{},
+		handler.TypedEnqueueRequestForOwner[*cnsv1alpha1.CnsNodeVmBatchAttachment](
+			mgr.GetScheme(),
+			mgr.GetRESTMapper(),
+			&vmopv1.VirtualMachine{},
+			handler.OnlyControllerOwner(),
+		),
+	)); err != nil {
+		return fmt.Errorf("failed to start CnsNodeVmBatchAttachment watch: %w", err)
+	}
+
+	return nil
+}
+
+func NewReconciler(
+	ctx context.Context,
+	client client.Client,
+	logger logr.Logger,
+	recorder record.Recorder) *Reconciler {
+	return &Reconciler{
+		Context:  ctx,
+		Client:   client,
+		logger:   logger,
+		recorder: recorder,
+	}
+}
+
+var _ reconcile.Reconciler = &Reconciler{}
+
+type Reconciler struct {
+	client.Client
+	Context  context.Context
+	logger   logr.Logger
+	recorder record.Recorder
+}
+
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=cns.vmware.com,resources=cnsnodevmbatchattachments,verbs=create;delete;get;list;watch;patch;update
+// +kubebuilder:rbac:groups=cns.vmware.com,resources=cnsnodevmbatchattachments/status,verbs=get;list
+
+// Reconcile reconciles a VirtualMachine object and processes the volumes for batch attachment.
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (_ ctrl.Result, reterr error) {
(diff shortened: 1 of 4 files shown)


controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,664 @@
...
+		Namespace: ctx.VM.Namespace,
+	}, attachment); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, err
+		}
+
+		return nil, nil
+	}
+
+	return attachment, nil
```
bryanv, line 246: Add note to validate that the batch attachment is actually owned by this VM

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+
+		// Add PVC capacity information
+		if err := r.updateVolumeStatusWithPVCInfo(ctx, volStatus.PersistentVolumeClaim.ClaimName, &vmVolStatus); err != nil {
+			ctx.Logger.Error(err, "Failed to get PVC capacity info for volume", "volume", volStatus.Name, "pvc", volStatus.PersistentVolumeClaim.ClaimName)
+		}
+
+		volumeStatuses = append(volumeStatuses, vmVolStatus)
+	}
+
+	ctx.VM.Status.Volumes = volumeStatuses
```
bryanv, line 531: Should call `SortVirtualMachineVolumeStatuses()`

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+	Context  context.Context
+	logger   logr.Logger
+	recorder record.Recorder
+}
+
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=vmoperator.vmware.com,resources=virtualmachines/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=cns.vmware.com,resources=cnsnodevmbatchattachments,verbs=create;delete;get;list;watch;patch;update
+// +kubebuilder:rbac:groups=cns.vmware.com,resources=cnsnodevmbatchattachments/status,verbs=get;list
+
```
lubronzhan, line 117: TODO: add marker for pvc for instanceStorage use case

```
// +kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=create;delete;get;list;watch;patch;update
```

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+	if !vm.DeletionTimestamp.IsZero() {
+		// TODO: Handle deletion/detachment in future iterations
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.ReconcileNormal(volCtx); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
```
lubronzhan, line 170: TODO: reconcile result if instanceStorage presented

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+			Name:     volStatus.Name,
+			Type:     vmopv1.VolumeTypeManaged,
+			Attached: volStatus.PersistentVolumeClaim.Attached,
+			DiskUUID: volStatus.PersistentVolumeClaim.Diskuuid,
+			Error:    volStatus.PersistentVolumeClaim.Error,
+		}
+
+		// Add PVC capacity information
+		if err := r.updateVolumeStatusWithPVCInfo(ctx, volStatus.PersistentVolumeClaim.ClaimName, &vmVolStatus); err != nil {
+			ctx.Logger.Error(err, "Failed to get PVC capacity info for volume", "volume", volStatus.Name, "pvc", volStatus.PersistentVolumeClaim.ClaimName)
```
lubronzhan:
```suggestion
			ctx.Logger.Error(err, "Failed to update PVC capacity info for volume", "volume", volStatus.Name, "pvc", volStatus.PersistentVolumeClaim.ClaimName)
```

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+	}
+
+	// Check if update is needed
+	if r.volumeSpecsEqual(existing.Spec.Volumes, volumeSpecs) {
+		// No update needed, just update VM status from existing attachment
+		return r.updateVMStatusFromBatchAttachment(ctx, existing)
+	}
+
+	// Update the batch attachment with new volume specs
+	existing.Spec.Volumes = volumeSpecs
```
lubronzhan, line 358: It should also validate quota and stuff as well? 
```
validateVolumeSpecs
```

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+		}
+
+		// Apply application type presets first
+		// Ideally, this would already have been mutated by the webhook, but just handle that here anyway.
+		if err := r.applyApplicationTypePresets(pvcSpec, &cnsVolumeSpec); err != nil {
+			return nil, fmt.Errorf("failed to apply application type presets for volume %s: %w", vol.Name, err)
+		}
+
+		// Map disk mode (can override application type presets)
+		if pvcSpec.DiskMode != "" {
```
zhangheliu, line 393: Would not DiskMode always be equal to Persistent when it is not defined by the user, which then overrides the application setting?

lubronzhan: It's set in the API https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha5/virtualmachine_storage_types.go#L177

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+		if vol.PersistentVolumeClaim != nil {
+			pvcVolumes = append(pvcVolumes, vol)
+		}
+	}
+
+	// Handle WFFC for each PVC volume
+	for _, vol := range pvcVolumes {
+		if err := r.handlePVCWithWFFC(ctx, vol); err != nil {
+			ctx.Logger.Error(err, "Failed to handle WFFC for volume", "volume", vol.Name)
+			// Continue processing other volumes
```
lubronzhan, line 255: Then we should keep track of error for each vol, so that later when we create BatchAttechment spec, we should skip adding these if there are error

lubronzhan, line 265: MM, I think this func could be put inside of `processVolumesWithErrorHandling `?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+	// - Remove orphaned volumes from batch attachment
+	// - Preserve status information during cleanup
+	// - Handle cleanup errors gracefully
+
+	return nil
+}
+
+// processVolumesWithErrorHandling processes volumes with robust error handling.
+// This ensures partial failures don't prevent other volumes from being processed.
+func (r *Reconciler) processVolumesWithErrorHandling(_ *pkgctx.VolumeContext, _ []vmopv1.VirtualMachineVolume) error {
```
lubronzhan, line 621: Do we still need to make sure only one `pending` attachement at a time as previous logic? `onlyAllowOnePendingAttachment` https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachine/volume/volume_controller.go#L653

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan: Or if we want to filter early, then can we add "TODO: filter pvcs with error"

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan, line 621: Also need to filter the PVCs with error here

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,663 @@
...
+		existingMap[vol.Name] = vol
+	}
+
+	for _, desiredVol := range desired {
+		existingVol, exists := existingMap[desiredVol.Name]
+		if !exists {
+			return false
+		}
+
+		// Compare relevant fields.  Compare fields explicitly to avoid relying on default values.
```
zhangheliu: jsut wondering why this is an issue?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: We can add it when we actually fetch those.  There are ongoing conversations about the future of Instance storage.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: Hmm.  Remember that these are PVCs that already exist and we are only handling their attachment with the VM via this new resource.  CSI would have handled the quota evaluation and enforcement during volume creation.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: Basically trying to differentiate between zero value coming from unset, and set by user.  I don't think I need to worry about that because all of these are strings.  Updated.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: I will leave a TODO for this.   Even that logic was a hack at best so I would want to explore what breaks if we get rid of that.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: I am going to file a story to make sure users don't shoot themselves in the foot.

- If applicationType is specified, _and_ the diskMode etc are specified in the volume spec, they _must_ match.  If they don't we return an error.
- If no disk mode is specified and applicationType is not specified (today) then we default to persistent.

To your question specifically, no?  As of now, the diskmode can be specified by either the application type, or disk mode, or both.  When the controller reconciles the volume, it is simply setting the corresponding disk mode in the CnsNodeVmBatchAttachment spec.  Let me know if I am missing anything.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -0,0 +1,678 @@
...
+	pvcVolumes := make([]vmopv1.VirtualMachineVolume, 0)
+	for _, vol := range ctx.VM.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil {
+			pvcVolumes = append(pvcVolumes, vol)
+		}
+	}
+
+	// Handle WFFC for each PVC volume
+	for _, vol := range pvcVolumes {
+		if err := r.handlePVCWithWFFC(ctx, vol); err != nil {
```
lubronzhan, line 263: MM, I think this func could be put inside of processVolumesWithErrorHandling ?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
zhangheliu: I was more concerned about the Diskmode, which is currently set to Persistent by default, and always enters the override logic here.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: Maybe in a followup.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: Discussing over chat. We will handle in a followup if anything needs to be fixed.

PR #1190, opened 2025-09-18
🌱 Encrypt disks on copy for fast deploy direct

akutz:

This patch updates Fast Deploy's direct mode to encrypt disks as part of the copy operation instead of after the fact.

Fixes `NA`

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmlifecycle/create_fastdeploy.go
+++ b/pkg/providers/vsphere/vmlifecycle/create_fastdeploy.go
@@ -66,7 +66,7 @@ func fastDeploy(
 	}
 	logger.Info("Got destination file paths", "dstFilePaths", dstFilePaths)
 
-	// Collect the disks and remove the storage profile from them.
+	// Collect the disks.
 	var (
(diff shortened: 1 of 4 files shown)


pkg/providers/vsphere/vmlifecycle/create_fastdeploy.go
```diff
@@ -330,45 +331,65 @@ func fastDeployDirect(
...
+				AdapterType: string(vimtypes.VirtualDiskAdapterTypeLsiLogic),
+				DiskType:    string(vimtypes.VirtualDiskTypeThin),
+			},
+			SectorFormat: string(diskFormat),
+			Profile:      profile,
+		}
+
+		// Copy the disks using the same crypto key as the VM if the storage
+		// class is encrypted.
+		if isEncryptedStorageProfile && configSpec.Crypto != nil {
```
lubronzhan, line 363: Looks like `configSpec.Crypto` doesn't have to be not nil, if `ds.Backing.Crypto` is not nil?

PR #1194, opened 2025-09-19
:seedling: Add backoffLimit to VM publish request and a fatal reason

abaruni:

This PR introduces a new field, `backoffLimit`, to the VirtualMachinePublishRequest API. This field will tell the controller to stop retrying after a number of attempts and mark the requested completed with a newly added `Fatal` reason. Additionally, when quota check fails for inventory-backed content libraries, the request will be marked completed with a fatal reason.

Fixes #

```release-note
Add spec.backoffLimit to VirtualMachinePublishRequest API.
```


controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -287,6 +287,18 @@ func (r *Reconciler) updateSourceAndTargetRef(ctx *pkgctx.VirtualMachinePublishR
 // publishVirtualMachine checks if source VM and target is valid. Publish a VM if all requirements are met.
 func (r *Reconciler) publishVirtualMachine(ctx *pkgctx.VirtualMachinePublishRequestContext) error {
 	vmPublishReq := ctx.VMPublishRequest
+
+	if vmPublishReq.Status.Attempts == vmPublishReq.Spec.BackoffLimit {
```
bryanv: How about only if `Spec.BackoffLimit != 0` to allow for unlimited retries, and I think with a new VMOP deployed at the right time could cause this to fail before the first attempt.

Also, shouldn't this check should really be `>=`? And shouldn't this be done after the condition checks below, especially `VirtualMachinePublishRequestConditionUploaded`?

abaruni: Yes, it most certainly should occur after the condition checks. I think allowing for unlimited retries is good.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -304,6 +305,18 @@ func (r *Reconciler) publishVirtualMachine(ctx *pkgctx.VirtualMachinePublishRequ
 	if conditions.IsTrue(ctx.VMPublishRequest, vmopv1.VirtualMachinePublishRequestConditionSourceValid) &&
 		conditions.IsTrue(ctx.VMPublishRequest, vmopv1.VirtualMachinePublishRequestConditionTargetValid) {
 
+		if vmPublishReq.Spec.BackoffLimit != 0 && vmPublishReq.Status.Attempts >= vmPublishReq.Spec.BackoffLimit {
+			err := pkgerr.NoRequeueError{
+				Message: "publish attempts limit has been reached",
```
dilyar85: Nit: add the backoff limit in error message for better visibility?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -531,21 +544,25 @@ func (r *Reconciler) checkContentLibraryQuota(ctx *pkgctx.VirtualMachinePublishR
...
+
+		conditions.MarkError(vmPubReq,
+			vmopv1.VirtualMachinePublishRequestConditionTargetValid,
+			vmopv1.TargetContentLibraryFailedQuotaCheckReason,
+			err)
+
+		conditions.MarkError(vmPubReq,
+			vmopv1.VirtualMachinePublishRequestConditionComplete,
+			vmopv1.FatalReason,
+			err)
```
dilyar85, line 563: Feel like this is kind of redundant since the target valid condition already contains a fatal reason?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -1048,7 +1065,8 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachinePublishRequestCon
 	}()
 
 	// In case the .spec.ttlSecondsAfterFinished is not set, we can return early and no need to do any reconcile.
-	isComplete = conditions.IsTrue(vmPublishReq, vmopv1.VirtualMachinePublishRequestConditionComplete)
+	isComplete = conditions.IsTrue(vmPublishReq, vmopv1.VirtualMachinePublishRequestConditionComplete) ||
+		(conditions.GetReason(vmPublishReq, vmopv1.VirtualMachinePublishRequestConditionComplete) == vmopv1.FatalReason)
 	if isComplete {
 		requeueAfter, deleted, err := r.removeVMPubResourceFromCluster(ctx)
```
dilyar85: Is it expected to delete the CR after TTL in fatal state? It might be confusing for users to see their VM publish request CR is gone without an image being published.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: this is needed to signal that we've completed this request and won't attempt to further process it. it is meant for those parties particularly interested in publishing to inventory backed libraries

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: i don't know. but if they don't set a ttl, then we won't delete, correct?

dilyar85: > i don't know. but if they don't set a ttl, then we won't delete, correct?

Right, but if users set a TTL, they likely expect the CR to be deleted after a _successful_ publish. If this is not a strict requirement, I'd rather avoid deleting the CR with TTL if it fails due to a fatal reason, so users can still access it for reference.

api/v1alpha1/virtualmachinepublishrequest_conversion.go
```diff
@@ -5,21 +5,38 @@
...
+	if err := Convert_v1alpha1_VirtualMachinePublishRequest_To_v1alpha5_VirtualMachinePublishRequest(src, dst, nil); err != nil {
+		return err
+	}
+
+	restored := &vmopv1.VirtualMachinePublishRequest{}
+	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
+		return err
+	}
+
+	dst.Spec.BackoffLimit = restored.Spec.BackoffLimit
```
bryanv, line 27: Can you add tests in api/tests for these?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -1055,6 +1072,12 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachinePublishRequestCon
 		return ctrl.Result{RequeueAfter: requeueAfter}, err
 	}
 
+	// If this request has a complete condition with fatal reason, then we do not process it any further. We
+	// return early with no need to do any reconcile.
+	if conditions.GetReason(vmPublishReq, vmopv1.VirtualMachinePublishRequestConditionComplete) == vmopv1.FatalReason {
```
dilyar85: Can we introduce a var `isFatal` here and pass it to the defer func, so that it will register the metrics as failed if it's true? Something like:

```go
    // Register VM publish request metrics based on the reconcile result.
	var isComplete, isDeleted, isFatal bool
	defer func() {
		if isDeleted {
			// If the vmPub is deleted, return immediately.
			// We don't need to call DeleteMetrics here, we will run this in ReconcileDelete().
			return
		}

		var res metrics.PublishResult
		switch {
		case isComplete:
			res = metrics.PublishSucceeded
		case reterr != nil || isFatal:
			res = metrics.PublishFailed
		default:
			res = metrics.PublishInProgress
		}

		r.Metrics.RegisterVMPublishRequest(r.Logger, vmPublishReq.Name, vmPublishReq.Namespace, res)
	}()

	...

	isFatal = conditions.GetReason(vmPublishReq, vmopv1.VirtualMachinePublishRequestConditionComplete) == vmopv1.FatalReason
	if isFatal {
		// If this request has a complete condition with fatal reason, then we do not process it any further. We
		// return early with no need to do any reconcile.
		return ctrl.Result{}, reterr
	}
```
Otherwise with the current code, it'll always register `PublishInProgress` for fatal case since the reterr is nil.

api/test/v1alpha1/virtualmachinepublishrequest_conversion_test.go
```diff
@@ -0,0 +1,69 @@
+package v1alpha1_test
```
bryanv, line 5: Missing the header boilerplate

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -1055,6 +1072,13 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachinePublishRequestCon
 		return ctrl.Result{RequeueAfter: requeueAfter}, err
 	}
 
+	// If this request has a complete condition with fatal reason, then we do not process it any further. We
+	// return early with no need to do any reconcile.
+	isFatal = conditions.GetReason(vmPublishReq, vmopv1.VirtualMachinePublishRequestConditionComplete) == vmopv1.FatalReason
```
bryanv: If you're feeling it could just `conditions.Get(... vmopv1.VirtualMachinePublishRequestConditionComplete)` once earlier since isCompleted and isFatal are both checking the same condition

PR #1195, opened 2025-09-19
 🌱  Modify ConfigSpec creation to specify tags outside of placement policies

dilyar85:

This PR updates the ConfigSpec for VM creation to set tags outside of placement policies, by using the new govmomi APIs introduced in https://github.com/vmware/govmomi/commit/a3b63a98bfb91398208c42414973d0c98f0ca9b4.

This ensures a VM that doesn't specify any placement policy can also specify these tags and be part of another policy in future.

N/A.

- Ran the updated tests using ginkgo directly with `-p` and passed.

```release-note
Modify ConfigSpec creation to specify tags outside of placement policies.
```


pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -275,15 +277,42 @@ func genConfigSpecAffinityPolicies(
...
+	// It could be specified by this, or any other VM's placement policy later.
+	for key, value := range filteredLabels {
+		tagsToAdd = append(tagsToAdd, vimtypes.TagSpec{
+			ArrayUpdateSpec: vimtypes.ArrayUpdateSpec{
+				Operation: vimtypes.ArrayUpdateOperationAdd,
+			},
+			Id: vimtypes.TagId{
+				NameId: &vimtypes.TagIdNameId{
+					Tag:      key + ":" + value,
+					Category: vmCtx.VM.NamespacedName(),
```
aruneshpa: This should be namespace name -- not the namespaced name of the VM. Remember that category + name of a tag uniquely identifies a tag.  And VMs within a namespace should be able to specify a tag that can be used to affine them.  Putting the name of the VM in the category means now each tag is unique for every VM.

dilyar85: Updated to VM's namespace here and in test code.

pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -275,15 +277,42 @@ func genConfigSpecAffinityPolicies(
 	if len(placementPols) > 0 {
 		configSpec.VmPlacementPolicies = placementPols
 	}
+}
```
varunsl-brcm, line 277: Pre-existing but I think it should be safe to remove L265-L267 now?

varunsl-brcm (commented):
The changes that consume `vim.TagSpec` and related structures looks good to me.

pkg/providers/vsphere/virtualmachine/configspec.go
dilyar85: Good catch! Removed those lines since they're not used anymore.

aruneshpa (approved):

PR #1198, opened 2025-09-23
🌱Report VM's snapshot usage separately in VM Spec

lubronzhan:
New API changes for VM.status.storage.usage to reflect Snapshot usage separately.

- This snapshot usage is only shown when the snapshot feature is enabled. When not enabled, snapshot's usage is still included in old fashion - in `other` and `disks`.
- Modified GetSnapshotSize to return both managed(FCD) and unmanaged(Classic) disk usage
- Added GetAllSnapshotSize to calculate usage for all snapshots of a VM

New API change is in api/v1alpha5/virtualmachine_storage_types.go

With snapshot enabled, this is the new view
```
    storage:
      requested:
        disks: 16Gi
      total: "21563626188"
      usage:
        disks: "109052218"
        other: "4383757004"
        snapshots:
          vm: "921986291"
          volume: "81789879"
```

After disabling snapshot
```
    storage:
      requested:
        disks: 16Gi
      total: "21561092527"
      usage:
        disks: "899678737"
        other: "4381223343"
```

```release-note
Report VM's snapshot usage separately if feature is enabled
```


api/v1alpha5/virtualmachine_storage_types.go
```diff
@@ -354,25 +354,40 @@ type VirtualMachineStorageStatus struct {
 	Used *VirtualMachineStorageStatusUsed `json:"usage,omitempty"`
 }
 
+type VirtualMachineStorageStatusUsedSnapshotDetails struct {
+	// +optional
+
+	VM *resource.Quantity `json:"vm,omitempty"`
```
akutz, line 363: Please document these fields. The example I sent you did not have docs because you were intended to write them as mine was just an example :)

api/v1alpha5/virtualmachine_storage_types.go
```diff
@@ -354,25 +354,40 @@ type VirtualMachineStorageStatus struct {
...
+
+	// +optional
+
+	Volume *resource.Quantity `json:"volume,omitempty"`
```
akutz, line 370: Please document these fields. The example I sent you did not have docs because you were intended to write them as mine was just an example :)

api/v1alpha5/virtualmachine_storage_types.go
```diff
@@ -354,25 +354,40 @@ type VirtualMachineStorageStatus struct {
 	Used *VirtualMachineStorageStatusUsed `json:"usage,omitempty"`
 }
 
+type VirtualMachineStorageStatusUsedSnapshotDetails struct {
```
akutz, line 357: The snapshot should have an `Other` category, no?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -188,25 +188,23 @@ func (m snapshotMap) add(parent string, tree []vimtypes.VirtualMachineSnapshotTr
...
 	moVM mo.VirtualMachine,
-	vmSnapshot *vimtypes.ManagedObjectReference) (int64, error) {
+	vmSnapshot *vimtypes.ManagedObjectReference) (int64, int64, error) {
 	if vmSnapshot == nil || vmSnapshot.Value == "" {
-		return 0, fmt.Errorf("moSnapshot is nil or empty")
-	}
-
-	if moVM.Config == nil {
-		return 0, fmt.Errorf("moVM.Config is nil")
+		return 0, 0, fmt.Errorf("moSnapshot is nil or empty")
```
akutz: If the error is a constant string, use `errors.New`.  That or consider defining sentinel errors.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,37 +216,37 @@ func GetSnapshotSize(logger logr.Logger,
...
 	if snapshot == nil {
-		return 0, fmt.Errorf("snapshot with reference %q"+
+		return 0, 0, fmt.Errorf("snapshot with reference %q"+
 			" not found in vmLayout.Snapshot", vmSnapshot.Value)
 	}
 
-	fcdDeviceKeySet := getFCDDeviceKeySet(moVM)
+	fcdDeviceKeySet, err := getFCDDeviceKeySet(moVM)
+	if err != nil {
+		return 0, 0, err
```
akutz: Always use the pattern `fmt.Errorf("failed to get FCD device key set: %w", err)` to ensure that the returned error is wrapped with context.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,37 +216,37 @@ func GetSnapshotSize(logger logr.Logger,
 	}
 
 	if snapshot == nil {
-		return 0, fmt.Errorf("snapshot with reference %q"+
+		return 0, 0, fmt.Errorf("snapshot with reference %q"+
```
akutz: Change to `snapshot ref %q`. No need to be so verbose in an error string.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,37 +216,37 @@ func GetSnapshotSize(logger logr.Logger,
...
+	)
 
 	// Add the file key for the snapshot memory (.vmem) file if present.
 	if snapshot.MemoryKey != -1 {
 		logger.V(5).Info("Adding memoryKey", "memoryKey", snapshot.MemoryKey)
 		total += fileKeyMap[snapshot.MemoryKey]
 	}
 
-	logger.V(5).Info("Adding the file key for snapshot (.vmsn) file", "dataKey", snapshot.DataKey)
+	logger.V(5).Info("Adding the file key for snapshot (.vmsn) file",
```
akutz: Don't use special chars in log entries. Get rid of the `(.vmsn)` and change to `vmsn`.

akutz: Also, why all the L5 logs? We should be using error, Info/L2, and L4.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -259,18 +257,61 @@ func GetSnapshotSize(logger logr.Logger,
...
+	return total, fcdTotal, nil
+}
+
+// GetAllSnapshotSize calculates the size of a all snapshots of the VM in bytes.
+// It returns two sizes, first one includes the memory file, the vmdk files,
+// and the vmsn file, excludes FCDs. Second one only includes FCDs.
+func GetAllSnapshotSize(
+	logger logr.Logger,
+	moVM mo.VirtualMachine) (int64, int64, error) {
+	vmLayout := moVM.LayoutEx
```
akutz, line 301: There's no reason to alias `moVM.LayoutEx`. And if you did, always keep a blank line between the function signature and the first line in it.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -259,18 +257,61 @@ func GetSnapshotSize(logger logr.Logger,
...
-	return total, nil
+	return total, fcdTotal, nil
+}
+
+// GetAllSnapshotSize calculates the size of a all snapshots of the VM in bytes.
+// It returns two sizes, first one includes the memory file, the vmdk files,
+// and the vmsn file, excludes FCDs. Second one only includes FCDs.
+func GetAllSnapshotSize(
+	logger logr.Logger,
+	moVM mo.VirtualMachine) (int64, int64, error) {
```
akutz: Change to `(total, fcdTotal int64, _ error)`.  This way the function signature itself indicates what is being returned.

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -188,25 +188,23 @@ func (m snapshotMap) add(parent string, tree []vimtypes.VirtualMachineSnapshotTr
...
+// two sizes, first one includes the memory file, the vmdk files, and the vmsn
+// file, excludes FCDs. Second one only includes FCDs.
+// The algorithm use almost the same logic as how VC UI calculates the snapshot
+// size. The only difference is that this function returns the size of FCDs
+// as a different variable, and returns an error if the snapshot is not found.
+func GetSnapshotSize(
+	logger logr.Logger,
 	moVM mo.VirtualMachine,
-	vmSnapshot *vimtypes.ManagedObjectReference) (int64, error) {
+	vmSnapshot *vimtypes.ManagedObjectReference) (int64, int64, error) {
```
akutz: Change to `(total, fcdTotal int64, _ error)`.  This way the function signature itself indicates what is being returned.

akutz (changes requested):

api/v1alpha5/virtualmachine_storage_types.go
lubronzhan: Yeah we discussed this and thought this is not a P0. The Snapshot detail can just be categorized into
- VM: space used by VirtualMachineSnapshot (which includes other + disks).
- Volume: space used by VolumeSnapshot.

The customer cares more about the total number of space that will be freed if the VM is deleted.

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: Ok, just thought these are super verbose. Let me change that

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,59 +220,98 @@ func GetSnapshotSize(logger logr.Logger,
...
 }
 
 // getFCDDeviceKeySet returns a set of device keys of disk devices that are FCDs.
-func getFCDDeviceKeySet(moVM mo.VirtualMachine) sets.Set[int32] {
+func getFCDDeviceKeySet(moVM mo.VirtualMachine) (sets.Set[int32], error) {
 	deviceKeysSet := sets.Set[int32]{}
+
+	if moVM.Config == nil {
+		return nil, errMoVMConfigNil
+	}
```
palnabarun: If we all we need from moVM is Config, then this function can only get the config. The return values won't need to be changed in that case.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1094,7 +1106,13 @@ func updateStorageUsage(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 				vimtypes.VirtualMachineFileLayoutExFileTypeSnapshotMemory,
 				vimtypes.VirtualMachineFileLayoutExFileTypeSnapshotData:
 
-				// Skip snapshot related non-disk files
+				// Storage consumed by snapshots is tracked separately from
+				// the VM's files if VM service ensable vmsnapshot feature.
```
palnabarun: nit
```suggestion
				// the VM's files if VM Service vmsnapshot feature is enabled.
```

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1116,8 +1134,20 @@ func updateStorageUsage(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 		}
 	}
 
-	if disksReqd == 0 && disksUsed == 0 && other == 0 {
-		return
+	// Get the storage consumed by snapshots.
+	if snapEnabled {
+		vmSnapshotUsed, volSnapshotUsed, err = virtualmachine.GetAllSnapshotSize(logger, moVM)
```
palnabarun: Just confirming, is it a bad idea to pass vmCtx? I am guessing separation of concerns was why it is not passed, but both this package and virtualmachine package is inside the vsphere provider.

lubronzhan: I'm just trying to pass as minimum as possible. For example https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmlifecycle/update_status.go#L388-L390 here it doesn't pass the whole object either

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: Why the linter didn't catch this typo `ensable `

api/v1alpha5/virtualmachine_storage_types.go
lubronzhan: Discussed this offline.
Summary:
Andrew is ok with current code if it doesn't impact how we report snapshot usage to quota framework.
We don't use this data to report quota.

api/v1alpha5/virtualmachine_storage_types.go
```diff
@@ -354,25 +354,46 @@ type VirtualMachineStorageStatus struct {
...
+	// +optional
+
+	// VM describes the total storage space used by the VirtualMachine's
+	// VirtualMachineSnapshot. Including the space for snapshot's metadata,
```
bryanv: This is probably clearer to have this be one sentence like (.. VirtualMachineSnapshot including the snapshot's ...)

api/v1alpha5/virtualmachine_storage_types.go
```diff
@@ -354,25 +354,46 @@ type VirtualMachineStorageStatus struct {
...
+	// memory file, delta disks, etc.
+	VM *resource.Quantity `json:"vm,omitempty"`
+
+	// +optional
+
+	// Volume describes the total storage space used by the VirtualMachine's
+	// VolumeSnapshot. Including the space for first class disk(FCD)'s
```
bryanv: Same here

pkg/providers/vsphere/virtualmachine/snapshot_test.go
```diff
@@ -311,293 +313,419 @@ func snapShotTests() {
...
+				Expect(vcVM.Properties(
+					ctx, vcVM.Reference(), []string{"snapshot", "layoutEx", "config.hardware.device"}, &moVM)).
+					To(Succeed())
 				Expect(moVM.Snapshot).ToNot(BeNil())
 				vmCtx.MoVM = moVM
 
 				sum = 0
 				for _, file := range moVM.LayoutEx.File {
-					if strings.HasSuffix(file.Name, ".vmdk") || strings.HasSuffix(file.Name, ".vmsn") || strings.HasSuffix(file.Name, ".vmem") {
+					if strings.HasSuffix(file.Name, ".vmdk") ||
```
bryanv: Could make this more concise: https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmprovider_vm_snapshot_test.go#L99-L102

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -43,9 +46,9 @@ func SnapshotVirtualMachine(
 
 	logger := args.VMCtx.Logger.WithValues("snapshotName", snapshotName)
 	snapNode, err := FindSnapshot(args.VMCtx.MoVM, snapshotName)
-	// If it's other error except ErrMultipleSnapshots, that means it's either
-	// ErrSnapshotNotFound or ErrNoSnapshots, continue.
-	if errors.Is(err, ErrMultipleSnapshots) {
+	// If it's other error except errMultipleSnapshots, that means it's either
```
bryanv, line 51: I know this was existing but IMO this is kind of brittle. Can we change this to a switch that enumerates the not-error errors here and return any unexpected errors?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,60 +220,99 @@ func GetSnapshotSize(logger logr.Logger,
...
 
-	return total, nil
+	return total, fcdTotal, nil
+}
+
+// GetAllSnapshotSize calculates the size of a all snapshots of the VM in bytes.
+// It returns two sizes, first one includes the memory file, the vmdk files,
+// and the vmsn file, excludes FCDs. Second one only includes FCDs.
+func GetAllSnapshotSize(
+	logger logr.Logger,
```
bryanv: Could we pass in a `ctx context.Context` instead and have `GetSnapshotSize()` get its logger from there instead?

api/v1alpha5/virtualmachine_storage_types.go
bryanv, line 342: While here, maybe instead of the "+" like "and" instead to read better?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1149,10 +1179,26 @@ func updateStorageUsage(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 		vm.Status.Storage.Used.Other = kubeutil.BytesToResource(other)
 	}
 
+	if vmSnapshotUsed > 0 || volSnapshotUsed > 0 {
+		if vm.Status.Storage.Used.Snapshots == nil {
+			vm.Status.Storage.Used.Snapshots =
+				&vmopv1.VirtualMachineStorageStatusUsedSnapshotDetails{}
+		}
+		vm.Status.Storage.Used.Snapshots.VM = kubeutil.BytesToResource(vmSnapshotUsed)
```
bryanv: Do we want to check each of the respective Used vars for > 0 here since it looks like elsewhere here avoid setting the field for  zero?

pkg/providers/vsphere/vmlifecycle/update_status.go
bryanv: For `moVM.LayoutEx == nil` GetAllSnapshotSize() will return an error but earlier in the function that is just check for nil. Is returning an error in that case here really OK?

lubronzhan: Looks like in other place, like `updateVolumeStatus`, if `moVM.LayoutEx == nil` it just return silently. Let me follow that pattern then

lubronzhan: Is there a case that VM's moVM.LayoutEx will be nil? I assume no?

bryanv: IDK what situations it could happen but it seems we take care to check for it

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -32,9 +33,10 @@ type SnapshotArgs struct {
 
 // Snapshot related errors.
 var (
-	ErrNoSnapshots       = errors.New("no snapshots for this VM")
-	ErrSnapshotNotFound  = errors.New("snapshot not found")
-	ErrMultipleSnapshots = errors.New("multiple snapshots found")
+	ErrSnapshotNotFound        = errors.New("snapshot not found")
+	errNoSnapshots             = errors.New("no snapshots for this VM")
```
akutz: Why make this a non-exportable sentinel error?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -32,9 +33,10 @@ type SnapshotArgs struct {
...
+	errMultipleSnapshots       = errors.New("multiple snapshots found")
```
akutz: Why make this a non-exportable sentinel error?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -32,9 +33,10 @@ type SnapshotArgs struct {
...
+	errSnapshotMoRefNilOrEmpty = errors.New("snapshotMoRef is nil or empty")
```
akutz: Why make this a non-exportable sentinel error?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1149,10 +1178,30 @@ func updateStorageUsage(vm *vmopv1.VirtualMachine, moVM mo.VirtualMachine) {
 		vm.Status.Storage.Used.Other = kubeutil.BytesToResource(other)
 	}
 
+	if vmSnapshotUsed > 0 || volSnapshotUsed > 0 {
```
akutz, line 1181: I feel like this entire if block would be more clear as a switch statement. It also maps to Effective Go per https://go.dev/doc/effective_go#switch

pkg/providers/vsphere/vmprovider_vm_snapshot.go
```diff
@@ -122,7 +122,8 @@ func (vs *vSphereVMProvider) GetSnapshotSize(
 		return 0, fmt.Errorf("failed to find snapshot %q: %w", vmSnapshotName, err)
 	}
 
-	return virtualmachine.GetSnapshotSize(logger, moVM, &vmSnapshot.Snapshot)
+	total, _, err := virtualmachine.GetSnapshotSize(vmCtx.Context, moVM, &vmSnapshot.Snapshot)
```
akutz: Just pass in `vmCtx`. It will go into a `context.Context` just fine.

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: It's not used by other package anymore, but still it's referenced in existing package more than once.

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: It's not used by other package anymore, but still it's referenced in existing package more than once.

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: Yeah this is only used once, let me change that

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: Actually, this is not good for switch statement IMHO. These conditions are not mutually exclusive.

One way I can think of using switch is like below, still not as clear as using if. 

WDYT

```
ensureSnapshot := func(){
	if vm.Status.Storage.Used.Snapshots == nil {
		vm.Status.Storage.Used.Snapshots =
			&vmopv1.VirtualMachineStorageStatusUsedSnapshotDetails{}
	}
}
setVMSnapshotUsed = func(){
	vm.Status.Storage.Used.Snapshots.VM = kubeutil.BytesToResource(vmSnapshotUsed)
}
setVolSnapshotUsed = func(){
	vm.Status.Storage.Used.Snapshots.Volume = kubeutil.BytesToResource(volSnapshotUsed)
}
switch{
	case vmSnapshotUsed > 0 && volSnapshotUsed > 0:
		ensureSnapshot()
		setVMSnapshotUsed()
		setVolSnapshotUsed()
	case vmSnapshotUsed > 0:
		ensureSnapshot()
		setVMSnapshotUsed()
	case volSnapshotUsed > 0:
		ensureSnapshot()
		setVolSnapshotUsed()
}
```

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,60 +229,100 @@ func GetSnapshotSize(logger logr.Logger,
...
 		fileKeyMap[file.Key] = file.Size
 	}
 
-	var total int64
+	logger := pkglog.FromContextOrDefault(ctx).
+		WithValues("snapshotRef", vmSnapshot.Value)
 
-	// Add the file key for the snapshot memory (.vmem) file if present.
+	// Add the file key for the snapshot memory (vmem) file if present.
 	if snapshot.MemoryKey != -1 {
```
zhangheliu: nit: const for the init/default value?

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,60 +229,100 @@ func GetSnapshotSize(logger logr.Logger,
...
-
 		if len(disk.Chain) == 0 {
-			logger.V(5).Info("Skipping the disk since its chain is empty", "diskKey", disk.Key)
+			logger.V(4).Info("Skipping the disk since its chain is empty", "diskKey", disk.Key)
 			continue
 		}
 
 		// File keys for the child most delta disk
 		childMostFileKeys := disk.Chain[len(disk.Chain)-1].FileKey
 		if len(childMostFileKeys) > 0 {
```
zhangheliu: is len(childMostFileKeys) > 0 check needed? the loop is noop is the len is 0, therefore curDiskTotal would still be 0

pkg/providers/vsphere/virtualmachine/snapshot.go
```diff
@@ -218,60 +229,100 @@ func GetSnapshotSize(logger logr.Logger,
...
 }
 
 // getFCDDeviceKeySet returns a set of device keys of disk devices that are FCDs.
-func getFCDDeviceKeySet(moVM mo.VirtualMachine) sets.Set[int32] {
+func getFCDDeviceKeySet(conf *vimtypes.VirtualMachineConfigInfo) sets.Set[int32] {
+
 	deviceKeysSet := sets.Set[int32]{}
-	device := object.VirtualDeviceList(moVM.Config.Hardware.Device)
+
+	if conf == nil {
```
zhangheliu, line 323: nit: nil check at beginning and then return nil?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -385,11 +385,14 @@ func reconcileStatusStorage(
 	_ *object.VirtualMachine,
 	_ ReconcileStatusData) []error { //nolint:unparam
 
+	var errs []error
```
zhangheliu, line 388: nit: declar name for the return value directly?

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: If I return nil, then later I need to do nil check of `fcdDeviceKeySet` if it's being used

pkg/providers/vsphere/virtualmachine/snapshot.go
lubronzhan: Logically not needed but just to add it to save some steps if it's empty. But yeah can be removed

pkg/providers/vsphere/vmlifecycle/update_status.go
lubronzhan: I'm following the same pattern that the rest of the repo is using. Looks like it's only used to avoid confusion. In this case it doesn't have much big benefit

akutz: Agreed with @lubronzhan . Named return values are tricky. We should have probably used them in the validator (which is where the `var errs []error` pattern originated), but I agree we should stick with convention for now.

PR #1200, opened 2025-09-24
🐛 Fix typecast for VM total memory

zhangheliu:
The patch fixes the integer overflow issue when setting vm.status.hardware.memory from MoVM. The memory is stored as int32 in MoVM which leads to an integer overflow when multiplying it by 10^6.

Before:
Screenshot 2025-09-23 at 6 46 01 PM

After:
Screenshot 2025-09-23 at 6 47 22 PM

Fixes # N/A

N/A

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmlifecycle/update_status_test.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status_test.go
@@ -7,6 +7,7 @@ package vmlifecycle_test
 import (
 	"context"
 	"fmt"
+	"math"
 	"slices"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


lubronzhan (approved):
Good catch, LGTM!

pkg/providers/vsphere/vmlifecycle/update_status_test.go
```diff
@@ -2756,7 +2756,7 @@ var _ = Describe("Hardware status", func() {
 				vmCtx.MoVM.Config = &vimtypes.VirtualMachineConfigInfo{
 					Hardware: vimtypes.VirtualHardware{
 						NumCPU:   4,
-						MemoryMB: 2048,
+						MemoryMB: 4096,
```
aruneshpa: Drop a comment here indicating that you are choosing this value so that the int32 boundary of 2,147,483,647 is crossed?

aruneshpa (approved):
LGTM.  Would you mind putting in a sample before and after status.memory output?

pkg/providers/vsphere/vmlifecycle/update_status_test.go
zhangheliu: good point. I added a new UT with a proper description.

PR #1201, opened 2025-09-24
[x-port] Remove caBundle from conversion webhook defs

aruneshpa:

Remove caBundle from conversion webhook defs
This is a x-port of #649 for the upcoming patch release on 8.0 U3.

```release-note
Remove caBundle from conversion webhook defs
```


aruneshpa:
We can ignore the test failures here since we are using the runners that use deps from main which do not work on the U3 branch.

I have run tests locally and it passes.

```
make test
...
Ginkgo ran 63 suites in 12m37.550413125s
Test Suite Passed
+ '[' 0 -eq 2 ']'
+ exit 0
```

I've also generated a build and ran pipelines with this change.  So, should be good to merge.

PR #1202, opened 2025-09-25
🌱Implement greenfield volume attach of WFFC type volumes

lubronzhan:

Add logic of handling PVC with WFFC, validate hardware, reconcile delete

- The logic was mostly borrowed from volume_controller.go.
- Add logic to handle PVC with WFFC, validate the hardware, reconcileDelete(does nothing)
	- Validate PVC with WFFC fails fast before creating BatchAttachment so we could incur less ReconfigureVM calls
- go.mod is updated to include `k8s.io/component-helpers` so we could use upstream label: `storagehelpers.AnnSelectedNode`

- Changed the unit test to test ReconcileNormal and ReconcileDelete

```release-note
Implement greenfield volume attach of WFFC type volumes, validate hardware through CnsNodeVmBatchAttachment.
```


controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -590,17 +598,90 @@ func (r *Reconciler) reconcileInstanceStoragePVCs(_ *pkgctx.VolumeContext) (bool
...
+	if err := r.Get(ctx, pvcKey, &pvc); err != nil {
+		return fmt.Errorf("cannot get PVC: %w", err)
+	}
+
+	if pvc.Status.Phase == corev1.ClaimBound {
+		// Regardless of the StorageClass binding mode there is nothing to done if already bound.
+		return nil
+	}
+
+	if pvc.Annotations[constants.KubernetesSelectedNodeAnnotationKey] != "" {
```
bryanv: Why was this added? Unlike pods, the value of this is the zone name because of the `CNSSelectedNodeIsZoneAnnotationKey` annotation

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -590,17 +598,90 @@ func (r *Reconciler) reconcileInstanceStoragePVCs(_ *pkgctx.VolumeContext) (bool
...
+// HandlePVCWithWFFC handles PVCs with WaitForFirstConsumer binding mode.
 // This ensures proper node selection for storage classes requiring it.
-func (r *Reconciler) handlePVCWithWFFC(_ *pkgctx.VolumeContext, _ vmopv1.VirtualMachineVolume) error {
-	// TODO: Implement WFFC (WaitForFirstConsumer) handling
-	// This method should:
-	// - Check storage class binding mode
-	// - Set selected node annotations for unbound PVCs
-	// - Handle node selection for proper PVC binding
-	// - Validate node availability and zone constraints
+func (r *Reconciler) HandlePVCWithWFFC(ctx *pkgctx.VolumeContext, volume vmopv1.VirtualMachineVolume) error {
```
akutz: No. This should be private and tested via ReconcileNormal.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -590,17 +598,90 @@ func (r *Reconciler) reconcileInstanceStoragePVCs(_ *pkgctx.VolumeContext) (bool
...
+	}
+	pvc.Annotations[volumebatchutils.CNSSelectedNodeIsZoneAnnotationKey] = "true"
+	pvc.Annotations[constants.KubernetesSelectedNodeAnnotationKey] = zoneName
+
+	if err := r.Client.Update(ctx, &pvc); err != nil {
+		return fmt.Errorf("cannot update PVC to add selected-node annotation: %w", err)
+	}
 
 	return nil
+
```
bryanv: Remove

akutz (changes requested):
Do not have any exported methods on the Reconciler other than `ReconcileNormal` and `ReconcileDelete`. All tests should be driven via those functions. Ideally via `Reconcile` itself except for the lack of true deletion in envtest. Tests are way too fragile if we start testing these functions directly. That's not how the controller calls it, and neither should/will we.

controllers/virtualmachine/volumebatch/utils/constants.go
```diff
@@ -0,0 +1,11 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package utils
```
aruneshpa: Can you move these out of the batch directory since this is shared by both version of controllers?

controllers/virtualmachine/volumebatch/utils/test_utils.go
```diff
@@ -0,0 +1,87 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	cnsv1alpha1 "github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/api/v1alpha1"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
+	"github.com/vmware-tanzu/vm-operator/pkg/util"
+	"github.com/vmware-tanzu/vm-operator/test/builder"
+)
+
+func AssertAttachmentSpecFromVMVol(
```
bryanv: I would not expect the batch controller to need this, which is for asserting the created node attachment is what is expected.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -590,17 +598,90 @@ func (r *Reconciler) reconcileInstanceStoragePVCs(_ *pkgctx.VolumeContext) (bool
...
+	}
+
+	if pvc.Annotations[constants.KubernetesSelectedNodeAnnotationKey] != "" {
+		// Try fetch the node object to check if it still exists. and the zone
+		// matches the VM's Zone.
+		node := &corev1.Node{}
+		nodeKey := client.ObjectKey{
+			Name: pvc.Annotations[constants.KubernetesSelectedNodeAnnotationKey],
+		}
+		if err := r.Get(ctx, nodeKey, node); err != nil {
```
aruneshpa: Comparing this to the current volume controller.  Has CSI started supporting changing the node of a PVC once the annotation is set?  I think we can still assume that if the node annotation is set, it will be a valid node.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -590,17 +598,90 @@ func (r *Reconciler) reconcileInstanceStoragePVCs(_ *pkgctx.VolumeContext) (bool
...
+		if zoneName == "" {
+			return fmt.Errorf("VM does not have Zone set")
+		}
+	}
+
+	if pvc.Annotations == nil {
+		pvc.Annotations = map[string]string{}
+	}
+	pvc.Annotations[volumebatchutils.CNSSelectedNodeIsZoneAnnotationKey] = "true"
+	pvc.Annotations[constants.KubernetesSelectedNodeAnnotationKey] = zoneName
```
aruneshpa: Nit: Consider using this const from upstream: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-helpers/storage/volume/pv_helpers.go#L51

controllers/virtualmachine/volumebatch/volumebatch_controller.go
aruneshpa: Nit: Remove this extra newline?

controllers/virtualmachine/volumebatch/utils/test_utils.go
aruneshpa: I don't see this being used anywhere other than volume_controller.  Are you moving this here because you anticipate us using it in a later change?
Even then, like BV said, this is specifically for the CnsNodeVmAttachment object (not batch).

.golangci.yml
```diff
@@ -197,6 +197,7 @@ linters:
           - "!$test"
           - "!**/test/builder/*.go"
           - "!**/matcher.go"
+          - "!**/test_utils.go"
```
aruneshpa: Once you move the assert method back, I think this can be removed.  Adding a placeholder comment.

controllers/virtualmachine/volume/volume_controller_unit_test.go
```diff
@@ -201,7 +202,10 @@ func unitTestsReconcile() {
 		instanceStorageTest = false
 	})
 
-	getCNSAttachmentForVolumeName := func(vm *vmopv1.VirtualMachine, volumeName string) *cnsv1alpha1.CnsNodeVmAttachment {
+	getCNSAttachmentForVolumeName := func(ctx *builder.UnitTestContextForController, vm *vmopv1.VirtualMachine, volumeName string) *cnsv1alpha1.CnsNodeVmAttachment {
```
bryanv: Is there a situation where the the captured context isn't sufficient anymore?

controllers/virtualmachine/volume/volume_controller_unit_test.go
```diff
@@ -401,9 +405,9 @@ func unitTestsReconcile() {
 				By("Ignores the CnsNodeVmAttachment for other VM", func() {
 					Expect(vm.Status.Volumes).To(HaveLen(1))
 
-					attachment := getCNSAttachmentForVolumeName(vm, vmVol.Name)
+					attachment := getCNSAttachmentForVolumeName(ctx, vm, vmVol.Name)
 					Expect(attachment).ToNot(BeNil())
-					assertVMVolStatusFromAttachment(vmVol, attachment, vm.Status.Volumes[0])
+					assertVMVolStatusFromAttachment(vmVol, attachment, volCtx.VM.Status.Volumes[0])
```
bryanv: Why convert to `volCtx.VM`? It is going to be the same value as before, and isn't this needless confusing otherwise since just on the prior change line it is still using `vm`?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -258,33 +267,33 @@ func (r *Reconciler) processBatchAttachment(
...
+				"attachment", existingAttachment.Name,
+				"namespace", existingAttachment.Namespace)
+			if err := r.Client.Delete(ctx, existingAttachment); err != nil {
+				return fmt.Errorf("failed to delete attachment: %w", err)
+			}
 		}
 		return nil
 	}
 
+	var creationErrs []error
```
bryanv: Maybe it will be done later but I think we're missing code that diffs what's in the batch and VM Spec to determine what is actually be created. I don't think we should need to look at a PVC again after its volume entry has been added to the Batch spec.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -592,26 +585,73 @@ func (r *Reconciler) reconcileInstanceStoragePVCs(_ *pkgctx.VolumeContext) (bool
...
+	}
+
+	if pvc.Status.Phase == corev1.ClaimBound {
+		// Regardless of the StorageClass binding mode there is nothing to done if already bound.
+		return nil
+	}
+
+	if pvc.Annotations[storagehelpers.AnnSelectedNode] != "" {
+		// Once set, this annotation cannot really be changed so just keep going.
+		// TBD: Make check if the Node still exists, and is in the VM's Zone, but
```
bryanv: While here you can remove this stale comment

controllers/virtualmachine/volume/volume_controller_unit_test.go
lubronzhan: Good catch, forgot to clean this up

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan: Make sense, if that volume is ~already attached~ and no PVC change, then no need to handle PVC

pkg/util/cns.go
```diff
@@ -16,3 +22,21 @@ package util
 func CNSAttachmentNameForVolume(vmName, volumeName string) string {
 	return vmName + "-" + volumeName
 }
+
+// CNSBatchAttachmentNameForVolume returns the name of the
+// CnsNodeVmBatchAttachment based on the VM name.
+func CNSBatchAttachmentNameForVolume(vmName string) string {
+	return vmName + "-batch"
```
aruneshpa: Why even have the "batch" suffix?  Why not just use the VM name?
I would either just use the VM naem, or do a "batchattach" suffix.  First preference is just VM name.
Also mentioned in getBatchAttachmentForVM method.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -117,6 +128,7 @@ type Reconciler struct {
 
 // Reconcile reconciles a VirtualMachine object and processes the volumes for batch attachment.
 func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (_ ctrl.Result, reterr error) {
+
```
aruneshpa: Accidental?  I don't think we consistently do this.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -192,38 +204,35 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VolumeContext) error {
...
+		if err := r.validateHardwareVersion(ctx); err != nil {
+			return fmt.Errorf("hardware version validation failed: %w", err)
 		}
 	}
 
 	// Process volumes and create/update batch attachment
 	if err := r.processBatchAttachment(ctx, batchAttachment); err != nil {
-		ctx.Logger.Error(err, "Error processing CnsNodeVmBatchAttachment")
-		return err
+		return fmt.Errorf("error processing CnsNodeVmBatchAttachment: %w", err)
```
aruneshpa, line 223: Heads up that this API is about to be renamed to `CNSNodeVMBatchAttachment`.  So, depending on when we merge this, we may have to rebase.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -192,38 +204,35 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VolumeContext) error {
...
 
 	return nil
 }
 
-// GetBatchAttachmentForVM returns the CnsNodeVmBatchAttachment
+// getBatchAttachmentForVM returns the CnsNodeVmBatchAttachment
 // resource for the VM. We assume that the name of the resource
 // matches the name of the VM.
-func (r *Reconciler) GetBatchAttachmentForVM(ctx *pkgctx.VolumeContext) (*cnsv1alpha1.CnsNodeVmBatchAttachment, error) {
+// This only return nil obj and doesn't return error
```
aruneshpa:
```suggestion
// Returns nil if no CNSNodeVMBatchAttachment resource exists for the VM.
```

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -258,31 +267,37 @@ func (r *Reconciler) processBatchAttachment(
...
-			ctx.Logger.Error(err, "Failed to handle WFFC for volume", "volume", vol.Name)
-			// Continue processing other volumes
+	if len(pvcVolumes) == 0 {
+		// No PVC volumes to process
+		if existingAttachment != nil {
+			ctx.Logger.Info("Delete existing CnsNodeVmBatchAttachment",
+				"attachment", existingAttachment.Name,
+				"namespace", existingAttachment.Namespace)
+			if err := r.Client.Delete(ctx, existingAttachment); err != nil {
+				return fmt.Errorf("failed to delete attachment: %w", err)
```
aruneshpa: IgnoreNotFound to handle races when the object is cleaned up after we get the resource, but before we get here.

aruneshpa: `attachment` -> `CnsNodeVmBatchAttachment`?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -258,31 +267,37 @@ func (r *Reconciler) processBatchAttachment(
...
+		return nil
 	}
 
-	// Handle orphaned volume cleanup
+	curVolumeAttachSpecMap := make(map[string]cnsv1alpha1.VolumeSpec)
 	if existingAttachment != nil {
-		if err := r.cleanupOrphanedVolumes(ctx, pvcVolumes); err != nil {
-			ctx.Logger.Error(err, "Failed to cleanup orphaned volumes")
-			// Continue with processing
+		for _, status := range existingAttachment.Spec.Volumes {
```
aruneshpa: This sounds weird.  Can you call the range variable something else, like `volumeSpec`?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -258,31 +267,37 @@ func (r *Reconciler) processBatchAttachment(
...
 	}
-
-	if len(pvcVolumes) == 0 {
-		// No PVC volumes to process
-		if existingAttachment != nil {
-			// TODO: AKP: Delete existing CnsNodeVmBatchAttachmet or
-			// CnsNodeVmAttachment resource(s) when no volumes.
-			ctx.Logger.Info("Delete existing CnsNodeVmBatchAttachment",
-				"attachment", existingAttachment.Name)
+	var creationErrs []error
```
aruneshpa:
```suggestion
	var createErrs []error
```

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -565,15 +568,11 @@ func (r *Reconciler) updateVolumeStatusWithPVCInfo(
...
-func (r *Reconciler) ReconcileDelete(ctx *pkgctx.VolumeContext) error {
-	// TODO: Implement VM deletion cleanup logic
-	// This method should:
-	// - Clean up CnsNodeVmBatchAttachment resources
-	// - Handle any pending volume operations
-	// - Ensure proper resource cleanup on VM deletion
-
+func (r *Reconciler) ReconcileDelete(_ *pkgctx.VolumeContext) error {
+	// Do nothing here since we depend on the Garbage Collector to do the deletion of the
+	// dependent CNSNodeVMAttachment objects when their owning VM is deleted.
```
aruneshpa: CNSNodeVMAttachment and CNSNodeVMBatchAttachment

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -565,15 +568,11 @@ func (r *Reconciler) updateVolumeStatusWithPVCInfo(
...
+	// We require the Volume provider to handle the situation where the VM is deleted before
```
aruneshpa: This can be removed, no?  Per the above comment, when the VM (owner) is deleted, the batch attachment (and all other brownfield single node attachments) will be deleted -- which will detach the volumes from the VM.

pkg/util/cns.go
lubronzhan: Ok, gonna just use the vm name

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan: Cool, will do a replaceAll later

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan: Looks like if we make the func signature several lines, then better add an empty line? At least what Andrew told me last time on my PR

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan: Forgot to clean this up. Good catch!

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan: Then why do we have that from the old controller, it also has the ownerRef

bryanv: This comment is more of a warning to not go do a Delete() of the resource here, since the owning VM needs to go away first, since we don't want to rip out the disks before the VM is actually deleted

lubronzhan: Ok let me add it back

pkg/util/cns.go
```diff
@@ -16,3 +22,21 @@ package util
...
+	return vmName
+}
+
+// BuildControllerKey builds a controller key that CNS can understand
+// Format: <type>:<busNumber> or just <type> if bus number not specified.
+func BuildControllerKey(controllerType vmopv1.VirtualControllerType, busNumber *int32) string {
+	if busNumber != nil {
+		return fmt.Sprintf("%s:%d", controllerType, *busNumber)
+	}
+	if controllerType != "" {
```
bryanv: Just `return string(controllerType)` ?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
bryanv, line 377: Shouldn't `r.updateVMStatusFromBatchAttachment()` also be called in this case? It seem odd to wait to update the status until there are no volume changes, since there could be updates from a prior batch update.

bryanv (approved):

controllers/virtualmachine/volumebatch/volumebatch_controller.go
lubronzhan: Make sense. Should either check based on status or always updates. let me add a todo

PR #1203, opened 2025-09-25
🌱 Update VM hardware SCSI/SATA controllers status

zhangheliu:

This patch updates vm.status.hardware.controllers, SCSI and SATA, based on their base classes. ParaVirtualSCSIController,
VirtualBusLogicController, VirtualLsiLogicController, VirtualLsiLogicSASController, VirtualSCSIController will display as SCSI controllers and VirtualSATAController and VirtualAHCIController will display as SATA controllers. If needed, the exact type can be retrieved by looking at the vm.spec.hardware with the controllerType and the busNumber.

This patch also changes VirtualDeviceStatus name to label since the label contains the display string that users see on the VC UI.

Manual Testing:
Screenshot 2025-09-24 at 6 53 22 PM

Fixes # N/A

N/A

```release-note
NONE
```


pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,89 @@
...
+	"strings"
+
+	"github.com/vmware/govmomi/object"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+)
+
+// GetVirtualDiskFileNameAndUUID extracts the file name and disk UUID from a
+// VirtualDisk device. Returns empty strings for unsupported
+// backing types or when UUID is not available.
+func GetVirtualDiskFileNameAndUUID(vd *vimtypes.VirtualDisk) (
```
lubronzhan, line 63: Nice work here. Just curious, do you want to follow the exact pattern here https://github.com/vmware-tanzu/vm-operator/pull/1206/files#diff-f614f9460c468e586e5b2dcac36be35321d790bfd04ce0c81fab5aab2e92ea48R162

zhangheliu: yeah, I was following that PR but it seems like the pattern has changed. I will wait until that PR is merged to see how we could refactor the code after

lubronzhan:
```
  [FAILED] in [It] - /home/runner/work/vm-operator/vm-operator/controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go:275 @ 10/06/25 18:36:55.893
• [FAILED] [0.021 seconds]
Unit tests Volume Batch Controller ReconcileNormal when VM Spec.Volumes is not empty when controller type and bus number are set [It] sets volme variables correctly [controller, api]
/home/runner/work/vm-operator/vm-operator/controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go:273

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc001289520>: 
      error processing CnsNodeVmBatchAttachment: volume spec validation failed: batch spec volumes do not match vm spec volumes
      {
          msg: "error processing CnsNodeVmBatchAttachment: volume spec validation failed: batch spec volumes do not match vm spec volumes",
          err: <*fmt.wrapError | 0xc001289500>{
              msg: "volume spec validation failed: batch spec volumes do not match vm spec volumes",
              err: <*errors.errorString | 0xc000d0d0f0>{
                  s: "batch spec volumes do not match vm spec volumes",
              },
          },
      }
  occurred
  In [It] at: /home/runner/work/vm-operator/vm-operator/controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go:275 @ 10/06/25 18:36:55.893
--------------------------
```
Looks like some tests need to be fixed

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -354,6 +355,12 @@ func (r *Reconciler) updateBatchAttachment(
 		return r.updateVMStatusFromBatchAttachment(ctx, existing)
 	}
 
+	// Validate the volume specs before creating the batch attachment
```
lubronzhan:
```suggestion
	// Validate the volume specs before updating the batch attachment.
```

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,594 @@
...
+
+type DeviceInfo struct {
+	CtrlKey
+	UnitNumber int32
+}
+
+type CtrlInfo struct {
+	ControllerKey int32
+	SharingMode   vmopv1.VirtualControllerSharingMode
+	SCCIType      vmopv1.SCSIControllerType
```
lubronzhan:
```suggestion
	SCSIType      vmopv1.SCSIControllerType
```

zhangheliu:
> ```
>   [FAILED] in [It] - /home/runner/work/vm-operator/vm-operator/controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go:275 @ 10/06/25 18:36:55.893
> • [FAILED] [0.021 seconds]
> Unit tests Volume Batch Controller ReconcileNormal when VM Spec.Volumes is not empty when controller type and bus number are set [It] sets volme variables correctly [controller, api]
> /home/runner/work/vm-operator/vm-operator/controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go:273
> 
>   [FAILED] Unexpected error:
>       <*fmt.wrapError | 0xc001289520>: 
>       error processing CnsNodeVmBatchAttachment: volume spec validation failed: batch spec volumes do not match vm spec volumes
>       {
>           msg: "error processing CnsNodeVmBatchAttachment: volume spec validation failed: batch spec volumes do not match vm spec volumes",
>           err: <*fmt.wrapError | 0xc001289500>{
>               msg: "volume spec validation failed: batch spec volumes do not match vm spec volumes",
>               err: <*errors.errorString | 0xc000d0d0f0>{
>                   s: "batch spec volumes do not match vm spec volumes",
>               },
>           },
>       }
>   occurred
>   In [It] at: /home/runner/work/vm-operator/vm-operator/controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go:275 @ 10/06/25 18:36:55.893
> --------------------------
> ```
> 
> Looks like some tests need to be fixed

yeah, the ControllerKey needs to be retrieved from VM status hardware instread of "ControllerType:BusNumber" string.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+			return fmt.Errorf(ErrControllerNotFound, pvc.ControllerType,
+				*pvc.ControllerBusNumber)
+		}
+
+		switch pvc.ApplicationType {
+		case vmopv1.VolumeApplicationTypeOracleRAC:
+			if pvc.SharingMode != vmopv1.VolumeSharingModeMultiWriter {
+				return errors.New(ErrPVCSharingModeMustBeMultiWriter)
+			}
+			if ctrlInfo.SharingMode != vmopv1.VirtualControllerSharingModeNone {
```
lubronzhan, line 420: Could you add more context here? The pvc's sharing mode must be multiWriter while the controller's sharing mode must be none?

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+				return errors.New(ErrControllerSharingModeMustBeNone)
+			}
+		case vmopv1.VolumeApplicationTypeMicrosoftWSFC:
+			if pvc.SharingMode != vmopv1.VolumeSharingModeNone {
+				return errors.New(ErrPVCSharingModeMustBeNone)
+			}
+			if ctrlInfo.SharingMode !=
```
lubronzhan, line 427: Same as above

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -149,18 +149,133 @@ type VirtualDeviceStatus struct {
...
+
+	// +optional
+
+	// DiskUUID represents the underlying virtual disk UUID and is present when
+	// attachment succeeds.
+	DiskUUID string `json:"diskUUID,omitempty"`
+}
+
+// IDEControllerStatus describes the observed state of an IDE controller.
+type IDEControllerStatus struct {
```
aruneshpa, line 161: Having all these different distinct types seems too verbose, specially given that they share most of the fields.  And I thought the issue with the current approach was the "virtual" vs "paravirtual" SCSI bit.  How about:

```
type VirtualControllerStatus struct {
    Type        VirtualControllerType            `json:"type"`
    BusNumber   int32                           `json:"busNumber"`
    DeviceKey   int32                           `json:"deviceKey"`
    
    // Optional fields based on controller type.. Obviously, we will have documentation on these fields indicating that they will only be set for a specific controller type.
    SharingMode *VirtualControllerSharingMode   `json:"sharingMode,omitempty"`
    SCSIType    *SCSIControllerType            `json:"scsiType,omitempty"`
    PCISlot     *int32                         `json:"pciSlot,omitempty"`
    
    Devices     []VirtualDeviceStatus          `json:"devices,omitempty"`
}
```

This also dedupes the `GetBusNumber` and other corresponding methods for the type.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+			unitNumber  string
+		)
+
+		switch pvc.DiskMode {
+		case vmopv1.VolumeDiskModeNonPersistent:
+			diskMode = cnsv1alpha1.IndependentPersistent
+		case vmopv1.VolumeDiskModePersistent:
+			diskMode = cnsv1alpha1.Persistent
+		case "":
+		case vmopv1.VolumeDiskModeIndependentNonPersistent,
```
lubronzhan: I forgot about the context, is this gonna be supported later? Wonder if we don't support it should we also document in our API

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+}
+
+type VolumeKey struct {
+	VolumeName string
+	PVCName    string
+}
+
+type CtrlKey struct {
+	CtrlType  vmopv1.VirtualControllerType
+	BusNumber int32
```
aruneshpa, line 183: This should be *int32 to differentiate when a bus has not been specified, vs "0" has been specified explicitly?

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+}
+
+type DeviceInfo struct {
+	CtrlKey
+	UnitNumber int32
```
aruneshpa, line 188: This should be *int32 to differentiate when a unit number has not been specified, vs "0" has been specified explicitly?

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+
+	"github.com/vmware/govmomi/object"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
+	cnsv1alpha1 "github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/api/v1alpha1"
+)
+
+// Controller unit number limits.
```
aruneshpa: Maybe add a line explaining that this essentially indicates the maximum number of devices that you can connect to a controller of a specific type?  

Also, you are also grouping the reserved unit number slot in this same block.  Please add a comment there explaining how it is used or why is it relevant.

api/v1alpha5/virtualmachine_hardware_types.go
aruneshpa: It sounds like we already have a precedence for splitting these in the VirtualMachineHardwareSpec in various controller specs.  @akutz , in that case, maybe the current diff is okay.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+		if d.DiskUUID != "" {
+			vmDiskUUIDToDeviceInfoMap[d.DiskUUID] = DeviceInfo{
+				CtrlKey:    ctrlKey,
+				UnitNumber: d.UnitNumber,
+			}
+		}
+	}
+}
+
+func GetExistingControllersAndDevices(
```
aruneshpa, line 269: vvNit: Add doc comment.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+	ctrlType vmopv1.VirtualControllerType,
+	busNumber int32,
+	deviceKey int32,
+	sharingMode vmopv1.VirtualControllerSharingMode,
+	scsiType vmopv1.SCSIControllerType,
+	devices []vmopv1.VirtualDeviceStatus,
+	ctrlInfoMap map[CtrlKey]CtrlInfo,
+	vmDiskUUIDToDeviceInfoMap map[string]DeviceInfo) {
+
+	ctrlKey := CtrlKey{CtrlType: ctrlType, BusNumber: busNumber}
```
aruneshpa, line 239: How do we differentiate between a bus number not being specified, and a bus number of 0?
Is HardwareStatus.NVMEControllers.busNumber always guaranteed to be populated?  If not, we need to check that in calling site.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+				DiskMode:      diskMode,
+				SharingMode:   sharingMode,
+				ControllerKey: ctrlKey,
+				UnitNumber:    unitNumber,
+			},
+		})
+	}
+
+	actual := sets.New(batchSpecVolumes...)
+	if !actual.Equal(expected) {
```
lubronzhan: To check duplicate elements, maybe compare the length `batchSpecVolumes` and `expected`?

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	if ctrlType == vmopv1.VirtualControllerTypeSCSI {
+		ctrlInfo.SCCIType = scsiType
+	}
+
+	ctrlInfoMap[ctrlKey] = ctrlInfo
+
+	for _, d := range devices {
+		if d.DiskUUID != "" {
+			vmDiskUUIDToDeviceInfoMap[d.DiskUUID] = DeviceInfo{
```
aruneshpa, line 255: Can you add a comment somewhere that the devices passed into this method are the ones that are connected to this controller already?

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -427,9 +435,16 @@ func (r *Reconciler) buildVolumeSpecs(
...
-			controllerKey := pkgutil.BuildControllerKey(pvcSpec.ControllerType, pvcSpec.ControllerBusNumber)
-			cnsVolumeSpec.PersistentVolumeClaim.ControllerKey = controllerKey
+		if pvcSpec.ControllerType != "" && pvcSpec.ControllerBusNumber != nil {
+			ctrlInfoMap, _ := vmopv1util.GetExistingControllersAndDevices(
+				hwStatus)
+			if ctrlInfo, exist := ctrlInfoMap[vmopv1util.CtrlKey{
+				CtrlType:  pvcSpec.ControllerType,
+				BusNumber: *pvcSpec.ControllerBusNumber,
+			}]; exist {
+				cnsVolumeSpec.PersistentVolumeClaim.ControllerKey =
```
aruneshpa, line 443: Add a TODO to come back and modify this once the CNSNodeVMBatchAttachment CRD changes are merged.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -427,9 +435,16 @@ func (r *Reconciler) buildVolumeSpecs(
 		}
 
 		// Map controller key (combination of controller type and bus number)
-		if pvcSpec.ControllerType != "" || pvcSpec.ControllerBusNumber != nil {
-			controllerKey := pkgutil.BuildControllerKey(pvcSpec.ControllerType, pvcSpec.ControllerBusNumber)
-			cnsVolumeSpec.PersistentVolumeClaim.ControllerKey = controllerKey
+		if pvcSpec.ControllerType != "" && pvcSpec.ControllerBusNumber != nil {
+			ctrlInfoMap, _ := vmopv1util.GetExistingControllersAndDevices(
```
aruneshpa: Hmm.  It sounds like we are doing this for _each_ volume?  I think we can construct this map once before we start processing volumes?

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+			c.SharingMode,
+			c.Type,
+			c.Devices,
+			ctrlInfoMap,
+			vmDiskUUIDToDeviceInfoMap)
+	}
+	return ctrlInfoMap, vmDiskUUIDToDeviceInfoMap
+}
+
+// GetExistingInfo constructs controller and device information from the current
```
aruneshpa: Can you change the name of this method to be more specific?  "GetInfo" seems vague.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+	// Sort controllers by BusNumber
+	slices.SortFunc(controllers, func(a, b T) int {
+		return int(a.GetBusNumber() - b.GetBusNumber())
+	})
+
+	return controllers
+}
+
+type VolumeKey struct {
+	VolumeName string
```
aruneshpa, line 177: VolumeName is always unique within a VM.  Is there a point of using PVC name in the key of the map?  I totally understand if you want to store this in the value but having it in key can lookup performance impact?

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+	for _, s := range volStatus {
+		p := s.PersistentVolumeClaim
+		if s.Name != "" && p.ClaimName != "" && p.Attached && p.Diskuuid != "" {
+			m[VolumeKey{VolumeName: s.Name, PVCName: p.ClaimName}] = p.Diskuuid
+		}
+	}
+	return m
+}
+
+// getExistingUnmanagedDiskUUIDSet returns a set with all the underlay disk
```
aruneshpa: "underlay" -> "underlying"?

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+	"errors"
+	"fmt"
+	"path"
+	"slices"
+	"strconv"
+	"strings"
+
+	"github.com/vmware/govmomi/object"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	"k8s.io/apimachinery/pkg/util/sets"
```
aruneshpa, line 16: So, this is apimachinery, so I am not too bothered.  But typically, we just use `map[T]struct{}` whenever we need to use Set data structure.  If we can be consistent, that will be great.  This doesn't have to block the review though.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+	unmanagedDiskUUIDSet := getExistingUnmanagedDiskUUIDSet(
+		volStatus)
+	ctrlInfoMap, vmDiskUUIDToDeviceInfoMap := GetExistingControllersAndDevices(
+		hwStatus)
+
+	managedDeviceInfoMap := make(map[VolumeKey]DeviceInfo)
+	for volumeKey, diskUUID := range cnsVolumeToDiskUUIDMap {
+		deviceInfo, ok := vmDiskUUIDToDeviceInfoMap[diskUUID]
+		if !ok {
+			return nil, nil, nil, errors.New(ErrHardwareOutOfSync)
```
aruneshpa, line 362: Add a comment here explaining what just happened?  Something like:
"One of the volumes is attached to a device that is not reported in VM's hardware status.  This can happen if the VM's status reporting is paused, or hasn't occurred yet."

Do more word massaging if needed.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -713,18 +728,46 @@ func (r *Reconciler) validateHardwareVersion(ctx *pkgctx.VolumeContext) error {
...
+	ctrlInfoMap, managedDeviceInfoMap, unmanagedDiskDeviceInfoSet, err :=
+		vmopv1util.GetExistingInfo(
+			existingAttachmentStatus,
+			volCtx.VM.Status.Volumes,
+			volCtx.VM.Status.Hardware)
+	if err != nil {
+		return err
+	}
+
+	// This ensures vm spec volumes to batch volumes is done properly;
```
aruneshpa: This comment is confusing.  We only have the vm.spec.volumes at this point since this is called both during create or update CNSNodeVMBatchAttachment request.  And VM spec.volumes is always the source of truth that gets translated to batch, so this is implicitly true? Trying to make sure if I am missing something.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+			unmanagedDiskDeviceInfoSet.Insert(deviceInfo)
+		}
+	}
+
+	return ctrlInfoMap, managedDeviceInfoMap, unmanagedDiskDeviceInfoSet, nil
+}
+
+// ValidateVMSpecVolumesToBatchSpecVolumes ensures VM spec volumes are properly
+// converted to batch spec volumes with correct disk modes and sharing modes.
+func ValidateVMSpecVolumesToBatchSpecVolumes(
```
aruneshpa: Maybe I am missing something, but I think this method is not needed.

We are basically testing that the CnsNodeVMBatchAttachment volume spec that we generated is valid.  That is not the goal of the validation.  We want to validate that the spec specified in the vm.spec.volumes is valid (e.g., invalid sharing mode, invalid unit number etc) -- which you already do as part of another method.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+			pvc       = vol.PersistentVolumeClaim
+			ctrlInfo  CtrlInfo
+			ctrlExist bool
+		)
+
+		// skip the volume validation if the applicationType or the controller
+		// is not specifically defined
+		if pvc.ControllerType == "" ||
+			pvc.ControllerBusNumber == nil ||
+			pvc.ApplicationType == "" {
```
aruneshpa: We should validate the correct sharingmodes if the applicationTypes are defined even if controller and bus numbers are skipped.
I agree that mutation and validation webhooks will handle defaulting and rejection.  But you should still check that here.

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,609 @@
...
+				VolumeName: vol.Name,
+				PVCName:    pvc.ClaimName,
+			}]
+			if exist {
+				devInfo = &dv
+			}
+		}
+
+		if devInfo != nil {
+			if expectedSet.Has(*devInfo) {
```
aruneshpa, line 539: vm.spec.volumes is a `listMap` on a volume name.  I can't think of a case where there would be two `VolumeKey` (volume name + PVC name) in the managedDeviceInfoMap?

pkg/util/vmopv1/volume.go
lubronzhan: I guess there is a case that PVC is changed? There could be a moment that attachment spec/ status and volume spec doesn't match

pkg/util/vmopv1/volume.go
zhangheliu: When a controller is attached to a VM, it should always have a bus number associated with this controller. It is a required field for the controller status, so there is no need to differentiate nil values.

pkg/util/vmopv1/volume.go
zhangheliu: When a device is attached to a VM, it should always have a unit number associated with this device. It is a required field for the device status, so there is no need to differentiate nil values.

pkg/util/vmopv1/volume.go
zhangheliu: Right, when populating hardware status from the VC VM, the bus number is always guaranteed to be populated.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
zhangheliu: added a story to keep track of this

pkg/util/vmopv1/volume.go
zhangheliu: Right, this is for when a PVC is changed for a volume while the existing CNS batch attachment status might still have the old data.

controllers/virtualmachine/volumebatch/volumebatch_controller.go
zhangheliu: No, you are not missing anything. Since only the spec.volumes will be validated, I wanted to double-check check batch volume CR is generated properly as well but yes, it can be removed.

pkg/util/vmopv1/volume.go
zhangheliu: right, same as above. I will remove this.

pkg/util/vmopv1/volume.go
zhangheliu: good catch, will update

pkg/util/vmopv1/volume.go
zhangheliu: This is to handle the case when vm.spec.volumes is updated, such as when a volume's PVC is updated, while CNS batch volume status might not have been updated. 
similar to https://github.com/vmware-tanzu/vm-operator/pull/1203#discussion_r2409087047

pkg/util/vmopv1/volume.go
zhangheliu: right, those are per our PersistentVolumeClaimVolumeSource ApplicationType definition

pkg/util/vmopv1/volume.go
zhangheliu: Same as above

pkg/util/vmopv1/volume.go
zhangheliu: set.equal makes sure all the feilds are equal as well. but per the conversation above. I think I am going to remove this function

pkg/util/vmopv1/volume.go
zhangheliu: We have started using Sets in other controllers as well (although for more complex set operations). I am going to leave it as it is right now since the overhead is negligible (memory and runtime) and the dependencies are already included. It might also be useful if we need to start using other more complex set operations as well in this helper function.

pkg/util/vmopv1/volume.go
lubronzhan: I was talking about between `batchSpecVolumes ` and `expected `, not `actual ` and `expected`. But yeah it's unrelevant now

pkg/util/vmopv1/volume.go
zhangheliu: updated in the function comment

controllers/virtualmachine/volumebatch/volumebatch_controller.go
zhangheliu: good catch. updated

pkg/util/vmopv1/volume.go
```diff
@@ -0,0 +1,547 @@
...
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
+	cnsv1alpha1 "github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/api/v1alpha1"
+)
+
+// These constants define the maximum number of devices that can be attached
+// to each controller type. Unit number 7 is reserved for SCSI controllers
+// as it's used internally by the SCSI controller and cannot be assigned
+// to user devices.
+const (
+	IDEControllerMaxUnitNumber   = 1
```
lubronzhan, line 27: If these constants(including error messages) are not used outside the package, they can just be defined as non export type.

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -149,18 +149,133 @@ type VirtualDeviceStatus struct {
 
 	// UnitNumber describes the observed unit number of the device.
 	UnitNumber int32 `json:"unitNumber"`
+
+	// +optional
+
+	// DiskUUID represents the underlying virtual disk UUID and is present when
+	// attachment succeeds.
+	DiskUUID string `json:"diskUUID,omitempty"`
```
akutz, line 157: I am considering changing the status field to just `UUID`, so why not do that here as well?

api/v1alpha5/virtualmachine_hardware_types.go
akutz: I considered both options when designing the API and was very careful about how I did it. I broke it down just enough IMO to keep it logical. It also matches what they did in VAPI as well as what different generators recommended when I asked options for representing ConfigSpec as K8s APIs.

That said, when I designed the _status_ for controllers, I kept things simplified as we did not need to surface all the same details as spec. 

I am torn on this. I feel we are only doing a lot of this as a response to our not implementing the VolumeV2 logic in the VM provider workflow where I had intended it to live...

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -149,18 +149,133 @@ type VirtualDeviceStatus struct {
...
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +listMapKey=type
+
+	// Devices describes the observed devices connected to the controller.
+	Devices []VirtualDeviceStatus `json:"devices,omitempty"`
+}
+
+// GetDevices returns the devices connected to the controller.
```
akutz, line 181: `GetDevices returns the observed devices connected to the controller.`

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -149,18 +149,133 @@ type VirtualDeviceStatus struct {
...
+func (c IDEControllerStatus) GetDevices() []VirtualDeviceStatus {
+	return c.Devices
+}
+
+// GetBusNumber returns the bus number of the controller.
```
akutz, line 186: I will not keep repeating this, but please add `observed` in the descriptions related to status.

api/v1alpha5/virtualmachine_hardware_types.go
akutz: Also, why do we have the helper functions on this type? Why not just the field access directly?

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -149,18 +149,133 @@ type VirtualDeviceStatus struct {
...
+
+	// +required
+
+	// DeviceKey describes the observed device key of the controller.
+	DeviceKey int32 `json:"deviceKey"`
+
+	// +optional
+
+	// SharingMode describes the observed sharing mode for the controller.
+	SharingMode VirtualControllerSharingMode `json:"sharingMode,omitempty"`
```
akutz, line 206: It should default to `None`.

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -149,18 +149,133 @@ type VirtualDeviceStatus struct {
...
```
akutz, line 278: It should default to `None`.

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -389,11 +514,34 @@ type VirtualMachineHardwareStatus struct {
...
-	// Controllers describes the observed list of virtual controllers for the
-	// VM.
-	Controllers []VirtualControllerStatus `json:"controllers,omitempty"`
+	// IDEControllers describes the observed list of IDE controllers for the VM.
+	IDEControllers []IDEControllerStatus `json:"ideControllers,omitempty"`
+
+	// +optional
+	// +listType=map
+	// +listMapKey=busNumber
+	// +kubebuilder:validation:MaxItems=4
```
akutz, line 525: The spec should have a validation for max elements, but if for some reason a VM does have more than this, we should report them. Remove this from status.

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -389,11 +514,34 @@ type VirtualMachineHardwareStatus struct {
...
+
+	// NVMEControllers describes the observed list of NVMe controllers for the VM.
+	NVMEControllers []NVMEControllerStatus `json:"nvmeControllers,omitempty"`
+
+	// +optional
+	// +listType=map
+	// +listMapKey=busNumber
+	// +kubebuilder:validation:MaxItems=4
```
akutz, line 533: The spec should have a validation for max elements, but if for some reason a VM does have more than this, we should report them. Remove this from status.

api/v1alpha5/virtualmachine_hardware_types.go
```diff
@@ -389,11 +514,34 @@ type VirtualMachineHardwareStatus struct {
...
+
+	// SATAControllers describes the observed list of SATA controllers for the VM.
+	SATAControllers []SATAControllerStatus `json:"sataControllers,omitempty"`
+
+	// +optional
+	// +listType=map
+	// +listMapKey=busNumber
+	// +kubebuilder:validation:MaxItems=4
```
akutz, line 541: The spec should have a validation for max elements, but if for some reason a VM does have more than this, we should report them. Remove this from status.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1653,106 +1653,133 @@ func reconcileStatusController(
...
-		// Map controllers to device key and to their busNumber+type.
-		controllerKeyMap = make(map[int32]*vmopv1.VirtualControllerStatus)
-		controllerMap    = make(map[string]*vmopv1.VirtualControllerStatus)
-	)
+	// collect all the controllers with deviceKey to *ControllerStatus
+	ideControllers := make(map[int32]*vmopv1.IDEControllerStatus)
+	nvmeControllers := make(map[int32]*vmopv1.NVMEControllerStatus)
+	sataControllers := make(map[int32]*vmopv1.SATAControllerStatus)
+	scsiControllers := make(map[int32]*vmopv1.SCSIControllerStatus)
+	// create a deviceKey to controllerType map for adding device later
```
akutz, line 1661: Blank line between the previous block and this one.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1653,106 +1653,133 @@ func reconcileStatusController(
 		vm.Status.Hardware = &vmopv1.VirtualMachineHardwareStatus{}
 	}
 
-	var (
-		// Map controllers to device key and to their busNumber+type.
-		controllerKeyMap = make(map[int32]*vmopv1.VirtualControllerStatus)
-		controllerMap    = make(map[string]*vmopv1.VirtualControllerStatus)
-	)
+	// collect all the controllers with deviceKey to *ControllerStatus
```
akutz, line 1656: Comments should be actual sentences that follow grammar rules. Ex.

```go
// Map controller status objects by their device key.
```

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1653,106 +1653,133 @@ func reconcileStatusController(
...
+	ideControllers := make(map[int32]*vmopv1.IDEControllerStatus)
+	nvmeControllers := make(map[int32]*vmopv1.NVMEControllerStatus)
+	sataControllers := make(map[int32]*vmopv1.SATAControllerStatus)
+	scsiControllers := make(map[int32]*vmopv1.SCSIControllerStatus)
+	// create a deviceKey to controllerType map for adding device later
+	controllerTypes := make(map[int32]vmopv1.VirtualControllerType)
```
akutz, line 1662: I am unclear on how this variable is still being used...

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1653,106 +1653,133 @@ func reconcileStatusController(
 		vm.Status.Hardware = &vmopv1.VirtualMachineHardwareStatus{}
 	}
 
-	var (
-		// Map controllers to device key and to their busNumber+type.
-		controllerKeyMap = make(map[int32]*vmopv1.VirtualControllerStatus)
-		controllerMap    = make(map[string]*vmopv1.VirtualControllerStatus)
-	)
+	// collect all the controllers with deviceKey to *ControllerStatus
+	ideControllers := make(map[int32]*vmopv1.IDEControllerStatus)
```
akutz, line 1657: Also, please continue to use the `var` block as it organizes things better:

```go
var (
	ideControllers  = map[int32]*vmopv1.IDEControllerStatus{}
	nvmeControllers = map[int32]*vmopv1.IDEControllerStatus{}
	sataControllers = map[int32]*vmopv1.IDEControllerStatus{}
	scsiControllers = map[int32]*vmopv1.IDEControllerStatus{}
)
```

I also prefer the use of direct init unless there is a known length. That way you obviate the need for the `make` keyword and keep things more succinct.

akutz (changes requested):

pkg/util/vmopv1/volume.go
zhangheliu: hmm, I am not sure either. we could start a thread asking about this.

PR #1205, opened 2025-09-25
🌱 Allow VM Spec.Bootstrap to be mutable

bryanv:

Previously we would restrict we restrict the bootstrap to only be changed with the VM was "off", but there isn't really a good reason to do so. The customization will be applied on the next powering on transition. Instead, just restrict the bootstrap provider type from being changed.

Remove a bogus "network can't change while powered on" test: this was just testing network mutability or not since we had no network related checks for just powered on.

Fixes #

```release-note
Allow the VM Spec.Bootstrap to be updated, but the provider type cannot be changed
```

### DIFF
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -93,6 +93,7 @@ const (
 	invalidClassInstanceReferenceOwnerMismatch = "VirtualMachineClassInstance must be an instance of the VM Class specified by spec.class"
 	labelSelectorCanNotContainVMOperatorLabels = "label selector can not contain VM Operator managed labels (vmoperator.vmware.com)"
 	guestCustomizationVCDParityNotEnabled      = "VC guest customization VCD parity capability is not enabled"
+	bootstrapProviderTypeCannotBeChanged       = "bootstrap provider type cannot be changed"
 )
 
 // +kubebuilder:webhook:verbs=create;update,path=/default-validate-vmoperator-vmware-com-v1alpha5-virtualmachine,mutating=false,failurePolicy=fail,groups=vmoperator.vmware.com,resources=virtualmachines,versions=v1alpha5,name=default.validating.virtualmachine.v1alpha5.vmoperator.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
@@ -224,7 +225,6 @@ func (v validator) validateImageOnUpdate(ctx *pkgctx.WebhookRequestContext, vm, 
 //   - Minimum VM Hardware Version
 //
 // Following fields can only be changed when the VM is powered off.
-//   - Bootstrap
 //   - GuestID
 //   - CD-ROM (updating connection state is allowed regardless of power state)
 func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.Response {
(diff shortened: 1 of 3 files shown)


bryanv:
> LGTM! Time to remove the `Immutable Network Configuration` note from https://github.com/vmware-tanzu/vm-operator/blob/main/docs/concepts/workloads/vm.md#networking now?

That isn't enabled yet so I don't think so.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -284,6 +286,38 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
 }
 
+func (v validator) validateBootstrapProviderImmutable(
+	_ *pkgctx.WebhookRequestContext,
+	vm, oldVM *vmopv1.VirtualMachine) field.ErrorList {
+
+	var fieldErrs field.ErrorList
+	p := field.NewPath("spec", "bootstrap")
+
+	bs := vm.Spec.Bootstrap
```
aruneshpa: The choice of variable name is hilarious 😆

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -284,6 +286,38 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
+	oldBS := oldVM.Spec.Bootstrap
+
+	if oldBS == nil {
+		return fieldErrs
+	}
+
+	if bs == nil {
+		return append(fieldErrs, field.Forbidden(p, bootstrapProviderTypeCannotBeChanged))
```
aruneshpa, line 307: Hmm.  I am wondering why is this an issue?  If we support changing the payload for a particular bootstrap type, a user should be able to unset / remove it?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -284,6 +286,38 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
+	}
+
+	// vApp can be combined with LinuxPrep and Sysprep, or used standalone, so don't
```
aruneshpa, line 310: Hmm.  I understand the reasoning somewhat but it does feel inconsistent.  vAppConfig _can_ be used as a standalone bootstrap mechanism.  So, we should block changing that?

Even for vAppConfig + LinuxPrep / SysPrep, I think it makes sense to block if the oldVM had both and the new one is omitting the vAppConfig.  Let me know what you think?

aruneshpa (commented):
Nit: in the commit message you have: "Previously we would restrict we restrict the bootstrap".

webhooks/virtualmachine/validation/virtualmachine_validator.go
bryanv: So changing vApp is already funky in the sense that like if you change the properties in the spec, since we don't necessarily know what was previously set, we can't go clear those potentially old values out. So if you would either clear the properties or just unset vApp, there isn't much of a meaningful difference.

aruneshpa: Alright, this makes sense.

webhooks/virtualmachine/validation/virtualmachine_validator.go
bryanv: Yah I see your point. We don't go clear an existing customization - like pending GOSC or CloudInit EC fields - so I worry that could be misleading. Let me merge this as is and think about it over the next few days.

PR #1206, opened 2025-09-25
✨ All disks are PVCs (APIs & skeleton)

akutz:

This commit represents the foundational work for the "All Disks Are PVCs" feature, which aims to convert all unmanaged virtual machine disks into Kubernetes Persistent Volume Claims (PVCs). Additionally, this commit provides the foundation for realizing _new_ VMs with PVCs created from the disks in VM images. This is a significant architectural change that will provide better integration with Kubernetes storage management and enable more sophisticated storage operations.

### API Changes
- **New Storage Types**: Added comprehensive storage-related types in `api/v1alpha5/virtualmachine_storage_types.go` including:
  - `VirtualMachineVolume` and `VirtualMachineVolumeSource` for volume specifications
  - `PersistentVolumeClaimVolumeSource` with support for unmanaged and instance volume claims
  - `UnmanagedVolumeClaimVolumeSource` to handle volumes from images or existing VMs
  - New enums for volume types, provisioning modes, disk modes, and sharing modes
  - Volume status tracking structures

### Virtual Machine Image Disk Information Updates
- **Disk Structure Refactoring**: Updated `VirtualMachineImageDiskInfo` structure:
  - Replaced `Capacity` and `Size` fields with `Limit` and `Requested` 
  - Added `Name` field to uniquely identify disks
  - Updated content library utilities to populate disk names from various backing types
  - Modified conversion functions between API versions to handle the new structure

### CNS Register Volume Integration
- **External API Addition**: Added `CnsRegisterVolume` types from vsphere-csi-driver for managing the registration of existing volumes as PVCs
- **Go Module Updates**: Updated vsphere-csi-driver module dependencies

### Feature Flag Support
- **Capability Management**: Added `CapabilityKeyAllDisksArePVCs` capability key to enable/disable the feature
- **Configuration Integration**: Added `AllDisksArePVCs` feature flag to the configuration system

### Controller Infrastructure
- **VM Controller Updates**: Added new reconciliation step for unmanaged to managed disk conversion
- **Storage Policy Updates**: Updated storage policy usage controller to handle new disk information structure
- **Webhook Updates**: Modified unified storage quota validation to use new disk limit fields

### Test Infrastructure
- **Conversion Testing**: Added fuzzer functions for virtual machine image conversions
- **Unit Test Updates**: Updated numerous unit tests to use new disk structure fields
- **Integration Test Updates**: Modified integration tests for storage quota validation

## Implementation Strategy

The implementation follows a phased approach:

1. **API Foundation** (This Commit): Establish the necessary API structures and types
2. **Migration Logic** (Future): Implement the actual conversion of unmanaged disks to PVCs
3. **PVC Management** (Future): Handle PVC lifecycle and data source references
4. **CNS Integration** (Future): Integrate with CNS for volume registration

## Backward Compatibility

The changes maintain backward compatibility through:
- Conversion functions between API versions
- Gradual migration approach controlled by feature flags
- Preservation of existing disk information in new format

Fixes `NA`

```release-note
Introduce APIs for PVC DataSourceRefs and All Disks are PVCs feature.
```


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1053,6 +1067,97 @@ func (vs *vSphereVMProvider) getTags(
...
+	//    PVC was created by VM Operator or by an external actor, such as
+	//    a DevOps user or Blueprint service. In the case of the latter,
+	//    they may pre-create the PVC object with the dataSourceRef field
+	//    pointing to this VM in order to ensure the unmanaged volume is
+	//    migrated to a new storage class and/or resized when being
+	//    converted to a managed volume (PVC). Therefore, we only want to
+	//    set those fields if VM Op was the one to create the PVC in the
+	//    first place.
+	//
+	// 4. If the PVC object's status.phase field is empty, then use the
```
xing-yang, line 1126: The PVC will be created before CnsRegisterVolume is created, right?  Otherwise, CnsRegisterVolume controller will create a PVC itself.

akutz: Yes, that is the order. The TODOs should highlight that above. Either VM Op creates the PVC or it finds the existing one.

api/v1alpha5/virtualmachine_storage_types.go
```diff
@@ -210,6 +216,33 @@ type PersistentVolumeClaimVolumeSource struct {
...
+
+type UnmanagedVolumeClaimVolumeType string
+
+const (
+	UnmanagedVolumeClaimVolumeTypeFromImage = "FromImage"
+	UnmanagedVolumeClaimVolumeTypeFromVM    = "FromVM"
+)
+
+type UnmanagedVolumeClaimVolumeSource struct {
+	// +required
```
aruneshpa, line 229: Historically, I think we have omitted the `required` tag, and only relied on the absence of `optional` to imply required.

api/v1alpha5/virtualmachine_storage_types.go
```diff
@@ -210,6 +216,33 @@ type PersistentVolumeClaimVolumeSource struct {
...
+
+	// +required
+
+	// Name describes the name of the unmanaged volume.
+	//
+	// For volumes from an image, the name is from the image's
+	// status.disks[].name field.
+	//
+	// For volumes from the VM, the name is from the VM's
+	// status.volumes[].name field.
```
aruneshpa, line 245: This sounds cyclical?  The `unmanagedVolumeClaim` is under spec.volumes, and it points a volume in spec.volumes?

api/v1alpha5/virtualmachineimage_types.go
```diff
@@ -92,15 +92,21 @@ type VirtualMachineImageProductInfo struct {
...
-	Capacity *resource.Quantity `json:"capacity,omitempty"`
+	// Limit is the total virtual disk capacity.
+	Limit *resource.Quantity `json:"limit,omitempty"`
 
 	// +optional
 
-	// Size is the estimated populated size of the virtual disk in bytes.
-	Size *resource.Quantity `json:"size,omitempty"`
+	// Requested is the minimum storage requirements for the virtual disk.
+	Requested *resource.Quantity `json:"requested,omitempty"`
```
aruneshpa, line 109: I am probably forgetting the discussion, but `Requested` will never be not equal to capacity because we reserve the entire capacity for each disk.  Is this correct?  If so, maybe we should also clarify this in the API comments?  Maybe even consider deduplicating them to a single field.

pkg/providers/vsphere/contentlibrary/content_library_utils.go
```diff
@@ -151,24 +152,49 @@ func UpdateVmiWithVirtualMachine(
...
 				capacity = kubeutil.BytesToResource(disk.CapacityInBytes)
+				if di := disk.DeviceInfo; di != nil {
+					if d := di.GetDescription(); d != nil {
+						name = d.Label
+					}
+				}
+			}
+
+			if name == "" {
+				name = strconv.Itoa(len(status.Disks))
```
aruneshpa: Do you want to add a prefix like "disks-" to make this more readable?

pkg/providers/vsphere/contentlibrary/content_library_utils.go
```diff
@@ -277,18 +304,26 @@ func populateImageStatusFromOVFDiskSection(imageStatus *vmopv1.VirtualMachineIma
 	imageStatus.Disks = make([]vmopv1.VirtualMachineImageDiskInfo, len(diskSection.Disks))
 
 	for i, disk := range diskSection.Disks {
-		diskDetail := vmopv1.VirtualMachineImageDiskInfo{}
+		name := disk.DiskID
+		if name == "" {
+			name = strconv.Itoa(i)
```
aruneshpa: Same comment here. 
Do you want to add a prefix like "disks-" to make this more readable?

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: @xing-yang , as of now, `CnsRegisterVolume` uses a [hardcoded](https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/pkg/syncer/cnsoperator/controller/cnsregistervolume/cnsregistervolume_controller.go#L315C2-L315C40) logic to get to the PV name from the CNS volume ID.  I am guessing that code will be modified to register this virtual disk as a CNS volume -- get the volume ID and then use some combination of the PVC name to generate the PV name?  Trying to make sure this workflow is idempotent.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1053,6 +1067,97 @@ func (vs *vSphereVMProvider) getTags(
...
+	//    controllerutil.CreateOrPatch function to ensure a
+	//    CnsRegisterVolume object exists for this disk.
+	//
+	//    See https://docs.google.com/document/d/1L6hGuRMY2Caci5OZyZMxp84Bpquw-WCX2ZQQ_2d5mlE/edit?usp=sharing
```
aruneshpa, line 1130: That link is internal, so isn't really helpful here unfortunately.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1053,6 +1067,97 @@ func (vs *vSphereVMProvider) getTags(
...
+	//    the PVC created/patched in the previous step.
+	//
+	//    The CnsRegisterVolume object should have an OwnerRef set to
+	//    this VM.
+	//
+	//    The CnsRegisterVolume object should have a label, ex.
+	//    vmoperator.vmware.com/created-by, with a value that is this
+	//    VM's name.
+	//
+	// 5. If the CnsRegisterVolume status indicates the operation is
```
aruneshpa, line 1143: @xing-yang , can we please have a `TtlAfterFinished` style garbage collection for CnsRegisterVolume resources?  Once this API is successful, it is not useful for consumers.  So, it will be a good value add.  RegisterVM API also has to handle the cleanup, so it will benefit there as well.

pkg/providers/vsphere/contentlibrary/content_library_utils.go
palnabarun: We should also make the logic an utility function.

api/v1alpha5/virtualmachine_storage_types.go
akutz: Yes, but the K8s API conventions have changed that. On new API changes I've started using `required` due to the new requirements. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required for more info.

api/v1alpha5/virtualmachine_storage_types.go
akutz: The disks that show up under `status.volumes` do not have to originate from `spec.volumes`. That is why this is written this way. I will add a note to the API docs that an `unmanagedVolumeClaim` should not be used to reference a `Managed` volume (i.e. a volume already in `spec.volumes`).

api/v1alpha5/virtualmachineimage_types.go
akutz: We do _not_ reserve the entire capacity. We report `requested` as the total usage, but that does not mean that is what the disk is actually using.

pkg/providers/vsphere/contentlibrary/content_library_utils.go
akutz: > We should also make the logic an utility function.

I disagree on the utility function as it is too much of a unicorn.

akutz: > Do you want to add a prefix like "disks-" to make this more readable?

Maybe. Not for readability, but perhaps because it prevents a type error when someone writes `1` without `"1"` quotes and `kubectl` gives an error.

pkg/providers/vsphere/vmprovider_vm.go
akutz: These TODO comments are not permanent. They are there to assist in the implementation.

PR #1208, opened 2025-09-29
🌱 Add FIPS-only mode enforcement for Linux builds and enable CGO

dilyar85:

This PR adds a `fips_only_linux.go` file with `import _ "crypto/tls/fipsonly"` to enforce FIPS-only mode for Linux builds. It also enables `CGO` and adds `GOEXPERIMENT= boringcrypto` to use BoringCrypto for Go FIPS compliance on Linux.

These changes are required to build VMOP with FIPS compliant.

Fixes N/A.

Built an internal cayman_vmop from this change and verified the binary is using BoringCrypto:

```console
$ go version manager
manager.1: go1.24.7 X:boringcrypto

$ strings manager | grep boringcrypto_FIPS_mode
_cgo_39a3e70c2c46_Cfunc__goboringcrypto_FIPS_mode
_goboringcrypto_FIPS_mode
crypto/internal/boring._Cfunc__goboringcrypto_FIPS_mode
```

```release-note
Add FIPS-only mode enforcement for Linux builds and enable CGO.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,9 @@ export BUILD_BRANCH
 export BUILD_COMMIT
 export BUILD_VERSION
 
-CGO_ENABLED ?= 0
+# Enable CGO to use BoringCrypto for Go FIPS compliance on Linux.
+CGO_ENABLED ?= 1
+GOEXPERIMENT ?= boringcrypto
 
 BUILDINFO_LDFLAGS = "\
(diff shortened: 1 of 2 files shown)


akutz (changes requested):
This is an internal requirement and should be implemented internally. Please do so in our internal repo where we introduce things like this.

akutz:
I am going to close this change and ask you to move it to our internal repo.

PR #1209, opened 2025-09-29
🌱 Support external VM mutation webhook funcs

akutz:

This patch updates the VM mutation webhook to allow external actors to register callback functions for create and update admission requests. These functions will be invoked as part of the standard VM mutation webhook for create/update.

While anyone can register a webhook for an object in K8s, this pattern allows the VM Service team to easily extend the current framework via an internal component.

Fixes `NA`

```release-note
Support external VM mutation webhooks.
```


dilyar85 (approved):
LGTM!

PR #1210, opened 2025-09-29
Handle publishing of encrypted VMs for inventory-backed ContentLibraries

abaruni:

This PR adds logic to the VirtualMachinePublishRequest controller to handle encrypted VMs when publishing to an inventory-back ContentLibrary. There are three pieces:

- Encrypted VMs cannot be published to content library whose storage class does not support encryption
- Publish requests whose source VMs have a mix of encrypted and unencrypted disks will be rejected
- For encrypted VMs which are allowed, we set the storage policy of the target content library on each of the disks in the configSpec that is passed into the cloneSpec

Fixes #

```release-note
Handle publishing of encrypted Virtual Machines for inventory-backed Content Libraries.
```

### DIFF
--- a/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
+++ b/controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
@@ -10,10 +10,12 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"slices"
 	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -44,6 +46,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
 	pkgutil "github.com/vmware-tanzu/vm-operator/pkg/util"
+	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/providers/vsphere/virtualmachine/publish.go
```diff
@@ -118,10 +123,33 @@ func CloneVM(
...
+					},
+				})
+			}
+
+			cloneSpec.Config.DeviceChange = deviceChanges
+		}
 	}
 
 	vmCtx.Logger.Info("Publishing VM as template",
-		"actId", actID,
```
lubronzhan, line 124: Is this intended?

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
```diff
@@ -568,6 +601,52 @@ func (r *Reconciler) checkContentLibraryQuota(ctx *pkgctx.VirtualMachinePublishR
...
+	}
+
+	isEncryptedProfile, err := r.VMProvider.DoesProfileSupportEncryption(ctx, storagePolicyID)
+	if err != nil {
+		return err
+	}
+
+	if !isEncryptedProfile {
+		err := pkgerr.NoRequeueError{
+			Message: "source VM is encrypted, but target storage class does not support encryption",
```
lubronzhan, line 637: The function implicates the source vm is encrypted, do you think is it better to move the `if ctx.VM.Status.Crypto != nil` check in this func as well?

pkg/providers/vsphere/virtualmachine/publish.go
abaruni: yes, we cannot specify an activation ID on clone. this was copy/pasta artifact from a while back

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller.go
abaruni: no. i don't really think it makes sense for that check to be in this function. this function is only called if the source vm is encrypted. there isn't a need to call the function in order to check if the vm is encrypted, when the function really only exists to validate that the encryption of the vm is valid with respect to the storage class specified in the content library.

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
```diff
@@ -413,6 +415,124 @@ func unitTestsReconcile() {
...
+			})
+
+			When("target content library's storage class doesn't exist", func() {
+				BeforeEach(func() {
+					clv1a2.Spec.StorageClass = builder.DummyStorageClassName
+				})
+
+				It("should return an error", func() {
+					_, err := reconciler.ReconcileNormal(vmpubCtx)
+					Expect(err).To(HaveOccurred())
```
dilyar85, line 511: Nit: check the error contains `failed to get storageClass` (to verify it's actually failed by this reason).

pkg/providers/vsphere/virtualmachine/publish.go
```diff
@@ -83,19 +84,23 @@ func CloneVM(
...
 	vm := object.NewVirtualMachine(
 		vimClient,
 		vimtypes.ManagedObjectReference{
 			Type:  string(vimtypes.ManagedObjectTypeVirtualMachine),
 			Value: vmCtx.VM.Status.UniqueID,
 		})
 
-	var filteredExtraConfig pkgutil.OptionValues
-	var err error
-	if filteredExtraConfig, err = GetFilteredExtraConfigFromObject(
```
dilyar85, line 95: Nit: can probably remove this function now since it's introduced for this call (not being used anywhere else).

dilyar85 (approved):
LGTM with some nit left, thanks @abaruni!

controllers/virtualmachinepublishrequest/virtualmachinepublishrequest_controller_unit_test.go
abaruni: will do

pkg/providers/vsphere/virtualmachine/publish.go
abaruni: good catch. i thought it was being used elsewhere, but it's not

bryanv (approved):

PR #1211, opened 2025-09-29
✨ Only GOSC customizations to be explicitly requested at next power on

bryanv:

Add CustomizeAtNextPowerOn field to LinuxPrep and Sysprep that allows a GOSC customization to be explicitly requested. When this field is set to true, a customization will be done, and after the successful customization this field will be updated to false.

Fixes #

```release-note
Add CustomizeAtNextPowerOn to GOSC LinuxPrep and Sysprep to explicitly request a customization will be done at the next power on.
```


pkg/providers/vsphere/vmlifecycle/bootstrap.go
```diff
@@ -137,57 +140,93 @@ func DoBootstrap(
...
 	}
 
-	return nil
+	if customizeLatch != nil {
+		// Set latch to false so that a later customization on has to be
+		// explicitly requested.
+		vmCtx.Logger.V(4).Info(
+			"Setting customization latch to false after successful customization")
+		*customizeLatch = false
+	}
```
dilyar85: Check `customizeLatch` value is actually true here, so that we don't log `successful customization` when there's no customization at all?

dilyar85 (approved):

PR #1213, opened 2025-09-30
🐛 Fix setting Sysprep password without AutoLogin

bryanv:

Setting the password with Sysprep does not require AutoLogin, but AutoLogon requires the password to be set. The validation webhook was correct but the check to later determine if the password was set was incorrect.

Fixes #

```release-note
Setting the administrator password with inline Sysprep should not require AutoLogin to be true 
```

### DIFF
--- a/pkg/providers/vsphere/sysprep/secret_test.go
+++ b/pkg/providers/vsphere/sysprep/secret_test.go
@@ -121,7 +121,6 @@ var _ = Describe("GetSysprepSecretData", func() {
 		BeforeEach(func() {
 			inlineSysprep = vmopv1sysprep.Sysprep{
 				GUIUnattended: &vmopv1sysprep.GUIUnattended{
-					AutoLogon: true,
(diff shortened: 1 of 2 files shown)


aruneshpa (approved):

pkg/providers/vsphere/sysprep/secret_test.go
```diff
@@ -158,6 +158,17 @@ var _ = Describe("GetSysprepSecretData", func() {
 					Expect(err.Error()).To(Equal(fmt.Sprintf(`no data found for key "%s" for secret default/%s`, anotherKey, pwdSecretName)))
 				})
 			})
+
+			Context("GUIUnattended without Password", func() {
```
aruneshpa, line 161: Do we add a test for `GUIUnattended` without Password failing?

bryanv: `GUIUnattended` without password isn't suppose to fail. The existing test have `GUIUnattended` with password

PR #1214, opened 2025-09-30
🌱 Skip VM with zone label and address recreated VM in group placement

dilyar85:

This PR updates the VM Group controller during group placement reconciliation by: 
- Skipping VM members that already have the `topology.kubernetes.io/zone` label, preventing unnecessary DRS placement (VM controller already ignores the group placement result for these VMs).
- Reassigning placement for VM members whose UID does not match the group member status, ensuring accurate placement status when a VM is recreated with the same name but not removed from the group.

All the required API changes are backported to v1alpha2 to maintain compatibility with VKS (still on v1a2).

Fixes N/A.

- Rewrote the group placement tests to verify all the cases that we skip and do placement by group.
- Deployed this change to an internal testbed and validated the following:

```console
# Create a VM with a zone label and verify it is not placed in the group member status.
members:
  - conditions:
    - lastTransitionTime: "2025-09-29T23:56:18Z"
      message: ""
      reason: "True"
      status: "True"
      type: GroupLinked
    - lastTransitionTime: "2025-09-29T23:56:18Z"
      message: ""
      reason: "True"
      status: "True"
      type: PlacementReady
    kind: VirtualMachine
    name: vm-af-1 --------------------------------> VM with "topology.kubernetes.io/zone" set
    uid: e6245eba-147c-42ae-85be-014c5f3d091b
    
  members:
  - conditions:
    - lastTransitionTime: "2025-09-29T23:56:14Z"
      message: ""
      reason: "True"
      status: "True"
      type: GroupLinked
    - lastTransitionTime: "2025-09-29T23:56:18Z"
      message: ""
      reason: "True"
      status: "True"
      type: PlacementReady
    kind: VirtualMachine
    name: vm-1 --------------------------------> VM without "topology.kubernetes.io/zone" set
    placement:
      name: ""
      pool: resgroup-77
      zoneID: domain-c36
    powerState: PoweredOn
    uid: a3ddb647-798a-4020-beb2-f683f7de0e24

# Check VMOP logs to confirm that VMs with matching UIDs are not being re-placed:
I0929 23:56:55.266221       1 virtualmachinegroup_controller.go:643] "Group already has placement condition ready for VM, skipping" logger="virtualmachinegroup-controller" VirtualMachineGroup="sdiliyaer-test/vmg-root" reconcileID="05c16719-8023-493c-abe3-0295d8173bfd" vmName="vm-af-1" vmUID="e6245eba-147c-42ae-85be-014c5f3d091b"
I0929 23:56:55.266400       1 virtualmachinegroup_controller.go:643] "Group already has placement condition ready for VM, skipping" logger="virtualmachinegroup-controller" VirtualMachineGroup="sdiliyaer-test/vmg-root" reconcileID="05c16719-8023-493c-abe3-0295d8173bfd" vmName="vm-1" vmUID="a3ddb647-798a-4020-beb2-f683f7de0e24"

# Recreate VM with the same name without removing it from the group and verify group placement member status is updated:
 members:
  - conditions:
    - lastTransitionTime: "2025-09-30T01:01:35Z"
      message: ""
      reason: "True"
      status: "True"
      type: GroupLinked
    - lastTransitionTime: "2025-09-30T00:59:07Z"
      message: ""
      reason: "True"
      status: "True"
      type: PlacementReady
    kind: VirtualMachine
    name: vm-af-1
    placement:
      name: ""
      pool: resgroup-77
      zoneID: domain-c36
    uid: fd2aaac9-cd69-4a0b-9822-39125e5a7883
```

```release-note
Skip VM with zone label and address recreated VM in group placement.
```

### DIFF
--- a/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
+++ b/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -406,18 +407,6 @@ func (r *Reconciler) reconcileMember(
 	if member.Kind == vmKind {
 		vm := obj.(*vmopv1.VirtualMachine)
 
-		// If the VM already exists, set the member condition appropriately.
-		if vm.Status.UniqueID != "" {
-			if !conditions.IsTrue(ms, vmopv1.VirtualMachineGroupMemberConditionPlacementReady) {
-				conditions.Set(ms, &metav1.Condition{
-					Type:    vmopv1.VirtualMachineGroupMemberConditionPlacementReady,
-					Status:  metav1.ConditionTrue,
-					Reason:  vmopv1.VirtualMachineGroupMemberAlreadyPlacedReason,
-					Message: "VM already placed",
-				})
-			}
-		}
-
 		vmStatusPowerState := obj.GetPowerState()
 
 		if vmStatusPowerState == "" {
@@ -516,35 +505,33 @@ func (r *Reconciler) reconcilePlacement(
 		return nil
 	}
 
-	groupPlacements, err := r.getPlacementMembers(ctx, ctx.VMGroup)
+	// Initialize groupPatches map to collect all child group patches that need
+	// to update their member placement status.
+	groupPatches := make(map[*vmopv1.VirtualMachineGroup]client.Patch)
+
+	groupPlacements, err := r.getPlacementMembers(ctx, ctx.VMGroup, groupPatches)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get all placement members: %w", err)
 	}
 
 	if len(groupPlacements) == 0 {
 		pkglog.FromContextOrDefault(ctx).V(5).Info("No group members need placement")
-		return nil
+	} else if err := r.VMProvider.PlaceVirtualMachineGroup(ctx, ctx.VMGroup, groupPlacements); err != nil {
+		return fmt.Errorf("failed to place group members: %w", err)
 	}
 
-	groupPatches := make([]client.Patch, len(groupPlacements))
-	for i, placement := range groupPlacements {
-		if ctx.VMGroup == placement.VMGroup {
+	// Patch all child groups regardless of whether they have members that need
+	// placement to ensure their member placement conditions are always updated.
+	var errs []error
+	for group, patch := range groupPatches {
+		if ctx.VMGroup.Name == group.Name {
 			// Let the outer patch helper update the root group.
 			continue
 		}
-		groupPatches[i] = client.MergeFrom(placement.VMGroup.DeepCopy())
-	}
-
-	if err := r.VMProvider.PlaceVirtualMachineGroup(ctx, ctx.VMGroup, groupPlacements); err != nil {
-		return err
-	}
-
-	var errs []error
-	for i, placement := range groupPlacements {
-		if gp := groupPatches[i]; gp != nil {
-			if err := r.Status().Patch(ctx, placement.VMGroup, gp); err != nil {
-				errs = append(errs, err)
-			}
+		if err := r.Status().Patch(ctx, group, patch); err != nil {
+			errs = append(errs, fmt.Errorf(
+				"failed to patch placement status for child group %q: %w",
+				group.Name, err))
 		}
 	}
 
@@ -554,6 +541,7 @@ func (r *Reconciler) reconcilePlacement(
 func (r *Reconciler) getPlacementMembers(
 	ctx context.Context,
 	vmGroup *vmopv1.VirtualMachineGroup,
+	groupPatches map[*vmopv1.VirtualMachineGroup]client.Patch,
 ) ([]providers.VMGroupPlacement, error) {
 
 	var groupPlacements []providers.VMGroupPlacement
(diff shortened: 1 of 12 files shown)


controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -623,6 +641,22 @@ func (r *Reconciler) getVMForPlacement(
...
+			"vmName", vmName,
+			"uniqueID", vm.Status.UniqueID,
+		)
+		return nil, nil
+	}
+
+	// If the VM has an explicit zone label, skip group placement to respect the
+	// zone override.
+	if zoneName := vm.Labels[corev1.LabelTopologyZone]; zoneName != "" {
+		pkglog.FromContextOrDefault(ctx).V(5).Info(
```
aruneshpa: Here and elsewhere: I thought we agreed that level 4 would be the lowest log level that we will use.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -593,28 +581,58 @@ func (r *Reconciler) getPlacementMembers(
...
-	if _, member := findMemberStatus(vmName, vmKind, vmGroup.Status.Members); member != nil {
-		if conditions.IsTrue(member, vmopv1.VirtualMachineGroupMemberConditionPlacementReady) {
-			return nil, nil
-		}
-		if !conditions.IsTrue(member, vmopv1.VirtualMachineGroupMemberConditionGroupLinked) {
-			return nil, fmt.Errorf("VM %q is not linked for group %q", vmName, vmGroup.Name)
-		}
-	} else {
+	_, memberStatus := findMemberStatus(vmName, vmKind, vmGroup.Status.Members)
+	if memberStatus == nil {
```
aruneshpa, line 587: So, the reason for not calling this method and handling the "AlreadyPlaced" in the parent method was because we also need to handle the situation where the VM is part of a group, but the Group status is wiped.  In that case, the VM will have a uniqueID, but since we won't be able to find it in member status, this method will always return an error?
Basically, the group status should always be reconstructable from the members.  Can you confirm that the reconcilemembers will always populate the group status before we get here?  
Maybe we need to requeue the reconcile after we patch the Group member statues so the placement always sees that member?

aruneshpa (approved):
I just have one comment.  Other than that, looks good.  Thanks Sai.

controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
```diff
@@ -451,7 +466,7 @@ var _ = Describe(
 				}
 				intgFakeVMProvider.Unlock()
 
-				setVMUniqueID := func(vmKey types.NamespacedName, uniqueID string) {
+				setVMUniqueID = func(vmKey types.NamespacedName, uniqueID string) {
```
lubronzhan, line 427: nit: I saw other folks just define this helper func after `AfterEach`, before tests

lubronzhan (approved):
LGTM!

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -647,8 +683,13 @@ func (r *Reconciler) getGroupsForPlacement(
 		return nil, fmt.Errorf("failed to get group member group %s: %w", groupName, err)
 	}
 
+	// Initialize patch for this child group if it will need status updates.
+	if _, exists := groupPatches[vmGroup]; !exists {
```
bryanv: If it does exist, can't we handle the cycles todo below?

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -593,28 +581,58 @@ func (r *Reconciler) getPlacementMembers(
...
+			conditions.MarkError(
+				memberStatus,
+				vmopv1.VirtualMachineGroupMemberConditionPlacementReady,
+				"Error",
+				err)
+		} else if vm == nil {
+			// If both error and vm are nil, it means the VM is already placed,
+			// or will be placed outside the group. Update the PlacementReady
+			// condition to True to be able to set the group as ready.
+			conditions.MarkTrue(
```
bryanv: For PlacementReady=true in these situations I thought we had agreed to include some extra details (like Reason=AlreadyPlaced) for extra details about why the placement status is otherwise empty?

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
dilyar85: > but since we won't be able to find it in member status, this method will always return an error?

No, the `reconcileMember` function is called before this `reconcilePlacement`, so the member status will always be populated to match `spec.members` before the placement call here.

> Basically, the group status should always be reconstructable from the members. Can you confirm that the reconcilemembers will always populate the group status before we get here?

Yes, this is the current implementation.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
dilyar85: I saw there're other places using V(5) hence copied over. Updated all V(5) to V(4).

controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
dilyar85: Yep, it should be outside BeforeEach/AfterEach. I didn't wanna move it to cause more diff lines for easier reviews. Updated in the latest commit to define all functions outside.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
dilyar85: Excellent point! Didn't realize this could also be leveraged for cycle detection. Updated to address the todo, thanks @bryanv!

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
dilyar85: Moved the VM.status.uniqueID check first to update the PlacementReady with AlreadyPlaced correctly.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
dilyar85: > No, the reconcileMember function is called before this reconcilePlacement, so the member status will always be populated to match spec.members before the placement call here.

Actually, for nested groups, there could be a chance that the member status doesn't have the VM member yet during placement (if the nested group's own reconciliation hasn't called reconcileMember function). However, the root group controller should requeue and eventually find the member in nested group status. So this should still work.

PR #1215, opened 2025-09-30
🐛 Fix data race in VirtualMachineImageCache test

dilyar85:

This PR updates the fakeClient in VMIC test with `sync.RWMutex` to prevent concurrent reads/writes between test cleanup and controller goroutines. See https://github.com/vmware-tanzu/vm-operator/actions/runs/18137320993/job/51619344403?pr=1214 for the data race error.

Fixes N/A.

Passed when locally running `for i in {1..10}; do  echo "Run $i"; go test -count=1 -race ./controllers/virtualmachineimagecache/... -ginkgo.focus="cannot cache storage uris" || break; done` with this change (pervasively failing on main).

```release-note
Fix data race in VirtualMachineImageCache test.
```

### DIFF
--- a/controllers/virtualmachineimagecache/virtualmachineimagecache_controller_test.go
+++ b/controllers/virtualmachineimagecache/virtualmachineimagecache_controller_test.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"path"
 	"strings"
+	"sync"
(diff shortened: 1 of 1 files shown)


lubronzhan (approved):
LGTM!

PR #1216, opened 2025-10-01
:seedling: Reintroduce removed CRDs to avoid breakage

aruneshpa:

In a prior commit #1158, we removed some CRDs that were never supposed to be shipped on the API version in which they were.  But, it turns out that you can't remove a CRD from a version where it was marked as a storage version.  Kubernetes API server validation webhooks reject that request.

One solution is to force delete these types from the versions where we think they are safe.  But in the interest of time, just re-introduce them for now.

```release-note
Reintroduce removed CRDs to avoid breakage
```


bryanv (approved):

lubronzhan (commented):
LGTM

PR #1217, opened 2025-10-01
🌱 Cleanup duplicate Sysprep secret fetching

bryanv:

Both sysprep GetSysprepSecretData() and GetSecretResources() (used for backup) had very similar fetching of the Sysprep secrets. Refactor them to use a common helper method.

Don't fetch the same secret multiple times: our controller-runtime client does not cache secrets so avoid needless round trips.

Move this into pkg/util so it is along side the similar CloudInit and LinuxPrep code.

Fixes #

```release-note
NONE
```


pkg/util/sysprep/secret.go
```diff
@@ -0,0 +1,116 @@
...
+	secretData := SecretData{}
+	secrets := map[string]*corev1.Secret{}
+
+	getSecret := func(name, key string, outData *string) error {
+		// Our ctrl-runtime client does not cache secrets, so avoid repeated
+		// fetches of the same secret.
+		secret, ok := secrets[name]
+		if !ok {
+			var err error
+			secret, err = util.GetSecretResource(ctx, k8sClient, secretNamespace, name)
```
dilyar85, line 43: Consider adding a quick sanity check here to skip if the secret name is empty (in case it's set to an empty value in the API)?

dilyar85 (approved):
LGTM with a minor comment. Thanks @bryanv for refactoring this!

pkg/util/sysprep/secret.go
bryanv: Yah so the name is a require field in all the password structs here so I'm not sure why the domain admin one had that check (and that's the one I copied from the script text). I guess I'll add that check but generally we're assuming it is set.

PR #1224, opened 2025-10-03
Add a log statement that prints UserInfo object

abaruni:

Add a line in auth.go that will print the the
UserInfo object that is used to perform the isPrivilegedUser check. Log level 6 to keep from cluttering logs but to give the option for visibility into this.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/builder/auth.go
+++ b/pkg/builder/auth.go
@@ -37,6 +37,15 @@ func IsPrivilegedAccount(
 	ctx *pkgctx.WebhookContext,
 	userInfo authv1.UserInfo) bool {
 
+	if ctx != nil {
+		ctx.Logger.V(6).Info("Checking if account is privileged",
+			"Webhook Name", ctx.Name,
+			"Username", userInfo.Username,
+			"UID", userInfo.UID,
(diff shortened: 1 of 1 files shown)


akutz (approved):
Thank you @abaruni !

akutz:
@abaruni I think some of the tests are failing because the context or Logger are not initialized in them, and so we are seeing NPEs in tests, ex. https://github.com/vmware-tanzu/vm-operator/actions/runs/18234377025/job/51925025052?pr=1224#step:5:6803.

pkg/builder/auth.go
```diff
@@ -37,6 +37,13 @@ func IsPrivilegedAccount(
 	ctx *pkgctx.WebhookContext,
 	userInfo authv1.UserInfo) bool {
 
+	ctx.Logger.V(6).Info("Checking if account is privileged",
```
akutz: You should check if `ctx` is `nil` before using this (see the other comment I left on the PR).

lubronzhan (approved):
LGTM

PR #1228, opened 2025-10-06
🌱 Remove GOSC VCD Parity fields from VM when not enabled

bryanv:

Remove the fields added in 7b12de5 when capability is not enabled. Update things to allow for removing of nested fields under spec and status. Use dot field separators in the tests since that is a little easier to grab as-is from like k explain, and to reduce the overall chances of fat fingering anything twice.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -78,7 +78,7 @@ func UnstructuredExternal() ([]unstructured.Unstructured, error) {
 // Install installs the CRDs into the provided Kubernetes environment based on
 // the current set of feature/capability flags. This will also remove any APIs
 // from the provided environment if their flags have been disabled.
-func Install(
+func Install( //nolint:gocyclo
 	ctx context.Context,
 	k8sClient ctrlclient.Client,
 	mutateFn func(kind string, obj *unstructured.Unstructured) error) error {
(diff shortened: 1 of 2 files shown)


pkg/crd/crd.go
```diff
@@ -484,24 +501,20 @@ func decode(fs embed.FS, fileName string, dst ctrlclient.Object) error {
...
-	return []string{
-		"schema",
-		"openAPIV3Schema",
-		"properties",
-		"spec",
-		"properties",
-		fieldName,
+func specFieldPath(fieldNames ...string) []string {
+	result := []string{"schema", "openAPIV3Schema", "properties", "spec", "properties"}
+	result = append(result, fieldNames[0])
```
lubronzhan, line 506: Maybe check whether fieldNames's len is lager than 0 first?

pkg/crd/crd.go
```diff
@@ -484,24 +501,20 @@ func decode(fs embed.FS, fileName string, dst ctrlclient.Object) error {
...
-	return []string{
-		"schema",
-		"openAPIV3Schema",
-		"properties",
-		"status",
-		"properties",
-		fieldName,
+func statusFieldPath(fieldNames ...string) []string {
+	result := []string{"schema", "openAPIV3Schema", "properties", "status", "properties"}
+	result = append(result, fieldNames[0])
```
lubronzhan, line 515: Same for this.

pkg/crd/crd.go
bryanv: So this was intentional since it is almost certainly what was not intended

PR #1229, opened 2025-10-07
🌱 Allow privilege users to set RequireDuringScheduling for VM-VM Host AAF

dilyar85:

This PR updates the VirtualMachine validating webhook to allow privilege users to set the `vmAntiAffinity.requiredDuringSchedulingPreferredDuringExecution` field. This is required for Mobility Operator to specify VM-VM Host AAF polices.

Fixes N/A.

None.

```release-note
Allow privilege users to set RequireDuringScheduling for VM-VM Host AAF.
```


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -2251,10 +2251,9 @@ func (v validator) validateVMAffinity(
 	if a := affinity.VMAntiAffinity; a != nil {
 		p := path.Child("vmAntiAffinity")
 
-		// TODO(sai): Allow this for privileged users.
-		if len(a.RequiredDuringSchedulingPreferredDuringExecution) > 0 {
-			allErrs = append(allErrs, field.Forbidden(p.Child("requiredDuringSchedulingPreferredDuringExecution"),
-				"VM anti-affinity with RequiredDuringSchedulingPreferredDuringExecution is not allowed"))
+		// Mobility Operator needs to specify RequireDuringScheduling type policies for VM-VM Host AAF.
```
aruneshpa: One of the AIs from the code walkthrough was that we will allow both "required" and "preferred" during scheduling variants for VM-VM zonal anti-affinity.  Can you handle that as well?

webhooks/virtualmachine/validation/virtualmachine_validator.go
aruneshpa: This is not your change, but this needs to be changed as well?  As of now, we are only allowing host / node label for preferred preferred?
Zones need to be allowed as well.

webhooks/virtualmachine/validation/virtualmachine_validator.go
dilyar85: Updated to allow VM-VM zonal anti-affinity for all users.

webhooks/virtualmachine/validation/virtualmachine_validator.go
dilyar85: Fixed.

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -2251,10 +2251,45 @@ func (v validator) validateVMAffinity(
...
+				}
+
+				// For non-privileged users, only zone topology key is allowed for anti-affinity required terms.
+				if !ctx.IsPrivilegedAccount && rs.TopologyKey != corev1.LabelTopologyZone {
+					allErrs = append(allErrs, field.NotSupported(
+						p.Child("topologyKey"), rs.TopologyKey, []string{corev1.LabelTopologyZone}))
+				}
+
+				// For privileged users (mobility operator), either zone or host topology key is allowed for anti-affinity required terms.
+				if ctx.IsPrivilegedAccount && rs.TopologyKey != corev1.LabelTopologyZone && rs.TopologyKey != corev1.LabelHostname {
```
aruneshpa, line 2288: TopologyKey can only be host, or zone, so if you validate in the beginning of the method that only two values are allowed, then you can just say:

```
if topologyKey == host {
        if !ctx.IsPriviligedAccount {
                 Nope.
        }
}
```

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -2284,9 +2319,10 @@ func (v validator) validateVMAffinity(
 					}
 				}
 
-				if rs.TopologyKey != "" && rs.TopologyKey != corev1.LabelHostname {
+				// Either zone or host topology key is allowed for anti-affinity preferred terms.
+				if rs.TopologyKey != corev1.LabelTopologyZone && rs.TopologyKey != corev1.LabelHostname {
```
aruneshpa, line 2323: This can just be the very entrypoint of each of the required / preferred blocks.  Any other topologyKey than host or zone and we reject the request.

webhooks/virtualmachine/validation/virtualmachine_validator.go
dilyar85: I kinda feel like the current code is more readable and easier to understand for admin & non-admin users. Will keep it as is if you don't feel strongly about this change.

webhooks/virtualmachine/validation/virtualmachine_validator.go
dilyar85: I think the current convention in our webhook is to collect all errors and return them together, so that all errors are visible to users in one request. I will keep it as is since it's just a matter of checking this before or after the label in the loop.

PR #1196, opened 2025-09-20
:seedling: Simplify VirtualMachineSnapshot and related APIs

aruneshpa:
**What does this PR do, and why is it needed?**

This commit refactors the VirtualMachineSnapshot API to use a simple string field for VM references instead of a complex LocalObjectRef structure, and resolves field conflicts in the VirtualMachine API.

API Changes:
- Replace VirtualMachineSnapshotSpec.VMRef (LocalObjectRef) with VMName (string)
- Rename VirtualMachine.Spec.CurrentSnapshot to CurrentSnapshotName to resolve JSON tag conflict with VirtualMachine.Status.CurrentSnapshot
- Simplify VirtualMachineSnapshotReference to use Name field directly
- VM status now includes the type (managed / unmanaged) and the name of the snapshot CR)

Other changes include:
- Update virtualmachinesnapshot controller to use VMName instead of VMRef
- Update storagepolicyusage and other references in code to handle the above change.
- Simplify the validation webhook to handle this change
- Add validation to prevent empty/whitespace-only currentSnapshotName
- Maintain VKS/TKG node snapshot prevention logic
- Provider changes to handle this change

**Are there any special notes for your reviewer**:

Note that I wanted to keep the `spec.currentSnapshot` for the desired snapshot but that's not possible because two fields with different type can't have the same JSON tag (spec being a string, and status being an object).

**Please add a release note if necessary**:

```release-note
Simplify VirtualMachineSnapshot and related APIs
```

PR #1231, opened 2025-10-07
🌱Fix snapshot related field name in VM spec removal logic

lubronzhan:

`currentSnapshot` was changed to `currentSnapshotName` in https://github.com/vmware-tanzu/vm-operator/pull/1196. Then we should also update the field when removing snapshot related field from vm spec.

Fix some of the test description.

```release-note
None
```


dilyar85 (approved):
LGTM!

PR #1337, opened 2025-11-22
:sparkles: Populate diskMode, provisioning mode and Scrubing of disks in VM status

aruneshpa:

This change modifies VM status to include additional properties of a virtual disk such as:
- Disk mode: this includes persistent, independent_persistent, independent_nonpersistent, nopersistent, undoable and append
- Provisioning mode: Thin vs thick provisioned
- Scrubing: Lazy vs eager zero-ing

Diskmode is already available in the Status but not being populated. This change not sets that property.  A new property called provisioningMode is introduced to handle zero-ing of disks.

```release-note
Populate diskMode, provisioning mode and Scrubing of disks in VM status
```


pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1414,6 +1467,45 @@ func updateVolumeStatus(vmCtx pkgctx.VirtualMachineContext) {
 			}
 		})
 
+	// Update disk attachment properties for ALL volumes (classic and managed).
+	// This is done in a single pass at the end to ensure consistency.
+	// For managed volumes, the volume controllers populate status from CnsNodeVmAttachment,
```
lubronzhan: Better to copy over the ProvisioningMode in volumeBatch controller as well

https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachine/volumebatch/volumebatch_controller.go#L650
https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachine/volumebatch/volumebatch_controller.go#L1020

Otherwise it will keep getting removing and adding

lubronzhan: This got me thinking, we are also patching status.volume with `detaching` suffix here, then we should also add these for detaching disks in volumeController. You don't have to include this as part of this PR

lubronzhan (approved):
LGTM

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1356,6 +1405,8 @@ func updateVolumeStatus(vmCtx pkgctx.VirtualMachineContext) {
 					KeyID:      ddi.CryptoKey.KeyID,
 				}
 			}
+			// Note: DiskMode, SharingMode, and ProvisioningMode are set later
```
bryanv: Why not in this loop?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1437,6 +1490,45 @@ func updateVolumeStatus(vmCtx pkgctx.VirtualMachineContext) {
...
+		if !ok {
+			vmCtx.Logger.V(4).Info("No disk info found for volume",
+				"volumeName", vol.Name,
+				"volumeType", vol.Type,
+				"diskUUID", vol.DiskUUID)
+			continue
+		}
+
+		// Populate disk attachment properties from vSphere disk info.
+		if di.DiskMode != "" {
```
bryanv, line 1736: If these are "", should the status field be cleared?

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1437,6 +1490,45 @@ func updateVolumeStatus(vmCtx pkgctx.VirtualMachineContext) {
 			}
 		})
 
+	// Update disk attachment properties for ALL volumes (classic and managed).
+	// This is done in a single pass at the end to ensure consistency.
```
bryanv: What do you mean by consistency here? And why not in the earlier loop?

pkg/providers/vsphere/vmlifecycle/update_status.go
aruneshpa: Didn't bother about that because a diskMode going from not-empty to empty is not a valid scenario unless the disk's backing is changed to the _two_ specific backing that doesn't have disk mode:
- VirtualDiskRawDiskVer2BackingInfo
- VirtualDiskPartitionedRawDiskVer2BackingInfo
Which are very rare.

https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/util/devices.go#L288-L335.  

Also, for regular disks, diskMode is immutable property, so there's no case of a disk's mode changing once it is non-empty.

LMK if this makes sense.

pkg/providers/vsphere/vmlifecycle/update_status.go
aruneshpa: Earlier loop has separate handling for classic / unmanaged disks and PVCs.  So, if I handle it there, I would need to duplicate it in both, if and else blocks.
Consistency may have been wrong choice of a word.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1700,6 +1701,53 @@ func updateVolumeStatus(vmCtx pkgctx.VirtualMachineContext) {
...
+	// For managed volumes, the volume controllers populate status from CnsNodeVmAttachment,
```
bryanv: Nit: update the comment for batch

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1700,6 +1701,53 @@ func updateVolumeStatus(vmCtx pkgctx.VirtualMachineContext) {
...
+	// which doesn't include DiskMode, SharingMode, or ProvisioningMode.
+	// We populate these properties here by matching DiskUUID with vSphere disk info.
+	diskInfoByUUID := make(map[string]pkgvol.VirtualDiskInfo)
```
bryanv: Change the value type of the `existingDisksInConfig` and use that map instead?

pkg/providers/vsphere/vmlifecycle/update_status.go
bryanv: If you do my other comment, prealloc this while here

bryanv (approved):

PR #1406, opened 2025-12-31
🌱 Don't always backfill VM Status.Zone from label

bryanv:

When either is unset, or they have different values, lookup the VM's zone and assign the zone to the label and status. To avoid a lookup if both the label and status are set to the same value, we'll just trust what is there.

The label value is protected by the validation webhook for non privileged users, but if the label does happen to get changed - like by kube-admin - relook the correct zone.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmlifecycle/update_status.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status.go
@@ -375,14 +375,18 @@ func reconcileStatusZone(
 
 	var errs []error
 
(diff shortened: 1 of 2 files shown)


dilyar85 (approved):
LGTM!

akutz (approved):
Approved, but I would ask you to please add some in-line comments in `update_status.go` that reflect the same messaging in the PR description. That way future folks will know the reason for this behavior. Thanks!

PR #1534, opened 2026-03-23
🐛 Add missing v1a6 version conversion bits

bryanv:

Add stuff that the old new-schema-version.py missed when a type was just added in the now prior Hub version.

While here, also add conversion for VirtualMachineClassInstance since that wasn't done between v1a4 and v1a5.

Fixes #

```release-note
NONE
```

### DIFF
--- a/api/v1alpha4/virtualmachineclassinstance_conversion.go
+++ b/api/v1alpha4/virtualmachineclassinstance_conversion.go
@@ -1,0 +1,23 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha4
+
+import (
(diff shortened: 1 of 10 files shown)


api/v1alpha4/virtualmachineclassinstance_conversion.go
```diff
@@ -0,0 +1,23 @@
...
+package v1alpha4
+
+import (
+	ctrlconversion "sigs.k8s.io/controller-runtime/pkg/conversion"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
+)
+
+// ConvertTo converts this VirtualMachineClassInstance to the Hub version.
+func (src *VirtualMachineClassInstance) ConvertTo(dstRaw ctrlconversion.Hub) error {
```
bryanv, line 14: Since this CRD is stripped when the feature is not enabled, I think we could remove it from v1a4 & v1a5 instead.

akutz (approved):
Thank you for catching this!

PR #1547, opened 2026-04-07
:bug: Revert pull request #1533 - VM Create and storage policy ID 

brito-rafa:
This PR reverts PR #1533 - VM Create to retrieve storage policy ID agnostically from StorageClass or VolumeAttributesClass wrapped by a capability flag.
Under the new retrieval simplified logic of storage profile, the reco tests failed to detect encryption from storage policy objects. Reinstating original code until more reco tests are validated. 

**What does this PR do, and why is it needed?**

Revert changes introduced on commit `5f2890c`.

**Which issue(s) is/are addressed by this PR?** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes #

Revert changes introduced on commit `5f2890c`.

**Please add a release note if necessary**:

```release-note

```

PR #1551, opened 2026-04-08
🐛 Reinstate Logic of #1533 and Fix bug introduced by #1540 on StoragePolicy Status Field

brito-rafa:

This patch reinstates the logic originally introduced in PR #1533 (and reversed by #1547), which allows the VM creation process to derive the storage policy ID from either a StorageClass or a VolumeAttributesClass (VAC) when the capability is enabled.
The original implementation was reverted due to failures in gce2e encryption tests.
On this PR, we resolved the root cause of the gce2e encryption test failures by adding a missing capability check around the status updates of Storage Policy objects for the VolumeAttributesClasses field (originally introduced in PR #1540).
During the investigation, we acknowledged that storage policy objects may not be immediately ready for encryption checks, as they require a multi-step lifecycle (creation by the StorageClass controller followed by population by the StoragePolicy controller).
We slightly updated the logic of the function `IsEncryptedStorageProfile` which we no longer "swallow" the NotFound error - as suggested (and agreed to address it eventually) during revision of  #1533 . This function is called by other components and to minimize impact and ensure backward compatibility for other callers, the IgnoreNotFound was added when applicable. We added the storage policy object as part of vcsim test context to represent the realist production scenario.

Fixes # N/A

Same code from #1533  + adding missing capability check from PR #1540 + no longer ignoring error if Storage Policy does not exist 

```release-note
Addresses bug when Storage Policy objects on VM Operator namespace were not properly reconciled when running on Kubernetes version v1.33 and below. 
Allows the VM creation process to derive the storage policy ID from either a StorageClass or a VolumeAttributesClass (VAC) when the capability is enabled
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -2296,11 +2296,11 @@ func (vs *vSphereVMProvider) vmCreateGetStoragePrereqs(
 		return err
 	}
 
-	// The storage profile ID can be obtained from the StorageClass or VolumeAttributesClass (VAC) object, depending if the StoragePolicyMutability feature is enabled.
 	storageProfileID, err := kubeutil.GetStoragePolicyID(vmCtx, vs.k8sClient, *vmCtx.VM)
 	if err != nil {
 		return err
 	}
+
 	isEnc, err := kubeutil.IsEncryptedStorageProfile(vmCtx, vs.k8sClient, storageProfileID)
 	if err != nil {
 		return err
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -2296,9 +2296,12 @@ func (vs *vSphereVMProvider) vmCreateGetStoragePrereqs(
 		return err
 	}
 
-	storageProfileID := vmStorage.StorageClassToPolicyID[vmStorageClass]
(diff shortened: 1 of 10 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2296,9 +2296,12 @@ func (vs *vSphereVMProvider) vmCreateGetStoragePrereqs(
 		return err
 	}
 
-	storageProfileID := vmStorage.StorageClassToPolicyID[vmStorageClass]
-	isEnc, _, err := kubeutil.IsEncryptedStorageClass(
-		vmCtx, vs.k8sClient, vmCtx.VM.Spec.StorageClass)
+	// The storage profile ID can be obtained from the StorageClass or VolumeAttributesClass (VAC) object, depending if the StoragePolicyMutability feature is enabled.
```
akutz: Do not exceed 80 chars please. Wrap this.

akutz (changes requested):
A nit on a comment block. Try not to exceed 80 chars when writing code or comments.

pkg/providers/vsphere/vmprovider_vm.go
faisalabujabal: I removed this comment since it is a duplicate of the go docs in the function definition

akutz (approved):

PR #1552, opened 2026-04-08
✨ Dual-stack networking: NetOP IPv6 assignment modes, VirtualMachineService changes

hpannem:

- **NetOP:** Bumps `net-operator-api` and derives IPv4 vs IPv6 assignment from `status.ipAssignmentMode` and `status.ipv6AssignmentMode` (unset IPv6 mode treated as none). Static `IPConfigs` are emitted only for families in static-pool mode. Updates tests and the bundled NetOP CRD reference as needed.
- **VirtualMachineService (v1alpha6):** Dual-stack `Service` fields (`ipFamilies`, `ipFamilyPolicy`) and controller behavior for endpoints; extends unit/integration coverage.
- **Older API versions:** `VirtualMachineService` uses conversion annotations so hub-only dual-stack fields round-trip. `VirtualMachine` restores `spec.network.interfaces[].ipFamilyPolicy` on spoke→hub (by interface name) together with existing VLAN/bootstrap/etc. restores.
- **conversion-gen:** Regenerated `zz_generated.conversion.go` so `VirtualMachineNetworkSpec.Interfaces` converts via per-element `Convert_*` (no unsafe slice reinterpretation for that slice).

**Which issue(s) is/are addressed by this PR?** *(optional)*

Fixes #

**Are there any special notes for your reviewer?**

- `IPFamilyPolicy` exists only on the **v1alpha6** interface type; spokes do not have the field—values are preserved via **annotation + restore** on `VirtualMachine.ConvertTo`, not via typed spoke fields.
- Explicit regression: `api/test/v1alpha5/virtualmachine_conversion_test.go` includes `spec.network.interfaces.ipFamilyPolicy` hub↔spoke↔hub; fuzz conversion tests still run per spoke.

**Please add a release note if necessary:**

```release-note
VirtualMachineService supports dual-stack Service fields. NetOP network results honor per-family assignment modes including `ipv6AssignmentMode`. Older API versions preserve hub dual-stack and per-interface IP family policy through conversion annotations.

### CODE REVIEW

File: api/v1alpha6/virtualmachine_network_types.go [hunk 1]
```diff
@@ -179,6 +191,19 @@ type VirtualMachineNetworkInterfaceSpec struct {
...
+
+	// IPFamilyPolicy specifies the IP family policy for this network interface.
+	// Values: IPv4Only, IPv6Only, DualStack.
+	// When set to IPv4Only, only an IPv4 address will be allocated.
+	// When set to IPv6Only, only an IPv6 address will be allocated.
+	// When set to DualStack, both IPv4 and IPv6 addresses will be allocated.
+	// This field is directly passed to the underlying network interface provider.
+	// If not specified, the field will be nil and no IPFamilyPolicy will be set and the
+	// default behavior of the network interface provider will be used.
+	IPFamilyPolicy *NetworkInterfaceIPFamilyPolicy `json:"ipFamilyPolicy,omitempty"`
```
silvery1622: The new added filed ipFamilyPolicy shares the same name as the Kubernetes Service field and the VirtualMachineService field but carries different semantic values (e.g., IPv4Only vs SingleStack). Would it make sense to rename this field to ipFamily to avoid potential confusion for users?

hpannem: Good observation. However, dependent net-operator already names it as IPFamilyPolicy. So, is it ok?

akutz:
I would like this to also handle changes to the ports. We can discuss offline, and I'll update here after we do.

api/v1alpha6/virtualmachine_network_types.go
silvery1622: Understood. Thanks for the context!

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -557,6 +575,16 @@ func (r *ReconcileVirtualMachineService) createOrUpdateEndpoints(ctx *pkgctx.Vir
 		return err
 	}
 	subsets := utils.RepackSubsets(unpackedSubsets)
+	// Split subsets by IP family to ensure IPv4 and IPv6 remain separate
+	// Above RepackSubsets merges subsets with identical ports, which combines IPv4 and IPv6
+	subsets = r.splitSubsetsByIPFamily(subsets)
```
bryanv: It isn't clear to me why we're doing this. A quick look at the EndpointSlice mirroring and I don't think need requires the EP to be this way. Is there something in SV that requires the legacy EP to be this way for dual-stack?

Similarity it isn't clear to me why we're then filtering after the fact instead of doing that up front when just building the thing.

This code is quite old but we had to import the k8s RepackSubsets since the k8s would repack in the EP subsets inline in the strategy routine so we needed to do that here to avoid constant churning. I don't think that is actually done anymore today but since we still depend on the CreateOrPatch here I'd like to keep the repacked subsets as-is so it is less likely anything after that would potentially reorder thing in a non-deterministic way.

hpannem: Fair point — generate → repack → split → filter on legacy Endpoints was not ideal.

I was partly thinking ahead to EndpointSlice (which configures subnets or slices per addressType/family type), but for Endpoints it may be unnecessary.

Now, I removed the split and moved the family filtering into generateSubsetsForService so disallowed families never enter the pipeline. We can revisit per-family structure when we actually implement EndpointSlice.

hpannem:
> I would like this to also handle changes to the ports. We can discuss offline, and I'll update here after we do.

Got your point. I will track this as separate ticket.

pkg/providers/vsphere/network/netop_assignment_mode_internal_test.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+
+	netopv1alpha1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
+)
+
+func TestEffectiveNetOPIPv4AssignmentMode(t *testing.T) {
+	t.Parallel()
```
bryanv, line 16: Use Ginkgo, for better or worse.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -426,6 +426,9 @@ func reconcileStatusGuest(
 	if config := vmCtx.MoVM.Config; config != nil {
 		extraConfig = object.OptionValueList(config.ExtraConfig).StringMap()
 	}
+	if extraConfig == nil {
```
bryanv, line 429: Why?

api/v1alpha6/virtualmachineservice_types.go
```diff
@@ -138,6 +139,39 @@ type VirtualMachineServiceSpec struct {
...
+	// IPFamilies is a list of IP families (e.g. IPv4, IPv6) assigned to this
+	// service. This field is usually assigned automatically based on cluster
+	// configuration and the ipFamilyPolicy field. If this field is specified
+	// manually, the requested family is available in the cluster,
+	// and ipFamilyPolicy allows it, it will be used; otherwise creation of
+	// the service will fail. This field is conditionally mutable: it allows
+	// for adding or removing a secondary IP family, but it does not allow
+	// changing the primary IP family of the Service. Valid values are "IPv4"
+	// and "IPv6".  This field only applies to Services of types ClusterIP,
+	// NodePort, and LoadBalancer, and does apply to "headless" services.
```
bryanv, line 155: NodePort does not make sense for how VirtualMachineService are setup or for the network segmentation so it is not supported. 

Capital S `Service` should be our actual type name here.

controllers/virtualmachineservice/virtualmachineservice_controller_intg_test.go
```diff
@@ -303,6 +303,142 @@ func intgTestsReconcile() {
...
+							for _, addr := range subset.Addresses {
+								allIPs = append(allIPs, addr.IP)
+							}
+						}
+
+						Expect(allIPs).To(ContainElement("192.168.1.100"), "Should contain IPv4 address")
+						Expect(allIPs).NotTo(ContainElement("2001:db8::100"), "Should not contain IPv6 address for SingleStack IPv4 service")
+					})
+
+					By("Cleanup", func() {
```
bryanv, line 372: Why isn't the cleanup done in the AfterEach?

controllers/virtualmachineservice/virtualmachineservice_controller_intg_test.go
```diff
@@ -303,6 +303,142 @@ func intgTestsReconcile() {
...
+					By("Create VirtualMachineService", func() {
+						Expect(ctx.Client.Create(ctx, vmService)).To(Succeed())
+					})
+
+					By("Endpoints should contain only IPv4 address for SingleStack service", func() {
+						// In an IPv4-only cluster, a SingleStack service gets an IPv4 ClusterIP
+						// An IPv4 service can only route to IPv4 endpoints, so only IPv4 is included
+						// even if the VM has both IPv4 and IPv6 addresses
+						endpoints := &corev1.Endpoints{}
+						Eventually(func() bool {
```
bryanv, line 353: In new code, use the func(g Gomega) { ... }).To(Succeed()) pattern for these. Feel free to update the existing code to use this too.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1119,6 +1171,40 @@ func updateGuestNetworkStatus(
 		}
 	}
 
+	// Fallback logic: If one IP is missing, try to find it from the interface containing the other IP
```
bryanv, line 1174: Historically we've always depended on the guest pushing down what it thinks is primary IPs are and think we should continue to honor that. Blindly using the IPv6 address from the primary IPv4 interface concerns me too since its scope make not really make sense for a primary IP.

api/v1alpha3/virtualmachine_conversion.go
```diff
@@ -466,11 +467,10 @@ func (src *VirtualMachine) ConvertTo(dstRaw ctrlconversion.Hub) error {
 	restore_v1alpha6_VirtualMachinePolicies(dst, restored)
 	restore_v1alpha6_VirtualMachineCryptoVTPM(dst, restored)
 	restore_v1alpha6_VirtualMachineAffinity(dst, restored)
-	restore_v1alpha6_VirtualMachineCryptoVTPM(dst, restored)
```
bryanv, line 469: Why?

api/test/v1alpha5/virtualmachineservice_conversion_test.go
```diff
@@ -0,0 +1,51 @@
...
+		g := NewWithT(t)
+		after := &vmopv1.VirtualMachineService{}
+		spoke := &vmopv1a5.VirtualMachineService{}
+
+		g.Expect(spoke.ConvertFrom(hub)).To(Succeed())
+		g.Expect(spoke.ConvertTo(after)).To(Succeed())
+		g.Expect(apiequality.Semantic.DeepEqual(hub, after)).To(BeTrue(), cmp.Diff(hub, after))
+	})
+
+	t.Run("implements Hub and Convertible", func(t *testing.T) {
```
bryanv, line 47: I don't understand why this is done as a test

api/v1alpha6/virtualmachine_network_types.go
```diff
@@ -223,6 +235,19 @@ type VirtualMachineNetworkInterfaceSpec struct {
...
```
bryanv, line 250: I agree with the point here - `IPFamilyPolicy` already feels out of place here, and don't like that we've overloaded the name with different values. Allocate or Request IP would be clearer to what is actually the effect of this field, especially how this interacts with directly specifying an IP or otherwise configuration the device.

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -409,6 +410,23 @@ func (r *ReconcileVirtualMachineService) createOrUpdateService(ctx *pkgctx.Virtu
 		service.Spec.ExternalName = vmService.Spec.ExternalName
 		service.Spec.LoadBalancerIP = vmService.Spec.LoadBalancerIP
 		service.Spec.LoadBalancerSourceRanges = vmService.Spec.LoadBalancerSourceRanges
+
+		// Set IPFamilies and IPFamilyPolicy for dual-stack support
+		// These fields only apply to ClusterIP, NodePort, and LoadBalancer types
+		// They are wiped when type is ExternalName
+		if vmService.Spec.Type != vmopv1.VirtualMachineServiceTypeExternalName {
+			if len(vmService.Spec.IPFamilies) > 0 {
```
bryanv, line 418: Should we be adding some validation webhooks checks for the VirtualMachineService for this field? I don't think we can catch everything that can cause the underlying Service update to fail but an early error for what we can might be nice.

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -702,20 +708,132 @@ func (r *ReconcileVirtualMachineService) generateSubsetsForService(
...
+	// If IPFamilies is explicitly set, use that (takes precedence for all policies)
+	if len(service.Spec.IPFamilies) > 0 {
+		for _, family := range service.Spec.IPFamilies {
+			allowedFamilies[family] = true
+		}
+		return allowedFamilies
+	}
+
+	// For SingleStack policy without explicit IPFamilies, use ClusterIP family as primary
+	if service.Spec.IPFamilyPolicy != nil && *service.Spec.IPFamilyPolicy == corev1.IPFamilyPolicySingleStack {
```
bryanv, line 777: Isn't the Service `Spec.IPFamilies` written back to as to what was selected?

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -702,20 +708,132 @@ func (r *ReconcileVirtualMachineService) generateSubsetsForService(
...
+		// Determine primary family from IPFamilies or ClusterIP
+		var primaryFamily corev1.IPFamily
+		if len(service.Spec.IPFamilies) > 0 {
+			primaryFamily = service.Spec.IPFamilies[0]
```
bryanv, line 781: Is this block reachable?

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -702,20 +708,132 @@ func (r *ReconcileVirtualMachineService) generateSubsetsForService(
...
+
+		// For SingleStack, only allow the primary family (ClusterIP/IPFamilies family)
+		// This ensures that endpoints match the Service's IP family, as an IPv4 Service
+		// cannot route to IPv6 endpoints and vice versa
+		allowedFamilies[primaryFamily] = true
+		return allowedFamilies
+	}
+
+	// For non-SingleStack policies, use IPFamilies if explicitly set
+	if len(service.Spec.IPFamilies) > 0 {
```
bryanv, line 813: Same check that is done earlier?

controllers/virtualmachineservice/virtualmachineservice_controller_intg_test.go
```diff
@@ -303,6 +303,142 @@ func intgTestsReconcile() {
 				vmServiceName = "test-vm-service-unmatched-vm"
 			})
 
+			Context("Dual-stack VM", func() {
```
bryanv, line 306: Is this placed under the appropriate parent Context?

api/v1alpha3/virtualmachine_conversion.go
hpannem: This is duplicated at line 468. It looks like accidental.

api/v1alpha6/virtualmachine_network_types.go
hpannem: Agree with overloading... however, the net operator already names it like that... We can rename VM CR field... which would deviate from net operator or nsx operator. Please suggest.
K8s service has two fields: IPFamily and IPFamilyPolicy. Here, we have overloaded both with one field.

api/test/v1alpha5/virtualmachineservice_conversion_test.go
hpannem: Agreed. Will remove it.

api/v1alpha6/virtualmachineservice_types.go
hpannem: You are right about the NodePort. I will remove this. Updated the godoc referred to VirtualMachineService consistently instead of generic “service” / Service.

controllers/virtualmachineservice/virtualmachineservice_controller.go
hpannem: Good catch! This is a refactoring issue! I moved this to early exit, but forgot to remove this.

controllers/virtualmachineservice/virtualmachineservice_controller.go
hpannem: I added kubebuilder checks.

controllers/virtualmachineservice/virtualmachineservice_controller.go
hpannem: Thats right! I added early exit in function when spec.IPFamilies are specified... this is not necessary.

controllers/virtualmachineservice/virtualmachineservice_controller.go
hpannem: Good point. I introduced this as a fallback in case if service.IPfamilies are not filled for some reason in the Service. We could default it to spec.Service.IPFamilies always. what do you suggest? Except legacy clusters, I believe the field would be filled by K8s.

pkg/providers/vsphere/vmlifecycle/update_status.go
hpannem: When vm config is nil, this will be nil. I know we are using map for read only down the line... So, nil map is fine. I just put to avoid any inadvertent usage. What do you suggest?

pkg/providers/vsphere/network/netop_assignment_mode_internal_test.go
hpannem: sure.

pkg/providers/vsphere/vmlifecycle/update_status.go
hpannem: This is for the fallback in dualstack environment. If the guest info tools correctly report that, we will use that. If not, the primary interface is considered as the default interface right. So, fetching the other IP from this interface sounds reasonable right? I have seen instances the guest didn't report correctly. (I will retest and confirm this). In the meanwhile, what do you recommend?

pkg/providers/vsphere/vmlifecycle/update_status.go
hpannem: I removed it anyways!

controllers/virtualmachineservice/virtualmachineservice_controller_intg_test.go
hpannem: Done

controllers/virtualmachineservice/virtualmachineservice_controller_intg_test.go
hpannem: Done

api/v1alpha6/virtualmachine_network_types.go
yangna0420: @bryanv  @silvery1622 @hpannem 

Good point, Bryan and Silvery — I just realized this issue, agree that reusing the name ipFamilyPolicy with a completely different value set (IPv4Only/IPv6Only/DualStack vs. K8s SingleStack/PreferDualStack/RequireDualStack) is confusing and maybe we need fix it on the vm-operator side.

**Proposed change for vm-operator (this PR):**  ==> I generated this proposal with cursor
Rename the field on VirtualMachineNetworkInterfaceSpec from `ipFamilyPolicy` to `requestedIPFamilies`, using the same `corev1.IPFamily` type that K8s and VirtualMachineServiceSpec.IPFamilies already use:
```
// RequestedIPFamilies specifies which IP address families to allocate
// for this network interface. Valid values: "IPv4", "IPv6".
// When ["IPv4"] is specified, only IPv4 will be allocated.
// When ["IPv6"] is specified, only IPv6 will be allocated.
// When both are specified, dual-stack addresses will be allocated.
// If empty, the allocation is determined by the backing network's capabilities.
// +optional
// +listType=set
RequestedIPFamilies []corev1.IPFamily `json:"requestedIPFamilies,omitempty"`
```
The vm-operator controller would then do a straightforward mapping when writing to the net-operator NetworkInterface CR:

```
func toNetOpIPFamilyPolicy(families []corev1.IPFamily) netopv1alpha1.NetworkInterfaceIPFamilyPolicy {
    hasV4, hasV6 := false, false
    for _, f := range families {
        if f == corev1.IPv4Protocol { hasV4 = true }
        if f == corev1.IPv6Protocol { hasV6 = true }
    }
    switch {
    case hasV4 && hasV6: return netopv1alpha1.NetworkInterfaceIPFamilyPolicyDualStack
    case hasV6:          return netopv1alpha1.NetworkInterfaceIPFamilyPolicyIPv6Only
    default:             return netopv1alpha1.NetworkInterfaceIPFamilyPolicyIPv4Only
    }
}

```
This way:

The user-facing vm-operator API uses `[]corev1.IPFamily` — consistent with `VirtualMachineServiceSpec.IPFamilies` and the broader K8s convention — no more naming collision.
The underlying net-operator CRD (NetworkInterface.spec.ipFamilyPolicy) doesn't need to change immediately, since it's an infrastructure-level CRD that end users rarely interact with directly — it's primarily written by the vm-operator controller. The confusion risk is much lower at that layer. That said, the same naming issue does exist there.

**On the net-operator-api side:**
The NetworkInterface.spec.ipFamilyPolicy field was merged recently in [net-operator-api#33](https://github.com/vmware-tanzu/net-operator-api/pull/33). Since it's an internal CRD consumed by controllers rather than end users, the urgency is lower — but the same overloaded-name concern applies. If you feel it should also be renamed (e.g. to requestedIPFamilies for consistency), we can coordinate that change — the merge was recent enough that the blast radius is still small. Happy to do both together if you'd prefer.

Let me know if you're aligned on this direction and I'll update the PR accordingly.

pkg/providers/vsphere/vmlifecycle/update_status.go
bryanv: It is hard for us to say what is primary or not. You can see what the VMware cloudinit Datasource is doing at https://github.com/canonical/cloud-init/blob/87cc1ae50616c651887122f1c465a605d89a53ef/cloudinit/sources/DataSourceVMware.py#L297. It is similar to what is being done here but additional things that IMO make it safer like using default route, using len()==1 instead of assuming the first one. I'm not so sure if we should override for when can't make that determination. 

For non-cloudinit, IDK what VM Tools does out of the top of my head, but I could view whatever it does as standardized & expected behavior, like say for an imported VM where as the user did get that info via VMODL and now expects to get the same in our status.

api/v1alpha6/virtualmachine_network_types.go
hpannem: Thanks @yangna0420 for taking a stab at this. One concern is whether the order of the requested array matter. [ipv6, ipv4] or [ipv4, ipv6]. In k8s, ipfamilies work along with ipfamilypolicy. In our case the order does not matter, unless the implementation of net op changes. Should we just put **requestedAddressFamilyMode** with same enums just to deviate from ipfamilypolicy?

@bryanv what do you think?

yangna0420: @hpannem  If you replace to use `RequestedIPFamilies`, the order doesn't matter. ====> Since I suppose you could use the function `toNetOpIPFamilyPolicy ` to transform the value to `IPv4Only, IPv6Only and DualStack` in the net-operator NetworkInterface CR. 

But i am not sure whether other resources (not net-operator) depends on ``RequestedIPFamilies`.

yangna0420: three options:

- Option A: requestedIPFamilies []corev1. IPFamily
Consistent with K8s conventions and VirtualMachineServiceSpec. IPFamilies
Reuses corev1. IPFamily type — no new enums
Add a godoc note: "Order is not significant"
Downside: list form implies ordering even if we document otherwise

- Option B: addressFamilyPolicy (enum: IPv4Only/IPv6Only/DualStack)
Concise, no ordering ambiguity
Still has "policy" in the name, but the addressFamily prefix clearly separates it from K8s ipFamilyPolicy

- Option C: requestedAddressFamilyMode (enum: same values)
Completely avoids any K8s naming overlap
Downside: the name feels disconnected from the underlying net-operator field (ipFamilyPolicy)

I lean toward A for K8s ecosystem alignment, but B is also clean. Maybe could talking with @bryanv  for his suggestions?

pkg/providers/vsphere/vmlifecycle/update_status.go
hpannem: Thanks @bryanv. Makes sense. I tested again... without this code, it fails in dualstack... primaryIPv6 is not getting assigned. May be a bug in cloudinit, not sure. So, we need a fallback. I can add len() == 1 (non loopback or non link local) check and take other ip from the interface.

if guest info is unable to report any primary ip, should we fail?

bryanv: I don't understand what we would fail here. For whatever you're seeing no PrimaryIPv6, that means both VM Tools and CloudInit did not report a primary IPv6 so IDK if I wouldn't really conclude that is a Cloudinit bug.

hpannem: am adding the fallback like this. I am testing it. Hopefully it's ok.
```
	if len(ifaceStatuses) > 0 {
		switch {
		case primaryIP4 != "" && primaryIP6 == "":
			// Try to find IPv6 on the same interface as IPv4
			if idx := findInterfaceContainingIP(primaryIP4, ifaceStatuses); idx >= 0 {
				ipv6Addrs := extractIPsFromInterface(ifaceStatuses[idx], false, validatePrimaryIP)
				if len(ipv6Addrs) == 1 {
					primaryIP6 = ipv6Addrs[0]
				}
			}
		case primaryIP6 != "" && primaryIP4 == "":
			// Try to find IPv4 on the same interface as IPv6
			if idx := findInterfaceContainingIP(primaryIP6, ifaceStatuses); idx >= 0 {
				ipv4Addrs := extractIPsFromInterface(ifaceStatuses[idx], true, validatePrimaryIP)
				if len(ipv4Addrs) == 1 {
					primaryIP4 = ipv4Addrs[0]
				}
			}
		case primaryIP4 == "" && primaryIP6 == "":
			// Both empty, try first interface, if its the only interface present.
			if len(ifaceStatuses) == 1 {
				ipv4Addrs := extractIPsFromInterface(ifaceStatuses[0], true, validatePrimaryIP)
				ipv6Addrs := extractIPsFromInterface(ifaceStatuses[0], false, validatePrimaryIP)
				if len(ipv4Addrs) == 1 {
					primaryIP4 = ipv4Addrs[0]
				}
				if len(ipv6Addrs) == 1 {
					primaryIP6 = ipv6Addrs[0]
				}
			}
		}
	}

Comment (hpannem) [reply pkg/providers/vsphere/vmlifecycle/update_status.go]: > I don't understand what we would fail here. For whatever you're seeing no PrimaryIPv6, that means both VM Tools and CloudInit did not report a primary IPv6 so IDK if I wouldn't really conclude that is a Cloudinit bug.

I traced it. The issue is: cloud init is filtering link local ips in other path. My vm has two IPv6 addresses (one is link local which should be skipped).

```
test@new-testvm-3 [ ~ ]$ ip --json a | jq
[
  {
    "ifindex": 1,
    "ifname": "lo",
    "flags": [
      "LOOPBACK",
      "UP",
      "LOWER_UP"
    ],
    "mtu": 65536,
    "qdisc": "noqueue",
    "operstate": "UNKNOWN",
    "group": "default",
    "txqlen": 1000,
    "link_type": "loopback",
    "address": "00:00:00:00:00:00",
    "broadcast": "00:00:00:00:00:00",
    "addr_info": [
      {
        "family": "inet",
        "local": "127.0.0.1",
        "prefixlen": 8,
        "scope": "host",
        "label": "lo",
        "valid_life_time": 4294967295,
        "preferred_life_time": 4294967295
      },
      {
        "family": "inet6",
        "local": "::1",
        "prefixlen": 128,
        "scope": "host",
        "noprefixroute": true,
        "valid_life_time": 4294967295,
        "preferred_life_time": 4294967295
      }
    ]
  },
  {
    "ifindex": 2,
    "ifname": "eth0",
    "flags": [
      "BROADCAST",
      "MULTICAST",
      "UP",
      "LOWER_UP"
    ],
    "mtu": 1500,
    "qdisc": "mq",
    "operstate": "UP",
    "group": "default",
    "txqlen": 1000,
    "link_type": "ether",
    "address": "00:50:56:a6:3f:ec",
    "broadcast": "ff:ff:ff:ff:ff:ff",
    "altnames": [
      "eno1",
      "enp2s1",
      "ens33"
    ],
    "addr_info": [
      {
        "family": "inet",
        "local": "192.168.128.10",
        "prefixlen": 16,
        "broadcast": "192.168.255.255",
        "scope": "global",
        "label": "eth0",
        "valid_life_time": 4294967295,
        "preferred_life_time": 4294967295
      },
      {
        "family": "inet6",
        "local": "fc00:192:168:1::10a",
        "prefixlen": 64,
        "scope": "global",
        "valid_life_time": 4294967295,
        "preferred_life_time": 4294967295
      },
      {
        "family": "inet6",
        "local": "fe80::250:56ff:fea6:3fec",
        "prefixlen": 64,
        "scope": "link",
        "valid_life_time": 4294967295,
        "preferred_life_time": 4294967295
      }
    ]
  }
]
test@new-testvm-3 [ ~ ]$ vmtoolsd --cmd "info-get guestinfo.metadata"
H4sIAAAAAAAA/3SRXXLrIAyF370KNoADmBCs3cggX3tugjOgxtPdd2icv7Z5Q5+EdKQzLYUTnghEolUyFb6cZNfMqTCmQHKOICKNB9V5LYc4emnJoBz2PcrDaGPn3eCj6ZvjEvAo37RLxOuS/0MjBPFEORGXGnyH6voSAkOgM8uMIEY8FrrhGDOVQuVWJ4XuTaudb7XxrVY77e6ZMSgFujegnQcNoBXunN3ScQpn+9q8IveK/iHTip8WHmNa/ZqrP34O2ipOyGG6Ka1h2PSDUAr2CvYO0EE3AoWtqh6sUL5QLo+Pv7a+79232vetVn9z88yNUxYGZVU9xBN/J74Qy6t91ZdGiKppXhII06w4s1ySfPJyPl8cCM4f1HwBAAD//wEAAP//Mw0W60gCAAA=
test@new-testvm-3 [ ~ ]$ vmtoolsd --cmd "info-get guestinfo.metadata" | base64 -d | gunzip
hostname: new-testvm-3
instance-id: def70381-bdf8-4e2a-b59a-7f4d386b8d29
local-hostname: new-testvm-3
network:
  ethernets:
    eth0:
      accept-ra: false
      addresses:
      - 192.168.128.10/16
      - fc00:192:168:1::10a/64
      dhcp4: false
      dhcp6: false
      gateway4: 192.168.1.1
      gateway6: fc00:192:168:1::1
      match:
        macaddress: 00:50:56:a6:3f:ec
      nameservers:
        addresses:
        - 192.19.189.10
        - 192.19.189.20
        - 2604:b040::10
        - fc00:192:168:1::1
      set-name: eth0
  version: 2
wait-on-network:
  ipv6: true
test@new-testvm-3 [ ~ ]$ vmtoolsd --cmd "info-get guestinfo.local-ipv6"
No value found
test@new-testvm-3 [ ~ ]$ vmtoolsd --cmd "info-get guestinfo.local-ipv4"
192.168.128.10

Proof of the bug: (same code of cloudinit)

test@new-testvm-3 [ ~ ]$ sudo python3 -c "from cloudinit.netinfo import netdev_info; n=netdev_info(); print({k: len(v.get('ipv6',[])) for k,v in n.items()})"
{'lo': 1, 'eth0': 2}
test@new-testvm-3 [ ~ ]$ sudo python3 -c "
import cloudinit.sources.DataSourceVMware as ds
print(ds.get_default_ip_addrs())
"
('192.168.128.10', None)

hpannem: GOSC linuxprep for dual stack has the similar issue. It only reports one IP in dual stack environments.

controllers/virtualmachineservice/virtualmachineservice_controller.go
hpannem: I have removed this and depend on the k8s Service ipfamilies.

pkg/providers/vsphere/vmlifecycle/update_status.go
hpannem: Based on the above details, this fallback is needed. Even if cloudinit gets fixed, we may need this for GOSC. I added some checks. PTAL.

api/v1alpha6/virtualmachine_network_types.go
hpannem: I used requestedAddressFamilyMode.

webhooks/virtualmachineservice/validation/virtualmachineservice_validator_intg_test.go
```diff
@@ -99,6 +102,88 @@ func intgTestsValidateCreate() {
 		Entry("should work", true, "", nil),
 		Entry("should not work for invalid", false, "spec.type: Required value", nil),
 	)
+
+	Describe("CRD schema and CEL (ipFamilies and ipFamilyPolicy)", func() {
+		It("allows LoadBalancer with dual-stack ipFamilies and PreferDualStack policy", func() {
+			vm := ctx.vmService.DeepCopy()
```
bryanv: Kind of nit but `vm` isn't the best var name for this.

pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1119,6 +1168,45 @@ func updateGuestNetworkStatus(
 		}
 	}
 
+	// See cloud-init issue: https://github.com/canonical/cloud-init/issues/6851
+	// LinuxPrep does not report dual stack IPs. It reports only one primary IP.
```
bryanv, line 1172: OK. I think this is much safer but I do like what cloudinit does in that it favors the interface with the default route if there is one. For advertising something as the primary that does make the most sense. What's the feasibility of doing the same here?

pkg/providers/vsphere/network/netop_assignment_mode_internal_test.go
bryanv: Please do convert these

pkg/providers/vsphere/network/network_test.go
```diff
@@ -1145,6 +1306,212 @@ var _ = Describe("CreateAndWaitForNetworkInterfaces", Label(testlabels.VCSim), f
...
+						},
+					}
+				})
+
+				It("creates NetworkInterface CRs with correct policies for each interface", func() {
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring("network interface is not ready yet"))
+
+					By("verify each NetworkInterface CR that was created has the correct policy", func() {
+						// The function processes interfaces sequentially, so we verify each one that exists
```
bryanv: These are unit tests so that is just going to always be the case, so some of these branches below shouldn't be reachable. 

I have some work in flight to remove the polling, and to separate the device from the bootstrap so this isn't so annoying.

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -271,6 +271,14 @@ func (r *ReconcileVirtualMachineService) reconcileVMService(ctx *pkgctx.VirtualM
 		return err
 	}
 
+	// Load the Service from the API server so endpoint filtering uses the same
```
bryanv: I'm confused by this. This does not read it from the api-server directly but from the client-side controller-runtime client cache. There is no guarantee that you'll read what was just wrote, although there are some changes coming to make that possible.

I'm not sure if that is really a problem, and still just use the Service that `createOrUpdateService` returns. We'll eventually reconcile again if the Service changed. An empty IPFamilies needs to result in the current IPv4 behavior.

pkg/providers/vsphere/vmlifecycle/update_status.go
hpannem: All the time, this returns the one IP at least. Thats the default route interface IP. So, we use the other IP on this interface. So, we should be good. The only case is when no IP is reported. In that case too, we only select one if its the only interface.  I have tested pure IPv6 only interface with link local IPs. We still get guest.ipaddress field properly populated.
There are ip routes at the guestInfo VMOMI level. We may be able to decipher it, if the info is populated. I think we may not need that.
```
      ipRoutes:
      - gateway:
          address: 192.168.1.1
          device: "0"
        networkAddress: 0.0.0.0/0
      - gateway:
          device: "0"
        networkAddress: 192.168.0.0/16
      - gateway:
          device: "0"
        networkAddress: fc00:192:168:1::/64
      - gateway:
          address: fc00:192:168:1::1
          device: "0"
        networkAddress: ::/0
      - gateway:
          device: "0"
        networkAddress: fc00:192:168:1::/128
      - gateway:
          device: "0"
        networkAddress: fc00:192:168:1::10f/128

File: pkg/providers/vsphere/network/network_test.go (hunk 19 above)
Comment (hpannem) [reply pkg/providers/vsphere/network/network_test.go]: ok

File: controllers/virtualmachineservice/virtualmachineservice_controller.go (hunk 20 above)
Comment (hpannem) [reply controllers/virtualmachineservice/virtualmachineservice_controller.go]: >>> An empty IPFamilies needs to result in the current IPv4 behavior.
I think we should rely on k8s service behavior. K8s Service will fill it (always contains value) based on the cluster IP family right? For e.g. if IPv6 cluster, we want to create IPv6 endpoints right? So, we dont need defaulting in empty case I believe.

File: api/v1alpha6/virtualmachine_network_types.go (hunk 11 above)
Comment (hpannem) [reply api/v1alpha6/virtualmachine_network_types.go]: Modified agreed upon ipamModes: [IPv4, IPv6]

File: api/v1alpha6/virtualmachine_network_types.go [hunk 21]
```diff
@@ -223,6 +225,21 @@ type VirtualMachineNetworkInterfaceSpec struct {
 	// The admission webhook rejects keys that duplicate a first-class field
 	// above (e.g. "ctxPerDev" is rejected because VMXNet3.CtxPerDev exists).
 	AdvancedProperties []vmopv1common.KeyValuePair `json:"advancedProperties,omitempty"`
+
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=2
+
+	// IPAMModes lists which Kubernetes IP families (IPv4 and/or IPv6) should be
+	// allocated for this interface. Values use corev1.IPFamily ("IPv4", "IPv6").
```
bryanv: I think it is important to explicilty call out this what is allocated by the network provider. I don't think we need to either call out that it is `corev1.IPFamily` here but just what are the values.

api/v1alpha6/virtualmachine_network_types.go
```diff
@@ -223,6 +225,21 @@ type VirtualMachineNetworkInterfaceSpec struct {
...
+	// Each family appears at most once, duplicate values are rejected by the
+	// API server, and order does not change meaning
+	// — [IPv4, IPv6] and [IPv6, IPv4] are the same dual-stack request.
+	//
+	// When empty, the network interface provider's default applies (for NetOP,
```
bryanv: I don't think our doc here should mention NetOP. That isn't something most users are aware of (to them, it is VDS) and not a CR they really every directly interact with (I'm not sure if our RBAC for them includes GET). Just call out the default behavior when unset.

controllers/virtualmachineservice/virtualmachineservice_controller_unit_test.go
```diff
@@ -642,6 +698,230 @@ func unitTestsReconcile() {
...
+					allAddresses := collectAllEndpointIPs(endpoints)
+					Expect(allAddresses).To(ContainElements("192.168.1.1", "192.168.1.2", "2001:db8::1", "2001:db8::2"))
+				})
+			})
+
+			Context("Endpoint filtering based on Service IPFamilyPolicy", func() {
+				var dualStackVM *vmopv1.VirtualMachine
+				var ipv4VM *vmopv1.VirtualMachine
+				var ipv6VM *vmopv1.VirtualMachine
+				var service *corev1.Service
```
bryanv: Does `service` need to be at this scope?

pkg/providers/vsphere/network/export_test.go
```diff
@@ -0,0 +1,11 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package network
+
+// Test-only exports for the network_test package (see Go export_test pattern).
+
+var EffectiveNetOPIPv4AssignmentModeForTest = EffectiveNetOPIPv4AssignmentMode
```
bryanv: Are these used?

pkg/providers/vsphere/vmlifecycle/update_status.go
bryanv: I don't think we should be assuming that the default for both will always be the same interface. File a ticket so we can track this later.

bryanv (approved):
Please take care of a few comments before mering. Thanks.

api/v1alpha6/virtualmachine_network_types.go
hpannem: Done.

pkg/providers/vsphere/network/export_test.go
hpannem: Nope. Stale file.

pkg/providers/vsphere/vmlifecycle/update_status.go
hpannem: Yes. Was thinking always dual stack. Tracking as a ticket.

PR #1558, opened 2026-04-16
Reconcile VMLocation when it's moved to different RP and set appropriate conditions

anwithaun-1:

Problem:
Currently, the VM Operator does not set appropriate conditions or halt reconciliation when a VM is moved out of its designated resource pool. This leads to two major issues:
Permission Creep: DevOps admin privileges are inadvertently extended to reconfigure VMs that have drifted into different inventory locations.
Lack of Visibility: There is no "stop-gap" or clear status indicator telling administrators that the VM is in an orphaned state, leading to unpredictable reconciliation behavior.

Solution:
-  The controller now searches the broader inventory to see if the VM exists. A VM is flagged as "Orphaned" if it is located outside its Namespace RP.
- Child ResourcePool is considered as valid locations. 
- Once an orphan is detected, the controller finalizes the status update and immediately exits the reconciliation loop with NoRequeErr.
- This prevents the system from using elevated privileges to reconfigure a VM that is no longer in a managed/expected location.
- Applies a clear condition to the VM object notifying the User/Administrator that the VM is in an unexpected location.
- Property watcher added to watch VM's resourcePool or Parent property change. This triggers the reconileLocation to validate the VM's location. 
- Once the VM is moved back to the expected location, mismatch condition set on VM will be removed.

State Conflict: Two instances of the same workload existing simultaneously in different locations.

Fixes #
VMSVC-3539
VMSVC-3541

```
Output:
Last Transition Time:  2026-05-11T05:43:40Z
    Message:               VM is in an unauthorized Resource Pool. Move the VM back to the Resource Pool hierarchy for namespace 'test-ns' to resume reconciliation.
    Reason:                LocationMismatch
    Status:                False
    Type:                  VirtualMachineConditionInAuthorizedLocation

```

```release-note

```

### DIFF
--- a/api/v1alpha6/virtualmachine_types.go
+++ b/api/v1alpha6/virtualmachine_types.go
@@ -115,6 +115,10 @@ const (
 	// configuration is synced to the desired state, determined by whether or
 	// not the guest is reporting the expected IP address(es).
 	VirtualMachineGuestNetworkConfigSynced = "VirtualMachineGuestNetworkConfigSynced"
+
+	// VirtualMachineConditionLocationMismatch indicates that the VM is not in the expected location in the vCenter inventory.
+	// This condition is set when a VM is detected to be in a different folder or resource pool than expected based on its namespace.
+	VirtualMachineConditionLocationMismatch = "VirtualMachineConditionLocationMismatch"
 )
 
 const (
--- a/api/v1alpha6/virtualmachine_types.go
+++ b/api/v1alpha6/virtualmachine_types.go
@@ -116,9 +116,9 @@ const (
 	// not the guest is reporting the expected IP address(es).
 	VirtualMachineGuestNetworkConfigSynced = "VirtualMachineGuestNetworkConfigSynced"
 
-	// VirtualMachineConditionLocationMismatch indicates that the VM is not in the expected location in the vCenter inventory.
-	// This condition is set when a VM is detected to be in a different folder or resource pool than expected based on its namespace.
-	VirtualMachineConditionLocationMismatch = "VirtualMachineConditionLocationMismatch"
+	// VirtualMachineConditionInAuthorizedLocation indicates that the VM is in the expected location in the vCenter inventory.
+	// This condition is set when a VM is detected to be in the incorrect folder or resource pool based on its namespace.
+	VirtualMachineConditionInAuthorizedLocation = "VirtualMachineConditionInAuthorizedLocation"
 )
 
 const (
--- a/api/v1alpha6/virtualmachine_types.go
+++ b/api/v1alpha6/virtualmachine_types.go
@@ -116,9 +116,9 @@ const (
 	// not the guest is reporting the expected IP address(es).
 	VirtualMachineGuestNetworkConfigSynced = "VirtualMachineGuestNetworkConfigSynced"
 
-	// VirtualMachineConditionInAuthorizedLocation indicates that the VM is in the expected location in the vCenter inventory.
+	// VirtualMachineInAuthorizedLocation indicates that the VM is in the expected location in the vCenter inventory.
 	// This condition is set when a VM is detected to be in the incorrect folder or resource pool based on its namespace.
-	VirtualMachineConditionInAuthorizedLocation = "VirtualMachineConditionInAuthorizedLocation"
+	VirtualMachineInAuthorizedLocation = "VirtualMachineInAuthorizedLocation"
 )
 
 const (
--- a/api/v1alpha6/virtualmachine_types.go
+++ b/api/v1alpha6/virtualmachine_types.go
@@ -116,9 +116,9 @@ const (
 	// not the guest is reporting the expected IP address(es).
 	VirtualMachineGuestNetworkConfigSynced = "VirtualMachineGuestNetworkConfigSynced"
 
-	// VirtualMachineInAuthorizedLocation indicates that the VM is in the expected location in the vCenter inventory.
+	// VirtualMachineInValidLocation indicates that the VM is in the expected location in the vCenter inventory.
 	// This condition is set when a VM is detected to be in the incorrect folder or resource pool based on its namespace.
-	VirtualMachineInAuthorizedLocation = "VirtualMachineInAuthorizedLocation"
+	VirtualMachineInValidLocation = "VirtualMachineInValidLocation"
 )
 
 const (
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -409,12 +409,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	return ctrl.Result{RequeueAfter: requeueDelay(vmCtx, err)}, nil
 }
 
-// Inject the halt signal into the context
+// Inject the halt signal into the context.
 func withHalt(ctx context.Context) context.Context {
 	return context.WithValue(ctx, haltReconcileKey{}, true)
 }
 
-// Check if we should stop
+// Check if we should stop.
 func isHalted(ctx context.Context) bool {
 	val := ctx.Value(haltReconcileKey{})
 	return val != nil && val.(bool)
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -62,6 +62,8 @@ const (
 // managers.
 var SkipNameValidation *bool
 
+type haltReconcileKey struct{}
+
 // AddToManager adds this package's controller to the provided manager.
 func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) error {
 	var (
@@ -392,6 +394,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		return pkgerr.ResultFromError(r.ReconcileDelete(vmCtx))
 	}
 	if err = r.ReconcileLocation(vmCtx); err != nil {
+		if isHalted(vmCtx.Context) {
+			return ctrl.Result{Requeue: false}, nil
+		}
 		return ctrl.Result{}, err
 	}
 
@@ -404,32 +409,41 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	return ctrl.Result{RequeueAfter: requeueDelay(vmCtx, err)}, nil
 }
 
+// Inject the halt signal into the context
+func withHalt(ctx context.Context) context.Context {
+	return context.WithValue(ctx, haltReconcileKey{}, true)
+}
+
+// Check if we should stop
+func isHalted(ctx context.Context) bool {
+	val := ctx.Value(haltReconcileKey{})
+	return val != nil && val.(bool)
+}
+
 func (r *Reconciler) ReconcileLocation(pkgctx *pkgctx.VirtualMachineContext) error {
 	if pkgctx.VM.Status.UniqueID == "" {
-		// This is a new VM creation request.
 		// Skip the check because the VM doesn't exist in vCenter yet.
 		return nil
 	}
 
-	// 1. Fetch the VM properties from vCenter
-	// We declare 'mvm' here so it is no longer undefined
-	actualNs, err := r.VMProvider.GetVMLocation(pkgctx.Context, pkgctx.VM.Name)
+	// Fetch the VM location from vCenter
+	currentLocation, err := r.VMProvider.GetVMLocation(pkgctx.Context, pkgctx.VM.Name)
 	if err != nil {
-		// If the VM is missing, we might want to handle it differently
 		return err
 	}
 
 	// 2. Compare Actual (vCenter) vs Expected (K8s)
-	if actualNs != pkgctx.VM.Namespace {
+	if currentLocation != pkgctx.VM.Namespace {
 		// Use the MarkFalse logic
 		pkgcond.MarkFalse(
-			pkgctx.VM, vmopv1.VirtualMachineConditionCreated, "NamespaceMismatch", "VM is physically in vCenter namespace '%s', but K8s expects '%s'",
-			actualNs, pkgctx.VM.Namespace)
+			pkgctx.VM, vmopv1.VirtualMachineConditionCreated, "LocationMismatch", "VM is moved to different Vcenter location , expected location is: '%s'", pkgctx.VM.Namespace)
 
 		if err := r.Status().Update(pkgctx.Context, pkgctx.VM); err != nil {
-			return fmt.Errorf("failed to persist namespace mismatch condition: %w", err)
+			return fmt.Errorf("failed to persist location mismatch condition: %w", err)
 		}
-		return fmt.Errorf("reconciliation halted: VM location mismatch")
+
+		pkgctx.Context = withHalt(pkgctx.Context)
+		return fmt.Errorf("reconciliation stopped : VM location mismatch")
 	}
 	return nil
 }
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -62,8 +62,6 @@ const (
 // managers.
 var SkipNameValidation *bool
 
-type haltReconcileKey struct{}
-
 // AddToManager adds this package's controller to the provided manager.
 func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) error {
 	var (
@@ -393,12 +391,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	if !vm.DeletionTimestamp.IsZero() {
 		return pkgerr.ResultFromError(r.ReconcileDelete(vmCtx))
 	}
-	if err = r.ReconcileLocation(vmCtx); err != nil {
-		if isHalted(vmCtx.Context) {
-			return ctrl.Result{Requeue: false}, nil
-		}
-		return ctrl.Result{}, err
-	}
 
 	err = r.ReconcileNormal(vmCtx)
 	if err != nil && !ignoredCreateErr(err) {
@@ -407,45 +399,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 
 	// Requeue after N amount of time according to the state of the VM.
 	return ctrl.Result{RequeueAfter: requeueDelay(vmCtx, err)}, nil
-}
-
-// Inject the halt signal into the context.
-func withHalt(ctx context.Context) context.Context {
-	return context.WithValue(ctx, haltReconcileKey{}, true)
-}
-
-// Check if we should stop.
-func isHalted(ctx context.Context) bool {
-	val := ctx.Value(haltReconcileKey{})
-	return val != nil && val.(bool)
-}
-
-func (r *Reconciler) ReconcileLocation(pkgctx *pkgctx.VirtualMachineContext) error {
-	if pkgctx.VM.Status.UniqueID == "" {
-		// Skip the check because the VM doesn't exist in vCenter yet.
-		return nil
-	}
-
-	// Fetch the VM location from vCenter
-	currentLocation, err := r.VMProvider.GetVMLocation(pkgctx.Context, pkgctx.VM.Name)
-	if err != nil {
-		return err
-	}
-
-	// 2. Compare Actual (vCenter) vs Expected (K8s)
-	if currentLocation != pkgctx.VM.Namespace {
-		// Use the MarkFalse logic
-		pkgcond.MarkFalse(
-			pkgctx.VM, vmopv1.VirtualMachineConditionCreated, "LocationMismatch", "VM is moved to different Vcenter location , expected location is: '%s'", pkgctx.VM.Namespace)
-
-		if err := r.Status().Update(pkgctx.Context, pkgctx.VM); err != nil {
-			return fmt.Errorf("failed to persist location mismatch condition: %w", err)
-		}
-
-		pkgctx.Context = withHalt(pkgctx.Context)
-		return fmt.Errorf("reconciliation stopped : VM location mismatch")
-	}
-	return nil
 }
 
 // Determine if we should request a non-zero requeue delay in order to trigger a
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -391,10 +391,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	if !vm.DeletionTimestamp.IsZero() {
 		return pkgerr.ResultFromError(r.ReconcileDelete(vmCtx))
 	}
-
-	err = r.VMProvider.ListNsAndNameoftheVM(ctx, vm.Namespace)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to list all the VM and validate namespaces")
+	if err = r.ReconcileLocation(vmCtx); err != nil {
+		return ctrl.Result{}, err
 	}
 
 	err = r.ReconcileNormal(vmCtx)
@@ -404,6 +402,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 
 	// Requeue after N amount of time according to the state of the VM.
 	return ctrl.Result{RequeueAfter: requeueDelay(vmCtx, err)}, nil
+}
+
+func (r *Reconciler) ReconcileLocation(pkgctx *pkgctx.VirtualMachineContext) error {
+	if pkgctx.VM.Status.UniqueID == "" {
+		// This is a new VM creation request.
+		// Skip the check because the VM doesn't exist in vCenter yet.
+		return nil
+	}
+
+	// 1. Fetch the VM properties from vCenter
+	// We declare 'mvm' here so it is no longer undefined
+	actualNs, err := r.VMProvider.GetVMLocation(pkgctx.Context, pkgctx.VM.Name)
+	if err != nil {
+		// If the VM is missing, we might want to handle it differently
+		return err
+	}
+
+	// 2. Compare Actual (vCenter) vs Expected (K8s)
+	if actualNs != pkgctx.VM.Namespace {
+		// Use the MarkFalse logic
+		pkgcond.MarkFalse(
+			pkgctx.VM, vmopv1.VirtualMachineConditionCreated, "NamespaceMismatch", "VM is physically in vCenter namespace '%s', but K8s expects '%s'",
+			actualNs, pkgctx.VM.Namespace)
+
+		if err := r.Status().Update(pkgctx.Context, pkgctx.VM); err != nil {
+			return fmt.Errorf("failed to persist namespace mismatch condition: %w", err)
+		}
+		return fmt.Errorf("reconciliation halted: VM location mismatch")
+	}
+	return nil
 }
 
 // Determine if we should request a non-zero requeue delay in order to trigger a
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -392,6 +392,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		return pkgerr.ResultFromError(r.ReconcileDelete(vmCtx))
 	}
 
+	err = r.VMProvider.ListNsAndNameoftheVM(ctx, vm.Namespace)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list all the VM and validate namespaces")
+	}
+
 	err = r.ReconcileNormal(vmCtx)
 	if err != nil && !ignoredCreateErr(err) {
 		return pkgerr.ResultFromError(err)
--- a/pkg/providers/vm_provider_interface.go
+++ b/pkg/providers/vm_provider_interface.go
@@ -91,5 +91,4 @@ type VirtualMachineProviderInterface interface {
 	GetSnapshotSize(ctx context.Context, vmSnapshotName string, vm *vmopv1.VirtualMachine) (int64, error)
 	// SyncVMSnapshotTreeStatus syncs the VM's current and root snapshots status.
 	SyncVMSnapshotTreeStatus(ctx context.Context, vm *vmopv1.VirtualMachine) error
-	GetVMLocation(ctx context.Context, vm *vmopv1.VirtualMachine) (string, error)
 }
--- a/pkg/providers/vm_provider_interface.go
+++ b/pkg/providers/vm_provider_interface.go
@@ -91,4 +91,6 @@ type VirtualMachineProviderInterface interface {
 	GetSnapshotSize(ctx context.Context, vmSnapshotName string, vm *vmopv1.VirtualMachine) (int64, error)
 	// SyncVMSnapshotTreeStatus syncs the VM's current and root snapshots status.
 	SyncVMSnapshotTreeStatus(ctx context.Context, vm *vmopv1.VirtualMachine) error
+	ListVMs(ctx context.Context) ([]*object.VirtualMachine, error)
+	ListNsAndNameoftheVM(ctx context.Context, expectedNs string) error
 }
--- a/pkg/providers/vm_provider_interface.go
+++ b/pkg/providers/vm_provider_interface.go
@@ -91,6 +91,5 @@ type VirtualMachineProviderInterface interface {
 	GetSnapshotSize(ctx context.Context, vmSnapshotName string, vm *vmopv1.VirtualMachine) (int64, error)
 	// SyncVMSnapshotTreeStatus syncs the VM's current and root snapshots status.
 	SyncVMSnapshotTreeStatus(ctx context.Context, vm *vmopv1.VirtualMachine) error
-	ListVMs(ctx context.Context) ([]*object.VirtualMachine, error)
-	ListNsAndNameoftheVM(ctx context.Context, expectedNs string) error
+	GetVMLocation(ctx context.Context, vmName string) (string, error)
 }
--- a/pkg/providers/vm_provider_interface.go
+++ b/pkg/providers/vm_provider_interface.go
@@ -91,5 +91,5 @@ type VirtualMachineProviderInterface interface {
 	GetSnapshotSize(ctx context.Context, vmSnapshotName string, vm *vmopv1.VirtualMachine) (int64, error)
 	// SyncVMSnapshotTreeStatus syncs the VM's current and root snapshots status.
 	SyncVMSnapshotTreeStatus(ctx context.Context, vm *vmopv1.VirtualMachine) error
-	GetVMLocation(ctx context.Context, vmName string) (string, error)
+	GetVMLocation(ctx context.Context, vm *vmopv1.VirtualMachine) (string, error)
 }
--- a/pkg/providers/vsphere/vcenter/resourcepool.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool.go
@@ -11,6 +11,7 @@ import (
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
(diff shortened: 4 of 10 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -149,6 +150,44 @@ func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachineAsync(
...
+func (vs *vSphereVMProvider) GetVMLocation(ctx context.Context, vmName string) (string, error) {
+	client, err := vs.VSphereClient(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	vimClient := client.VimClient()
+	m := view.NewManager(vimClient)
+	// Create view specifically for VMs
+	v, err := m.CreateContainerView(ctx, vimClient.ServiceContent.RootFolder, []string{"VirtualMachine"}, true)
```
aakashchan: We don't need ContainerView for this use case, usually its bad for vCenter performance in large scale environments.

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -401,6 +409,45 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	return ctrl.Result{RequeueAfter: requeueDelay(vmCtx, err)}, nil
 }
 
+// Inject the halt signal into the context.
+func withHalt(ctx context.Context) context.Context {
+	return context.WithValue(ctx, haltReconcileKey{}, true)
+}
+
+// Check if we should stop.
+func isHalted(ctx context.Context) bool {
```
aakashchan: Do we need this? Can we just add pause annotation to stop reconcilation?

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -391,6 +393,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 	if !vm.DeletionTimestamp.IsZero() {
 		return pkgerr.ResultFromError(r.ReconcileDelete(vmCtx))
 	}
+	if err = r.ReconcileLocation(vmCtx); err != nil {
```
aakashchan: I would suggest making changes to ReconcileNormal for paused annotation.
https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/providers/vsphere/vmprovider_utils.go#L913

Make use of this util method, probably I good idea is to introduce a new persona like, "vmop" indicating operator paused and can only unpaused by VI Admin. 
@faisalabujabal can take better call here.

pkg/providers/vm_provider_interface.go
```diff
@@ -91,4 +91,5 @@ type VirtualMachineProviderInterface interface {
 	GetSnapshotSize(ctx context.Context, vmSnapshotName string, vm *vmopv1.VirtualMachine) (int64, error)
 	// SyncVMSnapshotTreeStatus syncs the VM's current and root snapshots status.
 	SyncVMSnapshotTreeStatus(ctx context.Context, vm *vmopv1.VirtualMachine) error
+	GetVMLocation(ctx context.Context, vm *vmopv1.VirtualMachine) (string, error)
```
bryanv: There isn't a reason for this to be an interface method.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1143,30 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+	currentLocation, err := vs.GetVMLocation(vmCtx.Context, vmCtx.VM)
+	if err != nil {
+		return err
+	}
+
+	// 2. Compare Actual (vCenter) vs Expected (K8s)
+	if currentLocation != vmCtx.VM.Namespace {
+		// Use the MarkFalse logic
+		pkgcnd.MarkFalse(
+			vmCtx.VM, vmopv1.VirtualMachineConditionCreated, "LocationMismatch", "VM is moved to different Vcenter location , expected location is: '%s'", vmCtx.VM.Namespace)
```
bryanv: I don't think we should overload this condition for this. It is useful on its own to determine when created the VM.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1143,30 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
 }
 
+func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContext) error {
+	if vmCtx.VM.Status.UniqueID == "" {
+		// Skip the check because the VM doesn't exist in vCenter yet.
+		return nil
+	}
+
+	// Fetch the VM location from vCenter
+	currentLocation, err := vs.GetVMLocation(vmCtx.Context, vmCtx.VM)
```
bryanv: This isn't right - VKS VMs are placed in a child ResourcePool.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -149,6 +149,35 @@ func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachineAsync(
...
+		vm,
+	)
+	ctx = vmCtx.Context
+
+	client, err := vs.getVcClient(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	vcVM, err := vs.getVM(vmCtx, client, true)
```
bryanv: At the time you're calling this we already have gotten the VC VM.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Introduced new condition (VirtualMachineConditionLocationMismatch) for this scenario

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Could you pls elaborate this.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1105,9 +1119,50 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 		}
 	}
 
+	if err := vs.reconcileLocation(vmCtx, vcVM, vcClient); err != nil {
```
hpannem: Shouldn't the reconcileLocation much earlier than other reconciles? If it is not in supervisor location, we should not touch anything for this VM right?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -149,6 +149,20 @@ func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachineAsync(
 	return vs.createOrUpdateVirtualMachine(ctx, vm, true)
 }
 
+func (vs *vSphereVMProvider) GetVMLocation(vmCtx pkgctx.VirtualMachineContext, vm *object.VirtualMachine, vcClient *vcclient.Client) (string, error) {
+
+	var o mo.VirtualMachine
+	err := vm.Properties(vmCtx, vm.Reference(), []string{"resourcePool"}, &o)
```
hpannem: isn't this fetched in 

	if err := vcVM.Properties(
		vmCtx,
		vcVM.Reference(),
		VMUpdatePropertiesSelector,
		&vmCtx.MoVM); err != nil {

		return fmt.Errorf("failed to fetch vm properties: %w", err)
	}

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -149,6 +149,20 @@ func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachineAsync(
...
+	if err != nil {
+		return "", err
+	}
+	if o.ResourcePool != nil {
+		rp := object.NewResourcePool(vcClient.VimClient(), *o.ResourcePool)
+		return rp.ObjectName(vmCtx)
```
hpannem: If I move under similar named resource pool and what will happen at that time?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1105,9 +1119,50 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+		return err
+	}
+	// 2. Compare Actual (vCenter) vs Expected (K8s)
+
+	resourcePolicy, _ := GetVMSetResourcePolicy(vmCtx, vs.k8sClient) //
+
+	isValidLocation := false
+	if currentLocation == vmCtx.VM.Namespace {
+		isValidLocation = true //
+	}
```
hpannem, line 1144: else if resourcePolicy != nil ?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1105,9 +1119,50 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+			(childFolder != "" && currentLocation == childFolder) {
+			isValidLocation = true
+		}
+	}
+	if !isValidLocation {
+		// Use the MarkFalse logic
+		pkgcnd.MarkFalse(
+			vmCtx.VM, vmopv1.VirtualMachineConditionPlacementReady, "LocationMismatch", "VM is moved to different Vcenter location , expected location is: '%s'", vmCtx.VM.Namespace)
+
+		vmCtx.VM.Annotations[vmopv1.PauseAnnotation] = "true"
```
hpannem: what happens when the vm is moved back to the proper location? Someone has to edit the CR and unpause?

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: @bryanv  Updated the changes to validate the vm location in the namespace RP and it's child resource pools and raise the condition when the vm is not under these locations. pls check

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: We can move the reconcileLocation to the beginning before calling other reconcilation methods, but after reconcileStatus as the reconcileLocation is updating the status and that needs to be reconciled as part of reconcileStatus.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Fixed

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Currently the method is considering the VM's Namespace RP, and child resource pools as the valid location for the VM movement. Apart from these locations, if the VM is moved anywhere else then it is considered as invalid location.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Updated the changes wrt this. Please review again.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: @faisalabujabal @hpannem wrt the changes, PauseAnnotation is allowed to be removed by any user(devops or VI admin) , by then if the orphaned VM was moved back to the right location, the Mismatch Condition will be cleared , if not "PauseAnnotation" will be set back again on the VM which stops the reconcilation. Thoughts?

anwithaun-1 (commented):
Please take a look at the addressed comments.

pkg/providers/vsphere/vmprovider_vm.go
bryanv: This shouldn't set the pause annotation - the reconcile just needs to early return

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
anwithaun-1: Not using PauseAnnotation. Instead returning fast with NoRequeErr

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Done.

anwithaun-1 (commented):
Resolved all the comments. pls review again.

services/vm-watcher/vm_watcher_service.go
```diff
@@ -256,7 +256,8 @@ func (s Service) waitForChanges(ctx context.Context) error {
 	w, err := watcher.Start(
 		ctx,
 		vcClient.VimClient(),
-		nil,
+		[]string{"resourcePool",
```
bryanv: When set, the default DefaultWatchedPropertyPaths() is no longer used so that will break the watcher functionality that we need.

Also, the VM managed object doesn't have a parent field. What that one trying to watch?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,66 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+	}
+
+	// 4. Handle Mismatch
+	if !isValid {
+
+		pkgcnd.MarkFalse(
+			vmCtx.VM,
+			vmopv1.VirtualMachineConditionInAuthorizedLocation,
+			"LocationMismatch",
+			"VM is in an unauthorized Resource Pool. Move the VM back to the Resource Pool hierarchy for namespace '%s' to resume reconciliation.",
```
bryanv: NS can't have spaces in the name so I don't think quotes are needed

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,66 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+			vmCtx.VM.Namespace,
+		)
+
+		return pkgerr.NoRequeueError{Message: fmt.Errorf(
```
bryanv: Can't this use Sprintf instead of calling `Error()` on a error that was just built?

Doesn't the context of where this is use include the VM name?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,66 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+		return nil
+	}
+	// 3. Validate Ancestry
+	// Check if the VM is in the Root RP or a Child RP (common for VKS)
+	isValid := false
+	// Case A: VM is directly in the Namespace Root RP
+	if currentRPMoRef.Value == expectedRootRPMoID {
+		isValid = true
+	} else {
+		poolObj := object.NewResourcePool(vcClient.VimClient(), currentRPMoRef)
```
bryanv: We have some RP related helpers in vsphere/vcenter/resourcepool.go that this could be moved to.

services/vm-watcher/vm_watcher_service.go
anwithaun-1: @bryanv Moved this property watch to DefaultWatchedPropertyPaths().

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: We can use sprintf also in this case. I used Errorf as this has been used at many places with NoRequeue.

anwithaun-1:
@aruneshpa @bryanv @hpannem , pls review

api/v1alpha6/virtualmachine_types.go
```diff
@@ -115,6 +115,10 @@ const (
 	// configuration is synced to the desired state, determined by whether or
 	// not the guest is reporting the expected IP address(es).
 	VirtualMachineGuestNetworkConfigSynced = "VirtualMachineGuestNetworkConfigSynced"
+
+	// VirtualMachineConditionInAuthorizedLocation indicates that the VM is in the expected location in the vCenter inventory.
+	// This condition is set when a VM is detected to be in the incorrect folder or resource pool based on its namespace.
+	VirtualMachineConditionInAuthorizedLocation = "VirtualMachineConditionInAuthorizedLocation"
```
hpannem: we can remove the condition in the "name" on var name as well as the in the value

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+			)
+
+			return pkgerr.NoRequeueError{Message: fmt.Sprintf(
+				"reconciliation stopped for the VM %s because it is moved to unauthorized Resource Pool. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s",
+				vmCtx.VM.Name, expectedRootRPMoID, vmCtx.MoVM.ResourcePool.Value)}
+		}
+		return err
+	}
+
+	pkgcnd.Delete(vmCtx.VM, vmopv1.VirtualMachineConditionInAuthorizedLocation)
```
hpannem: do we need to set to true or delete?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+		vmCtx.VM.Labels[corev1.LabelTopologyZone],
+		vmCtx.VM.Namespace,
+	)
+	if err != nil {
+		logger.Error(err, "failed to get expected namespace resource pool")
+		return nil
+	}
+
+	// 3. Validate Ancestry
+	// Check if the VM is in the Root RP or a Child RP (common for VKS)
```
hpannem: what happens if the vm is in correct namespace however wrong child RP? Is this not under the PR's scope?

anwithaun-1: what is considered as wrong Child RP?
If the VM is in correct namespace and moved to any of the child RP of that namespace, then it's considered as a valid location as of now

hpannem: ok. However we need to confirm this with Bryan or Andrew.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+			)
+
+			return pkgerr.NoRequeueError{Message: fmt.Sprintf(
+				"reconciliation stopped for the VM %s because it is moved to unauthorized Resource Pool. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s",
+				vmCtx.VM.Name, expectedRootRPMoID, vmCtx.MoVM.ResourcePool.Value)}
+		}
+		return err
+	}
+
+	pkgcnd.Delete(vmCtx.VM, vmopv1.VirtualMachineInAuthorizedLocation)
```
hpannem: So, absence of this condition has two scenarios. 
1. the reconcile has not yet happened.
2. The reconcile happened and its in correct location. 

Would it be better to distinguish these two scenarios?
True: In correct location.
False: not in correct location.
Absent: Not yet run or unable to determine.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -993,6 +993,13 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 		return fmt.Errorf("failed to fetch vm properties: %w", err)
 	}
 
+	if err := vs.reconcileLocation(vmCtx, vcClient); err != nil {
```
aruneshpa, line 1000: Hmm.  The current code follows a convention of ordered bullets to ensure that adding or removing methods here are intentional.  That is by design.  Can we please follow that?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 	return reconcileErr
 }
 
+func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContext, vcClient *vcclient.Client) error {
+	logger := pkglog.FromContextOrDefault(vmCtx)
+	// 1. Get the VM's current Resource Pool from properties already fetched
+	if vmCtx.MoVM.ResourcePool == nil {
+		return fmt.Errorf("VM %s has no resource pool assigned", vmCtx.VM.Name)
```
aruneshpa: VM's are assigned _to_ resource pools.  I would update this as: "VM is not assigned to any resource pools" or something alike.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+	}
+
+	// 2. Get the Namespace's Root Resource Pool MoRef
+	// We use the topology utility to find the RP mapped to this K8s Namespace
```
aruneshpa: This comment is not needed.  The code itself is sufficient to deduce this. 

Here and elsewhere, a general feedback is to not have code comments that exactly describe _what_ the code is doing.  Ideally, the code should be easy to follow.  You should describe the "how" and / or the "why".

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+	_, expectedRootRPMoID, err := topology.GetNamespaceFolderAndRPMoID(
+		vmCtx,
+		vs.k8sClient,
+		vmCtx.VM.Labels[corev1.LabelTopologyZone],
+		vmCtx.VM.Namespace,
+	)
+	if err != nil {
+		logger.Error(err, "failed to get expected namespace resource pool")
+		return nil
```
aruneshpa: Hmm.  Should this error be ignored?  If we are can't be certain that we are in the RP that we should be in, it is okay to error and requeue IMO.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Not "common for VKS".  More like that's exactly what happens for VKS.  The word "common" makes it feel like that behavior is frequent but optional.

aruneshpa: Hmm.  I am with Hemanth on this.  The only valid use-case for child RP is VKS.  But if the VM is not in the VKS child RP, then we need to report it as an error / invalid location.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+	// Check if the VM is in the Root RP or a Child RP (common for VKS)
+	isValid, err := vcenter.IsVMInValidResourcePool(
+		vmCtx,
+		vcClient.VimClient(),
+		vmCtx.MoVM.ResourcePool.Value,
+		expectedRootRPMoID,
+	)
+
+	// 4. Handle Mismatch
+	if err != nil || !isValid {
```
aruneshpa: Why not break this down in two conditions:
```golang
if err != nil {

}

if !valid {

}
```
?

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: Correct.  The way the condition is named right now, it must be set to True during every reconcile if the VM is in a valid location.  

If it is only supposed to be set when the VM is in invalid location, then we can call it "VMInInvalidLocation" or something but that violates the principal that the condition types must have positive polarity.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: @bryanv any thoughts on this?

pkg/util/vsphere/watcher/watcher.go
```diff
@@ -75,6 +75,7 @@ func DefaultWatchedPropertyPaths() []string {
 		"summary.runtime.connectionState",
 		"summary.runtime.host",
 		"summary.runtime.powerState",
+		"resourcePool",
```
aruneshpa, line 78: Should we also watch the folder ("parent") here?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 	return reconcileErr
 }
 
+func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContext, vcClient *vcclient.Client) error {
+	logger := pkglog.FromContextOrDefault(vmCtx)
+	// 1. Get the VM's current Resource Pool from properties already fetched
+	if vmCtx.MoVM.ResourcePool == nil {
```
aruneshpa, line 1124: The PR title states both RP and Folder, but the implementation is only for the RP.  Is this intentional?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1115,57 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+			pkgcnd.MarkFalse(
+				vmCtx.VM,
+				vmopv1.VirtualMachineInAuthorizedLocation,
+				"LocationMismatch",
+				"VM is in an unauthorized Resource Pool. Move the VM back to the Resource Pool hierarchy for namespace %s to resume reconciliation.",
+				vmCtx.VM.Namespace,
+			)
+
+			return pkgerr.NoRequeueError{Message: fmt.Sprintf(
+				"reconciliation stopped for the VM %s because it is moved to unauthorized Resource Pool. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s",
```
aruneshpa: "Unauthorized" feels weird here.  It is totally possible that the user has access to and are authorized to the new namespace where the VM is.  Just that it's not the namespace where the VM should be.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: @hpannem @aruneshpa  
ReconcileLocation is called when VM is created or when resourcePool is changed. So in my opinion (Absent: Not yet run or unable to determine ) is not required. We can distinguish between the VM in correct location and in incorrect location. Also "unable to determine" can happen for many reasons. if we are not able to reconcile VM's location and just remove the condition in this case, then it leads to the impression that the VM is normal.

Are these states Ok ?
True: In correct location (when vm is created or or after creation when it's moved to valid location )
False: not in correct location (when vm is moved to invalid location)

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Fixed

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Sure

pkg/util/vsphere/watcher/watcher.go
anwithaun-1: @aruneshpa Initially I had added "parent" and removed it as  @bryanv mentioned  that  "the VM managed object doesn't have a parent field" in of the review thread above. I am not sure about this.  Could you please provide more clarity.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Updated the message

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: In this PR I am only validating the RP. It's a mistake adding folder in the description. I'll remove it. However I did some basic testing around moving VM to different folders and this is not extending the namespace quota privileges as the resourcePool will not be changed in this case. So going ahead with validating RP alone.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Continuing this conversation in another thread below

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Reasons why I am not returning the Err here:

1. In case if we fail to get namespace RP, then it'll block all the next VM reconcilations. it didn't make much sense to me to skip other reconcilations just because the reconcileLocation is failed to get the namespace. That is the reason , I went ahead with ignoring this error. Still I have handled err in case if I am not able to validate the location for some reason or if the VM is in the RP that it shouldn't be in. 
2. Certain Unit tests failed due to the empty zonename property sent which is expected by the GetNamespaceFolderAndRPMoID api. 

@aruneshpa  what's your suggestion on this.

pkg/providers/vsphere/vmprovider_vm.go
hpannem: Think of user perspective, who is monitoring the CR.
1. VM reconcile is not yet run. There could be many reasons for this. Operator down... or CR is queued and reconciler didn't get a chance to this. Condition absence == reconciler not run. User waits until condition is inserted and checks the status.
2. Once reconciled, condition is put. Usually, I believe conditions should not get deleted once they are inserted. Just change true or false or unknown. There may be exceptions in some cases like upgrade etc. However, here, it would be good to keep it. 

One correction: "Unable to determine" will go as an unknown on the condition. Condition will be set as "unknown" with a reason and message. Standard k8s pattern. aka. metav1.ConditionUnknown.

True: In correct location.
False: not in correct location.
Unknown: unable to determine.
Absent: Not yet run

anwithaun-1: True: In correct location.
False: not in correct location.

The above cases are handled. 

Unknown: unable to determine- About his condition, are you saying we need to specifically set markUnknown on the condition when there is any error from reconcileLocation ?

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: > Hmm. I am with Hemanth on this. The only valid use-case for child RP is VKS. But if the VM is not in the VKS child RP, then we need to report it as an error / invalid location.

@aruneshpa What is the reason we want to allow only VKS child RP and consider other child RP of that namespace as invalid..As long as the child RP is under namespace 
quota and not escalating any privileges, then we are good right? Is there any specific reason I am missing out here which is not following up with the intention of the jira. 
Also if we have to allow VKS only, then what's the way to differentiate between VKS and Non VKS child RP's. Any inputs on this?

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Updated the code to mark true when the VM is in correct location and mark False when it's in invalid location.

pkg/providers/vsphere/vmprovider_vm.go
aruneshpa: By "all the next VM reconciliations" do you mean the other reconcilers like policy, configuration, snapshot etc? If we are unable to assert that the VM is where it needs to be, then it is okay to not reconcile this VM at all.  

Looking at that method GetNamespaceFolderAndRPMoID, it can only error out if we are not able to fetch the Zone resource which can happen due to intermittent API server issues.  I would say let's say we return an error and requeue here.

pkg/util/vsphere/watcher/watcher.go
aruneshpa: Hmm.  VM definitely has a "parent" property.  We use that in automatic registration for backup / restore.  It comes from the ManagedObject type that is inherited by VirtualMachine: https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.ManagedEntity.html#parent

aruneshpa (approved):

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: @aruneshpa @bryanv @hpannem

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1119,54 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+		pkgcnd.MarkFalse(
+			vmCtx.VM,
+			vmopv1.VirtualMachineInAuthorizedLocation,
+			"LocationMismatch",
+			"VM is in an invalid Resource Pool. Move the VM back to the Resource Pool hierarchy for namespace %s to resume reconciliation.",
+			vmCtx.VM.Namespace,
+		)
+
+		return pkgerr.NoRequeueError{Message: fmt.Sprintf(
+			"reconciliation stopped for the VM %s because it is moved to invalid Resource Pool. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s",
```
hpannem: Nit: stopped ==> paused?

hpannem (approved):

pkg/util/vsphere/watcher/watcher.go
anwithaun-1: Sure. Adding parent also to the watcher

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: I'll address this as part of new Jira ticket.

api/v1alpha6/virtualmachine_types.go
```diff
@@ -115,6 +115,10 @@ const (
 	// configuration is synced to the desired state, determined by whether or
 	// not the guest is reporting the expected IP address(es).
 	VirtualMachineGuestNetworkConfigSynced = "VirtualMachineGuestNetworkConfigSynced"
+
+	// VirtualMachineInValidLocation indicates that the VM is in the expected location in the vCenter inventory.
+	// This condition is set when a VM is detected to be in the incorrect folder or resource pool based on its namespace.
```
bryanv, line 120: It is set regardless - true or false is what changes based on the VM's location.

pkg/providers/vsphere/vcenter/resourcepool.go
```diff
@@ -162,3 +163,30 @@ func findChildRP(
 
 	return childRP, nil
 }
+
+func IsVMInValidResourcePool(
```
bryanv, line 167: Anything added here should have a test or two.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -994,7 +995,17 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 	}
 
 	//
-	// 2. Get the recent tasks.
+	// 2. Reconcile location
```
bryanv, line 998: I might be misremembering but I thought the plan was to do this after updating status.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1119,61 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 	return reconcileErr
 }
 
+func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContext, vcClient *vcclient.Client) error {
+	logger := pkglog.FromContextOrDefault(vmCtx)
+	if vmCtx.MoVM.ResourcePool == nil {
+		return fmt.Errorf("VM %s is not assigned to any resource pools", vmCtx.VM.Name)
```
bryanv, line 1125: Generally we don't need to include the VM's name in its error because the name will be a part of the larger context

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1119,61 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+	}
+
+	// If the VM doesn't have a topology zone label, skip location validation.
+	// This can happen in unit tests or when the zone is being determined.
+	// Similar to how vmlifecycle/update_status.go handles missing zone labels.
+	zoneName := vmCtx.VM.Labels[corev1.LabelTopologyZone]
```
bryanv, line 1131: They will generally be the same but I would favor to use Status.Zone instead of the lablel. More so if we do this after updating status.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1108,6 +1119,61 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+		vs.k8sClient,
+		zoneName,
+		vmCtx.VM.Namespace,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to get expected namespace resource pool: %w", err)
+	}
+
+	// Check if the VM is in the Root RP or a Child RP
+	isValid, err := vcenter.IsVMInValidResourcePool(
```
bryanv, line 1147: This function's usage doesn't match its name. It doesn't take a VM and doesn't do anything really specify to VMs, and that package is really meant to be helper utils that wrap some of the verbosity of VC calls. There is a `GetChildResourcePool()` so I would more prefer something like `GetResourcePoolParent()` and the logic to determine a match lives in here.

pkg/util/vsphere/watcher/watcher.go
bryanv: My bad. It is in the base class. But if we're not going to take anything actions on it we shouldn't include it in what we're watching.

bryanv (commented):
This is very much missing tests that move a VM into the wrong location and into the right location and assert that it gets the expected conditions. This should also include a VM that simulates a VKS VM by having a VMSetResourcePolicy with a child pool set so we assert that this works there too. The cases where this condition is true is something implicitly tested elsewhere but we have something explicit for that behavior

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: the intention of adding this here is to skip all other reconciles if the VM itself is in invalid Location

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Right. But as of now, we are not calling the reconcileLocation after doing reconcileStatus. In this case, status.Zone wouldn't be updated right?

anwithaun-1:
Above comments have been addressed in the PR - https://github.com/vmware-tanzu/vm-operator/pull/1630

PR #1563, opened 2026-04-22
🌱 Make VirtualMachineService LB provider per namespace

bryanv:

Instead of setting up the LB provider once at controller init time, move it so it is done during the reconcile. Note that this still depends on the global network provider but eventually we'll need to support this per NS.

Fixes #

```release-note
NONE
```

### DIFF
--- a/controllers/virtualmachineservice/virtualmachineservice_controller.go
+++ b/controllers/virtualmachineservice/virtualmachineservice_controller.go
@@ -54,24 +54,11 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		controllerNameLong  = fmt.Sprintf("%s/%s/%s", ctx.Namespace, ctx.Name, controllerNameShort)
 	)
 
-	lbProviderType := pkgcfg.FromContext(ctx).LoadBalancerProvider
-	if lbProviderType == "" {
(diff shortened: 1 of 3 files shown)


controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -210,25 +194,47 @@ func (r *ReconcileVirtualMachineService) ReconcileNormal(ctx *pkgctx.VirtualMach
...
 func (r *ReconcileVirtualMachineService) reconcileVMService(ctx *pkgctx.VirtualMachineServiceContext) error {
 	ctx.Logger.Info("Reconcile VirtualMachineService")
 	defer ctx.Logger.Info("Finished Reconcile VirtualMachineService")
 
 	vmService := ctx.VMService
 
 	if vmService.Spec.Type == vmopv1.VirtualMachineServiceTypeLoadBalancer {
+		lbProvider, err := r.getLoadbalancerProvider(ctx)
+		if err != nil {
+			ctx.Logger.Error(err, "Failed to get load balancer provider")
```
faisalabujabal: general comment about this method is that we always log the error then return it. the caller of this method also logs the error and returns it. should we just have a pattern where the method that swallows the error is the one responsible for logging it to reduce duplicate error logs

faisalabujabal (approved):

PR #1566, opened 2026-04-23
🌱 fix: Moving the E2E Docker image to Photon

faisalabujabal:

This PR moves the E2E docker image to use photon instead of alpine because I am having issues with alpine in the build system.

Fixes #
NA

NA

```release-note
fix: Moving the E2E Docker image to Photon
```

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,8 +1,7 @@
 # VM Operator E2E Test Container
 # Self-contained E2E testing environment
 
-# Use Broadcom internal registry (Docker Hub mirror) 
-ARG BASE_IMAGE=mirror.gcr.io/library/golang:1.26.2-alpine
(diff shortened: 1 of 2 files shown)


hpannem (approved):

brito-rafa (approved):
LGTM

Dockerfile.e2e
```diff
@@ -43,26 +47,29 @@ LABEL branch="${BUILD_BRANCH}" \
       version="${BUILD_VERSION}"
 
 # Install required packages (alpine-based)
```
bryanv: I think this too

Dockerfile.e2e
```diff
@@ -2,7 +2,7 @@
 # Self-contained E2E testing environment
 
 # Use Broadcom internal registry (Docker Hub mirror)
```
bryanv: NBD but stale comment?

bryanv (approved):

PR #1570, opened 2026-04-27
✨ Persist VmPlacementPolicies and TagSpecs during VM creation

palnabarun:

Currently, VM placement policies and tag specifications are only
generated during placement recommendation calls via
`CreateConfigSpecForPlacement()`. This change extends the same logic
to `CreateConfigSpec()` to ensure placement policies are consistently
applied during actual VM creation.

Key changes:
- Add `genConfigSpecAffinityPolicies()` and `genConfigSpecTagSpecsFromVMLabels()`
  calls to `CreateConfigSpec()` when `VMAffinityDuringExecution` feature is enabled
- Implement `extractAffinityLabelsFromVM()` to optimize TagSpec generation by 
  only including VM labels that are referenced in affinity/anti-affinity rules
- Add comprehensive unit tests for placement policy generation and TagSpec
  optimization in `CreateConfigSpec` test suite
- Ensure consistent behavior between placement and creation flows

This ensures that VMs created with affinity/anti-affinity rules will have
the proper placement policies applied to the vSphere ConfigSpec during
creation. The TagSpec optimization reduces unnecessary vSphere tags by
only creating tags for labels that are actually used in DRS placement
policies.

Fixes #

```release-note

```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/affinity.go
+++ b/pkg/providers/vsphere/virtualmachine/affinity.go
@@ -11,12 +11,59 @@ import (
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 )
+
+// extractAffinityLabelsFromVM extracts all "key:value" labels referenced
+// in the VM's affinity and anti-affinity rules.
+// Returns a set where keys are "key:value" strings
+// that are used in affinity rules.
+func extractAffinityLabelsFromVM(vmCtx pkgctx.VirtualMachineContext) sets.Set[string] {
+	affinity := vmCtx.VM.Spec.Affinity
+	if affinity == nil {
+		return nil
+	}
+
+	affinityLabels := sets.New[string]()
+
+	// helper to extract "key:value" labels from VMAffinityTerm slice
+	extractFromTerms := func(terms []vmopv1.VMAffinityTerm) {
+		for _, term := range terms {
+			if term.LabelSelector == nil {
+				continue
+			}
+
+			labels, err := extractLabelsFromSelector(term.LabelSelector)
+			if err != nil {
+				vmCtx.Logger.V(4).Error(err, "invalid label selector")
+				continue
+			}
+
+			for _, label := range labels {
+				affinityLabels.Insert(label)
+			}
+		}
+	}
+
+	// process VM affinity rules
+	if affinity.VMAffinity != nil {
+		extractFromTerms(affinity.VMAffinity.RequiredDuringSchedulingPreferredDuringExecution)
+		extractFromTerms(affinity.VMAffinity.PreferredDuringSchedulingPreferredDuringExecution)
+	}
+
+	// process VM anti-affinity rules
+	if affinity.VMAntiAffinity != nil {
+		extractFromTerms(affinity.VMAntiAffinity.RequiredDuringSchedulingPreferredDuringExecution)
+		extractFromTerms(affinity.VMAntiAffinity.PreferredDuringSchedulingPreferredDuringExecution)
+	}
+
+	return affinityLabels
+}
 
 // genConfigSpecTagSpecsFromVMLabels generates tag specs from VM labels.
 // This is required when setting affinity/anti-affinity policies on the VM.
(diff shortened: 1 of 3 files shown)


pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -151,6 +151,12 @@ func CreateConfigSpec(
 		initResourceAllocation(&configSpec.MemoryAllocation)
 	}
 
+	// Apply placement policies and tag specs when feature is enabled
+	if pkgcfg.FromContext(vmCtx).Features.VMAffinityDuringExecution {
```
aniket-deole, line 155: We need to configure affinity iff vm is part of a vmgroup.

palnabarun: `spec.Affinity` can't be set when the VM is not part of a group. This is guarded by our validation webhook.

palnabarun: The block is a no-op when there are no AF/AAF rules.

pkg/providers/vsphere/virtualmachine/affinity.go
```diff
@@ -30,11 +78,27 @@ func genConfigSpecTagSpecsFromVMLabels(
...
+	// 	policies are added to the configSpec.TagSpecs.
+	// 	This is because DRS expects tags which are ONLY referenced in
+	// 	affinity/anti-affinity policies.
+	//
+	// TODO(for Day 2 operations):
+	// 	When VM is updated and policies are added which refer more labels,
+	// 	this method is called again and the tags are added to the
+	// 	configSpec.TagSpecs.
 	for key, value := range filteredLabels {
+		if pkgcfg.FromContext(vmCtx).Features.VMAffinityDuringExecution &&
```
aniket-deole, line 100: Should this filter only affect to affinity selectors with host topology?

palnabarun: IIUC the behavior is same for both host and zone topologies.

pkg/providers/vsphere/virtualmachine/affinity.go
```diff
@@ -11,13 +11,61 @@ import (
...
+// extractAffinityLabelsFromVM extracts all "key:value" labels referenced
+// in the VM's affinity and anti-affinity rules.
+// Returns a set where keys are "key:value" strings
+// that are used in affinity rules.
+func extractAffinityLabelsFromVM(vm *vmopv1.VirtualMachine) sets.Set[string] {
+	affinityLabels := make(sets.Set[string])
+
+	affinity := vm.Spec.Affinity
+	if affinity == nil {
+		return affinityLabels
```
bryanv: Kind of nit but we can return `nil` here instead since that is the common case?

pkg/providers/vsphere/virtualmachine/affinity.go
```diff
@@ -11,13 +11,61 @@ import (
...
+	}
+
+	// helper to extract "key:value" labels from VMAffinityTerm slice
+	extractFromTerms := func(terms []vmopv1.VMAffinityTerm) {
+		for _, term := range terms {
+			if term.LabelSelector == nil {
+				continue
+			}
+
+			// Extract labels from selector using the same logic as extractLabelsFromSelector
```
bryanv: This comment is a little confusing since it is already using `extractLabelsFromSelector`

pkg/providers/vsphere/virtualmachine/affinity.go
```diff
@@ -11,13 +11,61 @@ import (
...
+			labels, err := extractLabelsFromSelector(term.LabelSelector)
+			if err != nil {
+				// skip invalid selectors (same as in buildTagIDsFromTopology)
```
faisalabujabal: should we have a log message?

pkg/providers/vsphere/virtualmachine/affinity.go
```diff
@@ -30,11 +78,27 @@ func genConfigSpecTagSpecsFromVMLabels(
 	}
 
 	tagsToAdd := make([]vimtypes.TagSpec, 0, len(filteredLabels))
+	policyLabels := extractAffinityLabelsFromVM(vmCtx.VM)
```
faisalabujabal: shouldn't this be wrapped with the capability check?

pkg/providers/vsphere/virtualmachine/affinity.go
```diff
@@ -11,13 +11,60 @@ import (
...
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 )
 
+// extractAffinityLabelsFromVM extracts all "key:value" labels referenced
+// in the VM's affinity and anti-affinity rules.
+// Returns a set where keys are "key:value" strings
+// that are used in affinity rules.
+func extractAffinityLabelsFromVM(vmCtx pkgctx.VirtualMachineContext) sets.Set[string] {
+	affinityLabels := sets.New[string]()
```
bryanv: Move this down to L33 since it isn't used until then

pkg/providers/vsphere/virtualmachine/affinity.go
```diff
@@ -11,13 +11,60 @@ import (
...
+	// helper to extract "key:value" labels from VMAffinityTerm slice
+	extractFromTerms := func(terms []vmopv1.VMAffinityTerm) {
+		for _, term := range terms {
+			if term.LabelSelector == nil {
+				continue
+			}
+
+			labels, err := extractLabelsFromSelector(term.LabelSelector)
+			if err != nil {
+				vmCtx.Logger.Error(err, "invalid label selector")
```
bryanv: I might lower the verbosity level of this error because it could be spammy, even though it is not suppose to fail because of the validation webhook check.

bryanv (approved):
Please address a few minor comments before merging. Thanks.

pkg/providers/vsphere/virtualmachine/affinity.go
aruneshpa: Spoke to @aniket-deole.  This most likely needs to be changed.  He will follow up.

PR #1571, opened 2026-04-27
✨  Backfill VM network interface types from VM config

bryanv:

Instead of using the current VM's class, use the VC VM's actual hardware configuration to backfill the interfaces Types. This is so after the backfill, the VM Spec reflects the VM's current state, and then the user can adjust types as desired.

For VM create, don't mutate in the Types from the class but instead just let the schema upgrade do it if empty

For VM update, don't use the Type from the old VM. Instead, default to vmxnet3.

Fixes #

```release-note
For schema upgrade, backfill the VirtualMachineNetworkInterface Type from the VM current hardware configuration.
```

----
📚 Documentation preview 📚: https://vm-operator--1571.org.readthedocs.build/en/1571/


pkg/providers/vsphere/upgrade/virtualmachine/vm_schema_upgrade_test.go
```diff
@@ -328,22 +328,72 @@ var _ = Describe("ReconcileSchemaUpgrade", func() {
...
+							}
+						})
+						It("should backfill SRIOV from moVM", func() {
+							assertUpgraded()
+							assertFeatureVersion("9")
+							Expect(vm.Spec.Network.Interfaces[0].Type).To(Equal(vmopv1.VirtualMachineNetworkInterfaceTypeSRIOV))
+						})
+					})
+
+					When("VC VM has multiple NICs", func() {
```
hpannem: Should we add another case where moVM has extra nic than spec?

pkg/providers/vsphere/virtualmachine/network_interface_types_from_movm.go
```diff
@@ -0,0 +1,87 @@
...
+	return out
+}
+
+func mapVimEthernetToNetworkInterfaceType(
+	dev vimtypes.BaseVirtualDevice) vmopv1.VirtualMachineNetworkInterfaceType {
+
+	switch dev.(type) {
+	case *vimtypes.VirtualSriovEthernetCard:
+		return vmopv1.VirtualMachineNetworkInterfaceTypeSRIOV
+	case *vimtypes.VirtualVmxnet3:
```
hpannem: Nit: you can move this to first case as most of the nics are vmxnet3.

pkg/providers/vsphere/virtualmachine/network_interface_types_from_movm_test.go
```diff
@@ -0,0 +1,195 @@
...
+			BeforeEach(func() {
+				moVM = moVMWithEthernet(&vimtypes.VirtualPCIPassthrough{})
+			})
+
+			It("should default to VMXNet3", func() {
+				expectMutationSuccess()
+				Expect(vm.Spec.Network.Interfaces[0].Type).To(Equal(vmopv1.VirtualMachineNetworkInterfaceTypeVMXNet3))
+			})
+		})
+	})
```
hpannem, line 216: do we need a case where some of the spec interfaces types are filled and some are not.

webhooks/virtualmachine/mutation/virtualmachine_mutator_network_interface_type.go
```diff
@@ -9,68 +9,46 @@ import (
...
+	vmFeatureVersion := vmopv1util.ParseFeatureVersion(
+		vm.Annotations[pkgconst.UpgradedToFeatureVersionAnnotationKey])
 
-			t = oldVM.Spec.Network.Interfaces[i].Type
+	mutated := false
+	// The interface Type is backfilled during schema upgrade, so don't default until after
+	// that has been done.
+	// BV: I prefer doing this in the mutation but maybe this is kind of weird, and we should
+	// just do this during VM reconcile instead? Note that as-is with doing it in the mutation
+	// webhook we're missing some checks in the validation validateFieldsDuringSchemaUpgrade().
```
hpannem: One clarification:
If we don't check featureversion here and always update, user request will go through and backfill will fail and retry right or vice versa (user request fails due to backfill success)? (due to optimistic locks)

However, I agree that adding preventing the modification of the field during upgrade would give an opportunity to the user to check what is backfilled and modify if needed. I will add all the new fields to this method, including Type in my PR.

hpannem (commented):
This looks good to me. Couple of questions.
1. If you add a new network interface via spec for the existing vm, nothing is backfilled right? We will default it to type vmxnet3, if type is empty. The expectation is user needs to specify the proper config, if he needs it.
2. In general, all the config that am going to add needs to be backfilled right? 
3. What happens if there is a mismatch of types during backfill? I believe partial spec should not be allowed. E.g. In SRIOV spec, am not allowing partial spec values if type is provided by the user. i.e. if type=SRIOV is provided in VM spec, he has to provide the SRIOV device spec as well, even though class has the SRIOV card at that index. Is that ok?

akutz (approved):
Thanks @bryanv. Let's make sure we execute the follow-up ticket you filed internally tracking the backfill locking. cc @srajashekar-vmw, @aruneshpa 

This is basically what @hpannem was saying at https://github.com/vmware-tanzu/vm-operator/pull/1571#discussion_r3185243421. Good job on you @hpannem for catching this. You have come up to speed super quickly, and I appreciate it immensely!

hpannem (approved):
Thanks @bryanv for taking care of this.

webhooks/virtualmachine/mutation/virtualmachine_mutator_network_interface_type.go
bryanv: This is more so for the updates that the VM controller does between the VM CR being created, and actually creating the VC VM. If we do it before the schema is upgrade, then it will backfilled (defaulted) before then. But yes some other user update would hit this same code during that creating window. The locking here is somewhat separate: we need a consistent view of the VM Spec so that a user change isn't overwritten as a part of the backfill. Yes, that does introduce a window where this cannot be changed, if if something is invalid, that might mean the VM CR needs to be deleted and recreated. That is true of like other locked things like controllers, and I don't think we can entirely avoid it here either.

Issue #1, opened 2021-02-23
Create README and How To guide(s)


corrieb:
README should be concise and should talk about VM Service at a high level, linking to external blogs and information. It should also link to any Guides we add to the repo.

Given that we are designing a simple experience for build and test, we should create a How To Guide that takes you through the process of modifying, building and testing. This is separate from the "how to contribute" guide. It doesn't have to be elaborate, but it should cover the basics of getting an environment set up, building the code and what the different test targets are designed to cover

PR #1572, opened 2026-04-28
🐛 Classic disk reg: defer CRV until batch CSI signal

akutz:

When AllDisksArePVCs is enabled, ensure placeholder PVCs and VM claimNames in phase one, then reconcile pending until CnsNodeVMBatchAttachment status reflects the CSI driver's volume-ID cache miss for the PVC. Only then create CnsRegisterVolume, avoiding races with the batch attachment path.

The volume batch controller now includes unbound PVCs whose dataSourceRef targets the VM (same API group), so CSI can observe them in the batch spec before binding.

Adds tests for the deferred CRV lifecycle under this feature flag.

Fixes `NA`

@skogta helpfully, and clearly, outlined the possible race below:

T1: Batch attach Reconcile https://github.com/vmware-tanzu/vm-operator/issues/1 starts
- Reads current batch attach spec (without new PVC)

T2: VM Operator updates batch attach spec
- Adds unbound PVC to spec
- Immediately creates CnsRegisterVolume

T3: CnsRegisterVolume completes quickly
- PVC binds and gets volume ID
- Volume becomes "attached" to VM in vCenter

T4: Reconcile https://github.com/vmware-tanzu/vm-operator/issues/1 continues (still using old spec)
- Sees newly bound volume attached to VM
- But volume is NOT in the spec it read at T1
- Adds volume to "detach list"

T5: Reconcile https://github.com/vmware-tanzu/vm-operator/issues/1 detaches the volume
- Even though VM Operator intended it to stay attached

```release-note
Wait for signal from CSI that its seen unbounded volumes before doing CnsRegisterVolume.
```

### DIFF
--- a/pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
+++ b/pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
@@ -30,8 +30,10 @@ import (
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	pkgerr "github.com/vmware-tanzu/vm-operator/pkg/errors"
 	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
+	pkgutil "github.com/vmware-tanzu/vm-operator/pkg/util"
 	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 	"github.com/vmware-tanzu/vm-operator/pkg/util/ptr"
+	vmopv1util "github.com/vmware-tanzu/vm-operator/pkg/util/vmopv1"
 	pkgvol "github.com/vmware-tanzu/vm-operator/pkg/util/volumes"
 	pkgdatastore "github.com/vmware-tanzu/vm-operator/pkg/util/vsphere/datastore"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmconfig"
@@ -497,6 +499,12 @@ func ensureUnmanagedDisksHaveUpdatedCapacity(
 	return nil
 }
 
+// registerUnmanagedDisks uses a two-phase flow:
+//  1. Ensure every unmanaged classic disk has a PVC and claimName on the VM
+//     spec, then return pending so the volume batch controller can add unbound
+//     PVCs (dataSourceRef → VM) to CnsNodeVMBatchAttachment.
+//  2. On a later reconcile, wait until batch status shows the CSI volume-ID
+//     cache miss text, then ensure CnsRegisterVolume for pending PVCs.
 func registerUnmanagedDisks(
 	ctx context.Context,
 	k8sClient ctrlclient.Client,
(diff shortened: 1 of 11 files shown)


bryanv:
Just to confirm: the error that CSI will put in the BatchAttach will cause our Volume controller to then update the VM Status with SomeThing? Otherwise, I think we'll need the VM controller to Watch the BatchAttach (sigh), maybe filtering for "failed to find volumeID for PVC" in the conditions to avoid yet more noop reconciles.

pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
```diff
@@ -937,3 +978,62 @@ func cleanupVolumeStatus(vm *vmopv1.VirtualMachine) {
...
+	if err := k8sClient.Get(ctx, key, ba); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	if conditionsIncludePVCVolumeIDCacheMissMsg(ba.Status.Conditions) {
+		return true, nil
+	}
```
deepakkinni: I was thinking about this: volumebatch_controller adds the PVC to CnsNodeVMBatchAttachment when it observes a PVC with a VM source in the VM spec. You are currently checking for a specific condition to be set, but what if CSI crashes before setting that condition?

skogta:
I am wondering if we even need the "signal" from CSI here?

If unbound PVCs with datasourceref as VirtualMachine are successfully patched on the batch attach spec, then VM op can directly move onto CNS registration step.

Since the batch attach update has succeeded, the resource is stored in etcd which automatically means that CSI can see unbound PVCs on its spec.

skogta:
Okay I see why "the signal" is required. Particularly for this kind of race:

T1: Batch attach Reconcile #1 starts
    - Reads current batch attach spec (without new PVC)

T2: VM Operator updates batch attach spec 
    - Adds unbound PVC to spec
    - Immediately creates CnsRegisterVolume

T3: CnsRegisterVolume completes quickly
    - PVC binds and gets volume ID
    - Volume becomes "attached" to VM in vCenter

T4: Reconcile #1 continues (still using old spec)
    - Sees newly bound volume attached to VM
    - But volume is NOT in the spec it read at T1
    - Adds volume to "detach list"

T5: Reconcile #1 detaches the volume
    - Even though VM Operator intended it to stay attached

pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
```diff
@@ -497,79 +498,117 @@ func ensureUnmanagedDisksHaveUpdatedCapacity(
...
+	var hasPending bool
+
+	for _, di := range info.Disks {
+		volume, ok := info.Volumes[di.Target.String()]
+		if !ok {
+			continue
+		}
+
+		pvcName := volume.Name
+		if pvc := volume.PersistentVolumeClaim; pvc != nil {
```
bryanv, line 575: By this time through this code path, is there a case that `volume.PersistentVolumeClaim` can be nil?

akutz: I do not think so, but does it hurt?

pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
akutz: Then the VM is forever blocked from attaching/detaching disks. If upon restarting CSI does not update the condition appropriately, then CSI is broken and should be fixed.

akutz:
> Just to confirm: the error that CSI will put in the BatchAttach will cause our Volume controller to then update the VM Status with SomeThing? Otherwise, I think we'll need the VM controller to Watch the BatchAttach (sigh), maybe filtering for "failed to find volumeID for PVC" in the conditions to avoid yet more noop reconciles.

@bryanv I re-pushed the changes to add the watch on the batch attach object.

bryanv (approved):

pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
skogta: If CSI crashes before updating this error, then in the next reconcile it will definitely populate this error message unless the VM itself gets deleted - which is fine because that means VM op will not need to add any volumes to batchattach spec anyway (and hence register them).

@deepakkinni did you have some other case in mind?

deepakkinni: lgtm then.

PR #1573, opened 2026-04-28
✨ Simplify VM create network interface provisioning

bryanv:

Replace blocking network interface polling with watch-based reconciliation and introduce NetworkDevice type for create flow optimizations.

This is the first in several in addressing the cruft that has built up around the network interfaces. Issues in the Update path will be next.

- Add CreateNetworkDevices function for VM create that creates all network interface CRs and checks readiness without polling, returning NetworkDevices containing just what is needed for the ethernet card device
- Add VM controller watches for NetworkInterface, VirtualNetworkInterface, and SubnetPort so the polling can be removed.
- Keep existing CreateAndWaitForNetworkInterfaces for update path that needs full NetworkInterfaceResult with bootstrap customization fields The interface CR create code is shared between the two function

The create flow now creates all interface CRs, immediately checks readiness, and returns either devices or an error. Controller watches ensure prompt re-reconciliation when CRs become ready, eliminating the 15-second polling timeout that blocked VM creation. This will speed up creation of VMs with network interfaces.

Parts of this change facilitate later work such as mixed network provider setups, and BYO network interface.

Fixes #

```release-note
During VM create, do no poll each network interface CR. Instead, use a Watch() to reconcile the VM when the CR is updated.
```

### DIFF
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -22,10 +22,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	netopv1alpha1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
 	vpcv1alpha1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
 	byokv1 "github.com/vmware-tanzu/vm-operator/external/byok/api/v1alpha1"
+	ncpv1alpha1 "github.com/vmware-tanzu/vm-operator/external/ncp/api/v1alpha1"
 	cnsv1alpha1 "github.com/vmware-tanzu/vm-operator/external/vsphere-csi-driver/api/v1alpha1"
 	vspherepolv1 "github.com/vmware-tanzu/vm-operator/external/vsphere-policy/api/v1alpha1"
 	pkgcond "github.com/vmware-tanzu/vm-operator/pkg/conditions"
@@ -180,19 +182,36 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 			handler.EnqueueRequestForOwner(
 				mgr.GetScheme(),
 				mgr.GetRESTMapper(),
-				&vmopv1.VirtualMachine{},
-				handler.OnlyControllerOwner()),
+				&vmopv1.VirtualMachine{}),
 		)
 	}
 
-	if pkgcfg.FromContext(ctx).NetworkProviderType == pkgcfg.NetworkProviderTypeVPC {
+	switch pkgcfg.FromContext(ctx).NetworkProviderType {
+	case pkgcfg.NetworkProviderTypeVDS:
+		builder = builder.Watches(
+			&netopv1alpha1.NetworkInterface{},
+			handler.EnqueueRequestForOwner(
+				mgr.GetScheme(),
+				mgr.GetRESTMapper(),
+				&vmopv1.VirtualMachine{}),
+		)
+
+	case pkgcfg.NetworkProviderTypeNSXT:
+		builder = builder.Watches(
+			&ncpv1alpha1.VirtualNetworkInterface{},
+			handler.EnqueueRequestForOwner(
+				mgr.GetScheme(),
+				mgr.GetRESTMapper(),
+				&vmopv1.VirtualMachine{}),
+		)
+
+	case pkgcfg.NetworkProviderTypeVPC:
 		builder = builder.Watches(
 			&vpcv1alpha1.SubnetPort{},
 			handler.EnqueueRequestForOwner(
 				mgr.GetScheme(),
 				mgr.GetRESTMapper(),
-				&vmopv1.VirtualMachine{},
-				handler.OnlyControllerOwner()),
+				&vmopv1.VirtualMachine{}),
 		)
 	}
 
--- a/pkg/providers/vsphere/network/network.go
+++ b/pkg/providers/vsphere/network/network.go
@@ -31,6 +31,7 @@ import (
 	vpcv1alpha1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
 
 	ncpv1alpha1 "github.com/vmware-tanzu/vm-operator/external/ncp/api/v1alpha1"
+	pkgerr "github.com/vmware-tanzu/vm-operator/pkg/errors"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
 	"github.com/vmware-tanzu/vm-operator/pkg"
@@ -82,6 +83,17 @@ type NetworkInterfaceRoute struct {
 	Metric int32
 }
 
+// Device contains the information from the network interface CR needed to create or
+// configure a virtual ethernet card device on a VM.
+type Device struct {
+	InterfaceObj ctrlclient.Object
+
+	Backing    object.NetworkReference
+	NetworkID  string
+	MacAddress string
+	ExternalID string
+}
+
 const (
 	retryInterval           = 100 * time.Millisecond
 	defaultEthernetCardType = "vmxnet3"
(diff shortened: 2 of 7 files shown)


controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -180,19 +182,36 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 			handler.EnqueueRequestForOwner(
 				mgr.GetScheme(),
 				mgr.GetRESTMapper(),
-				&vmopv1.VirtualMachine{},
-				handler.OnlyControllerOwner()),
```
hpannem, line 184: is this because of old vms which didnt have controller owner?

hpannem (approved):
Nice improvement!

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
bryanv: Kind of. The network interface CRs that we create should have always been controller refs but weren't. I fixed that recently and all VPC SubnetPorts will be controller owner refs, but I changed this for consistency with the others and just for BYO network interface support (although the precise life cycle of those are still TBD)

akutz: Shouldn't we be patching existing objects with ControllerRefs then?

pkg/providers/vsphere/network/network.go
```diff
@@ -100,8 +101,351 @@ const (
...
+	finder *find.Finder,
+) ([]NetworkDevice, error) {
+
+	networkSpec := vm.Spec.Network
+	if networkSpec == nil {
+		return nil, nil
+	}
+
+	switch pkgcfg.FromContext(ctx).NetworkProviderType {
+	case pkgcfg.NetworkProviderTypeNamed:
```
akutz, line 143: Why we are we continuing to support this? Is it used anywhere?

pkg/providers/vsphere/network/network.go
```diff
@@ -100,8 +101,351 @@ const (
...
+		case *ncpv1alpha1.VirtualNetworkInterface:
+			dev, err = getNCPNetworkInterfaceDevice(ctx, vimClient, nil, ifaceCR)
+		case *vpcv1alpha1.SubnetPort:
+			dev, err = getVPCSubnetPortDevice(ctx, vimClient, nil, ifaceCR)
+		default:
+			err = fmt.Errorf("unsupported network interface CR type: %T", obj)
+		}
+
+		if err != nil {
+			// TODO: Update per-interface Status.
```
akutz, line 214: Why not do it here? Curious if you can expand on what parts of the status you want to update.

akutz (commented):
Requested some feedback, but not necessarily changes. Let me know what you think.

pkg/providers/vsphere/network/network.go
bryanv: We don't. I know that CAPV has asked about it before. But I'd rather pull it out in a separate commit instead of a refactor.

pkg/providers/vsphere/network/network.go
bryanv: It needs more thought. I want to at least reflect the readiness of it so there is more clue as to what is wrong so the initial triage goes more appropriate queue.

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
bryanv: Yes, I handled that with: https://github.com/vmware-tanzu/vm-operator/blob/f999886a4a076f1693e7718c693122fb05a2d419/pkg/providers/vsphere/network/network.go#L1128. But since in theory an existing network interface CR could already have a controller ref we can't always put ourselves as the owner.

VPC was done after I fixed us to always do controller refs so that is why the prior code used that in the handler but change here to match the others.

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -180,19 +182,36 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
...
+		)
+
+	case pkgcfg.NetworkProviderTypeVPC:
 		builder = builder.Watches(
 			&vpcv1alpha1.SubnetPort{},
 			handler.EnqueueRequestForOwner(
 				mgr.GetScheme(),
 				mgr.GetRESTMapper(),
-				&vmopv1.VirtualMachine{},
-				handler.OnlyControllerOwner()),
```
aruneshpa, line 195: I read your comment in the other thread, but I am still not sure I understand why we are removing this.  The SubnetPorts we create will always have controller owner refs set for the VM.  So why remove this line?   You say a NIF CR can have a controller ref and we can't update it to the VM.  But aren't we supposed to create the NIFs and own them always?

pkg/providers/vsphere/network/network.go
```diff
@@ -100,8 +112,317 @@ const (
...
+		return nil, fmt.Errorf("no network provider set")
+	}
+
+	objs := make([]ctrlclient.Object, 0, len(networkSpec.Interfaces))
+	var errs []error
+
+	// Create all the network interface CRs. The VM may have multiple interfaces so we
+	// want to create them all up front so they can be reconciled concurrently.
+	for i := range networkSpec.Interfaces {
+		interfaceSpec := &networkSpec.Interfaces[i]
```
aruneshpa: For my understanding, is the reason for not using for-range expression here is because you want to avoid the copy of each NIF struct on every iteration?

pkg/providers/vsphere/network/network.go
```diff
@@ -100,8 +112,317 @@ const (
...
+	if len(errs) != 0 {
+		return nil, errors.Join(errs...)
+	}
+
+	devices := make([]Device, 0, len(objs))
+
+	// For each interface, if its network interface CR is ready, get the information
+	// needed from it for the device.
+	for i, interfaceSpec := range networkSpec.Interfaces {
+		obj := objs[i]
```
aruneshpa, line 197: Is relying on the index based matching safe here?  IOW, should we create a mapping between the spec.network.interface and the CR or is it not worth the trouble?
I can't think of a case where index-based approach wouldn't work.  But double checking.

pkg/providers/vsphere/network/network.go
```diff
@@ -100,8 +112,317 @@ const (
...
+	clusterMoRef *vimtypes.ManagedObjectReference,
+	vnetIf *ncpv1alpha1.VirtualNetworkInterface) (Device, error) {
+
+	var dev Device
+
+	var readyCond *ncpv1alpha1.VirtualNetworkCondition
+	for _, cond := range vnetIf.Status.Conditions {
+		// NOTE: For whatever reason, the contributed NCP code used Contains.
+		if strings.Contains(cond.Type, "Ready") && strings.Contains(cond.Status, "True") {
+			readyCond = &cond
```
aruneshpa, line 322: Just making sure that we are not breaking out of the loop on first condition with Ready substring in the type?

pkg/providers/vsphere/network/network.go
```diff
@@ -100,8 +112,317 @@ const (
...
+		}
+	}
+
+	if readyCond == nil {
+		for _, cond := range vnetIf.Status.Conditions {
+			if strings.Contains(cond.Type, "Ready") && !strings.Contains(cond.Status, "True") {
```
aruneshpa, line 328: This really sounds like the original author might not have known the correct way to get conditions?  Can we check with them and simplify it here?

pkg/providers/vsphere/network/network.go
```diff
@@ -100,8 +112,317 @@ const (
...
+	}
+
+	if vnetIf.Status.ProviderStatus == nil {
+		return dev, pkgerr.NoRequeueNoErr("ready network interface does not have provider status")
+	}
+
+	networkID := vnetIf.Status.ProviderStatus.NsxLogicalSwitchID
+
+	var backing object.NetworkReference
+	if clusterMoRef != nil {
```
aruneshpa, line 348: The only caller in this diff is always sending nil for the clusterMoRef arg.  Is this method being used from elsewhere?

pkg/providers/vsphere/network/network.go
```diff
@@ -763,67 +1107,62 @@ func VPCCRName(vmName, networkName, interfaceName string) string {
...
+	switch networkType.Kind {
 	case "SubnetSet", "":
-		vpcSubnetPort.Spec.SubnetSet = networkRefName
+		subnetPort.Spec.SubnetSet = networkName
 	case "Subnet":
-		vpcSubnetPort.Spec.Subnet = networkRefName
+		subnetPort.Spec.Subnet = networkName
 	default:
-		return nil, fmt.Errorf("network kind %q is not supported for VPC", networkRefType.Kind)
+		return nil, pkgerr.NoRequeueNoErrorf("network kind %q is not supported for VPC", networkType.Kind)
```
aruneshpa: Is the intention of changing the error that in the unlikely scenario that we land at a kind that is not SubnetSet / Subnet, we can't do anything.  So no point re-queueing the reconcile?  Why would it not be an error though?

aruneshpa (approved):
LGTM overall, with a few minor questions / comments.

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
bryanv: BYO interface is likely coming and it isn't worth having anything special just for VPC here.

pkg/providers/vsphere/network/network.go
bryanv: Maybe that makes a difference but that isn't why it ended up this why. The existing CreateAndUpdate used pointers (mostly for worse) but I didn't want to directly change that code but I wanted to share the actual Create functions, so this ended up as a pointer. I'll pass the pointer of a local interfaceSpec var here and will eventually get the update path cleaned up.

pkg/providers/vsphere/network/network.go
bryanv: Yes. At this point there will be a 1-1 for the interface spec and the CR since any errors cause an early return before this point. I think this is a reasonable expectation of this code: it is just used during create, and we need all the interfaces to be able to create the VM. It could change later as we sort out what per-interface conditions.

pkg/providers/vsphere/network/network.go
bryanv: Yah, this is what the old code did and I didn't want to change it in this MR.

pkg/providers/vsphere/network/network.go
bryanv: Probably. This is old code and I didn't want to change it in this MR.

pkg/providers/vsphere/network/network.go
bryanv: The update path - when we'll know the VM's CCR - will eventually use this code too to get the device. Right now, the update path couples the device and bootstrap together that I want to break apart.

pkg/providers/vsphere/network/network.go
bryanv: The more important change here is the NoRequeue since there is no point in reconcile until this is updated. The NoError part doesn't matter so much as it is just what is in the logs but I change it.

PR #1577, opened 2026-04-29
🐛 xport-9.1 Defer CRV until batch CSI signal

akutz:

Cross-port of https://github.com/vmware-tanzu/vm-operator/pull/1572.

Unmanaged volume registration now gates CnsRegisterVolume on CnsNodeVMBatchAttachment status (CSI "failed to find volumeID for PVC").

- Add createBatchAttachWithPVCVolumeIDCacheMiss helper and a second createOrUpdateVM expecting ErrRegisterVolumes before asserting CRV.
- In the backfill flow, avoid Get(vm) before that second reconcile so claimName set only in memory is not dropped from the test VM object.
- After simulating PVC bound + VM status, Get/Update VM spec with claimName when the API copy lacks it so the final reconcile succeeds.

Fixes UnmanagedVolumes-focused tests under pkg/providers/vsphere.

This cross-port also includes updates to the vendored CSI CRD CnsNodeBatchVmAttachment API.

Fixes #

```release-note
Wait for signal from CSI that its seen unbounded volumes before doing CnsRegisterVolume.
```


akutz:
Merging sans review since this is an xport.

external/vsphere-csi-driver/api/v1alpha1/cnsnodebatchvmattachment_types.go
```diff
@@ -101,12 +118,19 @@ type PersistentVolumeClaimSpec struct {
...
 	// +listType=map
 	// +listMapKey=name
 	// VolumeStatus reflects the status for each volume.
 	VolumeStatus []VolumeStatus `json:"volumes,omitempty"`
+	// +optional
+
+	// Conditions describes any conditions associated with this CnsNodeVMBatchAttachment instance.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// TODO: remove this field once VM op changes are ready.
```
hickeng, line 131: unclear to me if the vmop changes being referred to are _these_ changes, and if so whether this should be removed in this PR or it's a second phase that has to happen in CSI first. (same for the other one below)

PR #1589, opened 2026-05-07
✨ Backfill ExtraConfig and NIC device properties from VM hardware during schema upgrade

hpannem:

During schema upgrade, populate the following spec fields from the live vSphere VM configuration so the spec reflects the VM's actual state:

- `spec.advanced.*` vmx-tagged first-class fields from `moVM.Config.ExtraConfig`
- `spec.network.interfaces[i].vmxnet3.*` from `moVM.Config.ExtraConfig` (using device-key-derived `ethernetX` index, not spec position)
- `spec.network.interfaces[i].vNUMANodeID` from `VirtualDevice.NumaNode`
- `spec.network.interfaces[i].vmxnet3.UPTv2Enabled` from `VirtualVmxnet3.Uptv2Enabled`

Adds `EthernetDeviceKeyBase` and `EthernetExtraConfigPrefix` to `pkg/util/vmopv1` to centralise the `deviceKey - 4000` → `ethernetX.` mapping.

Adds `validateAdvancedVMXFieldsNotChanged` and `validateNICBackfilledFieldsNotChanged` to `validateFieldsDuringSchemaUpgrade` to prevent users from modifying backfilled fields before schema upgrade completes.

Fixes #

- Spec wins over live vSphere state: only nil/zero fields are written; existing values are never overwritten.
- Spec interfaces are zipped by position to hardware devices (same convention as `FillEmptyNetworkInterfaceTypesFromMoVM`). A proper device-to-spec matching is a follow-up.

```release-note
During schema upgrade, backfill vmx-tagged advanced spec fields and per-NIC VMXNet3/vNUMANodeID properties from the live vSphere VM configuration.

### CODE REVIEW

File: pkg/providers/vsphere/upgrade/virtualmachine/vm_schema_upgrade.go [hunk 1]
```diff
@@ -174,6 +174,12 @@ func ReconcileSchemaUpgrade(
 
 				return fmt.Errorf("unexpected fill network interface types error: %w", err)
 			}
+			if _, err := virtualmachine.BackfillExtraConfigFromMoVM(vm, moVM); err != nil {
+				return fmt.Errorf("backfill extraconfig: %w", err)
+			}
+			if _, err := virtualmachine.BackfillNICDevicePropertiesFromMoVM(vm, moVM); err != nil {
```
bryanv: Can we fold this into the existing Type backfill, renaming that to something broader? There is a lot of overlap, and no need to later duplicate the matching logic.

hpannem: sure. that works!

pkg/providers/vsphere/upgrade/virtualmachine/backfill/nic.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// backfillNICSpec backfills vmxnet3.* spec fields from ExtraConfig using the
+// given prefix (e.g. "ethernet0."). Only applies to VMXNet3 or type-unset NICs.
+func backfillNICSpec(
+	prefix string,
+	iface *vmopv1.VirtualMachineNetworkInterfaceSpec,
+	extraConfig []vimtypes.BaseOptionValue) (bool, error) {
+
+	// Only backfill vmxnet3 fields for VMXNet3 (or type-unset) NICs.
```
bryanv: Type shouldn't be empty at this point, right?

pkg/providers/vsphere/upgrade/virtualmachine/vm_schema_upgrade.go
```diff
@@ -168,11 +169,11 @@ func ReconcileSchemaUpgrade(
...
-			if _, err := virtualmachine.FillEmptyNetworkInterfaceTypesFromMoVM(
-				vm,
-				moVM); err != nil {
-
-				return fmt.Errorf("unexpected fill network interface types error: %w", err)
+		if _, err := vmbackfill.ExtraConfigFromMoVM(vm, moVM); err != nil {
+					return fmt.Errorf("backfill extraconfig: %w", err)
+				}
+				if _, err := vmbackfill.NICConfigFromMoVM(vm, moVM); err != nil {
+				return fmt.Errorf("backfill nic: %w", err)
```
bryanv: Ident

pkg/util/vmopv1/extraconfig.go
```diff
@@ -0,0 +1,161 @@
...
+
+// DecodeVMXFieldValue decodes a raw vSphere ExtraConfig string value into rv.
+// rv must be settable and must be of kind Ptr or Slice.
+//
+// Supported pointer element types: bool, string (and string-based named types).
+// Supported slice element types: int32, string (and string-based named types).
+//
+// Bool decoding accepts "TRUE"/"FALSE" and "1"/"0" (case-insensitive).
+// []int32 is decoded from a comma-separated list of decimal integers.
+// []string-based slices are stored as a single-element slice containing raw.
```
bryanv: How strong is the vmx contract for the value of these fields?  Like "true" or empty is never, ever allowed? For an imported VM I wouldn't want us to get stuck here for what we consider an invalid value.

pkg/providers/vsphere/upgrade/virtualmachine/backfill/nic.go
hpannem: Correct.

pkg/util/vmopv1/extraconfig.go
hpannem: This needs a discussion. @bryanv I will talk to you offline.

hemanthk76in: Updated.

pkg/providers/vsphere/upgrade/virtualmachine/vm_schema_upgrade.go
```diff
@@ -168,11 +169,11 @@ func ReconcileSchemaUpgrade(
 
 	if features.TelcoVMServiceAPI {
 		if f := vmopv1util.FeatureVersionTelcoVMServiceAPI; !vmFeatureVersion.Has(f) {
-			if _, err := virtualmachine.FillEmptyNetworkInterfaceTypesFromMoVM(
-				vm,
-				moVM); err != nil {
-
-				return fmt.Errorf("unexpected fill network interface types error: %w", err)
```
bryanv, line 175: I had updated this error message to be "unexpected ..." because errors here are unexpected & bad, and it matches what earlier error are doing

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -3529,6 +3501,99 @@ func (v validator) validateFieldsDuringSchemaUpgrade(
...
+	if vm.Spec.Network != nil {
+		newIfaces = vm.Spec.Network.Interfaces
+	}
+	if oldVM.Spec.Network != nil {
+		oldIfaces = oldVM.Spec.Network.Interfaces
+	}
+
+	ifacesPath := specPath.Child("network").Child("interfaces")
+
+	for i := 0; i < min(len(newIfaces), len(oldIfaces)); i++ {
```
bryanv: I kind of think this should just be a deep equals on the entire interfaces. I don't think we want things added or removed, or swapping things. We could exclude bootstrap only fields like IP addresses but I don't think that is worth it, at least right now.

pkg/util/vmopv1/extraconfig.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+// NICVMXKeyMap returns the shared, lazily-built map of bare vmxnet3 property
+// names to field indices for VirtualMachineNetworkInterfaceVMXNet3Spec.
+func NICVMXKeyMap() map[string]int {
+	return cachedNICVMXKeyMap()
+}
+
+// BuildVMXKeyMap returns a map from vmx struct tag to field index for the given
+// struct type. Fields whose vmx tag contains "%d" (NIC-indexed) are excluded;
```
bryanv: For the splitting of these fields could we use like starts with `ethernet%d` instead?

bryanv (approved):

PR #1593, opened 2026-05-07
🌱 Add VirtualMachineService condition for underlying Service Ready

bryanv:

Reflect the underlying Service Ready condition on the VirtualMachineService conditions.

Note that most of our Service LB providers do not set the Ready condition. So it that case we'll report that as Unknown in our condition.

Add the special handling for the NSX nsx.vmware.com/hostnames annotation that is the driver for needing a ready condition in the first place.

Fixes # vmop-3559

```release-note
Add VirtualMachineServiceServiceReady condition to VirtualMachineService that reflects the Ready condition of the underlying Service.
```

----
📚 Documentation preview 📚: https://vm-operator--1593.org.readthedocs.build/en/1593/


api/v1alpha6/virtualmachineservice_types.go
```diff
@@ -30,6 +30,14 @@ const (
 	VirtualMachineServiceTypeExternalName VirtualMachineServiceType = "ExternalName"
 )
 
+const (
+	// ServiceReadyConditionType reflects the underlying Service Ready condition.
+	// Note that not all LoadBalancer providers set a Ready condition on the
+	// Service. When the condition is not present, this condition will have
+	// Status of Unknown with the Reason "ServiceReadyNotPresent".
+	ServiceReadyConditionType = "VirtualMachineServiceServiceReady"
```
faisalabujabal, line 38: can we not do it with one `Service`?

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -786,7 +786,57 @@ func (r *ReconcileVirtualMachineService) updateVMService(ctx *pkgctx.VirtualMach
 			}
 			vmService.Status.LoadBalancer.Ingress[idx] = vmIngress
 		}
+	} else {
+		vmService.Status.LoadBalancer.Ingress = nil
+	}
+
+	// Check for Ready condition in the underlying Service and reflect it on VirtualMachineService
+	if readyCond := getServiceReadyCondition(service); readyCond != nil {
+		switch readyCond.Status {
```
faisalabujabal, line 826: why not have the getServiceReadyCondition construct and update the condition making this method shorter. the getServiceReadyCondition doesn't do much as is

faisalabujabal (approved):

api/v1alpha6/virtualmachineservice_types.go
bryanv: I'm worried that name could be interpreted as the VirtualMachineService itself status. If we were to ever have that I'd prefer to use just Ready but we have other ready conditions with a type as the prefix.

controllers/virtualmachineservice/providers/loadbalancer_provider.go
```diff
@@ -24,6 +24,10 @@ const (
 	// this Service. Copied from kubernetes pkg/proxy/apis/well_known_labels.go to avoid
 	// k8s dependency.
 	LabelServiceProxyName = "service.kubernetes.io/service-proxy-name"
+
+	// AnnotationServiceNSXHostnamesKey is the NSX LB annotation used to specify
+	// the FQDN for certificate generation.
+	AnnotationServiceNSXHostnamesKey = "nsx.vmware.com/hostnames"
```
wenyingd: To align with external-dns behavior, annotation key "nsx.vmware.com/hostname" is used in nsx-operator at the moment, (no "s" in the suffix). Besides, according to the VCF rebranding requirement, the prefix domain "nsx.vmware.com" is possibly to be modified in the future. At the moment, we can use words "nsx.vmware.com/hostname"

controllers/virtualmachineservice/virtualmachineservice_controller.go
```diff
@@ -786,7 +786,64 @@ func (r *ReconcileVirtualMachineService) updateVMService(ctx *pkgctx.VirtualMach
...
+
+// syncServiceReadyCondition checks for Ready condition in the underlying Service
+// and reflects it on the VirtualMachineService.
+func syncServiceReadyCondition(
+	ctx *pkgctx.VirtualMachineServiceContext,
+	service *corev1.Service) {
+
+	var readyCond *metav1.Condition
+	for _, condition := range service.Status.Conditions {
+		if condition.Type == "Ready" {
```
ethanlynn-bc, line 819: Is it better to reuse vmop1.ReadyConditionType instead of hardcode "Ready" here?

controllers/virtualmachineservice/providers/loadbalancer_provider.go
bryanv: I've been given conflicting answers to what the annotation name is. I don't see it there - please add it to nsx-operator api then.

controllers/virtualmachineservice/virtualmachineservice_controller.go
bryanv: No. Even though the values are the same this is not a vmop condition so I'm not using a the const from our api

ethanlynn-bc (approved):

akutz (approved):
Thanks @bryanv !!!

PR #1602, opened 2026-05-13
:sparkles: Generate documentation from CRDs

aruneshpa:

This PR introduces the background work to generate API documentation from our types using crd-ref-docs.  The docs generated by this PR will be pushed to the developer portal via some automation alongside docs from other services.

A separate PR will integrate the doc generation with build process so we have reproducible documentation along with our binaries.

```release-note
Generate documentation from CRDs
```

### DIFF
--- a/api-docs/Makefile
+++ b/api-docs/Makefile
@@ -1,0 +1,84 @@
+# Makefile for API Documentation Generation
+
+# The api-docs image referenced here is being sourced from
+# https://github-vcf.devops.broadcom.net/vcf/wcp-ci/tree/master/api-doc-tool
+#
+# WARNING: The Go version baked into this image must be compatible with the
+# go directive in go.mod. If either is updated, verify the other still matches
+# or the vendor/docs-generation step will fail with a version mismatch.
+API_DOCS_IMAGE ?= wcp-docker-local.packages.vcfd.broadcom.net/tools/api-docs:699a6a4
+DOCKER_RUN := docker run --platform=linux/amd64 \
+  -v $(CURDIR):$(CURDIR) -w $(CURDIR) --rm $(API_DOCS_IMAGE)
+
+# Prefer native binaries when present on PATH (faster for local dev).
+# CI hosts that only ship the Docker image fall back to the container.
+CRD_REF_DOCS_NATIVE := $(shell command -v crd-ref-docs 2>/dev/null)
+ASCIIDOCTOR_NATIVE  := $(shell command -v asciidoctor 2>/dev/null)
+
+ifneq ($(CRD_REF_DOCS_NATIVE),)
+CRD_REF_DOCS := $(CRD_REF_DOCS_NATIVE)
+else
+CRD_REF_DOCS := $(DOCKER_RUN) crd-ref-docs
+endif
+
+ifneq ($(ASCIIDOCTOR_NATIVE),)
+ASCIIDOCTOR := $(ASCIIDOCTOR_NATIVE)
+else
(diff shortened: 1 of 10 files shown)


api-docs/go.mod
```diff
@@ -0,0 +1,29 @@
+module github.com/vmware-tanzu/vm-operator/api-docs
+
+go 1.24.4
```
faisalabujabal: we should sync with the version that we have

api-docs/main.go
```diff
@@ -0,0 +1,35 @@
...
+
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha4"
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha4/cloudinit"
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha4/common"
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha4/sysprep"
+
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha5/cloudinit"
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha5/common"
+	_ "github.com/vmware-tanzu/vm-operator/api/v1alpha5/sysprep"
```
faisalabujabal, line 30: what about v1alpha6? 

and curious if we are cross porting this to the 9.1 branch that we can remove v1alpha6 in the cherry picked commit

api-docs/Makefile
```diff
@@ -0,0 +1,80 @@
+# Makefile for API Documentation Generation
+
+# The api-docs image referenced here is being sourced from
+# https://github-vcf.devops.broadcom.net/vcf/wcp-ci/tree/master/api-doc-tool
+API_DOCS_IMAGE ?= wcp-docker-local.packages.vcfd.broadcom.net/tools/api-docs:92c1fac
```
faisalabujabal: 1. initially i thought that this domain should not be defined in an upstream repo, but I guess we already have it in other places. 
2. your plan is probably that you would atleast want a precommit pipeline to build this in upstream which will fail here, so we should replace this with a public image of asciidoctor and override this in the internal build systems makefile
3. remove the repo path in line 4

faisalabujabal (approved):

api-docs/go.mod
aruneshpa: Changed to keep it in sync with the api/ go.mod.

api-docs/Makefile
aruneshpa: I did consider either using asciidoctor from public artifactory, or re-building the image exactly like how it is being build internally.  However, the image being used from the artifactory here is not vanilla ascii doctor and contains additional fixes on top.  I don't intend to duplicate that here just yet.  I am discussing this with folks who push the image to our artifactory to see if we can move to an OSS version without custom fixes.  This can be a followup.

sidharthsurana:
LGTM

akutz:
Please revert this or collapse it into the current way the docs are generated. I do not want a new, top-level directory for this. In fact, due to it referencing internal code, this patch should be reverted and pushed into our internal Cayman repo. Please take care of this ASAP. Thanks!

aruneshpa:
> Please revert this or collapse it into the current way the docs are generated. I do not want a new, top-level directory for this. In fact, due to it referencing internal code, this patch should be reverted and pushed into our internal Cayman repo. Please take care of this ASAP. Thanks!

This is merged and crossported to the 9.1 release.  The the doc generation / tooling and process is governed by the org.

PR #1603, opened 2026-05-13
🐛 Validate max unit number for volume controllers

aruneshpa:

This PR adds validation to the VirtualMachine webhook to ensure that when a volume is explicitly attached to a controller, its unitNumber does not exceed the maximum allowed for that controller type.

For example, a SCSI controller (like PVSCSI) supports a maximum of 64 disks (unit numbers 0-64, skipping 7). If a user specifies a unitNumber of 65 or higher, the webhook will now reject the configuration.

Testing Done:
- Verified in a real env that when creating a VM that specifies invalid number of volumes (more than the available slots), API server rejects the request.  For all 4 types of controllers:
```
# SCSI
Error from server (spec.volumes[64].unitNumber: Invalid value: 65:
maximum unit number for SCSI controller is 64): error when creating
"65-vols-explicit.yaml": admission webhook
"default.validating.virtualmachine.v1alpha6.vmoperator.vmware.com"
denied the request: spec.volumes[64].unitNumber: Invalid value: 65:
maximum unit number for SCSI controller is 64

# NVME
Error from server (spec.volumes[0].unitNumber: Invalid value: 256:
maximum unit number for NVME controller is 255): error when creating
"invalid-nvme.yaml": admission webhook
"default.validating.virtualmachine.v1alpha6.vmoperator.vmware.com"
denied the request: spec.volumes[0].unitNumber: Invalid value: 256:
maximum unit number for NVME controller is 255

# IDE
Error from server (spec.volumes[0].unitNumber: Invalid value: 2:
maximum unit number for IDE controller is 1): error when creating
"invalid-ide.yaml": admission webhook
"default.validating.virtualmachine.v1alpha6.vmoperator.vmware.com"
denied the request: spec.volumes[0].unitNumber: Invalid value: 2:
maximum unit number for IDE controller is 1

# SATA
Error from server (spec.volumes[0].unitNumber: Invalid value: 30:
maximum unit number for SATA controller is 29): error when creating
"invalid-sata.yaml": admission webhook
"default.validating.virtualmachine.v1alpha6.vmoperator.vmware.com"
denied the request: spec.volumes[0].unitNumber: Invalid value: 30:
maximum unit number for SATA controller is 29
```

```release-note
Validate max unit number for volume controllers
```

### DIFF
--- a/webhooks/virtualmachine/validation/virtualmachine_validator.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator.go
@@ -1613,6 +1613,32 @@ func (v validator) validateControllerFields(
 		))
 	}
 
+	if vol.ControllerType != "" && vol.UnitNumber != nil {
+		var maxUnitNumber int32
+		switch vol.ControllerType {
+		case vmopv1.VirtualControllerTypeIDE:
+			maxUnitNumber = 1 // 2 slots: 0, 1
+		case vmopv1.VirtualControllerTypeNVME:
+			// NVMe supports 15 disks in older HW versions, but up to 255 in HW version 21.
+			// Since we don't know the HW version here, we use the maximum possible.
+			maxUnitNumber = 255
+		case vmopv1.VirtualControllerTypeSATA:
+			maxUnitNumber = 29 // 30 slots: 0-29
+		case vmopv1.VirtualControllerTypeSCSI:
+			// PVSCSI supports 64 disks (65 targets), others 15 disks (16 targets).
+			// We use 64 as the maximum possible unit number for SCSI to accommodate PVSCSI.
+			maxUnitNumber = 64
+		}
+
+		if *vol.UnitNumber > maxUnitNumber {
+			allErrs = append(allErrs, field.Invalid(
+				volPath.Child("unitNumber"),
+				*vol.UnitNumber,
+				fmt.Sprintf("maximum unit number for %s controller is %d", vol.ControllerType, maxUnitNumber),
+			))
+		}
+	}
+
 	return allErrs
 }
 
--- a/webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
+++ b/webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
@@ -482,6 +482,58 @@ func unitTestsValidateCreate() {
 					),
 				},
 			),
+			Entry("should deny PVC volume with unitNumber exceeding maximum for SCSI controller",
+				testParams{
+					setup: func(ctx *unitValidatingWebhookContext) {
+						ctx.vm.Spec.Volumes[0].ControllerType = vmopv1.VirtualControllerTypeSCSI
+						ctx.vm.Spec.Volumes[0].ControllerBusNumber = ptr.To(int32(0))
+						ctx.vm.Spec.Volumes[0].UnitNumber = ptr.To(int32(65))
+					},
+					expectAllowed: false,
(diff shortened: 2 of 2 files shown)


webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
```diff
@@ -482,6 +482,19 @@ func unitTestsValidateCreate() {
...
+			Entry("should deny PVC volume with unitNumber exceeding maximum for controllerType",
+				testParams{
+					setup: func(ctx *unitValidatingWebhookContext) {
+						ctx.vm.Spec.Volumes[0].ControllerType = vmopv1.VirtualControllerTypeSCSI
+						ctx.vm.Spec.Volumes[0].ControllerBusNumber = ptr.To(int32(0))
+						ctx.vm.Spec.Volumes[0].UnitNumber = ptr.To(int32(65))
+					},
+					expectAllowed: false,
+					validate: doValidateWithMsg(
+						field.Invalid(volPath.Index(0).Child("unitNumber"), int32(65), "maximum unit number for SCSI controller is 64").Error(),
```
bryanv: Can you please do the full error string? It think that makes it a whole lot easier to catch oopies in the field path like missing a field or something

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1611,6 +1611,32 @@ func (v validator) validateControllerFields(
 		))
 	}
 
+	if vol.ControllerType != "" && vol.UnitNumber != nil {
+		var maxUnitNumber int32
```
faisalabujabal, line 1617: we have MaxSlots() on the controller spec interface

faisalabujabal: and we already validate the unit number is within the range here https://github.com/vmware-tanzu/vm-operator/blob/main/webhooks/virtualmachine/validation/virtualmachine_validator_hardware_controllers.go#L306

aruneshpa: Yep, except all of those validations are skipped if the vm.spec.hardware is nil: https://github.com/vmware-tanzu/vm-operator/blob/main/webhooks/virtualmachine/validation/virtualmachine_validator_hardware_controllers.go#L66

And the HW is always nil at the time of Create because backfill hasn't happened yet.  Meaning effectively, we are not doing any per-slot validation during create.  So this is still needed?

bryanv: Instead of duplicating the magic numbers, we can create the ControllerSpec type and call its MaxSlots()? Might also want to see if this unit number is reserved?

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1613,6 +1613,32 @@ func (v validator) validateControllerFields(
...
+		switch vol.ControllerType {
+		case vmopv1.VirtualControllerTypeIDE:
+			maxUnitNumber = 1 // 2 slots: 0, 1
+		case vmopv1.VirtualControllerTypeNVME:
+			// NVMe supports 15 disks in older HW versions, but up to 255 in HW version 21.
+			// Since we don't know the HW version here, we use the maximum possible.
+			maxUnitNumber = 255
```
bryanv, line 1624: Isn't this 64? OTTOMH I that that additional was for PVSCSI with a special flag, not NVME?

PR #1604, opened 2026-05-14
🐛 fix: E2e tests gateway fixes

faisalabujabal:
Fix issues in the E2E tests for setting up encryption tests and content library tests:
- Fix how the gateway IP is resolved.
- Create a local content library for the content library tests.
- Fix govc environment variables

Fixes # NA

- I have moved one encryption test to experimental because I am having issues getting it to pass with the proxy setup. I'll have a follow-up story to remove it from the experimental suite

NA


aruneshpa (approved):

hack/e2e/kms.sh
```diff
@@ -8,23 +8,31 @@
...
 export GOVC_INSECURE=true
 GATEWAY_VM_USERNAME="${GATEWAY_VM_USERNAME:-root}"
-GATEWAY_VM_PASSWORD="${GATEWAY_VM_PASSWORD:-vmware}"
+# GATEWAY_VM_PASSWORD must be set by the caller (setup-e2e-testbed.sh passes
+# the discovered password). No default — empty fails fast.
+GATEWAY_VM_PASSWORD="${GATEWAY_VM_PASSWORD:-}"
 script_dir="$(dirname "$0")"
+
+# Common SSH/SCP options for all connections to the gateway VM.
+# -T: no PTY (avoids "Too many authentication failures" from the SSH agent)
```
bryanv, line 21: `-T` isn't in this list

hack/e2e/kms.sh
bryanv: Comment is now stale

hack/e2e/kms.sh
```diff
@@ -39,20 +47,29 @@ find_gateway_ip() {
...
-    openssl req -x509 -newkey rsa:4096 -sha256 -days 365 -nodes \
-            -subj "/C=US/ST=CA/L=PA/O=Broadcom/OU=VCF/CN=pykmip" \
-            -keyout "$crt_dir"/pykmip-key.pem -out "$crt_dir"/pykmip-crt.pem
-  fi
+  # gce2e-standard requires pykmip running on the gateway VM.
+  # Skip if already green (idempotent for parallel runners).
+  if kms_is_green "gce2e-standard"; then
+    echo "KMS provider gce2e-standard already green, skipping pykmip install"
+  else
+    if [ ! -e "$crt_dir/pykmip-crt.pem" ] ; then
```
bryanv, line 55: Can there be multiple runners doing this setup?

test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go
```diff
@@ -95,6 +100,11 @@ func VMEncryptionSpec(ctx context.Context, inputGetter func() VMEncryptionInput)
 		cancelPodWatches := framework.WatchPodLogsAndEventsInNamespaces(ctx, []string{config.GetVariable("VMOPNamespace")}, input.ClusterProxy.GetClientSet(), filepath.Join(input.ArtifactFolder, specName))
 		DeferCleanup(cancelPodWatches)
 
+		By("Ensure native key provider exists (create via WCP if not pre-configured)")
+		var nativeKPErr error
+		cleanupNativeKeyProvider, nativeKPErr = vcenter.EnsureNativeKeyProvider(ctx, vCenterClient, wcpClient, nativeKeyProviderID)
+		Expect(nativeKPErr).NotTo(HaveOccurred(), "failed to ensure native key provider %q", nativeKeyProviderID)
```
bryanv, line 106: Could we use `DeferCleanup(cleanupNativeKeyProvider)` here? I'd like to start using that more overall to make things more readable

test/e2e/vmservice/vmservice_suite_test.go
```diff
@@ -237,6 +245,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 			windowsVMIName, err := vmoperator.WaitForVirtualMachineImageName(ctx, &config.Config, svClusterProxy.GetClient(), wcpNamespaceName, windowsImageDisplayName)
 			Expect(err).ToNot(HaveOccurred(), "failed to get the VMI name in namespace %q", wcpNamespaceName)
 
+			By("Waiting for Windows image content to be fully downloaded")
+			vmoperator.WaitForVirtualMachineImageStatusDisks(ctx, &config.Config, svClusterProxy.GetClient(), wcpNamespaceName, windowsVMIName)
```
bryanv: If possible move this into the Sysprep test that uses the Windows image so the wait is only done when needed

Dockerfile.e2e
```diff
@@ -64,13 +71,28 @@ RUN tdnf update -y && \
...
 ARG GOVC_DOWNLOAD_URL=https://github.com/vmware/govmomi/releases/download/v0.53.0/govc_Linux_x86_64.tar.gz
 RUN curl -fsSL "${GOVC_DOWNLOAD_URL}" | tar -xz -C /usr/local/bin govc && \
     chmod +x /usr/local/bin/govc
 
+# packer — required by vmserviceapp E2E tests.
+ARG PACKER_DOWNLOAD_URL=https://releases.hashicorp.com/packer/1.9.1/packer_1.9.1_linux_amd64.zip
+RUN curl -fsSL "${PACKER_DOWNLOAD_URL}" -o /tmp/packer.zip && \
+    unzip /tmp/packer.zip -d /tmp/packer && \
+    mv /tmp/packer/packer /usr/local/bin/packer && \
+    chmod 0755 /usr/local/bin/packer && \
```
faisalabujabal, line 87: 🟡 **Medium**: External binaries (govc and packer) are downloaded and installed without checksum verification. This poses a security risk if the download source or the connection is compromised. Consider using a checksum (e.g., sha256sum) to verify the integrity of the downloaded files.

Makefile
```diff
@@ -959,6 +962,12 @@ e2e-image-build: ## Build E2E test container image
 		echo "Downloading kubectl $(E2E_KUBECTL_VERSION)..."; \
 		curl -fsSL "https://dl.k8s.io/release/$(E2E_KUBECTL_VERSION)/bin/linux/amd64/kubectl" -o kubectl && chmod +x kubectl; \
 	fi
+	@# Stage packer-plugin-vsphere source into the build context so the Dockerfile
+	@# can compile it without needing git credentials inside the Docker build.
+	@if [ ! -d packer-plugin-vsphere-src ]; then \
+		echo "Cloning packer-plugin-vsphere $(PACKER_PLUGIN_VSPHERE_REF)..."; \
+		git clone --depth=1 --branch $(PACKER_PLUGIN_VSPHERE_REF) $(PACKER_PLUGIN_VSPHERE_REPO) packer-plugin-vsphere-src; \
```
faisalabujabal, line 972: 🟡 **Medium**: The `git clone` logic does not detect if the existing `packer-plugin-vsphere-src` directory matches the current `PACKER_PLUGIN_VSPHERE_REF`. If the reference is updated in the Makefile, the build will continue to use stale source code unless the directory is manually deleted.

hack/e2e/install-pykmip.sh
```diff
@@ -17,10 +17,17 @@ set -x
...
+      pip
   fi
 
-  pip3 install pykmip
+  # PIP_INDEX_URL can be set by the caller to redirect pip to an internal
+  # package mirror (e.g. a corporate Artifactory instance). If unset, pip
+  # uses its default index (public PyPI).
+  pip3 install \
+    ${PIP_INDEX_URL:+--index-url "$PIP_INDEX_URL"} \
+    pykmip
```
faisalabujabal, line 30: 🟠 **High**: The shell expansion `${PIP_INDEX_URL:+--index-url "$PIP_INDEX_URL"}` is unquoted, which causes the internal double quotes to be treated as literal characters. This results in `pip` receiving a malformed URL (e.g., `"http://..."`). Additionally, it is vulnerable to word splitting and globbing if the URL contains spaces or special characters. Use `${PIP_INDEX_URL:+--index-url} ${PIP_INDEX_URL:+"$PIP_INDEX_URL"}` to correctly handle optional arguments with values.

hack/e2e/kms.sh
```diff
@@ -39,20 +47,30 @@ find_gateway_ip() {
...
-  sshpass -p "$password" ssh -o PubkeyAuthentication=no -o StrictHostKeyChecking=no "$target" /bin/bash ./install-pykmip.sh
+    sshpass -p "$password" scp $SSH_OPTS "$crt_dir"/pykmip-*.pem "$script_dir"/install-pykmip.sh "$target":
+    sshpass -p "$password" ssh -T $SSH_OPTS "$target" \
+      "PIP_INDEX_URL=${PIP_INDEX_URL:-} /bin/bash ./install-pykmip.sh" \
+      || echo "⚠ pykmip install failed — gce2e-standard KMS will not be available"
+  fi
 
-  setup "$2" || echo "KMS setup failed"
+  # setup() configures vCenter key providers; kms_is_green checks inside
+  # each block make it safe to call from multiple parallel runners.
```
faisalabujabal, line 76: 🟠 **High**: The `install` function crashes the script when no gateway IP is found. Because `set -o errexit` is active, the `scp` command to an invalid target (`root@`) terminates execution, preventing the `setup` call from configuring the `gce2e-native` provider, which is intended to be set up unconditionally.

```suggestion
    if [ -n "$2" ]; then
      target="$1@$2"
      password=$3
      sshpass -p "$password" scp $SSH_OPTS "$crt_dir"/pykmip-*.pem "$script_dir"/install-pykmip.sh "$target":
      sshpass -p "$password" ssh -T $SSH_OPTS "$target" \
        "PIP_INDEX_URL=${PIP_INDEX_URL:-} /bin/bash ./install-pykmip.sh" \
        || echo "⚠ pykmip install failed — gce2e-standard KMS will not be available"
    fi
```

hack/e2e/kms.sh
```diff
@@ -8,23 +8,31 @@
...
-  # NSX:
-  #  vm == external-vm-gateway
-  vm=$(govc find / -type m -name external*gateway 2>/dev/null || true)
+  # VDS:  vm == external-gateway-vds (or external-gateway)
+  # NSX:  vm == external-vm-gateway
+  # Use suffix wildcard to match all naming variants.
+  vm=$(govc find / -type m -name 'external-gateway*' 2>/dev/null || true)
+  if [ -z "$vm" ]; then
+    vm=$(govc find / -type m -name 'external-vm-gateway*' 2>/dev/null || true)
+  fi
```
faisalabujabal, line 35: 🟡 **Medium**: `govc find` with wildcards can return multiple VM paths if multiple matches exist in the inventory. If the `vm` variable contains multiple lines, the subsequent `govc vm.ip` call will fail to resolve an IP, causing the script to skip the `gce2e-standard` setup even if a valid gateway is available.

```suggestion
  vm=$(govc find / -type m -name 'external-gateway*' 2>/dev/null | head -n1 || true)
  if [ -z "$vm" ]; then
    vm=$(govc find / -type m -name 'external-vm-gateway*' 2>/dev/null | head -n1 || true)
  fi
```

hack/e2e/proxy.sh
```diff
@@ -14,16 +13,36 @@ set -x
 
 SCRIPT_DIR="$(dirname "${BASH_SOURCE}")"
 GOVC_INSECURE="${GOVC_INSECURE:-1}"
-GOVC_USERNAME="${GOVC_USERNAME:-administrator@vsphere.local}"
-GOVC_PASSWORD="${GOVC_PASSWORD:-vmware}"
+# GOVC_USERNAME and GOVC_PASSWORD are intentionally NOT defaulted here.
```
faisalabujabal, line 16: 🟡 **Medium**: The variable '${vmname}' is unquoted in the 'govc find' command. Since it now contains wildcards (e.g., 'external-gateway*'), the local shell might expand it to matching filenames in the current directory before passing it to govc, leading to incorrect results or syntax errors.

```suggestion
		vmpath=$(GOVC_URL=$vc GOVC_INSECURE=$GOVC_INSECURE GOVC_USERNAME=$GOVC_USERNAME GOVC_PASSWORD=$GOVC_PASSWORD govc find / -type m -name "${vmname}")
```

faisalabujabal, line 16: 🟡 **Medium**: With the introduction of wildcards, 'govc find' may return multiple VM paths. Returning them via unquoted 'echo ${vmpath}' joins them with spaces, which causes subsequent 'govc' commands (like 'vm.ip') to fail as they expect a single valid path.

```suggestion
			echo "${vmpath}" | head -n 1
```

hack/e2e/run-e2e.sh
```diff
@@ -37,15 +41,20 @@ PREBUILT_BIN="${E2E_PREBUILT_BINARY:-}"
...
 # Prebuilt binaries require the "--ginkgo." prefix for ginkgo-specific flags
 PREFIX=""
 [ "$MODE" = "prebuilt" ] && PREFIX="ginkgo."
 
-# 1. Initialize Ginkgo Args with verbosity and junit-report
-# Logic: --[ginkgo.]v --[ginkgo.]junit-report=...
-GINKGO_ARGS=("--${PREFIX}v" "--${PREFIX}junit-report=${REPORT_DIR}/test-results.xml")
+# 1. Initialize Ginkgo Args with verbosity, timeout, and junit-report
+# Logic: --[ginkgo.]v --[ginkgo.]timeout=... --[ginkgo.]junit-report=...
+GINKGO_ARGS=("--${PREFIX}v" "--${PREFIX}timeout=${GINKGO_TIMEOUT}" "--${PREFIX}junit-report=${REPORT_DIR}/test-results.xml")
```
faisalabujabal, line 57: 🟡 **Medium**: The JUnit report path uses 'REPORT_DIR', which defaults to '.' when 'E2E_ARTIFACT_FOLDER' is unset, while other test artifacts go to 'test_logs/'. This inconsistency can cause CI pipelines to miss the test results if they only collect the artifact folder.

```suggestion
GINKGO_ARGS=("--${PREFIX}v" "--${PREFIX}timeout=${GINKGO_TIMEOUT}" "--${PREFIX}junit-report=${ARTIFACT_FOLDER}/test-results.xml")
```

hack/e2e/setup-e2e-testbed.sh
```diff
@@ -26,23 +38,23 @@ echo "=== E2E Testbed Setup ==="
...
+    esac
+    _plugin_url="https://${WCP_IP}/wcp/plugin/${_plugin_os}/vsphere-plugin.zip"
+    _extract_dir="/tmp/vsphere-plugin-$$"
+    if curl --max-time 60 --retry 3 --retry-delay 5 -LOk "${_plugin_url}" 2>/dev/null; then
+        mkdir -p "${_extract_dir}"
+        unzip -o vsphere-plugin.zip -d "${_extract_dir}" >/dev/null 2>&1
+        # Just export PATH to wherever the binaries landed — no moving, no sudo.
+        export PATH="${_extract_dir}/bin:${PATH}"
+        rm -f vsphere-plugin.zip
+        echo "✓ kubectl-vsphere available at ${_extract_dir}/bin"
```
faisalabujabal: 🟡 **Medium**: The script updates the `PATH` and prints a success message even if the `unzip` command fails. The exit status of `unzip` should be verified to prevent adding a broken or empty directory to the search path.

```suggestion
        if unzip -o vsphere-plugin.zip -d "${_extract_dir}" >/dev/null 2>&1; then
            export PATH="${_extract_dir}/bin:${PATH}"
            echo "✓ kubectl-vsphere available at ${_extract_dir}/bin"
        fi
        rm -f vsphere-plugin.zip
```

hack/e2e/setup-testbed-env.sh
```diff
@@ -16,29 +16,9 @@
 #   source ./hack/e2e/setup-testbed-env.sh ./testbedinfo.json --e2e
 #
 
+set -u
```
faisalabujabal, line 19: 🟠 **High**: Hardcoding SCRIPT_SOURCED=true prevents the script from terminating when executed as a standalone script. The script_exit function will only return from itself, allowing the main script to continue execution with an invalid or incomplete configuration.

```suggestion
if [[ "${BASH_SOURCE[0]:-}" == "${0:-}" ]]; then SCRIPT_SOURCED=false; else SCRIPT_SOURCED=true; fi
```

hack/e2e/setup-testbed-env.sh
```diff
@@ -192,30 +178,30 @@ if [ -z "${VC_ROOT_PASSWORD}" ] || [ "${VC_ROOT_PASSWORD}" = "null" ]; then
...
 if [ -z "${VC_ROOT_USERNAME}" ] || [ "${VC_ROOT_USERNAME}" = "null" ]; then
-    export VC_ROOT_USERNAME="root"
-fi
-
-if [ -z "${VCSA_PASSWORD}" ] || [ "${VCSA_PASSWORD}" = "null" ]; then
-    export VCSA_PASSWORD="${VC_ROOT_PASSWORD}"
-    if [ -z "${VCSA_USERNAME}" ] || [ "${VCSA_USERNAME}" = "null" ]; then
-        export VCSA_USERNAME="administrator@vsphere.local"
-    fi
+    export VC_ROOT_USERNAME="administrator@vsphere.local"
```
faisalabujabal: 🟡 **Medium**: Changing the default VC_ROOT_USERNAME to 'administrator@vsphere.local' will likely break SSH-based credential extraction. This variable is used for SSH_USERNAME, and VCSA SSH access typically requires the 'root' user, whereas 'administrator@vsphere.local' is an API/SSO user.

```suggestion
    export VC_ROOT_USERNAME="root"
```

test/e2e/infrastructure/vsphere/vcenter/storage.go
```diff
@@ -311,3 +313,63 @@ func clusterConfiguredWithVsand(ctx context.Context, cluster *object.ClusterComp
...
+	DeleteKeyProvider(provider string) error
+	// BackupKeyProvider exports (backs up) the native key provider. vCenter
+	// requires a backup before the provider can be used for encryption.
+	BackupKeyProvider(provider string) error
+}
+
+// EnsureNativeKeyProvider ensures a native (vSphere-native) key provider with
+// the given providerID exists. It first checks for the provider via the vSphere
+// KMIP API (a read-only SOAP call), and if absent, creates it using the
+// supplied NativeKeyProviderManager. After creation it also backs up the
```
faisalabujabal, line 330: 🟡 **Medium**: The documentation for EnsureNativeKeyProvider states that it 'enables encryption on all hosts', but the implementation only calls BackupKeyProvider. If host enablement is a required step for vCenter encryption, it appears to be missing from the code.

test/e2e/vmservice/utils/encryption.go
```diff
@@ -0,0 +1,108 @@
...
+// absent we only warn since it is only available on VDS testbeds with a gateway.
+func EnsureEncryptionKeyProviders(ctx context.Context, clusterProxy wcpframework.WCPClusterProxyInterface) {
+	vCenterClient := vcenter.NewVimClientFromKubeconfig(ctx, clusterProxy.GetKubeconfigPath())
+	defer vcenter.LogoutVimClient(vCenterClient)
+
+	cryptoManager := crypto.NewManagerKmip(vCenterClient)
+
+	framework.Byf("Checking KMS key provider %q", nativeKeyProviderID)
+	if keyProviderExists(ctx, cryptoManager, nativeKeyProviderID) {
+		status, _ := cryptoManager.GetClusterStatus(ctx, nativeKeyProviderID)
```
faisalabujabal: 🟠 **High**: Potential nil pointer dereference: 'status' will be nil if 'GetClusterStatus' returns an error, causing a panic when accessing 'status.OverallStatus'.

```suggestion
		status, err := cryptoManager.GetClusterStatus(ctx, nativeKeyProviderID)
		if err == nil {
			framework.Byf("Native key provider %q status: %s", nativeKeyProviderID, status.OverallStatus)
		} else {
			framework.Byf("WARNING: failed to get status for native key provider %q: %v", nativeKeyProviderID, err)
		}
```

test/e2e/vmservice/utils/encryption.go
```diff
@@ -0,0 +1,108 @@
...
+		framework.Byf("Native key provider %q status: %s", nativeKeyProviderID, status.OverallStatus)
+	} else {
+		framework.Byf("WARNING: native key provider %q not found — "+
+			"encryption tests will attempt to create it in their own BeforeEach", nativeKeyProviderID)
+	}
+
+	framework.Byf("Checking KMS key provider %q", standardKeyProviderID)
+	if keyProviderExists(ctx, cryptoManager, standardKeyProviderID) {
+		status, _ := cryptoManager.GetClusterStatus(ctx, standardKeyProviderID)
```
faisalabujabal: 🟠 **High**: Potential nil pointer dereference: 'status' will be nil if 'GetClusterStatus' returns an error, causing a panic when accessing 'status.OverallStatus'.

```suggestion
		status, err := cryptoManager.GetClusterStatus(ctx, standardKeyProviderID)
		if err == nil {
			framework.Byf("Standard key provider %q status: %s", standardKeyProviderID, status.OverallStatus)
		} else {
			framework.Byf("WARNING: failed to get status for standard key provider %q: %v", standardKeyProviderID, err)
		}
```

test/e2e/vmservice/utils/encryption.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+	return false
+}
+
+// VerifyKeyProviderStatus asserts that the named key provider is registered and
+// has an acceptable status. Native providers may show "red" before first use.
+func VerifyKeyProviderStatus(ctx context.Context, cryptoManager *crypto.ManagerKmip, providerID, providerType string) {
+	framework.Byf("Verifying %s key provider %q", providerType, providerID)
+	Eventually(func(g Gomega) {
```
faisalabujabal, line 102: 🟡 **Medium**: Potential panic in 'Eventually' block: if 'GetClusterStatus' fails, 'status' is nil. The subsequent access to 'status.OverallStatus' will panic. Add a return or check after the error assertion.

```suggestion
		status, err := cryptoManager.GetClusterStatus(ctx, providerID)
		g.Expect(err).NotTo(HaveOccurred(), "failed to get status for key provider %s", providerID)
		if err != nil { return }
```

test/e2e/vmservice/vmservice/viadmin/contentlibraries.go
```diff
@@ -62,15 +66,51 @@ func VIAdminCLSpec(ctx context.Context, inputGetter func() VIAdminCLSpecInput) {
...
+						datastoreID = ds.Datastore
+						break
+					}
+				}
+				if datastoreID != "" {
+					break
+				}
+			}
+			Expect(datastoreID).NotTo(BeEmpty(), "no suitable datastore found for local content library")
+
```
faisalabujabal, line 93: 🟡 **Medium**: The datastore selection logic relies on hardcoded names ('vsanDatastore', 'sharedVmfs-0', 'nfs0-1'), which makes the test suite brittle and likely to fail in environments with different storage configurations. Consider falling back to the first available datastore if no preferred match is found.

```suggestion
			for _, dsName := range []string{"vsanDatastore", "sharedVmfs-0", "nfs0-1"} {
				for _, ds := range datastores {
					if ds.Name == dsName {
						datastoreID = ds.Datastore
						break
					}
				}
				if datastoreID != "" {
					break
				}
			}
			if datastoreID == "" && len(datastores) > 0 {
				datastoreID = datastores[0].Datastore
			}
```

test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go
```diff
@@ -358,30 +388,28 @@ func VMPublishRequestSpec(ctx context.Context, inputGetter func() VMPublishReque
 				Expect(publishedImageCRName).NotTo(BeEmpty(), "published VM Image resource name is empty")
 				vmoperator.WaitForVirtualMachineImageStatusDisks(ctx, &config.Config, svClusterClient, input.WCPNamespaceName, publishedImageCRName)
 
-				// Keep the published image attached to the namespace for the next test case.
-				deleteInventoryFolder = false
-
-				// Use the published the vmi to deploy a new VM should succeed with VM powered on and IP assigned.
+				// Deploy a new VM from the published image and verify it powers on with an IP.
```
faisalabujabal, line 391: 🟡 **Medium**: Inconsistent image readiness check. Use 'WaitForOVFVirtualMachineImageReady' instead of 'WaitForVirtualMachineImageStatusDisks' to ensure the image is fully ready before deployment, matching the update made on line 236 for the CLS context.

```suggestion
				vmoperator.WaitForOVFVirtualMachineImageReady(ctx, &config.Config, svClusterClient, input.WCPNamespaceName, publishedImageCRName)
```

test/e2e/vmservice/vmservice_suite_test.go
```diff
@@ -197,6 +199,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	Expect(vmservice.EnsureVMClassPresent(wcpClient, vmClassName)).To(Succeed())
 	Expect(vmservice.EnsureNamespaceHasAccess(wcpClient, vmClassName, wcpNamespaceName)).To(Succeed())
 
+	By("Ensure encryption key providers are configured for encryption tests")
```
faisalabujabal, line 202: 🟡 **Medium**: The return value of 'vmserviceutils.EnsureEncryptionKeyProviders' is ignored. This function likely performs vSphere operations that can fail; ignoring the error may lead to confusing failures in subsequent encryption tests.

```suggestion
	Expect(vmserviceutils.EnsureEncryptionKeyProviders(ctx, svClusterProxy)).To(Succeed())
```

test/e2e/vmservice/vmservice_suite_test.go
```diff
@@ -219,6 +224,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 		photonVMIName, err := vmoperator.WaitForVirtualMachineImageName(ctx, &config.Config, svClusterProxy.GetClient(), wcpNamespaceName, photonImageDisplayName)
 		Expect(err).ToNot(HaveOccurred(), "failed to get the VMI name in namespace %q", wcpNamespaceName)
 
+		By("Waiting for Photon image content to be fully downloaded")
```
faisalabujabal, line 227: 🟡 **Medium**: The return value of 'vmoperator.WaitForOVFVirtualMachineImageReady' is ignored. If the Photon image fails to become ready, the test will proceed to deploy a VM and likely fail with a less descriptive error.

```suggestion
		Expect(vmoperator.WaitForOVFVirtualMachineImageReady(ctx, &config.Config, svClusterProxy.GetClient(), wcpNamespaceName, photonVMIName)).To(Succeed())
```

test/e2e/vmservice/vmservice_suite_test.go
```diff
@@ -237,6 +245,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 			windowsVMIName, err := vmoperator.WaitForVirtualMachineImageName(ctx, &config.Config, svClusterProxy.GetClient(), wcpNamespaceName, windowsImageDisplayName)
 			Expect(err).ToNot(HaveOccurred(), "failed to get the VMI name in namespace %q", wcpNamespaceName)
 
+			By("Waiting for Windows image content to be fully downloaded")
```
faisalabujabal, line 248: 🟡 **Medium**: The return value of 'vmoperator.WaitForOVFVirtualMachineImageReady' is ignored for the Windows image setup, which may lead to race conditions during VM deployment.

```suggestion
			Expect(vmoperator.WaitForOVFVirtualMachineImageReady(ctx, &config.Config, svClusterProxy.GetClient(), wcpNamespaceName, windowsVMIName)).To(Succeed())
```

test/e2e/vmservice/vmserviceapp/templates/general-template.pkr.hcl
```diff
@@ -0,0 +1,119 @@
...
+  ssh_bastion_host = "${var.ssh_bastion_host}"
+  ssh_bastion_username = "${var.ssh_bastion_username}"
+  ssh_bastion_password = "${var.ssh_bastion_password}"
+  keep_input_artifact = "${var.keep_input_artifact}"
+  publish_location_name = "${var.publish_location_name}"
+  publish_image_name = "${var.publish_image_name}"
+  watch_source_timeout_sec = "${var.watch_source_timeout_sec}"
+  watch_publish_timeout_sec = "${var.watch_publish_timeout_sec}"
+}
+
```
faisalabujabal, line 111: 🟠 **High**: Using string interpolation for variables that can be null (e.g., 'source_name', 'kubeconfig_path', 'bootstrap_data_file') will cause a runtime error in Packer HCL2. Variables should be assigned directly (e.g., 'var.name') to allow null values to propagate correctly and to maintain type safety for boolean and numeric fields.

```suggestion
  kubeconfig_path      = var.kubeconfig_path
  supervisor_namespace = var.supervisor_namespace
  class_name           = var.class_name
```

test/e2e/vmservice/vmserviceapp/templates/jenkins-template.pkr.hcl
```diff
@@ -0,0 +1,156 @@
...
+      "until curl -s -o /dev/null -w '%%{http_code}' http://localhost:8080/ | grep -q '403\\|200'; do sleep 5; done",
+
+      "export JENKINS_URL=http://localhost:8080/",
+      "export USER=admin",
+      "export PASSWORD=$(sudo cat /var/lib/jenkins/secrets/initialAdminPassword)",
+      # Download Jenkins CLI to create and check job status.
+      "wget -q -O /tmp/jenkins-cli.jar $JENKINS_URL/jnlpJars/jenkins-cli.jar",      
+      # Create a new job from the above sample-job.xml file.
+      "java -jar /tmp/jenkins-cli.jar -s $JENKINS_URL -auth $USER:$PASSWORD create-job sample-job < /tmp/sample-job.xml",
+      # Build and wait for a successful completion of the job.
```
faisalabujabal, line 152: 🟠 **High**: The 'set -x' command enables shell tracing, which will leak the Jenkins admin password in the build logs when the 'PASSWORD' variable is expanded in the CLI commands on lines 150 and 152.

```suggestion
      "set +x",
      "export PASSWORD=$(sudo cat /var/lib/jenkins/secrets/initialAdminPassword)",
      "java -jar /tmp/jenkins-cli.jar -s $JENKINS_URL -auth $USER:$PASSWORD create-job sample-job < /tmp/sample-job.xml"
```

test/e2e/vmservice/vmserviceapp/templates/general-template.pkr.hcl
faisalabujabal: those were copied directly from the repo, so lets not touch them

test/e2e/vmservice/vmserviceapp/templates/jenkins-template.pkr.hcl
faisalabujabal: those were copied directly from the packer repo, so lets not touch them

test/e2e/vmservice/vmservice_suite_test.go
faisalabujabal: we are refactoring this in the next PR so we can ignore it for now

test/e2e/vmservice/vmservice_suite_test.go
faisalabujabal: i will refactor this in the next PR so would rather not touch it now

hack/e2e/kms.sh
faisalabujabal: yes and it is handled now

test/e2e/vmservice/vmservice/viadmin/contentlibraries.go
faisalabujabal: no we don't want the fallback

PR #1605, opened 2026-05-15
✨ Implement `spec.advanced.extraConfig` and first-class VMX property reconciliation

hpannem:

Implements the ExtraConfig reconciliation pipeline for `VirtualMachine` under the `TelcoVMServiceAPI` feature gate. This covers two layers of VM-level advanced configuration:

1. **First-class fields** (`spec.advanced.preferHTEnabled`, `hugePages1GEnabled`, etc.) — mapped to vSphere VMX keys via `vmx` struct tags and applied according to their `vmxmode` tag (`live` / `powercycle` / `poweroff`). A symmetric encode/decode path ensures no spurious reconfigurations when vSphere normalizes a value (e.g. `"true"` → `"TRUE"`).

2. **Bag keys** (`spec.advanced.extraConfig`) — arbitrary VMX key/value pairs tracked via a managed-keys annotation (`vmservice.extraconfig.managedKeys`) so stale entries are cleaned up when a key is removed from the spec.

The reconcile pipeline:
- Overlays bag keys (filtered for reserved/first-class keys), managed-key cleanup, and all first-class translations into `allDesiredEC`
- Runs a single `SemanticDiff` pass (decoding both sides) to suppress no-op updates
- Routes the result by `vmxmode`: PowerOff-mode entries that cannot be applied while the VM is on are deferred; PowerCycle-mode entries are applied but `vmx.reboot.powerCycle=TRUE` is injected
- Reports status via `VirtualMachineExtraConfigSynced` condition (`PowerOffRequired` / `PowerCyclePending` / `True`) and `status.extraConfig` (reflecting the VM's current observed values for user-controlled keys)

Reserved key validation is centralised in `vmopv1util.IsSystemReservedExtraConfigKey` and shared by both the webhook and the reconciler.

Also normalises JSON field name casing for three fields as per k8s standards: `pNUMANodeAffinity` → `pnumaNodeAffinity`, `vNUMANodeID` → `vnumaNodeID`, `preferHTEnabled` → `preferHtEnabled`, `udpRSSEnabled` → `udpRssEnabled`.

Fixes #

- `TelcoVMServiceAPI` feature gate must be enabled; the reconciler is a no-op otherwise.
- `status.extraConfig` reflects the VM's **observed** state (one-cycle lag), not the desired spec state.
- `HugePages1GEnabled` is `poweroff` mode (VM must be off to apply); the other five first-class fields are `powercycle` mode.

```release-note
Add ExtraConfig reconciliation for VirtualMachine spec.advanced fields and
spec.advanced.extraConfig bag keys under the TelcoVMServiceAPI feature gate.
Reports sync state via VirtualMachineExtraConfigSynced condition.

----
📚 Documentation preview 📚: https://vm-operator--1605.org.readthedocs.build/en/1605/

### CODE REVIEW

File: pkg/crd/crd_test.go [hunk 1]
```diff
@@ -427,15 +427,15 @@ var _ = Describe("Install", func() {
 					fields := specFieldPath(field)
 					assertField(true, fields...)
 				},
-				Entry("advanced preferHTEnabled", "advanced.preferHTEnabled"),
+				Entry("advanced preferHtEnabled", "advanced.preferHtEnabled"),
```
faisalabujabal, line 511: 🟡 **Medium**: The 'advanced' fields (e.g., preferHtEnabled, pnumaNodeAffinity) are being checked in the 'vSphere policies are enabled' test block. These fields are gated by 'TelcoVMServiceAPI', not 'VSpherePolicies', which may lead to incorrect test results or hidden gating bugs.

pkg/crd/crd_test.go
```diff
@@ -850,15 +850,15 @@ var _ = Describe("Install", func() {
 						Entry("volumes diskMode", "volumes.[].diskMode"),
 						Entry("volumes sharingMode", "volumes.[].sharingMode"),
 						Entry("volumes unitNumber", "volumes.[].unitNumber"),
-						Entry("advanced preferHTEnabled", "advanced.preferHTEnabled"),
+						Entry("advanced preferHtEnabled", "advanced.preferHtEnabled"),
```
faisalabujabal, line 989: 🟡 **Medium**: The 'advanced' fields are redundantly checked in the 'VM shared disks (OracleRAC) is enabled' test block. This duplication is unnecessary and risks verifying fields under the wrong feature gate context.

pkg/providers/vsphere/virtualmachine/extraconfig/managed.go
```diff
@@ -0,0 +1,112 @@
...
+		decodedDesired := reflect.New(fieldType).Elem()
+		if err := vmopv1util.DecodeVMXFieldValue(ctx, decodedDesired, desiredStr); err != nil {
+			log.V(1).Error(err, "unsupported field type for VMX decode; emitting entry unconditionally",
+				"key", kv.Key)
+		}
+		desiredValue := decodedDesired.Interface()
+
+		decodedObserved := reflect.New(fieldType).Elem()
+		if isObserved && observedStr != "" {
+			if err := vmopv1util.DecodeVMXFieldValue(ctx, decodedObserved, observedStr); err != nil {
```
faisalabujabal, line 100: 🟡 **Medium**: The logic does not match the log message when decoding fails. If 'DecodeVMXFieldValue' returns an error, the entry is not necessarily emitted; it may be filtered out if the zero-value comparison matches the observed state. To match the stated intent, the entry should be appended and the loop continued.

```suggestion
if err := vmopv1util.DecodeVMXFieldValue(ctx, decodedDesired, desiredStr); err != nil {
	log.V(1).Error(err, "unsupported field type for VMX decode; emitting entry unconditionally",
		"key", kv.Key)
	out = append(out, &vimtypes.OptionValue{Key: kv.Key, Value: kv.Value})
	continue
}
```

pkg/providers/vsphere/virtualmachine/extraconfig/translate_vm.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+		elem := fv.Elem()
+		switch elem.Kind() {
+		case reflect.Bool:
+			return vmopv1util.EncodeVMXBoolField(elem.Type(), elem.Bool())
+		case reflect.String:
+			return vmopv1util.EncodeVMXStringField(elem.Type(), elem.String())
+		}
+	case reflect.Slice:
+		if fv.IsNil() || fv.Len() == 0 {
```
faisalabujabal, line 73: 🟠 **High**: The `Ptr` kind switch is missing a case for `Int32`. Since `vNUMANodeID` in `VirtualMachineAdvancedSpec` is a `*int32`, it will be incorrectly translated to an empty string even when set, causing the reconciler to clear the value on the VM.

```suggestion
		case reflect.Bool:
			return vmopv1util.EncodeVMXBoolField(elem.Type(), elem.Bool())
		case reflect.String:
			return vmopv1util.EncodeVMXStringField(elem.Type(), elem.String())
		case reflect.Int32:
			return fmt.Sprintf("%d", elem.Int())
		}
```

pkg/providers/vsphere/virtualmachine/extraconfig/translate_vm.go
```diff
@@ -0,0 +1,82 @@
...
+//
+// Each non-zero field is converted to vSphere wire format via EncodeVMX*Field,
+// which applies a registered custom encoder for the field's type when available,
+// or falls back to the default encoding (*bool → "TRUE"/"FALSE", []int32 →
+// comma-separated decimals, []string → comma-joined). nil/zero fields emit ""
+// to signal clear-if-present to the caller.
+func TranslateFirstClass(advanced *vmopv1.VirtualMachineAdvancedSpec) pkgutil.OptionValues {
+	if advanced == nil {
+		return nil
+	}
```
faisalabujabal, line 34: 🟡 **Medium**: Returning `nil` when `advanced` is `nil` may prevent the reconciler from clearing previously set first-class VMX keys. If the intention is to sync the state, a `nil` spec should likely result in all mapped VMX keys being set to an empty string to ensure they are removed from the VM.

pkg/providers/vsphere/virtualmachine/extraconfig/translate_vm.go
```diff
@@ -0,0 +1,82 @@
...
+			return strings.Join(parts, ",")
+		case reflect.String:
+			strs := make([]string, fv.Len())
+			for i := range strs {
+				strs[i] = fv.Index(i).String()
+			}
+			return vmopv1util.EncodeVMXSliceStringField(elem, strs)
+		}
+	}
+	return ""
```
faisalabujabal: 🟡 **Medium**: The `translateFieldValue` function returns an empty string for any unhandled `reflect.Kind`. This is interpreted as a signal to delete the VMX key. It should ideally handle all expected types or provide a way to distinguish between an intentional 'clear' and an unsupported type to avoid accidental configuration loss.

pkg/util/vmopv1/extraconfig.go
```diff
@@ -114,29 +185,92 @@ type vmxSliceDecoder func(raw string) []string
...
+	return "FALSE"
+}
+
+// EncodeVMXSliceStringField converts a spec slice of string values to its
+// vSphere wire format. For types with a registered encoder (e.g. PNICQueueFeature
+// uses a bitmask integer) the encoded value is returned; for unregistered types
+// the values are joined with commas.
+func EncodeVMXSliceStringField(elemType reflect.Type, vals []string) string {
+	if enc, ok := vmxSliceEncoders[elemType]; ok {
+		return enc(vals)
```
faisalabujabal, line 271: 🟡 **Medium**: EncodeVMXSliceStringField joins slice elements with commas, but the corresponding decoder (decodeSliceVMXField) does not split by commas for unregistered string slice types. This inconsistency will cause reconciliation loops for any such field.

pkg/util/vmopv1/extraconfig.go
```diff
@@ -392,6 +559,37 @@ func decodePNICQueueFeatures(raw string) []string {
...
+			if err == nil {
+				n |= power
+			}
+		}
+	}
+	if n == 0 {
+		return ""
+	}
+	return strconv.FormatUint(n, 10)
+}
```
faisalabujabal, line 595: 🟡 **Medium**: The encoder for PNICQueueFeatures does not apply the same bitmask as the decoder (PNICFeaturesMaxItems), which can lead to reconciliation loops if the spec contains values outside the supported bit range.

```suggestion
func encodePNICQueueFeatures(vals []string) string {
	var n uint64
	for _, v := range vals {
		switch v {
		case string(vmopv1.PNICQueueFeatureLargeReceiveOffload):
			n |= 1
		case string(vmopv1.PNICQueueFeatureReceiveSideScaling):
			n |= 4
		default:
			power, err := strconv.ParseUint(v, 10, 64)
			if err == nil {
				n |= power
			}
		}
	}
	n &= (1 << vmopv1.PNICFeaturesMaxItems) - 1
	if n == 0 {
		return ""
	}
	return strconv.FormatUint(n, 10)
}
```

pkg/vmconfig/extraconfig/reconciler.go
```diff
@@ -0,0 +1,301 @@
...
+		conditions.MarkFalse(
+			vm,
+			vmopv1.VirtualMachineExtraConfigSynced,
+			ReasonError,
+			"%v", resultErr)
+		return nil
+	}
+
+	// Compute status.ExtraConfig: reflect what the VM currently has for user-controlled keys.
+	if moVM.Config != nil {
```
faisalabujabal, line 261: 🟡 **Medium**: Iterating over firstClassKeys (a map) to populate vm.Status.ExtraConfig results in non-deterministic ordering. This causes unnecessary status updates to the VirtualMachine object on every reconciliation cycle. Collect and sort the keys before iterating.

```suggestion
		keys := make([]string, 0, len(firstClassKeys))
		for k := range firstClassKeys {
			keys = append(keys, k)
		}
		sort.Strings(keys)
		for _, k := range keys {
			if v, ok := observed.GetString(k); ok && v != "" {
				statusEC = append(statusEC, vmopv1common.KeyValuePair{Key: k, Value: v})
			}
		}
```

pkg/crd/crd_test.go
hpannem: No. This looks like stale comment and not correct.

pkg/crd/crd_test.go
hpannem: general test for all the fields. Not applicable.

pkg/providers/vsphere/virtualmachine/extraconfig/managed.go
hpannem: In practice, decoding desired should never fail, since we store in the spec with the correct type. So, this is a non-issue. If both desired and observed decode fails, we can't do anyway.

Changing the log message though...

pkg/providers/vsphere/virtualmachine/extraconfig/translate_vm.go
hpannem: vNUMANodeID is not a vmx property. We only support *bool and *string. So, this is a non-issue.

pkg/util/vmopv1/extraconfig.go
hpannem: Yes. This is reasonable comment, but this part of the code is not exercised now, as all the props have decoders. Will fix it for completeness.

pkg/util/vmopv1/extraconfig.go
hpannem: makes sense to keep the encode and decode in sync with supported max items.

pkg/vmconfig/extraconfig/reconciler.go
hpannem: Good point. Fixed it.

pkg/providers/vsphere/virtualmachine/extraconfig/translate_vm.go
hpannem: The caller already sets empty object if nil. So, this is a non-issue.

pkg/providers/vsphere/virtualmachine/extraconfig/translate_vm.go
hpannem: Does not apply as existing types are already supported. However, I have made a fix to skip the prop if a new type introduced.

pkg/vmconfig/extraconfig/reconciler_test.go
```diff
@@ -0,0 +1,436 @@
...
+	It("injects vmx.reboot.powerCycle when powered-on VM has pending first-class update", func() {
+		vm.Status.PowerState = vmopv1.VirtualMachinePowerStateOn
+		vm.Spec.Advanced = &vmopv1.VirtualMachineAdvancedSpec{PreferHTEnabled: ptr.To(false)}
+		moVM = moVMWithEC("numa.vcpu.preferHT", "TRUE")
+		Expect(r.Reconcile(ctx, nil, nil, vm, moVM, configSpec)).To(Succeed())
+		Expect(getECVal(configSpec.ExtraConfig,
+			vsphereconst.ExtraConfigReservedKeyVMXRebootPowerCycle)).To(Equal("TRUE"))
+	})
+
+	It("does NOT inject vmx.reboot.powerCycle for bag-key-only changes", func() {
```
bryanv, line 156: Some bag tests like this one don't assert ExtraConfigManagedKeysKey - shouldn't they?

pkg/vmconfig/extraconfig/reconciler_test.go
```diff
@@ -0,0 +1,436 @@
...
+			vsphereconst.ExtraConfigReservedKeyVMXRebootPowerCycle)).To(BeEmpty())
+	})
+
+	It("emits update when bag key value changes", func() {
+		vm.Spec.Advanced = &vmopv1.VirtualMachineAdvancedSpec{
+			ExtraConfig: []vmopv1common.KeyValuePair{{Key: "foo", Value: "new"}},
+		}
+		moVM = moVMWithEC("foo", "old", vsphereconst.ExtraConfigManagedKeysKey, "foo")
+		Expect(r.Reconcile(ctx, nil, nil, vm, moVM, configSpec)).To(Succeed())
+		Expect(getECVal(configSpec.ExtraConfig, "foo")).To(Equal("new"))
```
bryanv, line 204: I added a `Expect(getECVal(configSpec.ExtraConfig, vsphereconst.ExtraConfigManagedKeysKey)).To(Equal("foo"))
` here and it fails. We don't consider that managed at this point like "emits new bag key and writes managed-keys tracking entry" test? (I think the two tests between this one and that one should be moved to elsewhere in this file)

pkg/vmconfig/extraconfig/reconciler_test.go
```diff
@@ -0,0 +1,436 @@
...
+		kv := bov.GetOptionValue()
+		if kv != nil && kv.Key == key {
+			v, _ := kv.Value.(string)
+			return v
+		}
+	}
+	return ""
+}
+
+type assertError string
```
bryanv: Can't we just use `errors.New()` instead?

pkg/vmconfig/extraconfig/reconciler.go
```diff
@@ -0,0 +1,306 @@
...
+
+	// Collect overlay entries to apply on top of configSpec.ExtraConfig.
+	var overlay pkgutil.OptionValues
+
+	// Overlay bag keys from spec.advanced.extraConfig, skipping reserved and first-class keys.
+	specBagKeys := make(map[string]bool)
+	if vm.Spec.Advanced != nil {
+		for _, kv := range vm.Spec.Advanced.ExtraConfig {
+			if vmopv1util.IsSystemReservedExtraConfigKey(kv.Key) || isFirstClassKey(kv.Key, firstClassKeys) {
+				continue
```
bryanv, line 134: Could comment there that the validation webhook is suppose to prevent this (right?)

pkg/vmconfig/extraconfig/reconciler.go
```diff
@@ -0,0 +1,306 @@
...
+			ReasonError,
+			"%v", resultErr)
+		return nil
+	}
+
+	// Compute status.ExtraConfig: reflect what the VM currently has for user-controlled keys.
+	if moVM.Config != nil {
+		observed := pkgutil.OptionValues(moVM.Config.ExtraConfig)
+		firstClassKeys := vmopv1util.AdvancedVMXKeyMap()
+		sortedKeys := make([]string, 0, len(firstClassKeys))
```
bryanv: Isn't this something we could just precompute?

pkg/providers/vsphere/constants/constants.go
```diff
@@ -57,9 +57,14 @@ const (
...
+	// ExtraConfigManagedKeysKey is the ExtraConfig key that tracks which
+	// spec.advanced.extraConfig keys have been written by vm-operator, so they
+	// can be removed when the user deletes them from the spec.
+	ExtraConfigManagedKeysKey = "vmservice.extraconfig.managedKeys"
+
 	// ExtraConfig reserved prefixes — vm-operator controls these and users must not set them directly.
-	ExtraConfigReservedPrefixVMService          = "vmservice."
-	ExtraConfigReservedKeyVMXRebootPowerCycle   = "vmx.reboot.powerCycle"
+	ExtraConfigReservedPrefixVMService        = "vmservice."
+	ExtraConfigReservedKeyVMXRebootPowerCycle = "vmx.reboot.powerCycle"
```
bryanv, line 76: Might be good for ExtraConfigReservedKeyVMXRebootPowerCycle to get its own comment about what it is doing & works

bryanv (commented):
I think we should add a vmprovider_vm_extraconfig_test.go that does a few basic sanity checks on the flow through the vm CreateOrUpdate.

hpannem:
> I think we should add a vmprovider_vm_extraconfig_test.go that does a few basic sanity checks on the flow through the vm CreateOrUpdate.

sure. Added.

pkg/vmconfig/extraconfig/reconciler_test.go
hpannem: agree. Done.

pkg/vmconfig/extraconfig/reconciler.go
hpannem: Added. Defense in Depth!

pkg/vmconfig/extraconfig/reconciler_test.go
hpannem: Probably there is confusion. No, as we track only key names as managed keys, that will not change if only value changes. So, the resulting config spec will not contain this a nothing is changed. I added an assert that the key should not be emitted. Reordered tests so that they are placed together.

bryanv (approved):

PR #1607, opened 2026-05-16
✨ api/v1alpha6: add VM compute reconfiguration schema

hpannem:

Adds three new optional fields to `VirtualMachineSpec`, gated behind the
`TelcoVMServiceAPI` capability, for Telco VNF compute reconfiguration:

- `spec.resources` — vCPU/memory size, host-level MHz/byte reservations and limits
- `spec.cpuAdvanced` — latency sensitivity, CPU topology, hot-add, IOMMU, nested HV, vPMC
- `spec.memoryAdvanced` — memory hot-add, reservation-locked-to-max

Includes: kubebuilder validation (CEL milli-suffix guard, Minimum=1), deepcopy
regeneration, hub–spoke restore functions for all five spoke versions
(v1alpha1–v1alpha5), hub–spoke–hub roundtrip tests for all new fields, CRD
capability gating with test coverage, and regenerated CRD manifest.

Fixes #

`requests/limits.cpu` values are in **MHz** (not millicores). A CEL rule blocks
the `m` suffix (`"2000m"` = 2 MHz, not 2000 MHz).

```release-note
Add spec.resources, spec.cpuAdvanced, and spec.memoryAdvanced to VirtualMachine
for Telco compute reconfiguration. Requires the TelcoVMServiceAPI capability.

----
📚 Documentation preview 📚: https://vm-operator--1607.org.readthedocs.build/en/1607/

### CODE REVIEW

File: api/test/v1alpha1/conversion_test.go [hunk 1]
```diff
@@ -287,6 +287,10 @@ func overrideVirtualMachineFieldsFuncs(codecs runtimeserializer.CodecFactory) []
 				vmSpec.Class = nil
 			}
 
+			vmSpec.Resources = nil
```
bryanv: What about the fuzzer trips up these fields?

api/v1alpha6/virtualmachine_compute_types.go
```diff
@@ -0,0 +1,219 @@
...
+	//
+	// For spec.resources.requests and spec.resources.limits, the value
+	// represents a host-level CPU allocation in MHz (e.g. "2000" for
+	// 2 GHz). Maps to CpuAllocation.Reservation and CpuAllocation.Limit
+	// respectively.
+	CPU resource.Quantity `json:"cpu,omitempty"`
+
+	// +optional
+
+	// Memory is a memory resource quantity (e.g. "8Gi").
```
bryanv: Maybe it is little confusing to use `(e.g. "8Gi")` here when that only applies when it is used in the `size` context?

api/v1alpha6/virtualmachine_compute_types.go
```diff
@@ -0,0 +1,219 @@
...
+	//
+	// For spec.resources.size, the value represents the guest-visible vCPU
+	// count and must be a whole number (e.g. "4"). Maps to
+	// ConfigSpec.NumCPUs.
+	//
+	// For spec.resources.requests and spec.resources.limits, the value
+	// represents a host-level CPU allocation in MHz (e.g. "2000" for
+	// 2 GHz). Maps to CpuAllocation.Reservation and CpuAllocation.Limit
+	// respectively.
+	CPU resource.Quantity `json:"cpu,omitempty"`
```
bryanv: Do these need to be pointers?

api/v1alpha6/virtualmachine_compute_types.go
```diff
@@ -0,0 +1,219 @@
...
+	// Maps to ConfigSpec.NumCPUs and ConfigSpec.MemoryMB.
+	// When set, overrides the corresponding VMClass hardware values.
+	Size *VirtualMachineResourceQuantity `json:"size,omitempty"`
+
+	// +optional
+
+	// Requests is the host-level resource reservation (host guarantee).
+	// requests.cpu is in MHz; requests.memory is in bytes.
+	// Maps to CpuAllocation.Reservation and MemoryAllocation.Reservation.
+	// Changes apply immediately on a powered-on VM.
```
bryanv: I'm worried "immediately" might give the wrong impression. Maybe just say something about the VM being powered on?

api/v1alpha6/virtualmachine_compute_types.go
```diff
@@ -0,0 +1,219 @@
...
+	// allowing users to tune the NUMA topology visible to the guest OS.
+	// Maps to ConfigSpec.NumCoresPerSocket.
+	// Requires power-off to apply.
+	CoresPerSocket *int32 `json:"coresPerSocket,omitempty"`
+
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+
+	// CoresPerNUMANode sets the number of cores per virtual NUMA node.
+	// When set, virtual NUMA node count = total vCPU count / CoresPerNUMANode.
```
bryanv: Just wondering what happens when CPU count isn't divisible by per numa?

api/test/v1alpha1/conversion_test.go
hpannem: Done.

api/v1alpha6/virtualmachine_compute_types.go
hpannem: This applies all the memory fields requests and limits as well.

api/v1alpha6/virtualmachine_compute_types.go
hpannem: Yes. You are right. Modfiied.

api/v1alpha6/virtualmachine_compute_types.go
hpannem: rephrased.

api/v1alpha6/virtualmachine_compute_types.go
hpannem: I modified this field to say vNUMAnodeCount as shown in vsphereUI instead of coresPerNumaNode. Also, I added description and constraints as well. To your question, the condition is corespersocket is a divisor or multiple of corespernumanode.

api/v1alpha6/virtualmachine_compute_types.go
bryanv: OK. Yes. I had in my mind that the reservations were percentages.

api/test/v1alpha1/conversion_test.go
```diff
@@ -352,6 +354,18 @@ func overrideVirtualMachineFieldsFuncs(codecs runtimeserializer.CodecFactory) []
 		func(msg *json.RawMessage, c randfill.Continue) {
 			*msg = []byte(`{"foo":"bar"}`)
 		},
+		func(q *vmopv1.VirtualMachineResourceQuantity, c randfill.Continue) {
```
bryanv: Did things fail without this? `resource.Quantity` is a part of the upstream fuzzer func added at   https://github.com/vmware-tanzu/vm-operator/blob/8aae3333be7812d0ae360b3d7e39e6614a53f9df/api/test/utilconversion/fuzztests/conversion_fuzz.go#L40

api/v1alpha2/virtualmachine_conversion.go
```diff
@@ -98,6 +98,24 @@ func restore_v1alpha6_VirtualMachineAdvanced(dst, src *vmopv1.VirtualMachine) {
 	dst.Spec.Advanced.ExtraConfig = adv.ExtraConfig
 }
 
+func restore_v1alpha6_VirtualMachineResources(dst, src *vmopv1.VirtualMachine) {
+	if src.Spec.Resources != nil {
+		dst.Spec.Resources = src.Spec.Resources.DeepCopy()
```
bryanv: These DeepCopy's shouldn't be needed. src is the just unmarshalled saved object so it is fine just overiding. I know some other DeepCopy/Clones have cropped in recently I'd like to not add to it since these conversion webhooks can be called very, very frequently.

api/test/v1alpha1/conversion_test.go
hpannem: You are right. I have not looked inside the metafuzzer. I thought the quantity has unexposed fields and hence need a custom fuzzer. Looks like there is a custom fuzzer that tests resource.Quantity.  
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/fuzzer/fuzzer.go

So, we are good without this.

api/v1alpha2/virtualmachine_conversion.go
hpannem: correct. Probably AI slop!

bryanv (approved):
Thanks!

PR #1613, opened 2026-05-20
🐛  Add ChangeTrackingEnabled to watched VM property

bryanv:

We'll reconfigure the VM to update the ChangeTrackingEnabled prop when the VM Spec ChangeBlockTracking is updated. We need to include this field in the watcher so that we will reconcile the VM after the reconfigure.

Without this, if the VM is powered on and this was the only field in the reconfigure, we will not timely reconcile the VM to do the InvokeFSR task to actually enable CBT.

Fixes #

```release-note
Add the VM config.ChangeTrackingEnabled property path to the VM watcher service so Reconfigures that change this property will result in an event and later reconciliation. 
```

### DIFF
--- a/pkg/util/vsphere/watcher/watcher.go
+++ b/pkg/util/vsphere/watcher/watcher.go
@@ -37,6 +37,7 @@ func DefaultWatchedPropertyPaths() []string {
 	return []string{
 		"config.annotation",
 
+		"config.changeTrackingEnabled",
 		"config.extraConfig",
 		"config.hardware.device",
 		"config.keyId",


pkg/util/vsphere/watcher/watcher.go
```diff
@@ -37,6 +37,7 @@ func DefaultWatchedPropertyPaths() []string {
 	return []string{
 		"config.annotation",
 
+		"config.changeTrackingEnabled",
```
aruneshpa, line 40: Consider adding a test for this?

aruneshpa (approved):
LGTM.

pkg/util/vsphere/watcher/watcher.go
bryanv: The existing tests for this generally don't have it for individual watched properties - annotation, devices, keyId, guestHeartbeatStatus, guest, summary - and I agree tests for that don't add much value here. The lack of CBT was already identified as an E2E gap, and that is the best way to really test this.

PR #1614, opened 2026-05-21
✨ Vim Config Options API

akutz:

This patch introduces the `vim.vmware.com/v1alpha1` API under `external/vim/api`. The new types include:

| API | Scope | Description |
|---|---|---|
| `ConfigTarget` | Cluster | The physical hardware/devices available in a vSphere cluster |
| `VirtualMachineConfigOptions` | Cluster | The virtual hardware options for a given hardware version, ex. `vmx-22` |
| `VirtualMachineGuestOptions` | Cluster | The recommended guest options for a given guest ID / hardware version combination. |
| `VirtualMachineConfigPolicy` | Namespace | The hardware available in a given namespace/zone combination for users to create / modify VMs without VM classes. |

There is also a specification-driven-development spec for this work at `.sdd/specs/001-class-policy-resize/` covering the `VirtualMachineConfigPolicy` pipeline. The key design is a two-hop hardware version/SR-IOV discovery where the `ConfigTarget` controller queries the vSphere cluster's hosts to gather information about default/max hardware versions and the hosts' SR-IOV devices. This information is aggregated in the `ConfigTarget` status.

Fixes `vmop-3331`

```release-note
Introduce the vim.vmware.com/v1alpha1 API.
```


.sdd/specs/001-class-policy-resize/plan.md
```diff
@@ -0,0 +1,400 @@
...
+
+// DVXClass is the Device Virtualization Extensions class name for this
+// NIC, populated when the NIC's device class appears in
+// HostHardwareInfo.dvxClasses with sriovNic=true (vSphere 8.0.0.1+).
+// Empty when this NIC does not belong to any DVX class.
+
+// +optional
+DVXClass string `json:"dvxClass,omitempty"`
+
+// DVXCheckpointSupported indicates whether live migration with
```
faisalabujabal, line 234: 🟡 **Medium**: The `+listMapKey=pciDevice.id` marker is invalid. Kubernetes `listMapKey` only supports top-level scalar fields of the list element; nested fields are not supported and will cause CRD generation or validation errors.

.sdd/specs/001-class-policy-resize/plan.md
```diff
@@ -0,0 +1,400 @@
...
+
+**Cluster-scope path** (S5.b):
+1. Resolve `spec.id` to a cluster via `pkg/providers/vsphere`.
+2. Call `EnvironmentBrowser.QueryConfigTarget` → populate `status.*`.
+3. Call `EnvironmentBrowser.QueryConfigOptionDescriptor` → enumerate HW version keys.
+4. For each HW version key: `CreateOrPatch` a `VirtualMachineConfigOptions`.
+
+**Per-host iteration** (S5.c, vmop-3759):
+1. Enumerate cluster hosts via `ClusterComputeResource.host`.
+2. For each host, issue a single `PropertyCollector` RPC fetching:
```
faisalabujabal, line 276: 🟡 **Medium**: Issuing individual `PropertyCollector` RPCs for every host in a cluster is inefficient and does not scale. For large clusters, this can lead to reconcile timeouts and excessive vCenter load. Use a single `PropertyCollector` call with a list of host references to fetch properties in bulk.

.sdd/specs/001-class-policy-resize/plan.md
```diff
@@ -0,0 +1,400 @@
...
+3. Aggregate `max(config.defaultHardwareVersion)` across the responding hosts and write it to `ConfigTarget.status.maxHardwareVersion`.
+4. Build per-host `VirtualMachineSriovInfo` entries from the SR-IOV / DVX data and write them to `ConfigTarget.status.sriov`. Each entry carries `hostMoID` for attribution.
+5. Per-host RPC failures are logged + emitted as warning events; they are not fatal. The reconcile is requeued so the absent host's data lands on a later pass. The successful hosts' data is still written.
+
+No per-host CRD is created; the per-host data is owned entirely by `ConfigTarget.status`. There is no second controller and no watch wiring between `ConfigTarget` and any per-host kind.
+
+**Garbage collection** (`controllers/configtarget/gc.go`): After the cluster-scope queries succeed:
+```go
+func GCVirtualMachineConfigOptions(ctx, client, clusterMoID, liveKeys []string) error
+```
```
faisalabujabal: 🟠 **High**: The garbage collection logic in the `ConfigTarget` controller is dangerous if `VirtualMachineConfigOptions` names are shared across clusters. A controller for one cluster will delete resources required by other clusters if they aren't in its local `liveKeys` list. GC must be scoped to resources owned by the specific cluster.

.sdd/specs/001-class-policy-resize/plan.md
```diff
@@ -0,0 +1,400 @@
...
+4. Update `VirtualMachineGuestOptions.status.hardwareVersions` listMap (keyed by `hardwareVersion`) with this HW version's guest OS data.
+
+#### I5. VirtualMachineGuestOptions plumbing (Story S7 / vmop-3744)
+
+No standalone controller; reconciled by the `VirtualMachineConfigOptions` controller. Deliverables:
+- Validation webhook: immutable `spec.id`; DNS-safe `metadata.name`.
+- RBAC, manifest registration.
+- Integration + E2E tests for the listMap merge.
+
+#### I6. VirtualMachineConfigPolicy controller (Story S8 / vmop-3745)
```
faisalabujabal, line 311: 🟠 **High**: Using shared names (e.g., hardware version or guest ID) for cluster-scoped resources like `VirtualMachineConfigOptions` and `VirtualMachineGuestOptions` will lead to race conditions. Multiple controllers will overwrite each other's status data, causing inconsistent state. These resources should be uniquely named per cluster or use an aggregation pattern.

.sdd/specs/001-class-policy-resize/plan.md
```diff
@@ -0,0 +1,400 @@
...
+    return admission.Errored(http.StatusInternalServerError, err)
+}
+if vmHardwareVersion(vm) > ct.Status.MaxHardwareVersion {
+    return admission.Denied(fmt.Sprintf(
+        "hardware version %s exceeds cluster maximum %s",
+        vmHardwareVersion(vm), ct.Status.MaxHardwareVersion))
+}
+```
+
+Single informer-cached `Get`. No list across per-host objects, no label selector match.
```
faisalabujabal, line 353: 🟡 **Medium**: Comparing hardware versions as strings (e.g., `vmHardwareVersion(vm) > ct.Status.MaxHardwareVersion`) is unreliable as lexicographical comparison fails for versions like 'vmx-9' vs 'vmx-10'. Hardware versions should be compared using numeric values or a version-aware utility.

```suggestion
if parseHardwareVersion(vmHardwareVersion(vm)) > parseHardwareVersion(ct.Status.MaxHardwareVersion) {
```

.sdd/specs/001-class-policy-resize/research.md
```diff
@@ -0,0 +1,259 @@
...
+
+---
+
+## Finding 3: Garbage collection design
+
+### Problem
+
+After the `ConfigTarget` controller reconciles, the following stale objects may exist in Kubernetes but no longer correspond to real vSphere inventory:
+
+- `VirtualMachineConfigOptions` for a hardware version that `QueryConfigOptionDescriptor` no longer returns (e.g., the vSphere cluster has been upgraded past `vmx-17`).
```
faisalabujabal, line 85: 🟡 **Medium**: The admission webhook snippet uses a string comparison ('>') for hardware versions. This is incorrect for versions like 'vmx-9' vs 'vmx-10' because 'vmx-9' is lexicographically greater than 'vmx-10'. The versions should be compared numerically by parsing the integer suffix.

.sdd/specs/001-class-policy-resize/spec.md
```diff
@@ -0,0 +1,185 @@
...
+5. **Given** a cluster with SR-IOV-capable hosts, **when** the ConfigTarget controller reconciles, **then** `ConfigTarget.status.sriov` contains one entry per (host, SR-IOV NIC) pair, populated by walking each host's `HostConfigInfo.pciPassthruInfo` (entries where `sriovCapable=true`), with each entry enriched from `HostHardwareInfo.dvxClasses` where `sriovNic=true`, and each entry carrying `hostMoID`, `active`, `maxVFs`, `numVFs`, and `dvxClass*` fields. The cluster-scope `QueryConfigTarget` call is never used for SR-IOV discovery.
+6. **Given** the per-host RPC for one host transiently fails while the others succeed, **when** the ConfigTarget controller reconciles, **then** the failing host's SR-IOV entries are absent from `ConfigTarget.status.sriov` for this pass, the controller records a warning event referencing the host, and the reconcile is requeued so the missing data lands on a later pass. The successful hosts' data is still written.
+
+---
+
```
faisalabujabal, line 68: 🟡 **Medium**: Transient RPC failures for a single host cause its SR-IOV data to be omitted from `ConfigTarget.status`. This results in fluctuating capacity reports which can cause the admission webhook to incorrectly reject VM requests. It is recommended to persist the last known good state and use a status condition to flag stale data.

```suggestion
6. **Given** the per-host RPC for one host transiently fails while the others succeed, **when** the ConfigTarget controller reconciles, **then** the failing host's SR-IOV entries are retained in `ConfigTarget.status.sriov` but marked as stale, and the controller records a warning event.
```

.sdd/specs/001-class-policy-resize/tasks.md
```diff
@@ -0,0 +1,120 @@
...
+- [ ] T066a [S5.c / vmop-3797] Add `ConfigTargetStatus.MaxHardwareVersion` to `external/vim/api/v1alpha1/config_target_types.go`; extend `VirtualMachineSriovInfo` in `external/vim/api/v1alpha1/config_target_devices_types.go` with `HostMoID` (required), `Active`, `MaxVFs`, `NumVFs`, `DVXClass`, `DVXCheckpointSupported`, `DVXSWDMATracingSupported`; add `+listType=map` with `+listMapKey=hostMoID` and `+listMapKey=pciDevice.id` to `status.sriov`; regenerate `zz_generated.deepcopy.go` and `vim.vmware.com_configtargets.yaml`. T066 cannot be merged before T066a.
```
faisalabujabal, line 65: 🟡 **Medium**: Kubernetes `controller-gen` markers for `listMapKey` do not support nested fields or dot notation (e.g., `pciDevice.id`). The key must be a direct field of the list element's struct. Consider flattening the PCI ID into a top-level field in `VirtualMachineSriovInfo`.

.sdd/specs/001-class-policy-resize/tasks.md
```diff
@@ -0,0 +1,120 @@
...
+- [ ] T107 [S8.d] E2E test — policy spec filled from real cluster capabilities after Zone creation
+
+### Story S9 — VM admission webhook enforcement (vmop-3746)
+
+- [ ] T110 [S9.a] Modify `webhooks/virtualmachine/validation_webhook.go` — on create/update/powerOn, load namespace VirtualMachineConfigPolicy; apply createMode/updateMode/powerOnMode; respect vmClassMode=AsPolicy vs AsConfig
+- [ ] T111 [S9.a] Unit tests for mode enforcement
+- [ ] T112 [S9.b] Implement ExtraConfig Allow/Deny enforcement — Fixed (exact), Regex (regexp), Glob (filepath.Match); Denied takes precedence over Allowed
+- [ ] T113 [S9.b] Unit tests for all three match types and precedence rules
+- [ ] T114 [S9.c] Implement `ConfigTarget`-based hardware-version check — resolve the VM's zone to a cluster MoID, `Get` the cluster's `ConfigTarget`, reject if the VM's effective hardware version exceeds `ConfigTarget.status.maxHardwareVersion`. No `HostSystem` list, no label selectors.
```
faisalabujabal: 🟡 **Medium**: Task T114 assumes a 1:1 mapping between a Zone and a Cluster MoID, which contradicts the multi-cluster zone support mentioned in T052 and T106. The admission logic should define how to handle multiple `ConfigTarget` objects within a single zone (e.g., using the minimum common hardware version).

api/v1alpha6/virtualmachinereservedprofile_types.go
```diff
@@ -0,0 +1,85 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha6
```
faisalabujabal, line 5: 🟠 **High**: The package name is declared as 'v1alpha6', but the PR description and related context indicate this should be part of the 'v1alpha1' API. This is likely a copy-paste error from another versioned API.

```suggestion
package v1alpha1
```

api/v1alpha6/virtualmachinereservedprofile_types.go
```diff
@@ -0,0 +1,85 @@
...
+	// +optional
+
+	// Conditions describes the observed conditions of the
+	// VirtualMachineReservedProfile.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// +optional
+
+	// Zones describes the observed status of the class in one or more zones.
+	Zones []VirtualMachineReservedProfileZoneStatus `json:"zones,omitempty"`
```
faisalabujabal, line 56: 🟡 **Medium**: Slices of conditions and zones are missing kubebuilder markers (e.g., '+listType=map', '+listMapKey') required for proper merging during Server-Side Apply (SSA). Conditions should use 'type' as the key, and zones should use 'name'.

config/crd/bases/vmoperator.vmware.com_virtualmachinereservedprofiles.yaml
```diff
@@ -0,0 +1,199 @@
...
+                      minimum: 0
+                      type: integer
+                  required:
+                  - name
+                  type: object
+                type: array
+            type: object
+        type: object
+    served: true
+    storage: true
```
faisalabujabal, line 206: 🟡 **Medium**: The 'conditions' and 'zones' arrays lack 'x-kubernetes-list-type: map' and 'x-kubernetes-list-map-keys' markers. This prevents the Kubernetes API server from correctly merging list items during server-side apply or strategic merge patch operations (use 'type' for conditions and 'name' for zones).

external/vim/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,54 @@
...
+// ManagedObjectID is a unique ID used to identify a managed object on a given
+// vSphere instance
+type ManagedObjectID struct {
+	// +required
+
+	// ID is the object's ID.
+	ID string `json:"id"`
+
+	// +optional
+
```
faisalabujabal, line 42: 🟡 **Medium**: Kubernetes API markers (e.g., +required, +optional) should not be separated from their fields by blank lines. This separation can cause the markers to be ignored by controller-gen during CRD generation.

external/vim/api/v1alpha1/config_target_devices_types.go
```diff
@@ -0,0 +1,870 @@
...
+	// The value is defined by vendors of the DVX device class as part of
+	// their localizable messages.
+	VendorName string `json:"vendorName"`
+
+	// +required
+
+	// SriovNic indicates whether the devices of this class are SR-IOV NICs.
+	SriovNic bool `json:"sriovNic"`
+
+	// +optional
```
faisalabujabal, line 778: 🟡 **Medium**: The 'Capacity' field in 'VirtualMachineIdeDiskDevicePartitionInfo' uses 'int32', which is insufficient for representing partition sizes in bytes (limited to 2GB) and is inconsistent with 'VirtualMachineIdeDiskDeviceInfo.Capacity' which uses 'resource.Quantity'.

```suggestion
	Capacity *resource.Quantity `json:"capacity"`
```

external/vim/api/v1alpha1/enum_types.go
```diff
@@ -0,0 +1,212 @@
...
+// FromVimType parses the VIM value.
+func (t *LatencySensitivityLevel) FromVimType(s string) {
+	switch s {
+	case "high":
+		*t = LatencySensitivityLevelHigh
+	case "low":
+		*t = LatencySensitivityLevelLow
+	case "normal":
+		*t = LatencySensitivityLevelNormal
+	}
```
faisalabujabal, line 134: 🟠 **High**: The `FromVimType` method for `LatencySensitivityLevel` incorrectly overwrites the mapped constant with the raw input string at the end of the function, preventing proper normalization to the defined constants (e.g., 'high' becomes 'high' instead of 'High').

```suggestion
func (t *LatencySensitivityLevel) FromVimType(s string) {
	switch s {
	case "high":
		*t = LatencySensitivityLevelHigh
	case "low":
		*t = LatencySensitivityLevelLow
	case "normal":
		*t = LatencySensitivityLevelNormal
	default:
		*t = LatencySensitivityLevel(s)
	}
}
```

external/vim/api/v1alpha1/enum_types.go
```diff
@@ -0,0 +1,212 @@
...
+
+// FromVimType parses the VIM value.
+func (t *PhysicalNICFeature) FromVimType(s uint8) {
+	switch s {
+	case 4:
+		*t = PhysicalNICFeatureRSS
+	case 5:
+		*t = PhysicalNICFeatureLRO
+	}
+	*t = PhysicalNICFeature("")
```
faisalabujabal: 🟠 **High**: The `FromVimType` method for `PhysicalNICFeature` unconditionally assigns an empty string to the receiver at the end, causing the function to always return an empty value and ignoring the switch statement results.

```suggestion
func (t *PhysicalNICFeature) FromVimType(s uint8) {
	switch s {
	case 4:
		*t = PhysicalNICFeatureRSS
	case 5:
		*t = PhysicalNICFeatureLRO
	default:
		*t = PhysicalNICFeature("")
	}
}
```

external/vim/api/v1alpha1/enum_types.go
```diff
@@ -0,0 +1,212 @@
...
+	switch s {
+	case 1:
+		*t = TxRxThreadModelPerVNIC
+	case 2:
+		*t = TxRxThreadModelPerVM
+	case 3:
+		*t = TxRxThreadModelPerQueue
+	}
+	*t = TxRxThreadModel("")
+}
```
faisalabujabal, line 217: 🟠 **High**: The `FromVimType` method for `TxRxThreadModel` unconditionally assigns an empty string to the receiver at the end, causing the function to always return an empty value and ignoring the switch statement results.

```suggestion
func (t *TxRxThreadModel) FromVimType(s uint8) {
	switch s {
	case 1:
		*t = TxRxThreadModelPerVNIC
	case 2:
		*t = TxRxThreadModelPerVM
	case 3:
		*t = TxRxThreadModelPerQueue
	default:
		*t = TxRxThreadModel("")
	}
}
```

external/vim/api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+	// GroupVersion is group version used to register these objects.
+	GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1alpha1"}
+
+	// schemeBuilder is used to add go types to the GroupVersionKind scheme.
+	schemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+
+	// AddToScheme adds the types in this group-version to the given scheme.
+	AddToScheme = schemeBuilder.AddToScheme
+
+	objectTypes = []runtime.Object{}
```
faisalabujabal, line 26: 🟠 **High**: The 'objectTypes' slice is empty. The new API types mentioned in the PR description (ConfigTarget, VirtualMachineConfigOptions, VirtualMachineGuestOptions, and VirtualMachineConfigPolicy) must be added here to be registered with the Kubernetes scheme.

```suggestion
	objectTypes = []runtime.Object{
		&ConfigTarget{},
		&ConfigTargetList{},
		&VirtualMachineConfigOptions{},
		&VirtualMachineConfigOptionsList{},
		&VirtualMachineGuestOptions{},
		&VirtualMachineGuestOptionsList{},
		&VirtualMachineConfigPolicy{},
		&VirtualMachineConfigPolicyList{},
	}
```

external/vim/api/v1alpha1/testdata/xml/vgpus-and-sriov/config-target-HostSystem-ha-host.xml
```diff
@@ -0,0 +1,666 @@
...
+        <backing _XMLSchema-instance:type="VirtualPCIPassthroughVmiopBackingInfo">
+          <vgpu>grid_a100-40c</vgpu>
+        </backing>
+      </device>
+    </componentDeviceInfo>
+  </vendorDeviceGroupInfo>
+  <maxSimultaneousThreads>2</maxSimultaneousThreads>
+  <dvxClassInfo>
+    <deviceClass xmlns:_XMLSchema-instance="http://www.w3.org/2001/XMLSchema-instance" _XMLSchema-instance:type="ElementDescription">
+      <label>DVX Test [Checkpoint, RP]</label>
```
faisalabujabal, line 380: 🟡 **Medium**: Duplicate 'key' values (0) found for multiple devices within 'vendorDeviceGroupInfo'. Device keys must be unique within a virtual machine configuration to avoid validation errors.

external/vim/api/v1alpha1/virtual_device_backing_types.go
```diff
@@ -0,0 +1,587 @@
...
+// VirtualPCIPassthroughDvxBackingInfo defines DVX (Device Virtualization
+// Extensions) backing for a PCI passthrough device.
+// It corresponds to vim.vm.device.VirtualPCIPassthrough.DvxBackingInfo.
+type VirtualPCIPassthroughDvxBackingInfo struct {
+	// +optional
+
+	// DeviceClass is the device class that backs this DVX device.
+	DeviceClass string `json:"deviceClass,omitempty"`
+
+	// +optional
```
faisalabujabal, line 323: 🟠 **High**: The `VendorId` field is defined as `int16`, which will cause unmarshaling errors for PCI Vendor IDs greater than 32767 (e.g., Intel's 0x8086). PCI IDs are 16-bit unsigned values; use `int32` to safely accommodate the full range.

```suggestion
	VendorId int32 `json:"vendorId"`
```

external/vim/api/v1alpha1/virtual_device_backing_types.go
```diff
@@ -0,0 +1,587 @@
...
+
+// VirtualPCIPassthroughDeviceBackingInfo contains information about the host
+// PCI device backing for a PCI passthrough device.
+// It corresponds to vim.vm.device.VirtualPCIPassthrough.DeviceBackingInfo.
+type VirtualPCIPassthroughDeviceBackingInfo struct {
+	// Id is the PCI name ID, composed of "bus:slot.function".
+	Id string `json:"id"`
+
+	// DeviceId is the PCI device ID.
+	DeviceId string `json:"deviceId"`
```
faisalabujabal, line 282: 🟡 **Medium**: The `SerialPort` field in `VirtualDeviceURIBackingInfo` creates a redundant recursive definition because `VirtualSerialPortURIBackingInfo` embeds the base struct. Since the subclass adds no new fields, this nested field is unnecessary and potentially confusing.

external/vim/api/v1alpha1/virtual_device_controller_types.go
```diff
@@ -0,0 +1,192 @@
...
+
+const (
+	// VirtualNVMESharingNone disables NVME bus sharing.
+	VirtualNVMESharingNone VirtualSCSISharing = "None"
+
+	// VirtualNVMESharingPhysical enables physical NVME bus sharing.
+	VirtualNVMESharingPhysical VirtualSCSISharing = "Physical"
+)
+
+// VirtualNVMEController represents an NVMe controller in a virtual machine.
```
faisalabujabal: 🟡 **Medium**: Constants for VirtualNVMESharing are incorrectly typed as VirtualSCSISharing, which will cause compilation errors when assigned to VirtualNVMEController.SharedBus.

```suggestion
	// VirtualNVMESharingNone disables NVME bus sharing.
	VirtualNVMESharingNone VirtualNVMESharing = "None"

	// VirtualNVMESharingPhysical enables physical NVME bus sharing.
	VirtualNVMESharingPhysical VirtualNVMESharing = "Physical"
```

external/vim/api/v1alpha1/virtual_device_disk_types.go
```diff
@@ -0,0 +1,500 @@
...
+
+	// DiskMode is the disk persistence mode.
+	DiskMode string `json:"diskMode"`
+
+	// +optional
+
+	// Split indicates whether the virtual disk is stored in multiple 2 GB
+	// files.
+	Split *bool `json:"split,omitempty"`
+
```
faisalabujabal, line 437: 🟠 **High**: Inconsistency in backing info structs: 'VirtualDiskFlatVer1BackingInfo', 'VirtualDiskFlatVer2BackingInfo', 'VirtualDiskRawDiskMappingVer1BackingInfo', 'VirtualDiskSeSparseBackingInfo', and 'VirtualDiskSparseVer1BackingInfo' are missing the 'VirtualDeviceFileBackingInfo' inline. This is inconsistent with 'VirtualDiskSparseVer2BackingInfo' and prevents these types from specifying essential fields like 'FileName' and 'Datastore' inherited from the base VIM type.

external/vim/api/v1alpha1/virtual_device_ethernet_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// +optional
+
+	// TxDataRingDescSize is the transmit data ring descriptor size
+	// (128-2048, multiple of 64).
+	TxDataRingDescSize int32 `json:"txDataRingDescSize,omitempty"`
+
+	// +optional
+
+	// RxDataRingDescSize is the receive data ring descriptor size
+	// (128-2048, multiple of 64).
```
faisalabujabal, line 180: 🟡 **Medium**: Optional integer fields in 'VirtualVmxnet3StrictLatencyConfig' should be defined as pointers. Using non-pointer types with 'omitempty' makes it impossible to distinguish between a zero value and an unset field, which is problematic since 0 is outside the documented valid ranges (e.g., 1-32).

```suggestion
	MaxTxQueues *int32 `json:"maxTxQueues,omitempty"`
```

external/vim/api/v1alpha1/virtual_device_types.go
```diff
@@ -0,0 +1,579 @@
...
+// +noop:kubebuilder:validation:XValidation:rule="!has(self.unitNumber) || self.type != 'VirtualUSBController' || self.unitNumber == 22",message="USB (UHCI/EHCI) controller unit number must be 22"
+// +noop:kubebuilder:validation:XValidation:rule="!has(self.unitNumber) || self.type != 'VirtualUSBXHCIController' || self.unitNumber == 23",message="XHCI controller unit number must be 23"
+// +noop:kubebuilder:validation:XValidation:rule="!has(self.unitNumber) || (self.type != 'VirtualSATAController' && self.type != 'VirtualAHCIController') || (self.unitNumber >= 24 && self.unitNumber <= 27)",message="SATA HBA unit number must be in [24, 27]"
+// +noop:kubebuilder:validation:XValidation:rule="!has(self.unitNumber) || self.type != 'VirtualNVMeController' || (self.unitNumber >= 30 && self.unitNumber <= 33)",message="NVMe HBA unit number must be in [30, 33]"
+
+// VirtualDevice is the base data object type for devices in a virtual machine.
+// It corresponds to vim.vm.device.VirtualDevice.
+type VirtualDevice struct {
+	// +optional
```
faisalabujabal: 🟡 **Medium**: The 'noop:' prefix on 'kubebuilder:validation:XValidation' markers is non-standard and will cause these validation rules to be ignored by 'controller-gen'. If these constraints (such as the 'oneOf' check for device fields and unit number range validations) are intended to be active, the 'noop:' prefix should be removed.

```suggestion
// +kubebuilder:validation:XValidation:rule="[has(self.controller), ...].filter(x, x).size() <= 1"
```

external/vim/api/v1alpha1/virtualmachine_config_options_choice_types.go
```diff
@@ -0,0 +1,348 @@
...
+
+	// +optional
+
+	// ResourceQuantity contains a resource quantity range option.
+	ResourceQuantity *ResourceQuantityOption `json:"resourceQuantity,omitempty"`
+
+	// +optional
+
+	// Choice contains a string choice option.
+	Choice *ChoiceOption `json:"choice,omitempty"`
```
faisalabujabal, line 54: 🟡 **Medium**: The 'OptionType' struct is missing a field for 'StringOption' (defined on line 52), which prevents string-based options from being represented in the union used by 'OptionDef'.

```suggestion
	// +optional

	// String contains a string option.
	String *StringOption `json:"string,omitempty"` 
```

external/vim/api/v1alpha1/virtualmachine_config_options_choice_types.go
```diff
@@ -0,0 +1,348 @@
...
+// ResourceQuantityOption describes a range of resource.Quantity values with
+// min, max, and default.
+type ResourceQuantityOption struct {
+	BaseOptionType        `json:",inline"`
+	ResourceQuantityRange `json:",inline"`
+
+	// +required
+
+	// Default is the default value.
+	Default resource.Quantity `json:"default"`
```
faisalabujabal, line 123: 🟡 **Medium**: In 'ChoiceOption', 'DefaultIndex' should be a pointer (*int32) to distinguish between a default index of 0 and the field being unset, especially since 'omitempty' is used. Additionally, the JSON tag 'default' deviates from the VIM property name 'defaultIndex'.

```suggestion
	DefaultIndex *int32 `json:"defaultIndex,omitempty"`
```

external/vim/api/v1alpha1/virtualmachine_config_options_device_types.go
```diff
@@ -0,0 +1,332 @@
...
+	SupportedDeviceTypeVirtualVideoCardOption = "VirtualVideoCardOption"
+)
+
+// VirtualDeviceOption describes the options for a virtual device type,
+// including configuration options and its relationship to other devices.
+// It corresponds to vim.vm.device.VirtualDeviceOption.
+type VirtualDeviceOption struct {
+	// +optional
+
+	// Type is the name of the run-time class to instantiate for this device.
```
faisalabujabal: 🟡 **Medium**: The `Type` field uses `VirtualDeviceType`, but the type defined at the top of this file is `SupportedDeviceType`. This mismatch should be resolved to ensure consistency with the constants defined in this package.

```suggestion
	Type SupportedDeviceType `json:"type,omitempty"`
```

external/vim/api/v1alpha1/virtualmachine_config_options_device_types.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// +optional
+
+	// AutoAssignController indicates whether this device will be
+	// auto-assigned a controller if one is required.
+	AutoAssignController *BoolOption `json:"autoAssignController,omitempty"`
+
+	// +optional
+
+	// DefaultBackingOptionIndex is an index into the backing option list that
```
faisalabujabal, line 104: 🟡 **Medium**: `DefaultBackingOptionIndex` is an `int32` with `omitempty`. Since `0` is a valid index, it will be omitted from the JSON output, making it impossible to distinguish between index 0 and an unset value. Use `*int32` for optional primitive fields where zero is a valid value.

```suggestion
	DefaultBackingOptionIndex *int32 `json:"defaultBackingOptionIndex,omitempty"`
```

external/vim/api/v1alpha1/virtualmachine_config_options_types.go
```diff
@@ -0,0 +1,633 @@
...
+	// virtualization is supported.
+	NestedHVSupported bool `json:"nestedHVSupported,omitempty"`
+
+	// +optional
+
+	// VPMCSupported indicates whether virtualized CPU performance counters
+	// are supported.
+	VPMCSupported bool `json:"vpmcSupported,omitempty"`
+
+	// +optional
```
faisalabujabal, line 516: 🟡 **Medium**: Inconsistent use of pointer types for optional primitive fields in Status. Fields such as `GuestOSDefaultIndex`, `DeviceListReadonly`, `NumCpuReadonly`, and many within `VirtualMachineConfigCapability` use `bool` or `int32` instead of `*bool` or `*int32`. This creates ambiguity in the observed state, as `false` or `0` cannot be distinguished from an unpopulated or unknown field.

```suggestion
GuestOSDefaultIndex *int32 `json:"guestOSDefaultIndex,omitempty"`
```

external/vim/api/v1alpha1/virtualmachine_guest_options_types.go
```diff
@@ -0,0 +1,423 @@
...
+	// guest while the virtual machine is running.
+	SupportsMemoryHotAdd bool `json:"supportsMemoryHotAdd,omitempty"`
+
+	// +optional
+
+	// SupportsPvscsiControllerForBoot indicates whether this guest can use
+	// a PVSCSI controller as the boot adapter.
+	SupportsPvscsiControllerForBoot bool `json:"supportsPvscsiControllerForBoot,omitempty"`
+
+	// +optional
```
faisalabujabal, line 249: 🟡 **Medium**: Inconsistent use of pointers for optional boolean fields in API status. Fields like 'SmcRequired' and 'SupportsCpuHotAdd' use 'bool' with 'omitempty', which prevents explicitly setting them to 'false' if the default is 'true'.

```suggestion
SmcRequired *bool `json:"smcRequired,omitempty"`
```

external/vim/api/v1alpha1/virtualmachine_guest_options_types.go
```diff
@@ -0,0 +1,423 @@
...
+
+	// NumRecommendedCoresPerSocket is the recommended number of cores per
+	// socket for this guest.
+	NumRecommendedCoresPerSocket int32 `json:"numRecommendedCoresPerSocket,omitempty"`
+
+	// +optional
+
+	// SupportsSlaveDisk indicates whether this guest supports a disk
```
faisalabujabal, line 158: 🟡 **Medium**: Optional integer fields should use pointers to distinguish between a zero value (0) and an unset field. This is particularly important for fields like 'RecommendedColorDepth' or 'NumSupportedFloppyDevices' where 0 might be a valid recommendation.

```suggestion
SupportedMaxCPUs *int32 `json:"supportedMaxCPUs,omitempty"`
```

external/vim/doc/controller-workflows.md
```diff
@@ -0,0 +1,133 @@
...
+    VMCP->>K: Patch VirtualMachineConfigPolicy spec from ConfigTarget status
+  else spec.syncMode = Disabled
+    Note over VMCP,K: No sync from vSphere, spec is operator-managed
+  end
+```
+
+## Controllers
+
+### `Zone` controller
+
```
faisalabujabal, line 104: 🟠 **High**: The `VirtualMachineConfigOptions` resource is cluster-scoped but named only by hardware version (e.g., `vmx-22`). If the management plane handles multiple vSphere clusters or vCenters with different ESXi versions, these resources will collide and overwrite each other. The name should include a cluster identifier to ensure uniqueness.

external/vim/doc/controller-workflows.md
```diff
@@ -0,0 +1,133 @@
...
+### `ConfigTarget` controller
+
+1. The new `ConfigTarget` controller watches for the `ConfigTarget` resources. Upon seeing a `ConfigTarget` resource, the controller calls [`QueryConfigTarget`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.EnvironmentBrowser.html#queryConfigTarget) on the cluster's environment browser. The controller uses the [result](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.ConfigTarget.html) to populate the `ConfigTarget`'s status information.
+2. The new `ConfigTarget` controller also calls the [`QueryConfigOptionDescriptor`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.EnvironmentBrowser.html#queryConfigOptionDescriptor) API, getting a list of the hardware versions supported by the cluster (the [`key`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.vm.ConfigOptionDescriptor.html#key) field).
```
faisalabujabal, line 114: 🟠 **High**: The `VirtualMachineGuestOptions` resource is cluster-scoped and named by guest ID, leading to collisions across different vSphere clusters. Since guest options can vary by cluster or hardware version, using a global name for a cluster-scoped resource is unsafe in multi-cluster environments.

external/vim/doc/controller-workflows.md
```diff
@@ -0,0 +1,133 @@
...
+3. For each hardware version, the `ConfigTarget` controller creates or patches a cluster-scoped `VirtualMachineConfigOptions` resource, with its `metadata.name` field being the hardware version, ex. vmx-22.
+
+### `VirtualMachineConfigOptions` controller
```
faisalabujabal, line 117: 🟡 **Medium**: The workflow describes multiple controllers (or reconcile loops) concurrently appending to the same `status.hardwareVersions` list in `VirtualMachineGuestOptions`. This will cause race conditions and lost updates unless Server-Side Apply is used with distinct field managers for each list entry.

external/vim/doc/controller-workflows.md
```diff
@@ -0,0 +1,133 @@
...
+
+### `VirtualMachineConfigPolicy` controller
+
+1. The new `VirtualMachineConfigPolicy` controller watches the new, namespace-scoped `VirtualMachineConfigPolicy` API.
+2. The `VirtualMachineConfigPolicy` controller checks the value of the `VirtualMachineConfigPolicy` field `spec.syncMode`.
+3. If the value of `spec.syncMode` is `Disabled`, the controller does not copy data from vSphere into the policy; reconciliation may still update status (for example conditions). The policy **spec** is treated as operator-managed.
+4. If the value of `spec.syncMode` is `ConfigTarget`, the controller continues to reconcile the resource against the cluster `ConfigTarget`.
+5. The `VirtualMachineConfigPolicy` controller looks up the `Zone` resource named in `spec.zone`.
+6. Using the same pool-to-cluster derivation as the `Zone` controller, the controller obtains the vSphere cluster managed object ID(s) from the zone's `spec.namespace.poolMoIDs`. For each relevant cluster, it reads the cluster-scoped `ConfigTarget` whose `metadata.name` is that cluster MoID.
```
faisalabujabal: 🟡 **Medium**: The documentation does not define how the `VirtualMachineConfigPolicy` controller handles zones spanning multiple clusters. It is unclear how statuses from multiple `ConfigTarget` resources are merged into a single policy spec (e.g., whether it uses the most restrictive limits or a first-match approach).

external/vim/doc/deploying-a-vm.md
```diff
@@ -0,0 +1,147 @@
...
+- **`storageClassOptions`**, **`supportedMonitorTypes`**, **`supportedOvfEnvironmentTransports`**, **`supportedOvfInstallTransports`** — additional option metadata from the config option.
+- **`defaultDevices`** — virtual devices vSphere creates by default for this version; clients should not recreate them explicitly.
+- **`propertyRelations`** — relationships between VM config spec properties (for example, a guest ID implying a required firmware type).
+
+Per-guest min/max memory, disk counts, recommended controllers, hot-plug behavior, and related fields **moved** to `VirtualMachineGuestOptions` (see below).
+
+## VirtualMachineGuestOptions
+
+`VirtualMachineGuestOptions` is a **cluster-scoped** resource: **one object per guest ID**. The object is **named after the lowercased guest ID** (e.g., `otherlinux64` for `OtherLinux64`). Short name: `vmguestoptions`. Its spec sets `id` to the canonical guest identifier (matching the enum / vSphere guest ID casing).
+
```
faisalabujabal, line 45: 🟡 **Medium**: vSphere Guest IDs (e.g., 'rhel7_64Guest') frequently contain underscores, which are prohibited in Kubernetes resource names. The documentation should specify if underscores are replaced (e.g., with hyphens) to ensure the resulting names are valid DNS-safe labels.

pkg/crd/crd_test.go
```diff
@@ -35,6 +35,7 @@ var (
 		"contentsources.vmoperator.vmware.com",
 		"virtualmachineclassbindings.vmoperator.vmware.com",
 		"virtualmachineclasses.vmoperator.vmware.com",
+		"virtualmachinereservedprofiles.vmoperator.vmware.com",
```
faisalabujabal, line 38: 🟡 **Medium**: The added CRD uses the 'vmoperator.vmware.com' group, but the PR description states the new API group is 'vim.vmware.com'. Additionally, 'virtualmachinereservedprofiles' is not listed among the new types in the PR, and the four mentioned VIM CRDs are missing from the test lists, which may lead to incomplete verification of the CRD installation logic.

.sdd/specs/001-class-policy-resize/plan.md
akutz: This should not be an issue as the ConfigOptions are virtual and should be the same from any host/cluster that provides them for a given hardware version.

api/v1alpha6/virtualmachinereservedprofile_types.go
akutz: This file is in a different package, so it is fine.

external/vim/api/v1alpha1/common_types.go
akutz: This is *not* true, and separation by the blank line is on purpose to avoid contaminating the GoDocs.

.sdd/specs/001-class-policy-resize/plan.md
akutz: I confirmed this offline with Utkarsh.

.sdd/specs/001-class-policy-resize/research.md
akutz: The snippet has been updated.

.sdd/specs/001-class-policy-resize/spec.md
akutz: There is no way to know if the error is transient. Last good state could lead to bad placement decicions.

external/vim/api/v1alpha1/groupversion_info.go
akutz: These are added in init blocks in the object source files. Bad AI.

akutz: cc @faisalabujabal

external/vim/api/v1alpha1/testdata/xml/vgpus-and-sriov/config-target-HostSystem-ha-host.xml
akutz: AI should ignore files in a `testdata` directory.

external/vim/api/v1alpha1/virtual_device_ethernet_types.go
akutz: AI, you are drunk, go home. The fact that zero is not in the range is *why* it is a valid empty value.

external/vim/doc/controller-workflows.md
akutz: All ESXi hosts share the same potential guest IDs, so this is safe.

external/vim/api/v1alpha1/virtualmachine_config_options_device_types.go
akutz: This comment did not know what it was talking about, not really, but it does raise an interesting problem I am now having to tackle.

external/vim/api/v1alpha1/virtualmachine_config_options_device_types.go
akutz: Since the default index in an array *is* zero, having the default value as `0` makes perfect sense. This suggestion is rejected.

external/vim/api/v1alpha1/virtualmachine_config_options_types.go
akutz: Ignoring suggestion as the `...Supported` patterns are fine to default to `false`.

external/vim/api/v1alpha1/virtualmachine_guest_options_types.go
akutz: Except the default is *not* true, and if it was, we would see that via the kubebuilder annotations. AI, stop wasting my time.

external/vim/api/v1alpha1/virtualmachine_guest_options_types.go
akutz: Invalid comment.

external/vim/doc/controller-workflows.md
akutz: The ConfigOptions are virtual, and can be shared my multiple clusters. The AI model clearly does not understand how vSphere APIs work. cc @faisalabujabal

external/vim/doc/controller-workflows.md
akutz: This is not racey because status.hardwareVersions is a listMap and these are virtual options, so each HW version info will be the same across clusters.

external/vim/doc/deploying-a-vm.md
akutz: AI, you are wasting my time. The VIM CRD APIs clearly show I have converted these strings to K8s-compliant names.

pkg/crd/crd_test.go
akutz: The dumb AI model did not recognize there are *multiple* API schema groups present.

external/vim/doc/deploying-a-vm.md
```diff
@@ -0,0 +1,147 @@
...
+## VirtualMachineConfigPolicy
+
+`VirtualMachineConfigPolicy` is **namespace-scoped** (short name: `vmconfigpolicy`). It is **not** a projection of a single vSphere method; think of it as a **namespace-scoped companion to `ConfigTarget`** whose fields are interpreted **per virtual machine**.
+
+There is **one `VirtualMachineConfigPolicy` object per zone** in a namespace. Each object is **named after that zone** (for example `us-east-1`, `us-west-1`). If namespace `my-namespace-1` spans three zones—`us-west-1`, `us-east-1`, and `us-south-1`—then there are **three** policy objects in that namespace. Validators and placement-aware components **must consider every zone’s policy** when deciding whether a configuration is allowed: **all zone policies in the namespace are consulted**, and the outcome depends on whether **any** of them permits the requested VM configuration (for example, at least one zone allows the requested memory, CPU, devices, and ExtraConfig rules so the VM could legally run there). If **no** zone’s policy allows the request, admission or scheduling should reject it.
```
bryanv, line 65: If the VM already specifies a zone then only that zone should be consulted, right?

bryanv (approved):

external/vim/doc/deploying-a-vm.md
akutz: Yes.

PR #1617, opened 2026-05-21
✨ test: add E2E coverage for placement policy combinations

bryanv:

Add a new `PLACEMENT-POLICY-COMBINATIONS` test suite under `VM-LCM` to validate that mandatory and optional VM-Host affinity policies (compute + infrastructure) are correctly evaluated when applied to VMs on WCP.

Two scenarios are covered:
1. **Mandatory + optional policies applied at VM creation**: creates a VM with an explicit optional `ComputePolicy` reference, then asserts that a `PolicyEvaluation` object is created for the VM containing both the mandatory and optional policies, and that the `Ready` condition is `True` (compliant).
2. **Retroactive policy application**: creates a VM before a new mandatory infrastructure policy is applied to the namespace, then asserts that the existing VM's `PolicyEvaluation` is updated to include the new policy and remains compliant.

Supporting fixes included in this change:
- Register `vsphere-policy/api/v1alpha1` types in the shared test scheme (`common/scheme.go`) so `svClusterClient` can deserialize `PolicyEvaluation` objects.
- Fix `dcli_client.go` `CreateTagCategory` / `CreateTag` to quote the `description` argument with `%q`, preventing dcli parse errors when the description contains spaces.
- Resolve the namespace-scoped `vmi-xxxxx` image name via `WaitForVirtualMachineImageName` instead of using the bare display name (`photon-5.0`), avoiding webhook rejection when both namespace-scoped and cluster-scoped images share the same display name.
- Use per-namespace-unique names for tags, tag categories, and policies to prevent `AlreadyExists` failures on repeated runs.

Tests are labeled `placement-policy` and `experimental`; run with:
    make test-e2e LABEL_FILTER="placement-policy"

Fixes # vmop-3609

```release-note
NONE
```

### DIFF
--- a/test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
@@ -20,6 +20,7 @@ import (
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	e2eframework "k8s.io/kubernetes/test/e2e/framework"
 	capiutil "sigs.k8s.io/cluster-api/util"
@@ -28,6 +29,7 @@ import (
 	vmopv1a2 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 	vmopv1a5 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
 	imgregv1a1 "github.com/vmware-tanzu/vm-operator/external/image-registry-operator/api/v1alpha1"
+	vspherepolv1 "github.com/vmware-tanzu/vm-operator/external/vsphere-policy/api/v1alpha1"
 
 	"github.com/vmware-tanzu/vm-operator/test/e2e/framework"
 	"github.com/vmware-tanzu/vm-operator/test/e2e/infrastructure/vsphere/testbed"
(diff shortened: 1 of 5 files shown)


test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
```diff
@@ -1028,6 +1029,300 @@ func VMSpec(ctx context.Context, inputGetter func() VMSpecInput) {
 			vmservice.VerifyVMTagsAndPolicyAssignment(ctx, config, svClusterClient, tagManager, tmpNamespaceName, vmName, policyNameToVMTagID, expectedPolicyNames)
 		})
 	})
+
+	Context("PLACEMENT-POLICY-COMBINATIONS", Label("placement-policy", "experimental"), func() {
```
faisalabujabal: would you eventually want this in core,  extended functional, or smoke? we should add that now as well as experimental

test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
```diff
@@ -1028,6 +1029,300 @@ func VMSpec(ctx context.Context, inputGetter func() VMSpecInput) {
...
+			tagCategoryID            string
+			mandatoryHostTagID       string
+			mandatoryVMTagID         string
+			optionalHostTagID        string
+			optionalVMTagID          string
+			hostID                   string
+			existingVMName           string
+		)
+
+		BeforeEach(func() {
```
faisalabujabal, line 1075: we should have an aftereach that cleans up the namespace

test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
```diff
@@ -1028,6 +1029,300 @@ func VMSpec(ctx context.Context, inputGetter func() VMSpecInput) {
...
+					Namespace:        tmpNamespaceName,
+					Name:             vmName,
+					ImageName:        tmpNamespaceVMIName,
+					VMClassName:      clusterResources.VMClassName,
+					StorageClassName: clusterResources.StorageClassName,
+					ResourcePolicy:   clusterResources.VMResourcePolicyName,
+					PowerState:       "PoweredOn",
+					Policies: []vmopv1a5.PolicySpec{
+						{
+							APIVersion: "vsphere.policy.vmware.com/v1alpha1",
```
faisalabujabal, line 1194: should those be in constants for when we update the versions

test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
```diff
@@ -1028,6 +1029,300 @@ func VMSpec(ctx context.Context, inputGetter func() VMSpecInput) {
...
+				By("Verifying PolicyEvaluation object is created for the VM")
+				var policyEvaluation vspherepolv1.PolicyEvaluation
+				policyEvaluationName := fmt.Sprintf("vm-%s", vmName)
+				Eventually(func(g Gomega) {
+					err := svClusterClient.Get(ctx, ctrlclient.ObjectKey{
+						Namespace: tmpNamespaceName,
+						Name:      policyEvaluationName,
+					}, &policyEvaluation)
+					g.Expect(err).NotTo(HaveOccurred(), "PolicyEvaluation object should exist")
+				}, 2*time.Minute, 10*time.Second).Should(Succeed(), "PolicyEvaluation should be created")
```
faisalabujabal: should we use a defined timeout from the wcp.yaml file? same comment for all the eventually blocks

faisalabujabal (approved):

test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
bryanv: The version here doesn't really matter - if/when it is bumped, this version must continue to keep working.

test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
bryanv: One of many reasons this setup stuff is horrible but there is an outer AfterEach that does this already (it depends on tmpNamespaceCtx being set)

PR #1620, opened 2026-05-22
🐛 Fix E2E testbed setup script for VPC testbeds and shell sourcing

aruneshpa:

This commit addresses several issues in the E2E testbed setup scripts:
1. Adds support for parsing VPC testbed JSON format where the `vc` field is an object rather than an array.
2. Distinguishes between OS root credentials (for SSH) and SSO credentials (for vCenter API/govc/dcli) to fix authentication errors during E2E tests.
3. Detects NSX networking topology automatically from `TESTBED_TOPOLOGY` for VPC testbeds.
4. Removes `set -u` and improves script sourcing detection to prevent unbound variable errors (e.g., `PS1_PWD_SUB`) when sourced in interactive zsh sessions.

```release-note
Fix E2E testbed setup script for VPC testbeds and zsh sourcing
```

### DIFF
--- a/hack/e2e/setup-e2e-testbed.sh
+++ b/hack/e2e/setup-e2e-testbed.sh
@@ -14,8 +14,6 @@
 #   source ./hack/e2e/setup-e2e-testbed.sh
 #
 
-set -uo pipefail
-
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 MGMT_CIDR='10.0.0.0/8'
 
@@ -56,11 +54,11 @@ fi
 # ---------------------------------------------------------------------------
 GATEWAY_IP=""
 if [ -n "${VC_URL:-}" ] && command -v govc >/dev/null 2>&1; then
-    export GOVC_URL="${VC_ROOT_USERNAME}:${VC_ROOT_PASSWORD}@${VC_URL}"
+    export GOVC_URL="${VC_URL}"
     export GOVC_INSECURE=true
 
     echo "Discovering gateway VM IP via govc..."
-    GATEWAY_IP=$(GOVC_USERNAME="${VC_ROOT_USERNAME}" GOVC_PASSWORD="${VC_ROOT_PASSWORD}" \
+    GATEWAY_IP=$(GOVC_USERNAME="${GOVC_USERNAME}" GOVC_PASSWORD="${GOVC_PASSWORD}" \
         "${SCRIPT_DIR}/proxy.sh" gateway "${VC_URL}" "${MGMT_CIDR}" 2>/dev/null || true)
 
     if [ -n "${GATEWAY_IP:-}" ] && [ "${GATEWAY_IP}" != "null" ]; then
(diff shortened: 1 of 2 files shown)


hack/e2e/setup-testbed-env.sh
```diff
@@ -158,22 +192,31 @@ if [ -z "${VC_ROOT_PASSWORD}" ] || [ "${VC_ROOT_PASSWORD}" = "null" ]; then
...
-    export VC_ROOT_USERNAME="administrator@vsphere.local"
+    export VC_ROOT_USERNAME="root"
+fi
+
+if [ -z "${VCSA_PASSWORD}" ] || [ "${VCSA_PASSWORD}" = "null" ]; then
+    export VCSA_PASSWORD="${VC_ROOT_PASSWORD}"
+fi
+
+if [ -z "${VCSA_USERNAME}" ] || [ "${VCSA_USERNAME}" = "null" ]; then
+    export VCSA_USERNAME="administrator@vsphere.local"
```
faisalabujabal: nit: can be part of the previous if condition

faisalabujabal (approved):

hack/e2e/setup-e2e-testbed.sh
```diff
@@ -56,11 +54,11 @@ fi
 # ---------------------------------------------------------------------------
 GATEWAY_IP=""
 if [ -n "${VC_URL:-}" ] && command -v govc >/dev/null 2>&1; then
-    export GOVC_URL="${VC_ROOT_USERNAME}:${VC_ROOT_PASSWORD}@${VC_URL}"
+    export GOVC_URL="${GOVC_USERNAME}:${GOVC_PASSWORD}@${VC_URL}"
```
faisalabujabal: 🟡 **Medium**: Embedding credentials directly into GOVC_URL is fragile and will fail if the username or password contains URL-reserved characters (e.g., '@', ':', '/'). Since GOVC_USERNAME and GOVC_PASSWORD are already exported in the environment, GOVC_URL should only contain the base URL.

```suggestion
    export GOVC_URL="${VC_URL}"
```

hack/e2e/setup-testbed-env.sh
```diff
@@ -125,23 +145,37 @@ fi
 
 # Extract vCenter connection details - try multiple JSON path patterns
 # Pattern 1: WCP testbed format with nested vc array
-if echo "$testbed_info" | jq -e '.vc[0]' >/dev/null 2>&1; then
-    echo "Using vc[0] format for vCenter details"
+if echo "$testbed_info" | jq -e '.vc | type == "array" and length > 0' >/dev/null 2>&1; then
```
faisalabujabal, line 148: 🟡 **Medium**: Networking type detection will fail for UTS testbeds using NSX-T. When `testbed_info` is extracted from `.deliverable_blob` (line 140), it loses access to the top-level `.TESTBED_TOPOLOGY` field, causing the script to incorrectly default to 'vds' even when the topology is 'NIMBUS_NSXT'.

```suggestion
NETWORKING_TYPE=$(echo "$testbed_data" | jq -r '.deliverable_blob.networking // .networking // if .TESTBED_TOPOLOGY == "NIMBUS_NSXT" then "nsx" else "vds" end')
```

PR #1624, opened 2026-05-26
✨ Separate network interface bootstrap from device

bryanv:

The bootstrap related fields in NetworkInterfaceResult were combined with the device levels fields in a single, large Results struct. The device related fields were separated in 608f62951. This commit does it for the bootstrap fields.

Create new standalone bootstrap functions that take the network interface CR and the interface spec to determine the bootstrap for this interface.  This allows for more straight forward testing since previously we had to go through vcsim tests to test the bootstrap.

The createAndWait* functions now call the appropriate wrapper and convert the result to NetworkInterfaceResult via a new bootstrapToNetworkInterfaceResult helper. Later, the NetworkInterfaceResult will go away and the VM bootstrap will consume just the new Bootstrap struct.

Replace the removed VCSim tests with 65 focused unit tests in bootstrap_test.go covering all paths in InterfaceBootstrap including gateway backfilling, DHCP enable-only semantics, address parsing, global DNS fallback opt-out, legacy NetOP assignment mode detection, mixed IPv4/IPv6 configurations, and IPv6 CIDR handling for NCP and VPC.

Fixes #

```release-note
NONE
```


pkg/providers/vsphere/network/bootstrap_test.go
```diff
@@ -0,0 +1,980 @@
...
+	Context("MTU", func() {
+		When("interfaceSpec.MTU is non-nil", func() {
+			BeforeEach(func() {
+				interfaceSpec.MTU = new(int64(9000))
+			})
+			It("copies the value", func() {
+				Expect(bootstrap.MTU).To(BeEquivalentTo(9000))
+			})
+		})
+
```
faisalabujabal, line 111: 🟠 **High**: Invalid use of the built-in `new()` function with values (e.g., `new(int64(9000))`, `new(false)`, `new(int32(64))`). The built-in `new(T)` expects a type as its argument, not a value. This will cause compilation errors. Use local variables or pointer helper functions to assign pointers to literals.

```suggestion
val := int64(9000)
				interfaceSpec.MTU = &val
```

pkg/providers/vsphere/network/bootstrap_test.go
```diff
@@ -0,0 +1,980 @@
...
+					{IPCIDR: "10.0.0.100/24", IsIPv4: true, Gateway: "10.0.0.1"},
+				}
+				interfaceSpec.Addresses = []string{"192.168.1.100/24", "2001:db8::200/64"}
+			})
+			It("backfills Gateway4 from the later IPv4 entry; Gateway6 is NOT backfilled (locks current quirk)", func() {
+				// The loop encounters IPv6 first while Gateway4==""; it skips setting Gateway4.
+				// It then encounters IPv4 and sets Gateway4. By then Gateway6 is never set.
+				Expect(bootstrap.IPConfigs).To(HaveLen(2))
+				Expect(bootstrap.IPConfigs[0].Gateway).To(Equal("10.0.0.1")) // IPv4 backfilled
+				Expect(bootstrap.IPConfigs[1].Gateway).To(BeEmpty())         // IPv6 not backfilled
```
faisalabujabal: 🟡 **Medium**: This test case asserts and 'locks in' a logic bug in the gateway backfilling implementation where the order of initial IP configurations incorrectly prevents IPv6 gateway assignment. The underlying implementation should be corrected to handle both IP families independently instead of codifying the broken behavior as a 'quirk'.

pkg/providers/vsphere/network/network.go
```diff
@@ -1223,68 +1024,28 @@ func createAndWaitVPCNetworkInterface(
...
-
-func vpcSubnetPortToResult(
-	ctx context.Context,
-	vimClient *vim25.Client,
-	clusterMoRef *vimtypes.ManagedObjectReference,
-	subnetPort *vpcv1alpha1.SubnetPort) (*NetworkInterfaceResult, error) {
-
-	var backing object.NetworkReference
 	networkID := subnetPort.Status.NetworkInterfaceConfig.LogicalSwitchUUID
+	var backing object.NetworkReference
```
faisalabujabal: 🟡 **Medium**: Potential nil pointer dereference if NetworkInterfaceConfig is not populated in the SubnetPort status.

```suggestion
if subnetPort.Status.NetworkInterfaceConfig == nil {
    return nil, fmt.Errorf("ready network interface does not have NetworkInterfaceConfig")
}
networkID := subnetPort.Status.NetworkInterfaceConfig.LogicalSwitchUUID
```

hpannem:
@bryanv while this looks good, I am thinking of the factory pattern, which forces the callers to define the network provider specific bootstrapper. Will share the details on chat.

pkg/providers/vsphere/network/network_test.go
```diff
@@ -893,421 +761,6 @@ var _ = Describe("CreateAndWaitForNetworkInterfaces", Label(testlabels.VCSim), f
 			})
 		})
 
-		Context("User Addresses Override NetOP StaticPool", func() {
```
hpannem, line 896: Should we add any e2e test for testing with user overrides? Cursor says no e2e tests present with overrides.

hpannem (approved):
Looks good. Overtime, I believe we can encapsulate specific provider logic into a NetworkProvider interface keyed on the interface CR type, with all provider-specific behavior as methods on it.

pkg/providers/vsphere/network/bootstrap.go
```diff
@@ -0,0 +1,389 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+//nolint:revive
```
hpannem: we may need a tracker to fix this.

bryanv: It was in network.go and copied along the bootstrap related code over here. I've fixed or pushed down to just the required spots this is still needed.

pkg/providers/vsphere/network/network_test.go
bryanv: Probably. There are so many combinations that it isn't practical to have it be very exhaustive but at least couple to handle common and maybe a comprehensive one too would be good. We'd need to improve the type of e2e asserts we have for this since I don't think we do any validation beyond "does it have an IP?".

pkg/providers/vsphere/network/bootstrap.go
```diff
@@ -0,0 +1,391 @@
...
+					break
+				}
+			}
+		}
+
+		bootstrap.IPConfigs = make([]NetworkInterfaceIPConfig, 0, len(interfaceSpec.Addresses))
+		for _, addr := range interfaceSpec.Addresses {
+			ip, _, err := net.ParseCIDR(addr)
+			if err != nil {
+				continue
```
akutz, line 152: It may be worth logging the error here for debugging purposes.

akutz (approved):
Approved with one comment. Thanks @bryanv!

PR #1627, opened 2026-05-27
🐛 fix(vmlifecycle): prevent premature VirtualMachineGuestNetworkConfigSynced=True

zhangheliu:

gi.IpAddress (the vCenter summary scalar) is updated by vCenter/NSX
the moment a portgroup assignment changes, while gi.Net[] (per-NIC
detail) is only updated once VMware Tools inside the guest publishes
the new configuration.  During a VM re-import or live-migration these
two fields can diverge across reconcile loops: gi.IpAddress already
carries the new IP while gi.Net[] still reflects the old one.

The previous condition-setting logic checked only gi.IpAddress, so it
could set VirtualMachineGuestNetworkConfigSynced=True before the guest
had actually converged, giving consumers a false signal.

This change introduces a cross-check: when gi.Net[] is non-empty,
every non-empty primary IP (IPv4 and/or IPv6) must be present in the
per-NIC interface statuses before the condition is set to True.  When
gi.Net[] is absent (VMware Tools not yet running), the check is
skipped so existing behaviour for those VMs is preserved.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmlifecycle/update_status.go
+++ b/pkg/providers/vsphere/vmlifecycle/update_status.go
@@ -599,6 +599,27 @@ func findInterfaceContainingIP(ip string, ifaces []vmopv1.VirtualMachineNetworkI
 	return -1
 }
 
+// primaryIPsInNICData returns true when the per-NIC data from gi.Net[] is
+// either absent (nothing to cross-check) or already contains every non-empty
+// primary IP. When gi.IpAddress (fast, summary scalar set by vCenter) and
+// gi.Net[] (slow, per-NIC data published by VMware Tools) diverge across a
+// reconcile loop this returns false, preventing a premature Synced condition.
+func primaryIPsInNICData(
+	ip4, ip6 string,
+	ifaces []vmopv1.VirtualMachineNetworkInterfaceStatus) bool {
+
+	if len(ifaces) == 0 {
+		return true
+	}
+	if ip4 != "" && findInterfaceContainingIP(ip4, ifaces) < 0 {
+		return false
+	}
+	if ip6 != "" && findInterfaceContainingIP(ip6, ifaces) < 0 {
+		return false
+	}
+	return true
+}
+
 // extractIPsFromInterface extracts IP addresses of the specified family from an interface.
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/vmlifecycle/update_status.go
```diff
@@ -1228,11 +1249,19 @@ func updateGuestNetworkStatus(
...
+					vmopv1.VirtualMachineGuestNetworkConfigSynced)
+				c.Reason = "Synced"
+				c.Message = fmt.Sprintf(
+					"IPv4=%q, IPv6=%q", primaryIP4, primaryIP6)
+				conditions.Set(vm, c)
+			} else {
+				conditions.MarkFalse(vm,
+					vmopv1.VirtualMachineGuestNetworkConfigSynced,
+					"NotSynced", "%s",
+					"Guest IP addresses not yet reflected in interface status")
```
bryanv: I'd mention guest primary IPs here in the message

bryanv (approved):
I'm not sure if this is going to make a meaningful difference in the issue this is addressing. When not using CloudInit, this is all just pushed down via VM Tools, and I'm not sure how or why it would report a primary IP w/o the rest. CloudInit is different in that the VMware Datasource pushes down what it views as the primary IPs, which will happen separately from VM Tools.

Longer term, this condition should be eval'd with the expected spec data (basically, the bootstrap data) instead of with the primary IPs.

zhangheliu:
> I'm not sure if this is going to make a meaningful difference in the issue this is addressing. When not using CloudInit, this is all just pushed down via VM Tools, and I'm not sure how or why it would report a primary IP w/o the rest. CloudInit is different in that the VMware Datasource pushes down what it views as the primary IPs, which will happen separately from VM Tools.
> 
> Longer term, this condition should be eval'd with the expected spec data (basically, the bootstrap data) instead of with the primary IPs.

Right, I discussed this with the mobility operator team. This change is not meant to address the issue as a long term solution but more to correctly surface the issue when their pipeline test code checking against our condition. At least they would identify the issue faster with this change.

pkg/providers/vsphere/vmlifecycle/update_status.go
zhangheliu: addressed

PR #1630, opened 2026-05-29
Reconcile VM's location when VM is moved outside of expected folder

anwithaun-1:

- validateVMFolder: A new method added to assess if the VM's active folder is the designated namespace folder or a valid child folder associated with a VirtualMachineSetResourcePolicy.
- validateVMResourcePool: Refactored to tightly restrict valid locations to explicit VKS child resource pools derived from active namespace resource policies.
-  Integrated an API helper (vcenter.GetParentFolder) to navigate and verify the structural vSphere inventory folder paths.
- Refactored the reconcileLocation workflow in vmprovider_vm.go to evaluate both the resource pool (isVMInValidRP) and the inventory folder (isVMInValidFolder).
- Unit Tests and e2e tests are added 

Fixes #
vmop-3844

Unaddressed comments from  - [https://github.com/vmware-tanzu/vm-operator/pull/1558](url) have been addressed in this PR. 

```
Last Transition Time:  2026-06-19T06:39:27Z
    Message:               VM was moved to an unexpected ResourcePool or Folder
    Reason:                ResourcePoolMismatch
    Status:                False
    Type:                  VirtualMachineLocationValid
```
```
Last Transition Time:  2026-06-19T06:33:22Z
    Message:               
    Reason:                True
    Status:                True
    Type:                  VirtualMachineLocationValid

```
```
Last Transition Time:  2026-06-19T06:46:47Z
    Message:               VM was moved to an unexpected ResourcePool or Folder
    Reason:                FolderMismatch
    Status:                False
    Type:                  VirtualMachineLocationValid
```

```release-note

```

### DIFF
--- a/pkg/providers/vsphere/vcenter/resourcepool.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool.go
@@ -180,3 +180,20 @@ func GetResourcePoolParent(
 
 	return moPool, nil
 }
+
+func GetFolderParent(
+	ctx context.Context,
+	vimClient *vim25.Client,
+	currentFolderMoID string) (mo.Folder, error) {
+
+	folderObj := object.NewFolder(vimClient,
+		vimtypes.ManagedObjectReference{Type: "Folder", Value: currentFolderMoID})
+
+	var moFolder mo.Folder
+	err := folderObj.Properties(ctx, folderObj.Reference(), []string{"parent"}, &moFolder)
+	if err != nil {
+		return mo.Folder{}, fmt.Errorf("failed to fetch parent properties for folder %s: %w",
+			currentFolderMoID, err)
+	}
+	return moFolder, nil
+}
--- a/pkg/providers/vsphere/vcenter/resourcepool.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool.go
@@ -173,7 +173,7 @@ func GetResourcePoolParent(
 		vimtypes.ManagedObjectReference{Type: "ResourcePool", Value: currentRPMoID})
 
 	var moPool mo.ResourcePool
-	err := poolObj.Properties(ctx, poolObj.Reference(), []string{"parent"}, &moPool)
+	err := poolObj.Properties(ctx, poolObj.Reference(), []string{"parent", "name"}, &moPool)
 	if err != nil {
 		return mo.ResourcePool{}, err
 	}
@@ -190,7 +190,7 @@ func GetFolderParent(
 		vimtypes.ManagedObjectReference{Type: "Folder", Value: currentFolderMoID})
 
 	var moFolder mo.Folder
-	err := folderObj.Properties(ctx, folderObj.Reference(), []string{"parent"}, &moFolder)
+	err := folderObj.Properties(ctx, folderObj.Reference(), []string{"parent", "name"}, &moFolder)
 	if err != nil {
 		return mo.Folder{}, fmt.Errorf("failed to fetch parent properties for folder %s: %w",
 			currentFolderMoID, err)
--- a/pkg/providers/vsphere/vcenter/resourcepool.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool.go
@@ -164,7 +164,7 @@ func findChildRP(
 	return childRP, nil
 }
 
-func GetResourcePoolParent(
+func GetParentResourcePool(
 	ctx context.Context,
 	vimClient *vim25.Client,
 	currentRPMoID string) (mo.ResourcePool, error) {
@@ -181,7 +181,7 @@ func GetResourcePoolParent(
 	return moPool, nil
 }
 
-func GetFolderParent(
+func GetParentFolder(
 	ctx context.Context,
 	vimClient *vim25.Client,
 	currentFolderMoID string) (mo.Folder, error) {
--- a/pkg/providers/vsphere/vcenter/resourcepool.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool.go
@@ -164,29 +164,19 @@ func findChildRP(
 	return childRP, nil
 }
 
-func IsVMInValidResourcePool(
+func GetResourcePoolParent(
 	ctx context.Context,
 	vimClient *vim25.Client,
-	currentRPMoID string,
-	expectedRootRPMoID string) (bool, error) {
+	currentRPMoID string) (mo.ResourcePool, error) {
 
-	// Case A: Direct match
-	if currentRPMoID == expectedRootRPMoID {
-		return true, nil
-	}
-	// Case B: Check if current pool's parent is the root
 	poolObj := object.NewResourcePool(vimClient,
 		vimtypes.ManagedObjectReference{Type: "ResourcePool", Value: currentRPMoID})
 
 	var moPool mo.ResourcePool
 	err := poolObj.Properties(ctx, poolObj.Reference(), []string{"parent"}, &moPool)
 	if err != nil {
-		return false, err
+		return mo.ResourcePool{}, err
 	}
 
-	if moPool.Parent != nil && moPool.Parent.Value == expectedRootRPMoID {
-		return true, nil
-	}
-
-	return false, nil
+	return moPool, nil
 }
--- a/pkg/providers/vsphere/vcenter/resourcepool.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool.go
@@ -164,36 +164,38 @@ func findChildRP(
 	return childRP, nil
 }
 
-func GetParentResourcePool(
+// GetResourcePoolProperties fetches the "parent" and "name" properties of the
+// given ResourcePool MoID and returns the populated mo.ResourcePool.
+func GetResourcePoolProperties(
 	ctx context.Context,
 	vimClient *vim25.Client,
-	currentRPMoID string) (mo.ResourcePool, error) {
+	rpMoID string) (mo.ResourcePool, error) {
 
 	poolObj := object.NewResourcePool(vimClient,
-		vimtypes.ManagedObjectReference{Type: "ResourcePool", Value: currentRPMoID})
+		vimtypes.ManagedObjectReference{Type: "ResourcePool", Value: rpMoID})
 
 	var moPool mo.ResourcePool
-	err := poolObj.Properties(ctx, poolObj.Reference(), []string{"parent", "name"}, &moPool)
-	if err != nil {
+	if err := poolObj.Properties(ctx, poolObj.Reference(), []string{"parent", "name"}, &moPool); err != nil {
 		return mo.ResourcePool{}, err
 	}
 
 	return moPool, nil
 }
 
-func GetParentFolder(
+// GetFolderProperties fetches the "parent" and "name" properties of the given
+// Folder MoID and returns the populated mo.Folder.
+func GetFolderProperties(
 	ctx context.Context,
 	vimClient *vim25.Client,
-	currentFolderMoID string) (mo.Folder, error) {
+	folderMoID string) (mo.Folder, error) {
 
 	folderObj := object.NewFolder(vimClient,
-		vimtypes.ManagedObjectReference{Type: "Folder", Value: currentFolderMoID})
+		vimtypes.ManagedObjectReference{Type: "Folder", Value: folderMoID})
 
 	var moFolder mo.Folder
-	err := folderObj.Properties(ctx, folderObj.Reference(), []string{"parent", "name"}, &moFolder)
-	if err != nil {
-		return mo.Folder{}, fmt.Errorf("failed to fetch parent properties for folder %s: %w",
-			currentFolderMoID, err)
+	if err := folderObj.Properties(ctx, folderObj.Reference(), []string{"parent", "name"}, &moFolder); err != nil {
+		return mo.Folder{}, err
 	}
+
 	return moFolder, nil
 }
--- a/pkg/providers/vsphere/vcenter/resourcepool_test.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool_test.go
@@ -198,12 +198,12 @@ func createDeleteExistResourcePoolChild() {
 				Expect(err).To(HaveOccurred())
 			})
 
-		By("NoOp when child does not exist", func() {
-			err := vcenter.DeleteChildResourcePool(ctx, ctx.VCClient.Client, parentRPMoID, childName)
-			Expect(err).ToNot(HaveOccurred())
+			By("NoOp when child does not exist", func() {
+				err := vcenter.DeleteChildResourcePool(ctx, ctx.VCClient.Client, parentRPMoID, childName)
+				Expect(err).ToNot(HaveOccurred())
+			})
 		})
 	})
-})
 }
 
 func getParentResourcePoolTests() {
--- a/pkg/providers/vsphere/vcenter/resourcepool_test.go
+++ b/pkg/providers/vsphere/vcenter/resourcepool_test.go
@@ -19,6 +19,8 @@ import (
 func resourcePoolTests() {
 	Describe("GetResourcePool", getResourcePoolTests)
 	Describe("CreateDeleteExistResourcePoolChild", createDeleteExistResourcePoolChild)
+	Describe("GetParentResourcePool", getParentResourcePoolTests)
+	Describe("GetParentFolder", getParentFolderTests)
 }
 
 func getResourcePoolTests() {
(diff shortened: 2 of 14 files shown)


pkg/providers/vsphere/vcenter/resourcepool_test.go
```diff
@@ -196,10 +198,97 @@ func createDeleteExistResourcePoolChild() {
 				Expect(err).To(HaveOccurred())
 			})
 
-			By("NoOp when child does not exist", func() {
-				err := vcenter.DeleteChildResourcePool(ctx, ctx.VCClient.Client, parentRPMoID, childName)
-				Expect(err).ToNot(HaveOccurred())
-			})
+		By("NoOp when child does not exist", func() {
```
bryanv: Incorrect indent

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,104 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
 
-	// Check if the VM is in the Root RP or a Child RP
-	isValid, err := vcenter.IsVMInValidResourcePool(
-		vmCtx,
-		vcClient.VimClient(),
-		vmCtx.MoVM.ResourcePool.Value,
-		expectedRootRPMoID,
-	)
+	// 2. List policies in the namespace
+	resourcePolicies, err := GetVMSetResourcePolicies(vmCtx, vs.k8sClient)
```
bryanv: There is no reason to list here - the VM's VMSetRP is in its Spec, if it has one.

But the lifecycle of those resources aren't that of the VM so I've made the intentional decision that we do not try to reference this for a VM after it has been created.

anwithaun-1: @bryanv The intention of getting resourcePolicies is to validate if the VM is moved to any of the Child VKS. If the VM is in child RP , but that is not referenced by VMSetRP, then it is considered as invalid Child RP for the VM to be in.
cc: @aruneshpa

bryanv: I don't understand the rational for viewing that is OK for what this is trying to enforce. I'm not sure of the evidence for saying what this is trying to prevent is meaningful issue. Either way, we cannot required that the VMSetRP exists for the successful reconcile of a VM post creation.

anwithaun-1: @bryanv 
We are addressing the issue of a virtual machine being manually moved outside its assigned namespace resource pool (RP) in vCenter. The impact of this is that the VM escapes its Kubernetes resource quotas, desyncs the operator's expected management state, and inadvertently extends DevOps user privileges to update VMs outside their designated boundaries. Because quotas remain intact if the VM is moved to a child RP within the namespace, the operator will only flag and disallow VMs found completely outside the namespace RP hierarchy.

The changes that are added here is specific to allowing only VKS child RP of the assigned namespace as valid child RP for the VM to be present. So how do you suggest we handle VKS VM's to solve this case?

cc: @aruneshpa

anwithaun-1: @aruneshpa @bryanv

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
+
+	return false, nil
+}
+
+func validateVMFolder(
+	vmCtx pkgctx.VirtualMachineContext,
+	expectedFolder string,
+	vcClient *vcclient.Client) (bool, error) {
+
+	if vmCtx.MoVM.Parent.Value == expectedFolder {
```
aruneshpa, line 1248: Parent of a MoVM is a pointer so it can be nil.  Guard against that to avoid NPEs.

pkg/providers/vsphere/vcenter/resourcepool.go
```diff
@@ -164,29 +164,36 @@ func findChildRP(
 	return childRP, nil
 }
 
-func IsVMInValidResourcePool(
+func GetParentResourcePool(
```
aruneshpa: Hmm.  Even though this method is called GetParentResourcePool this doesn't really return the parent?  It simply populates it (using prop collector fetch).  I would name it appropriately, or add doc string for this.
Maybe call it GetResourcePoolProperties, or actually return the parent RP.

aruneshpa: Same comment for GetParentFolder

pkg/providers/vsphere/vcenter/resourcepool.go
```diff
@@ -164,29 +164,36 @@ func findChildRP(
...
+	currentFolderMoID string) (mo.Folder, error) {
+
+	folderObj := object.NewFolder(vimClient,
+		vimtypes.ManagedObjectReference{Type: "Folder", Value: currentFolderMoID})
 
-	return false, nil
+	var moFolder mo.Folder
+	err := folderObj.Properties(ctx, folderObj.Reference(), []string{"parent", "name"}, &moFolder)
+	if err != nil {
+		return mo.Folder{}, fmt.Errorf("failed to fetch parent properties for folder %s: %w",
```
aruneshpa: GetParentResourcePool is nearly identical and yet we don't return a wrapped error there.  Either wrap the error in the `GetParentResourcePool` method as well, or lets drop this guy too so things are consistent.

My preference is to drop this from here since the caller of this method (reconcileLocation) already wraps helper text around returned errors.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
 	}
 
-	pkgcond.MarkTrue(vmCtx.VM, vmopv1.VirtualMachineInValidLocation)
+	if currentCond == nil || currentCond.Status != metav1.ConditionTrue {
+		pkgcond.MarkTrue(vmCtx.VM, vmopv1.VirtualMachineInValidLocation)
+	}
 	return nil
 }
 
+func validateVMResourcePool(
```
aruneshpa, line 1214: Comment here (or the caller) explaining why we are validating only that the current RP = namespace RP or parent RP = namespace RP (for VKS).
I would also add a line about why we are only validating at one level of depth.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
 		return pkgerr.NoRequeueError{Message: fmt.Sprintf(
-			"reconciliation paused for the VM %s because it is moved to invalid Resource Pool. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s",
-			vmCtx.VM.Name, expectedRootRPMoID, vmCtx.MoVM.ResourcePool.Value)}
+			"reconciliation paused for the VM %s because it is moved to invalid ResourcePool/Folder. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s,Expected Folder MoRef: %s, Current Folder MoRef: %s. Move the VM back to the Right RP/Folder hierarchy for namespace %s to resume reconciliation.",
+			vmCtx.VM.Name, expectedRootRPMoID, vmCtx.MoVM.ResourcePool.Value, expectedFolder, vmCtx.MoVM.Parent.Value, vmCtx.VM.Namespace)}
 	}
 
-	pkgcond.MarkTrue(vmCtx.VM, vmopv1.VirtualMachineInValidLocation)
+	if currentCond == nil || currentCond.Status != metav1.ConditionTrue {
+		pkgcond.MarkTrue(vmCtx.VM, vmopv1.VirtualMachineInValidLocation)
```
aruneshpa: I am now realizing that this name is very confusing.
`VirtualMacine *In Valid* Location`
vs 
`VirtualMachine *Invalid* Location`

and the difference is only in casing of a letter.

Let's change this to a different name.  How about:
`VirtualMachineLocationValid`?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
-	if !isValid {
-		pkgcond.MarkFalse(
-			vmCtx.VM,
-			vmopv1.VirtualMachineInValidLocation,
-			"LocationMismatch",
-			"VM is in an invalid Resource Pool. Move the VM back to the Resource Pool hierarchy for namespace %s to resume reconciliation.",
-			vmCtx.VM.Namespace,
-		)
+	if !isVMInValidRP || !isVMInValidFolder {
+		if currentCond == nil || currentCond.Status != metav1.ConditionFalse {
```
aruneshpa: I am curious about the reason for this change.
Everywhere in the code, we set Conditions unconditionaly (MarkTrue / MarkFalse) and let PatchHelper figure out if a patch needs to be sent to the API server.
Unless we have a strong reason not to, let's follow that pattern.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
+			pkgcond.MarkFalse(
+				vmCtx.VM,
+				vmopv1.VirtualMachineInValidLocation,
+				"LocationMismatch",
```
aruneshpa: Why not have distinct reasons indicating if this is a RP mismatch or Folder mismatch?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1120,20 +1121,20 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
 }
 
 func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContext, vcClient *vcclient.Client) error {
-	logger := pkglog.FromContextOrDefault(vmCtx)
+	vmCtx.Logger.V(4).Info("Reconciling VirtualMachine location")
```
aruneshpa, line 1143: Can you also add a comment here indicating that if a VM is not in its intended RP or Folder, we will return and requeue the reconcile meaning the VM will never be reconciled.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1120,20 +1121,20 @@ func (vs *vSphereVMProvider) updateVirtualMachine(
...
+		return fmt.Errorf("VM is not assigned to any resource pools")
 	}
 
 	// If the VM doesn't have a topology zone label, skip location validation.
 	// This can happen in unit tests or when the zone is being determined.
 	// Similar to how vmlifecycle/update_status.go handles missing zone labels.
-	zoneName := vmCtx.VM.Labels[corev1.LabelTopologyZone]
+	zoneName := vmCtx.VM.Status.Zone
 	if zoneName == "" {
 		return nil
```
aruneshpa, line 1154: Can we log something here (maybe a warning)?  VM's zone is almost always supposed to be set once it is created.  So if this happens in steady state, we would want to know.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
+				"LocationMismatch",
+				"VM is in an invalid ResourcePool or Folder. Move the VM back to the Right RP/Folder hierarchy for namespace %s to resume reconciliation.",
+				vmCtx.VM.Namespace,
+			)
+		}
 
 		return pkgerr.NoRequeueError{Message: fmt.Sprintf(
-			"reconciliation paused for the VM %s because it is moved to invalid Resource Pool. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s",
-			vmCtx.VM.Name, expectedRootRPMoID, vmCtx.MoVM.ResourcePool.Value)}
+			"reconciliation paused for the VM %s because it is moved to invalid ResourcePool/Folder. Expected Resource Pool MoRef: %s, Current Resource Pool MoRef: %s,Expected Folder MoRef: %s, Current Folder MoRef: %s. Move the VM back to the Right RP/Folder hierarchy for namespace %s to resume reconciliation.",
```
aruneshpa: This feels too verbose to me.  Can we make it:

```suggestion
return pkgerr.NoRequeueError{Message: fmt.Sprintf(
    "VM %s/%s relocated outside expected hierarchy: "+
        "expected RP %s got %s, expected folder %s got %s",
    vmCtx.VM.Namespace, vmCtx.VM.Name,
    expectedRootRPMoID, vmCtx.MoVM.ResourcePool.Value,
    expectedFolder, vmCtx.MoVM.Parent.Value,
)}
```

I am also not sure if putting the MoRef in the Conditions have any value.  DevOps users can't do anything about it other than talking to their admin.  Maybe @bryanv has thoughts.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
-			"VM is in an invalid Resource Pool. Move the VM back to the Resource Pool hierarchy for namespace %s to resume reconciliation.",
-			vmCtx.VM.Namespace,
-		)
+	if !isVMInValidRP || !isVMInValidFolder {
+		if currentCond == nil || currentCond.Status != metav1.ConditionFalse {
+			pkgcond.MarkFalse(
+				vmCtx.VM,
+				vmopv1.VirtualMachineInValidLocation,
+				"LocationMismatch",
+				"VM is in an invalid ResourcePool or Folder. Move the VM back to the Right RP/Folder hierarchy for namespace %s to resume reconciliation.",
```
aruneshpa:
```suggestion
pkgcond.MarkFalse(
    vmCtx.VM,
    vmopv1.VirtualMachineInValidLocation,
    "LocationMismatch",
    "VM was moved to an unexpected ResourcePool or Folder",
)
```

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1143,37 +1144,88 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
+	}
+	moFolder, err := vcenter.GetParentFolder(
+		vmCtx,
+		vcClient.VimClient(),
+		vmCtx.MoVM.Parent.Value,
+	)
+	if err != nil {
+		return false, err
+	}
+	// this is to verify if the folder is child of parent Namespace Folder
```
aruneshpa: Is this a leftover from debugging?  If not can we rephrase this to make it more readable (e.g, Capitalization etc).  Same comment for L1223.

pkg/providers/vsphere/vmprovider_vm_location_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+				// Second call: condition is already False — must not touch LastTransitionTime.
+				Expect(callProviderOnce()).Error().To(HaveOccurred())
+				cond2 := conditions.Get(vm, vmopv1.VirtualMachineInValidLocation)
+				Expect(cond2).ToNot(BeNil())
+				Expect(cond2.LastTransitionTime).To(Equal(ltt1))
+			})
+		})
+
+		Context("condition idempotency (True)", func() {
```
aruneshpa, line 292: Can you also add a test that validates that the Condition gets reset when the VM is moved back to the correct RP and / or Folder?

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Agree.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: The reason it's added here is to avoid updating the condition again if it's already set. I observed that multiple updates happening on the condition even though it was set already.

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: Fixed it.

pkg/providers/vsphere/vmprovider_utils.go
```diff
@@ -551,6 +551,18 @@ func GetVMSetResourcePolicy(
 	return resourcePolicy, nil
 }
 
+func GetVMSetResourcePolicies(
```
bryanv: Unused?

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1144,37 +1148,102 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
+	vcClient *vcclient.Client) (bool, error) {
+
+	if vmCtx.MoVM.ResourcePool != nil && vmCtx.MoVM.ResourcePool.Value == expectedRootRPMoID {
+		return true, nil
+	}
+
+	moPool, err := vcenter.GetResourcePoolProperties(
+		vmCtx,
+		vcClient.VimClient(),
+		vmCtx.MoVM.ResourcePool.Value,
```
bryanv, line 1226: You check `vmCtx.MoVM.ResourcePool` for nil above but then not here

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1144,37 +1148,102 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
+	expectedFolder string,
+	vcClient *vcclient.Client) (bool, error) {
+
+	if vmCtx.MoVM.Parent != nil && vmCtx.MoVM.Parent.Value == expectedFolder {
+		return true, nil
+	}
+	moFolder, err := vcenter.GetFolderProperties(
+		vmCtx,
+		vcClient.VimClient(),
+		vmCtx.MoVM.Parent.Value,
```
bryanv, line 1254: Similar for `vmCtx.MoVM.Parent`

pkg/providers/vsphere/vmprovider_vm_location_test.go
```diff
@@ -0,0 +1,315 @@
...
+			Expect(task.Wait(ctx)).To(Succeed())
+
+			err = callProviderOnce()
+			Expect(err).To(HaveOccurred())
+			var noRequeueErr pkgerr.NoRequeueError
+			Expect(errors.As(err, &noRequeueErr)).To(BeTrue())
+
+			Expect(conditions.IsFalse(vm, vmopv1.VirtualMachineLocationValid)).To(BeTrue())
+			cond := conditions.Get(vm, vmopv1.VirtualMachineLocationValid)
+			Expect(cond).ToNot(BeNil())
```
bryanv, line 224: This is never going to fail because of the earlier assertion. You could check the condition status here instead

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1144,37 +1148,102 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
 	}
 
-	// Handle Mismatch
-	if !isValid {
+	// If the VM is outside its expected RP or Folder hierarchy, block all further
+	// reconciliation by returning a NoRequeueError. The VM will not be reconciled
+	// again until its inventory location changes.
+	if !isVMInValidRP || !isVMInValidFolder {
+		var reason string
+		if !isVMInValidRP {
```
bryanv: I don't quite get this - if both are invalid we know that and the condition should reflect that.

bryanv (approved):

PR #1635, opened 2026-06-01
  🐛 Call VM InvokeFSR after ChangeBlockTracking reconfigure of powered on VM

bryanv:

For a CBT change to actually take effect on a powered on VM, InvokeFSR needs called. Since doReconfigure will return ErrReconfigure when a reconfigure is done, in the next reconcile, ConfigSpec CBT will not be set since it was updated on the prior reconcile, but that means we wouldn't call InvokeFSR.

Prior to ErrReconfigure, the InvokeFSR called would have just fallen through after the successful reconfigure. This change restores that effective behavior, but leaves the issue of InvokeFSR failing in which we won't then won't know to call InvokeFSR on a later reconcile unaddressed.

vcsim doesn't implement the InvokeFSR task so we just assert that our handler was called.

Fixes #

```release-note
After a ChangeBlockTracking change for a powered on VM, call InvokeFSR so the change takes effect.
```

### DIFF
--- a/pkg/providers/vsphere/session/session_vm_update.go
+++ b/pkg/providers/vsphere/session/session_vm_update.go
@@ -430,7 +430,7 @@ func (s *Session) poweredOnReconfigure(
 		return fmt.Errorf("update CD-ROM device connection error: %w", err)
 	}
 
-	if err := doReconfigure(
+	err := doReconfigure(
 		logr.NewContext(
 			vmCtx,
 			vmCtx.Logger.WithName("poweredOnReconfigure"),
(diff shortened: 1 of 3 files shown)


pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -439,12 +439,22 @@ func (s *Session) poweredOnReconfigure(
...
-		*configSpec); err != nil {
+		*configSpec)
 
-		return err
+	if errors.Is(err, ErrReconfigure) {
+		if cbtErr := s.reconcilePoweredOnChangeBlockTracking(vmCtx, configSpec); cbtErr != nil {
+			// The Reconfigure has been performed but we have no way to later determine if
+			// this needs to be called. We used to just this best effort after a reconfigure
+			// so continue to do that here for now. That the reconfigure was not just handle
+			// this is a broken VC API. A few options for later:
```
aruneshpa: > That the reconfigure was not just handle
			// this is a broken VC API.

This seems confusing.  Can you rephrase this?

aruneshpa (approved):

PR #1636, opened 2026-06-01
✨  Add initial per namespace scope network provider

bryanv:

When the PerNamespaceNetworkProvider capability is enabled, the new NetworkSettings CR is used to obtain the network provider for the namespace. When disabled, the old envvar is used. Introduce helper to wrap getting the network provider in the namespace.

Update the VM mutation & validation webhooks to use the new helper, and reorganize the tests so both legacy and capability ways have the same assertions.

Fixes #

vmop-3553, vmop-3699

```release-note
NONE
```


ihgann (approved):
LGTM!

config/crd/external-crds/netoperator.vmware.com_networksettings.yaml
```diff
@@ -41,7 +41,7 @@ spec:
             type: object
           provider:
             description: |-
-              Provider is the active network provider in this namespace. Workloads and network-aware
+              provider is the active network provider in this namespace. Workloads and network-aware
```
aruneshpa, line 44: I am surprised with this casing.  Shouldn't this be capitalized like other fields since this is a field in the type?

pkg/util/kube/networksettings/networksettings.go
```diff
@@ -0,0 +1,69 @@
...
+	// ErrNetworkSettingsNotFound is returned when the NetworkSettings CR for the
+	// namespace does not exist. This CR is supposed to be created "immediately" after
+	// namespace creation, so it is an unfortunate race if actually observed and
+	// should be transient.
+	ErrNetworkSettingsNotFound = errors.New("NetworkSettings has not been created for this namespace")
+)
+
+// GetProviderType returns the NetworkProviderType for the given namespace. When the
+// PerNamespaceNetworkProvider capability is not enabled, return provider type that is
+// specified globally by the NETWORK_PROVIDER envvar. When the capability is enabled
```
aruneshpa, line 34: Out of curiosity, when the cap is enabled, what will the value of this env var be in case of mixed network providers in namespaces?  Are clients no longer supposed to rely on this var?

pkg/util/kube/networksettings/networksettings.go
```diff
@@ -0,0 +1,69 @@
...
+	ctx context.Context,
+	reader ctrlclient.Reader,
+	namespace string) (pkgcfg.NetworkProviderType, error) {
+
+	if !pkgcfg.FromContext(ctx).Features.PerNamespaceNetworkProvider {
+		return pkgcfg.FromContext(ctx).NetworkProviderType, nil
+	}
+
+	var ns netopv1alpha1.NetworkSettings
+	err := reader.Get(ctx, ctrlclient.ObjectKey{Name: defaultNetworkSettingsName, Namespace: namespace}, &ns)
```
aruneshpa, line 47: Are you advocating that we follow this pattern of always reading the reader if no writes are to be made to an object?  (I am onboard, just want to make sure this isn't a one off due to some reason that I don't understand).

pkg/util/kube/networksettings/networksettings.go
```diff
@@ -0,0 +1,69 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	netopv1alpha1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
+
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+)
+
+const defaultNetworkSettingsName = "default"
```
aruneshpa, line 22: Is this guaranteed to be the name of the default settings?  If there's only ever a singleton resource in the namespace, maybe that can be how we get the resource with a second validation that the name of that resource is "default"?  Or do you think that's an overkill for this?

At the very least, consider adding a comment here that describes the expectation of the resource always being present in the namespace with this name?

pkg/util/kube/networksettings/networksettings.go
```diff
@@ -0,0 +1,69 @@
...
+func GetProviderType(
+	ctx context.Context,
+	reader ctrlclient.Reader,
+	namespace string) (pkgcfg.NetworkProviderType, error) {
+
+	if !pkgcfg.FromContext(ctx).Features.PerNamespaceNetworkProvider {
+		return pkgcfg.FromContext(ctx).NetworkProviderType, nil
+	}
+
+	var ns netopv1alpha1.NetworkSettings
```
aruneshpa, line 46: vvvvvvNit: `ns` is almost always used for a variable that points to a namespace.  Maybe consider using some other name?  Consider "netset"?

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -541,75 +553,94 @@ func getDefaultNetworkRef(
...
-			return false
+			return false, nil
+		}
+
+		ok, networkRef, err := getDefaultNetworkRef(ctx, client, vm)
+		if err != nil {
+			return false, err
+		}
+		if !ok {
+			return false, nil
```
aruneshpa, line 592: The `getDefaultNetworkRef` method doesn't return an error if the NetworkSettings.Spec.Provider is invalid (the default switch case).  And this method seems to ignore the error as well.  Should we consider denying the request in that case?

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
```diff
@@ -541,75 +553,94 @@ func getDefaultNetworkRef(
...
 			{
 				Name:    defaultInterfaceName,
 				Network: &networkRef,
 			},
 		}
+		return true, nil
+	}
 
-		return true
+	// Back-fill the network ref on existing interfaces, even when Disabled.
```
aruneshpa, line 608: Is this just to maintain prior behavior?  Because I would expect this to be an invalid request if a user specifies spec.network.disabled and network.interfaces[x] at the same time.

config/crd/external-crds/netoperator.vmware.com_networksettings.yaml
bryanv: It is what is in upstream repo. I think this a result of running kube lint and I think does match upstream convention, which isn't something that we're doing.

pkg/util/kube/networksettings/networksettings.go
bryanv: I would expect it (well, the underlying infra value that ours reflects) to still be set to the primary (VPC) because lots of things in other contexts still need it w/o being k8s aware, but we should not be using it

pkg/util/kube/networksettings/networksettings.go
bryanv: It is always the name.

pkg/util/kube/networksettings/networksettings.go
bryanv: It doesn't really matter but we only need the Reader interface so that's what is passed in here.

pkg/util/kube/networksettings/networksettings.go
bryanv: IDK in this context it is used right there and meh

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
bryanv: It is the prior behavior. Disabled just means we don't add them to the VC VM. I don't see a reason to not set these. I removed the Disabled check in the update path since that flag does not indicate us mutating in, just that the VM won't have interfaces.

webhooks/virtualmachine/mutation/virtualmachine_mutator.go
bryanv: We can treat this as a sealed enum, and we've got validation webhook checks.

aruneshpa (approved):

PR #1641, opened 2026-06-05
✨ api/v1alpha6: make `DHCP4`/`DHCP6` tri-state (`*bool`)

hpannem:

`VirtualMachineNetworkInterfaceSpec.DHCP4` and `DHCP6` are changed from `bool` to `*bool` in v1alpha6 (hub). The three values carry distinct semantics:

| Value | Meaning |
|-------|---------|
| `nil` | Unset — network provider determines the allocation method from subnet configuration |
| `*false` | Explicitly disabled — static allocation only |
| `*true` | Explicitly enabled — request an address via DHCP |

Previously these were plain booleans, making it impossible to distinguish "user never set this" from "user explicitly disabled DHCP". The pointer type closes that gap and also allows user intent to override a provider-reported DHCP lease (setting `*false` now disables DHCP even when the provider enabled it).

**Conversion semantics (old versions → v1alpha6):**

Old spoke versions (v1alpha1–v1alpha5) use `bool` where `false` is the zero value and carries no explicit intent. The conversion maps:
- spoke `false` → hub `nil` (unset, defer to provider)
- spoke `true` → hub `*true` (explicitly enabled)

**Changes:**

- `api/v1alpha6`: `DHCP4`/`DHCP6` fields changed to `*bool`; doc comments updated to describe all three states
- `api/v1alpha{2,3,4,5}/virtualmachine_conversion.go`: manual `Convert_v1alphaX_VirtualMachineNetworkInterfaceSpec_To_v1alpha6_...` override added; calls `autoConvert_` then remaps `&false → nil`; existing restore logic in `restore_v1alpha6_VirtualMachineNetworkInterfaces` correctly handles `*false` restoration from the annotation
- `api/v1alpha{2,3,4,5}/zz_generated.conversion.go`: generated public `Convert_` wrapper removed (superseded by the manual override above; the generator will skip regenerating it when a manual function with the same name is present)
- `pkg/providers/vsphere/network/bootstrap.go`: DHCP spec overlay updated — `nil` defers to provider result, non-nil overrides; resolves the long-standing TODO about being unable to unset provider-reported DHCP
- CRD YAML and deepcopy regenerated

**Which issue(s) is/are addressed by this PR?** *(optional)*

Fixes #

**Are there any special notes for your reviewer:**

```release-note
`VirtualMachineNetworkInterfaceSpec.DHCP4` and `DHCP6` are now `*bool` in v1alpha6, enabling explicit tri-state control: nil (provider decides), false (disabled), true (enabled). Older API versions continue to work unchanged; their false value maps to nil in v1alpha6.

### CODE REVIEW

File: api/v1alpha6/virtualmachine_network_types.go [hunk 1]
```diff
@@ -91,27 +91,37 @@ type VirtualMachineNetworkInterfaceSpec struct {
...
 	// +optional
 
-	// DHCP4 indicates whether or not this interface uses DHCP for IP4
-	// networking.
+	// DHCP4 controls IPv4 DHCP for this interface.
+	//
+	// When set to true, the interface requests an IPv4 address via DHCP.
+	// When set to false, DHCP for IPv4 is explicitly disabled; the interface
+	// relies on static allocation.
+	// When unset (nil), the network provider determines the allocation method
```
bryanv: I'd prefer we don't use `nil` here - that's more of a programing thing - and say `null` or empty. This documentation is better focused on yaml/json. Was this cribbed from prior pointer fields?

bryanv (approved):
One comment to resolve before merging. Thanks.

api/v1alpha6/virtualmachine_network_types.go
hpannem: Sure. Will change.  There are other new fields as well that has this language. Will change it in the respective PRs.

PR #1643, opened 2026-06-09
🐛 Allow non-VKS node VMs with zone label to participate in group placement

palnabarun:

When VMAffinityDuringExecution feature is enabled, zone-labeled VMs can now
participate in group placement to receive DRS host recommendations while still
respecting their zone labels. VKS nodes continue to bypass group placement
to maintain their zone pinning behavior unchanged.

Key changes:
- Zone-labeled VMs get optimal host placement within their assigned zones
- VKS nodes preserve existing zone override behavior
- Zone mismatch validation prevents VMs from being placed in wrong zones
- Backward compatibility maintained when feature flag is disabled

Changes by file:
- controllers/virtualmachinegroup_controller.go: Updated getVMForPlacement
  to allow non-VKS zone-labeled VMs in group placement when feature enabled
- pkg/providers/vsphere/vmprovider_vm.go: Modified placement result path
  to honor recommendations from DRS placement logic even for zone-labeled
  VMs which are not VKS nodes, added zone mismatch validation in
  processPlacementResult to prevent VMs from being placed in wrong zones,
  made error handling condition-aware to preserve ZoneMismatch reasons
- pkg/providers/vsphere/virtualmachine/configspec.go: Refactored
  CalculateAffinityConstraints to disable zone rules for VKS nodes,
  and VMAffinityDuringExecution+zone-labeled VMs.
- pkg/providers/vsphere/virtualmachine/configspec_test.go: Updated test
  expectations for VKS node zone rule behavior, added test cases for
  VMAffinityDuringExecution scenarios with zone-labeled VMs
- controllers/virtualmachinegroup_controller_test.go: Added test coverage
  for VKS vs non-VKS behavior with feature flag enabled
- pkg/providers/vsphere/vmprovider_vm_group_test.go: Added test scenarios
  for zone match, mismatch, and VKS bypass cases
- pkg/util/kube/label.go: Added HasZoneLabel helper function
- pkg/util/kube/label_test.go: Added test cases for HasZoneLabel helper function

Fixes vmop-3796

- VMG + VM creation with Zone Label + VM-VM Host AF
- VMG + VM creation with Zone Label + VM-VM Host AAF
- VMG + VM creation with VKS Node labels + Zone Label
- VMGroup E2E Suite was run with the changes on a testbed with FF enabled

```release-note
Allow non-VKS node VMs with zone label to participate in group placement
```

### DIFF
--- a/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
+++ b/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
@@ -36,6 +36,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/pkg/patch"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers"
 	"github.com/vmware-tanzu/vm-operator/pkg/record"
+	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 	vmopv1util "github.com/vmware-tanzu/vm-operator/pkg/util/vmopv1"
 )
 
@@ -689,18 +690,35 @@ func (r *Reconciler) getVMForPlacement(
 		return nil, nil
 	}
 
-	// If VM has a zone label, skip group placement to respect zone override.
 	if zoneName := vm.Labels[corev1.LabelTopologyZone]; zoneName != "" {
+		isVksNodeVM := kubeutil.HasCAPILabels(vm.Labels)
+		// A zone-labeled VM bypasses group placement to honor its zone override
+		// when VMAffinityDuringExecution is disabled, or when it is a VKS node
+		// (CAPI-managed) which must stay pinned to its assigned zone.
+		if !pkgcfg.FromContext(ctx).Features.VMAffinityDuringExecution ||
+			isVksNodeVM {
+			pkglog.FromContextOrDefault(ctx).V(4).Info(
+				"VM has explicit zone label, skipping group placement",
+				"vmName", vmName,
+				"zoneName", zoneName,
+			)
+			return nil, nil
+		}
+
+		// Feature enabled + non-VKS: fall through so the zone-labeled VM still
+		// participates in group placement for DRS host recommendations.
 		pkglog.FromContextOrDefault(ctx).V(4).Info(
-			"VM has explicit zone label, skipping group placement",
+			"VM has explicit zone label but participates in group placement "+
+				"(VMAffinityDuringExecution enabled, non-VKS)",
 			"vmName", vmName,
 			"zoneName", zoneName,
 		)
-		return nil, nil
 	}
 
 	// VMG doesn't have a PlacementReady condition true for this VM (UID).
-	// VM is not already placed, nor has a zone label override.
+	// VM is not already placed, and either has no zone label override or is a
+	// non-VKS zone-labeled VM participating in group placement for VM-VM
+	// Affinity/Anti-Affinity at host host level within the zone.
 	// Return this VM to get it placed by the group.
 	return vm, nil
 }
--- a/controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
+++ b/controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
@@ -23,9 +23,11 @@ import (
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
 	vspherepolv1 "github.com/vmware-tanzu/vm-operator/external/vsphere-policy/api/v1alpha1"
 	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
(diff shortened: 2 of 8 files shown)


controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -689,18 +690,33 @@ func (r *Reconciler) getVMForPlacement(
...
 			"vmName", vmName,
 			"zoneName", zoneName,
 		)
-		return nil, nil
 	}
 
 	// VMG doesn't have a PlacementReady condition true for this VM (UID).
-	// VM is not already placed, nor has a zone label override.
+	// VM is not already placed, and either has no zone label override or is a
+	// non-VKS zone-labeled VM participating in group placement.
```
aruneshpa: Here and elsewhere:  It is still not clear from "participating in group placement.".  I would be more explicit and say that participating in group placement for VM-VM affinity / anti-affinity at host level within the zone".

controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
```diff
@@ -703,6 +717,71 @@ var _ = Describe(
...
+					})
+
+					It("should skip group placement to respect zone override", func() {
+						By("verifying provider was not called for group placement")
+						Consistently(func(g Gomega) {
+							g.Expect(groupPlacementCallCount.Load()).To(BeZero(),
+								"VKS nodes with zone label should not be placed by group")
+						}, "3s", "100ms").Should(Succeed())
+
+						By("verifying PlacementReady is not set on member status")
```
aruneshpa, line 755: Hmm.  I thought we set this Condition for the member anyway with the zone that the user has specified?  Is that not the case?

pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -347,9 +350,19 @@ func CalculateAffinityConstraints(vmCtx pkgctx.VirtualMachineContext, isCreateVM
...
-		// Zonal rules are ignored during VM Creation as DRS does not differentiate between topologies for persisted policies.
-		// Persistent policies are treated at Host Topology.
+	// ConfigureZoneRules = false when:
+	//   1. VM Creation: DRS doesn't differentiate topologies for persisted policies.
+	//   2. VKS Node: Never get zonal policies (VMs are already zone-constrained).
+	//   3. VMAffinityDuringExecution + Zone Label: Already zone-constrained
+	//
+	//   isCreateVM? → isVksNodeVM? → (VMAffinityDuringExecution + ZoneLabel)?
+	//    ↓YES (any)       ↓YES              ↓YES     		↓NO
+	//    FALSE            FALSE             FALSE    		FALSE
```
aniket-deole: Limiting to the comment, should the last case be TRUE? : 
- Placement VM Workflow
- Is not capi
- Has VMAffinityDuringExecution
- Does not have zone label

palnabarun: actually, yes. there's a typo. if all of the cases of the ladder mentioned above fall through, then it's a true at the end for 9.1+ and false for releases before that.

palnabarun: the logic is correct. only the docstring has an issue.

controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -689,18 +690,33 @@ func (r *Reconciler) getVMForPlacement(
 		return nil, nil
 	}
 
-	// If VM has a zone label, skip group placement to respect zone override.
 	if zoneName := vm.Labels[corev1.LabelTopologyZone]; zoneName != "" {
+		// A zone-labeled VM bypasses group placement to honor its zone override
+		// when VMAffinityDuringExecution is disabled, or when it is a VKS node
+		// (CAPI-managed) which must stay pinned to its assigned zone.
+		if !pkgcfg.FromContext(ctx).Features.VMAffinityDuringExecution ||
+			kubeutil.HasCAPILabels(vm.Labels) {
```
aruneshpa: vvNit: define a isVKSNode :=`kubeutil.HasCAPILabels(vm.Labels)` and then use that?  That's more readable IMO.

aruneshpa (approved):

controllers/virtualmachinegroup/virtualmachinegroup_controller_test.go
palnabarun: In the group reconciliation workflow, we set the status for those members whose have user specified zones when the UniqueID is stamped on the status. This test only checks whether the VKS node VM has no placement ready if it just has the zone label. There are existing tests which set the UniqueID and test whether the condition is set.

PR #1644, opened 2026-06-09
✨ Implement compute config reconciliation for VirtualMachine

hpannem:
✨ Implement compute config reconciliation for VirtualMachine spec

This PR introduces a unified compute configuration reconciliation path for`VirtualMachine` objects. Previously, compute fields (CPU topology, resource allocation, CPU/memory flags, latency sensitivity) were either not managed or applied inconsistently across powered-on and powered-off reconfigure paths.

Key changes:

- **`OverwriteSpecComputeConfig`** — new core function using a `computeFieldDef` function-table pattern. Each field carries its own `minHWVer`, `hotPluggable`,`differs`, and `apply` callbacks. A 3-way comparison (spec vs. live vSphere state vs. class-derived ConfigSpec) determines whether to emit a change. `hwVer` is derived internally from `liveCI.Version` rather than passed as a parameter.

- **Spec semantics** — VM Operator fully owns every compute field it exposes. `nil` means "reset to platform default." The only exception is  `resources.size.cpu/memory` where `nil` defers to the VirtualMachineClass. `0` is a sentinel for `coresPerSocket` / `vnumaNodeCount` meaning "auto." `-1` is a sentinel for `limits.cpu` / `limits.memory` meaning "unlimited."

- **API validation** (`api/v1alpha6`, webhooks) —
  - `coresPerSocket` and `vnumaNodeCount` minimum changed to `0` (allow "auto" sentinel).
  - XValidation rule tightened: `vnumaNodeCount > 0` requires `coresPerSocket > 0` (an explicit non-zero value).
  - `-1` allowed for `limits.cpu` and `limits.memory` as the "unlimited" sentinel; `0` is still rejected for limits.

- **`SyncClassComputeToSpec` / `SyncClassSizeAndAllocationToSpec`** — new functions that copy the class's compute fields into `vm.Spec` when a class-based resize is triggered (`ResizeNeeded` returns true). This restores backward compatibility: during a resize the class is authoritative and overrides all spec fields (backfilled or user-specified) that it defines. The sync is performed in the reconciler — not the mutating webhook — because the VirtualMachineClass may not exist yet when the webhook fires.
  - Full class resize (`VMResize` feature): `SyncClassComputeToSpec` syncs all compute fields, and device changes are applied atomically in the same vSphere reconfigure call. The subsequent reconcile is a no-op.
  - CPU/memory-only resize (`VMResizeCPUMemory` feature): `SyncClassSizeAndAllocationToSpec` syncs only size and allocation fields, matching the narrow scope of that path.  
- **Removed `NUMAFixedAutoAffinityEnabled` & `cpu.ReservationLockedToMax`** which are not yet available in vSphere. `ExposeVNUMAOnCPUHotAdd` is API only field in vSphere. (not exposed in vSphere UI). Will expose whenever there is a need for that in VMOP.

Fixes #

- All new sync calls are gated by `TelcoVMServiceAPI` feature flag since `OverwriteSpecComputeConfig` only runs when that flag is enabled.

```release-note
VirtualMachine compute fields (CPU topology, resource allocation, CPU/memory flags, latency sensitivity) are now fully reconciled via `vm.Spec`. Setting a field to `nil` resets it to the platform default. `0` is an "auto" sentinel for `coresPerSocket`/`vnumaNodeCount`; `-1` is an "unlimited" sentinel for `limits.cpu`/`limits.memory`. Class-based resize continues to be authoritative: when a resize is triggered the class overrides all affected spec fields.

----
📚 Documentation preview 📚: https://vm-operator--1644.org.readthedocs.build/en/1644/

### CODE REVIEW

File: api/v1alpha6/virtualmachine_compute_types.go [hunk 1]
```diff
@@ -192,6 +173,8 @@ type VirtualMachineCPUAdvancedSpec struct {
 	// (VT-d / IOMMU) for this VM. Required for SR-IOV and PCI passthrough
 	// workloads. Requires EFI firmware (spec.bootOptions.firmware = "efi").
 	// Maps to ConfigSpec.Flags.VvtdEnabled.
+	// Minimum hardware version depends on host CPU vendor:
+	// vmx-14 or later on Intel hosts; vmx-18 or later on AMD hosts.
```
akutz: Does it make sense to call out host information like this? I would rather say "Intel CPUs" and "AMD CPUs".

api/v1alpha6/virtualmachine_compute_types.go
```diff
@@ -34,14 +38,16 @@ type VirtualMachineResourceQuantity struct {
 	// For spec.resources.size, maps to ConfigSpec.MemoryMB (guest-visible).
 	// For spec.resources.requests and spec.resources.limits, maps to
 	// MemoryAllocation.Reservation and MemoryAllocation.Limit respectively.
+	//
+	// For spec.resources.size, an increase can be applied while the VM is
+	// powered on when memoryAdvanced.hotAddEnabled is already enabled on the
```
akutz, line 41: I think we need more/better info here. Not all guest OSs support hot add. Therefore setting `memoryAdvanced.hotAddEnabled` will not work for every OS. How are we handling this?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -425,14 +425,19 @@ func (s *Session) poweredOnReconfigure(
...
+
 	UpdateConfigSpecExtraConfig(vmCtx, config, configSpec, vmCtx.VM, nil)
 	UpdateConfigSpecChangeBlockTracking(vmCtx, config, configSpec, vmCtx.VM.Spec)
 
 	if err := virtualmachine.UpdateConfigSpecCdromDeviceConnection(vmCtx, s.Client.RestClient(), s.K8sClient, config, configSpec); err != nil {
 		return fmt.Errorf("update CD-ROM device connection error: %w", err)
 	}
 
-	err := doReconfigure(
+	reconfigErr := doReconfigure(
```
akutz, line 445: Why change the var name?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -425,14 +425,19 @@ func (s *Session) poweredOnReconfigure(
 		return err
 	}
 
+	// Apply hot-pluggable compute config changes while the VM is powered on.
```
akutz: I would like this comment to be more generic since the function you are calling is.

akutz (changes requested):
This is a fairly large change with some interesting bits. A lot of the most useful insight is in util/vmopv1/resize_class_sync.go. I would like to see this PR updated with an SDD spec that describes this change clearly, along with diagrams for the expected flows. Then let's see how that matches the current code. Please do not use AI to write the SDD prose. You can for the diagrams, but I want you to manually craft the SDD spec's workflows to make sure we can then use AI to match the expected behavior against the implementation.

api/v1alpha6/virtualmachine_compute_types.go
hpannem: sure

api/v1alpha6/virtualmachine_compute_types.go
hpannem: right now, am not checking for any OS... Any idea? I can modify the comment to explicitly specify this.

pkg/providers/vsphere/session/session_vm_update.go
hpannem: just to make it explicit and consistent as other functions. should i revert?

hpannem:
> This is a fairly large change with some interesting bits. A lot of the most useful insight is in util/vmopv1/resize_class_sync.go. I would like to see this PR updated with an SDD spec that describes this change clearly, along with diagrams for the expected flows. Then let's see how that matches the current code. Please do not use AI to write the SDD prose. You can for the diagrams, but I want you to manually craft the SDD spec's workflows to make sure we can then use AI to match the expected behavior against the implementation.

I will work on this.

hpannem:
> > This is a fairly large change with some interesting bits. A lot of the most useful insight is in util/vmopv1/resize_class_sync.go. I would like to see this PR updated with an SDD spec that describes this change clearly, along with diagrams for the expected flows. Then let's see how that matches the current code. Please do not use AI to write the SDD prose. You can for the diagrams, but I want you to manually craft the SDD spec's workflows to make sure we can then use AI to match the expected behavior against the implementation.
> 
> I will work on this.

Added sdd specs. @akutz Please take a look.

docs/concepts/workloads/vm.md
```diff
@@ -435,21 +435,19 @@ If the condition is ever false, please refer first to the condition's `reason` f
 
 ### Compute Reconfiguration
 
-!!! note "Requires `TelcoVMServiceAPI` capability"
+Three optional fields are available for workloads that require fine-grained, direct control over compute resources: `spec.resources`, `spec.cpuAdvanced`, and `spec.memoryAdvanced`. These fields allow compute allocation and advanced CPU/memory settings to be specified directly on the VM, independent of the `VirtualMachineClass`.
```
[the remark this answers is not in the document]
hpannem: capabilities are not doc'd.

pkg/util/vmopv1/compute_overwrite.go
```diff
@@ -0,0 +1,651 @@
...
+			// Reservation-only ConfigSpec that vSphere validates and rejects when
+			// the value does not equal the locked memory size.
+			if ptr.DerefWithDefault(ci.MemoryReservationLockedToMax, false) {
+				if desiredLim != liveLim {
+					return true
+				}
+				if cs.MemoryAllocation == nil {
+					return false
+				}
+				return desiredLim != ptr.DerefWithDefault(cs.MemoryAllocation.Limit, int64(-1))
```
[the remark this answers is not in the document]
hpannem: Fixed.

pkg/util/vmopv1/compute_overwrite.go
```diff
@@ -0,0 +1,651 @@
...
+			level, _ := desiredLatencySensitivity(vm)
+			if level != vimtypes.LatencySensitivitySensitivityLevelHigh {
+				return false, ""
+			}
+			if !FullMemReservationSpecMet(&vm) {
+				return true, "full memory reservation required: " +
+					"set spec.memoryAdvanced.reservationLockedToMax=true, " +
+					"or set spec.resources.requests.memory equal to spec.resources.size.memory"
+			}
+			// TODO: verify requests.cpu equals numCPUs × hostCpuMHz (full reservation).
```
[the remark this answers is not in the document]
hpannem: This is good one. Fixed.

bryanv (approved):

api/v1alpha6/virtualmachine_types.go
```diff
@@ -208,6 +208,11 @@ const (
 	// VirtualMachinePowerCyclePendingReason). This is expected to resolve to
 	// True once a later reconcile observes the applied change.
 	VirtualMachineExtraConfigMismatchReason = "ExtraConfigMismatch"
+
+	// VirtualMachineComputeConfigMismatchReason is used on VirtualMachineConditionComputeConfigSynced
+	// when the VM's spec compute fields differ from the live vSphere configuration.
+	// The controller will apply the change on the current or next reconcile loop.
+	VirtualMachineComputeConfigMismatchReason = "ComputeConfigMismatch"
```
akutz, line 215: Condition constants should be positive. All conditions, when present, should have a `true` value that represents a good thing.

Please invert the meaning of this condition, ex. `ComputeConfigValidation`.

akutz (changes requested):
Other than the condition truthiness, this lgtm. Thank you @hpannem !!

api/v1alpha6/virtualmachine_types.go
akutz: Mea culpa. This is not the condition name, but the reason! Doh!

akutz (approved):

PR #1646, opened 2026-06-10
[x-port] [9.1] Handle PVCs in crypto reconciler

aruneshpa:

This is an x-port of #1460 from main branch.

When a PVC is specified in the VM Spec, CSI will not handle the crypto for that disk. Instead, VMOP must recrypt the disk if its encryption class changes. This is because the VM cannot mix key provider types. So if changing the key provider type of a disk, in that same reconfigure the VM's key provider must be also changed to one of that type.

This is a late change and choices were made to reduce the scope and mostly confine the code changes into one place. There are many improvements that we'll need to make later.

```release-note
Handle PVCs in crypto reconciler
```

### DIFF
--- a/pkg/vmconfig/crypto/crypto_reconciler_pre.go
+++ b/pkg/vmconfig/crypto/crypto_reconciler_pre.go
@@ -7,21 +7,26 @@ package crypto
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/vmware/govmomi/crypto"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	byokv1 "github.com/vmware-tanzu/vm-operator/external/byok/api/v1alpha1"
 	"github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	pkgconst "github.com/vmware-tanzu/vm-operator/pkg/constants"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
 	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
 	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
 	"github.com/vmware-tanzu/vm-operator/pkg/util/paused"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/ptr"
+	pkgvol "github.com/vmware-tanzu/vm-operator/pkg/util/volumes"
 	"github.com/vmware-tanzu/vm-operator/pkg/vmconfig/crypto/internal"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
@@ -78,6 +83,9 @@ var (
 	// ErrInvalidKeyID is returned if the key id specified by an
 	// EncryptionClass is invalid.
 	ErrInvalidKeyID = errors.New("invalid key id")
+
+	// ErrPVCNotFound is returned if the PVC is not found.
+	ErrPVCNotFound = errors.New("PVC not found")
 )
 
 func (r reconciler) Reconcile(
@@ -121,11 +129,7 @@ func (r reconciler) Reconcile(
 		args.useDefaultKeyProvider = true
 	} else {
 		args.encryptionClassName = c.EncryptionClassName
-		if c.UseDefaultKeyProvider == nil {
-			args.useDefaultKeyProvider = true
-		} else {
-			args.useDefaultKeyProvider = *c.UseDefaultKeyProvider
-		}
+		args.useDefaultKeyProvider = ptr.DerefWithDefault(c.UseDefaultKeyProvider, true)
 	}
 
 	// Check to if the VM is currently encrypted and record the current provider
@@ -174,7 +178,7 @@ func (r reconciler) reconcileCreate(
 		//
 
 		// Get the provider and key from the EncryptionClass.
-		args.newKey, err = getCryptoKeyFromEncryptionClass(ctx, args)
+		args.newKey, err = getCryptoKeyFromEncryptionClass(ctx, args, args.encryptionClassName)
 		if err != nil {
 			return setConditionAndReturnErr(args, err, ReasonInternalError)
 		}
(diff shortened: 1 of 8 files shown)


faisalabujabal (approved):

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -396,6 +405,171 @@ func (r reconciler) reconcileUpdateDefaultKeyProvider(
...
+					needsRecrypt = true
+				}
+			}
+
+			if needsRecrypt {
+				if updateFCDBackingForRecrypt(args, disk, desiredKey) {
+					logger.Info(
+						"Recrypt FCD",
+						"disk", diskInfo.FileName,
+						"currentProviderID", currentKey.ProviderId.Id,
```
hickeng, line 559: potential nil panic given we're checking `if currentKey.ProviderId != nil ` on line 543

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -396,6 +405,171 @@ func (r reconciler) reconcileUpdateDefaultKeyProvider(
...
+					args,
+					ErrNoDefaultKeyProvider,
+					ReasonNoDefaultKeyProvider)
+			}
+		}
+
+		currentKey := diskInfo.CryptoKey
+		disk := diskInfo.Device
+
+		logger.Info("PVC encrypted",
```
hickeng, line 520: This logs at info level unconditionally on every entry to this function.

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -396,6 +405,171 @@ func (r reconciler) reconcileUpdateDefaultKeyProvider(
...
+			"pvcName", pvc.Name, "currentKey", currentKey,
+			"desiredKey", fmt.Sprintf("%+v", desiredKey)) // NB: Keep diff small - fields aren't public.
```
hickeng, line 522: I'm not sure what this includes - please confirm we're not logging sensitive data, regardless of log level.

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -396,6 +405,171 @@ func (r reconciler) reconcileUpdateDefaultKeyProvider(
...
+		pvc := &corev1.PersistentVolumeClaim{}
+		pvcKey := ctrlclient.ObjectKey{
+			Namespace: args.vm.Namespace,
+			Name:      volume.PersistentVolumeClaim.ClaimName,
+		}
+		if err := args.k8sClient.Get(ctx, pvcKey, pvc); err != nil {
+			// NOTE: I don't think we'll even get here if PVC is missing.
+			return false, setConditionAndReturnErr(
+				args,
+				ErrPVCNotFound,
```
hickeng, line 459: setConditionAndReturnErr already handles IsNotFound and other error types - i think we should be passing the err directly?
```
setConditionAndReturnErr(args, err, ReasonInternalError)
```

pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
```diff
@@ -671,6 +671,19 @@ func ensurePVCForUnmanagedDisk(
...
 
+		// BMV: Make the registration of the disk match that of the VM's. This is a
+		// workaround for when the VM encryption class's provider  is not the same
+		// type as the default provider.
+		// Ultimately though, similar to the processing that is done in
+		// vmCreateGenConfigSpecImagePVCDataSourceRefs(), the crypto reconciler
+		// on create, should check if there is a placeholder PVC for an image's disk
+		// and use that - not the VM's - encryption class for the disk's crypto. This
+		// isn't ideal since we'll always use the VM's class here, but for the
+		// short term after registration the PVC's encryption class can be updated.
```
hickeng, line 682: I'm not sure - does the disk actually exist at this point? if it does are we potentially re-encrypting and potentially repeatedly hitting mixedKeys errors?

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -396,6 +405,171 @@ func (r reconciler) reconcileUpdateDefaultKeyProvider(
...
+					"keyID", desiredKey.id,
+					"isDefaultProvider", desiredKey.isDefaultProvider)
+				changed = true
+			}
+		} else {
+			// FCD is encrypted -> check if recrypt needed
+			needsRecrypt := false
+
+			// Provider comparison
+			if currentKey.ProviderId != nil && currentKey.ProviderId.Id != desiredKey.provider {
```
hickeng, line 543: `currentKey.ProviderId == nil || currentKey.ProviderId.Id != desiredKey.provider`  ? seems if currentKey is nil and we cannot get here with a nil desiredKey then we want to encrypt.

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -396,6 +405,171 @@ func (r reconciler) reconcileUpdateDefaultKeyProvider(
...
+		if desiredEncClassName != "" {
+			// Use the specified encryption class.
+			desiredKey, err = getCryptoKeyFromEncryptionClass(ctx, args, desiredEncClassName)
+			if err != nil {
+				logger.Error(err, "Failed to get crypto key from encryption class",
+					"encryptionClassName", desiredEncClassName)
+				return false, setConditionAndReturnErr(args, err, ReasonInternalError)
+			}
+		} else {
+			desiredKey = getCryptoKeyFromDefaultProvider(ctx, args)
```
hickeng, line 507: I'm not sure whether setting `UseDefaultKeyProvider = false` on the VM should mean you _cannot_ use a PvC with an encrypted storage class without the PVC having an explicit annotation.

I'm not sure how this interacts with https://github.com/vmware-tanzu/vm-operator/pull/1646/changes#r3483381798

Should we be returning an error here, or should we just be skipping it? eg.
```go
    if !args.useDefaultKeyProvider {
        // VM has explicitly opted out of default-provider management.
        // An unannotated PVC on an encrypted StorageClass has no designated key source.
        // Skip rather than fail — CSI owns this disk when it is detached from a VM.
        logger.Info(
            "Skipping unannotated FCD: VM opted out of default key provider",
            "storageClass", storageClassName)
        continue
    }
    desiredKey = getCryptoKeyFromDefaultProvider(ctx, args)
    if desiredKey.provider == "" {
        return false, setConditionAndReturnErr(
            args,
            ErrNoDefaultKeyProvider,
            ReasonNoDefaultKeyProvider)
    }
```

Given I'm reviewing this mostly without context the question may be non-sensical.

pkg/vmconfig/crypto/crypto_reconciler_pre.go
```diff
@@ -396,6 +405,171 @@ func (r reconciler) reconcileUpdateDefaultKeyProvider(
...
+
+		// TODO: We're using the target ID to match disks/volumes between the VM Spec and
+		// the VM's current HW config. I think this gives us a possible race: if the PVC
+		// at a target ID is changed, the detach of old, attach of new happens async from
+		// us via CNS. I think this may just be a general issue of how we're using this
+		// even elsewhere, but the effects may be limited because we're generally read-only
+		// only so things will clear up on a later reconcile. Here that is different because
+		// we're doing a reconfigure.
+		// This might not be a huge issue in practice, because once the old PVC stops being
+		// referenced in a VM Spec, CSI will take over the encryption of the PVC.
```
hickeng, line 471: I'm not sure CSI picks up the disk immediately on detach - iiuc the disk is simply no longer referenced in the vm spec so unless they're looking back at prior generation to do a diff i'm not sure how CSI knows a newly detached PVC is in need of reconcilation?
Do we need to do something explicit to provide CSI with a prompt to reconcile the newly detached PVC?

hickeng (commented):
Added related comments on https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/3867/

PR #1648, opened 2026-06-10
🐛 fix: More E2E test flakiness

faisalabujabal:
- Wrapping E2E assertions in eventually to reduce flakiness
- Updating the Dockerfile to support GOPROXY env variables and to reduce its size

Fixes # NA

NA

```release-note
fix: More E2E test flakiness
```

### DIFF
--- a/test/e2e/vmservice/vmservice/viadmin/registervm.go
+++ b/test/e2e/vmservice/vmservice/viadmin/registervm.go
@@ -75,8 +75,8 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
 		svClusterClientSet            *kubernetes.Clientset
 		vmServiceBackupRestoreEnabled bool
 		incrementalRestoreEnabled     bool
-		linuxImageDisplayName string
-		linuxVMIName          string
+		linuxImageDisplayName         string
+		linuxVMIName                  string
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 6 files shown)


test/e2e/vmservice/vmservice/viadmin/registervm.go
```diff
@@ -296,24 +296,25 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
...
+				if vm.Annotations == nil {
+					vm.Annotations = make(map[string]string)
 				}
-			}
+				vm.Annotations[vmopv1a3.PauseAnnotation] = trueString
+				g.Expect(svClusterClient.Update(ctx, vm)).To(Succeed())
+
+				for _, volume := range vm.Spec.Volumes {
+					if volume.PersistentVolumeClaim != nil {
+						pvcNames = append(pvcNames, volume.PersistentVolumeClaim.ClaimName)
```
bryanv, line 314: There isn't anything after it but if anybody does add something and it could fail, pvcNames would get appended again. So I'd prefer that pvcNames is nil out above this, but IMO it would be easier to just change these Update() into Patch() to address the race. Or there is a retry.RetryOnConflict() helper in client-go

bryanv (approved):

test/e2e/vmservice/vmservice/viadmin/registervm.go
faisalabujabal: good catch

PR #1651, opened 2026-06-11
🐛  Fix test resiliency 

aakashchan:
Bare Get → annotate/modify → Update sequences in e2e tests were vulnerable to HTTP 409 optimistic concurrency conflicts: the controller reconciler frequently updates the VM object after state transitions (e.g. after PoweredOff), so the resourceVersion seen at Get time is often stale by the time Update is issued.

Wrap every such bare update in an Eventually block that re-fetches the object on each attempt, matching the pattern already used elsewhere in the test suite.

Additionally fix the incremental-restore test:

- Set Removable=ptr.To(true) on all volumes immediately after pvcNameA is attached and *before* WaitForBackupToComplete, so the backup YAML in ExtraConfig carries removable:true. The CSI CnsRegisterVolume controller reads spec.volumes[*].removable when processing newly registered FCDs; without an explicit value the disk can be mis-classified during a restore pass.

- Use v1alpha5 for the pause-annotation step (which exposes the Removable field) and set Removable=true on every volume at the same time. The validating webhook rejects removal of any volume with removable=false; the reconciler's backfill path can set this flag on discovered disks, so it must be cleared before InvokeRegisterVM attempts to restore the backup state.

Files changed:
- test/e2e/vmservice/vmservice/util.go
- test/e2e/vmservice/vmservice/viadmin/registervm.go

Changes in vm_snapshot is an indentation noise. Please ignore, if needed, I can revert them to original state.

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/vmservice/lib/vmoperator/vmoperator.go
+++ b/test/e2e/vmservice/lib/vmoperator/vmoperator.go
@@ -19,6 +19,8 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	e2eframework "k8s.io/kubernetes/test/e2e/framework"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


test/e2e/vmservice/vmservice/viadmin/registervm.go
```diff
@@ -299,12 +301,22 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
 			vm, err := utils.GetVirtualMachine(ctx, svClusterClient, input.WCPNamespaceName, vmName)
 			Expect(err).ToNot(HaveOccurred())
 
-			if vm.Annotations == nil {
-				vm.Annotations = make(map[string]string)
-			}
-
-			vm.Annotations[vmopv1a3.PauseAnnotation] = trueString
-			Expect(svClusterClient.Update(ctx, vm)).To(Succeed())
+			Eventually(func() bool {
```
bryanv: This is very verbose. Let's switch to using Patch() here first before doing this route.

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -1439,3 +1442,136 @@ func EventuallyBootDiskStoragePolicyMatchesVMStorageClass(
...
+// VirtualMachineDiskPromotionSynced, ensuring the hardware was re-verified
+// after diskpromo, not just during the initial linked-clone check.
+func WaitForVMReadyForSnapshot(
+	ctx context.Context,
+	config *config.E2EConfig,
+	client ctrlclient.Client,
+	ns, vmName string,
+) {
+	const (
+		diskPromoSyncedCond = "VirtualMachineDiskPromotionSynced"
```
bryanv: You should be able to use the const from the api vmopv1 module

bryanv (approved):

aniket-deole (approved):

PR #1653, opened 2026-06-11
🌱Initial support for multiple network types

bryanv:

In the future, a namespace may support multiple providers, and a NetworkInterface may front an VCP SubnetPort. Do the initial plumbing of supporting that here.

Pull out the getting of the network interface backing into a common function.

Instead of a single field to configure the type of provider when creating a vcsim test context, allow for multiple ones to be provided. Maps all the existing usage on to this new way, and they will be migrated later.

Fixes # vmop-3554

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/network/network.go
+++ b/pkg/providers/vsphere/network/network.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


faisalabujabal (approved):
LGTM! only comment i have is about the title of the PR. it is not just updating the vcsim to support this but also includes the initial work as you mentioned in the description

PR #1663, opened 2026-06-15
🌱 Update VM validation webhook to support multiple network providers

bryanv:

When a namespace has been migrated, the prior (legacy) provider must still be supported as to not break existing VMs or VKS clusters.

Fixes # vmop-3880

```release-note
NONE
```


webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1044,18 +1045,16 @@ func (v validator) validateNetworkVLANs(
...
-// getNetworkProviderType returns the network provider type for the given
-// namespace. When PerNamespaceNetworkProvider is disabled it reads the global
-// config; when enabled it fetches NetworkSettings/default from that namespace.
-func (v validator) getNetworkProviderType(
+// getSupportedNetworkProviderTypes returns all network provider types for the
+// given namespace. When PerNamespaceNetworkProvider is disabled it returns the
+// global config value as a single-element slice; when enabled it fetches
+// NetworkSettings/default from that namespace, returning the primary provider
+// and, if set, the legacy provider as well.
+func (v validator) getSupportedNetworkProviderTypes(
```
faisalabujabal: this method is not really doing anything

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1091,35 +1090,76 @@ var networkProviderValidations = map[pkgcfg.NetworkProviderType]networkProviderV
...
 
-	if networkKind != "" && !slices.Contains(supported.kinds, networkKind) {
-		allErrs = append(allErrs, field.NotSupported(
-			interfacePath.Child("network", "kind"),
-			networkKind,
-			supported.kinds))
+	if networkKind != "" {
+		var supportedKinds []string
+		kindMatchesAny := false
+		for _, val := range validationsForKind {
```
faisalabujabal: nit: we can use slices indexoffunc here especially. it would also exit early. i know that is not a concern because the slices are small, but a good convention to start following.

faisalabujabal (approved):

PR #1671, opened 2026-06-17
vmop-3762: Add VirtualMachineConfigOptions validation webhook

faisalabujabal:
**What does this PR do, and why is it needed?**

Implements the admission webhook for `VirtualMachineConfigOptions` (Story S6.a / Tasks T080, T081, T086 from the [class-policy-resize spec](.sdd/specs/001-class-policy-resize/)), per VMSVC-3762.

The webhook enforces three rules on the cluster-scoped `VirtualMachineConfigOptions` CRD:
- **Create**: `spec.hardwareVersion` must match `^vmx-\d+$` (e.g. `vmx-21`). Empty value is also rejected.
- **Create**: `metadata.name` must equal `spec.hardwareVersion`. This protects the naming invariant the ConfigTarget controller (Story S5) relies on when it does `CreateOrPatch` per hardware-version key and garbage-collects via owner reference, both keyed by name.
- **Update**: `spec.hardwareVersion` is immutable after creation.

The webhook is gated behind the `VirtualMachineConfigPolicy` feature flag (wired in `webhooks/webhooks.go` via `pkgcfg.FromContext(ctx).Features.VirtualMachineConfigPolicy`); registration is skipped entirely when the capability is off, matching the pattern used by other feature-gated webhooks in this repo (e.g. VirtualMachineSnapshot).

**Which issue(s) is/are addressed by this PR?**

vmop-3762 (Story S6.a under Epic vmop-3331)

**Are there any special notes for your reviewer**:

This PR is based directly on `main` and does not depend on #1695 (vmop-3740: Zone controller fan-out) — the `VirtualMachineConfigOptions` CRD is already in `main`, and the webhook/validator code is fully self-contained.

The E2E test (`test/e2e/vmservice/vmservice/configpolicy/vmconfigoptions.go`) touches `test/e2e/vmservice/consts/consts.go` and `test/e2e/vmservice/vmservice_test.go`, which #1695 also touches (both add a capability constant / test registration in the same feature area). Expect a small, easily-resolved merge conflict in those two files whichever of #1671/#1695 merges second — no need to sequence the PRs.

**Test coverage**:
- 13 unit tests covering: valid create (vmx-N formats with matching name), empty/invalid format rejection (6 cases), name/hardwareVersion mismatch rejection, immutable update, unchanged update, delete
- Integration tests via envtest (require envtest binary, consistent with all other webhooks in this repo)
[...]

PR #1672, opened 2026-06-17
✨ vmop-3743: Add VirtualMachineConfigOptions controller

faisalabujabal:

Implements Story S6 (vmop-3743) from the 001-class-policy-resize spec: the `VirtualMachineConfigOptions` controller. This is a self-contained vertical slice (product code + unit tests + E2E tests) for the ConfigTarget → VirtualMachineConfigOptions pipeline stage. The admission webhook for this CRD is delivered separately in PR #1671 (vmop-3762) and is not part of this PR.

**Controller** (`controllers/virtualmachineconfigoptions/`):
- Watches cluster-scoped `VirtualMachineConfigOptions` objects
- Resolves the owning `ConfigTarget` via owner references to obtain the cluster MoID
- Calls `QueryConfigOptionEx` on the vSphere provider to retrieve hardware version capabilities from the cluster's `EnvironmentBrowser`
- Maps the result into `status` (description, guest OS identifiers, default index, hardware options, etc.)
- Fans out one `VirtualMachineGuestOptions` child object per guest OS descriptor, upserting the per-hardware-version status entry
- Tracks `status.observedGeneration` and sets a `Ready` condition; requeues with `RequeueAfter: 10s` when no ConfigTarget owner is found yet

**vSphere provider plumbing** (`pkg/providers/vsphere/environment_browser.go`):
- Wraps govmomi `EnvironmentBrowser.QueryConfigOption` using a `ClusterComputeResource` property fetch to resolve the browser reference

**RBAC**: `+kubebuilder:rbac` markers in the controller; `config/rbac/role.yaml` regenerated via `make generate-manifests`.

**Test coverage**:
- Controller unit tests (envtest-backed, ~75% coverage) plus a vcsim-backed test exercising the real `QueryConfigOptionEx` govmomi call path against a simulated vSphere backend.
- Rebased onto main to pick up the ConfigTarget controller (S5, PR #1704) and Zone fan-out (S3, PR #1695), which unblocked real E2E coverage. Extended the existing `test/e2e/vmservice/vmservice/configpolicy/configpolicy.go` spec (rather than the disconnected placeholder suite that previously lived here) with assertions that `VirtualMachineConfigOptions.status.guestOSIdentifiers` is populated and `Ready=True`, and that a `VirtualMachineGuestOptions` is fanned out per reported guest OS.

**Known limitation — multi-cluster `VirtualMachineConfigOptions` co-ownership (temporary, tracked in vmop-3797):**

`VirtualMachineConfigOptions` is keyed purely by hardware-version string (`metadata.name = vmx-N`), so it can legitimately be co-owned by multiple `ConfigTarget`s (clusters) that report supporting the same version — this is intentional, already-merged design from the `ConfigTarget` controller (S5). When co-owned, `findClusterMoID` currently queries `QueryConfigOptionEx` against **the lexicographically lowest cluster MoID** among the co-owners. This is an arbitrary-but-stable tie-break, not a considered "most correct" choice — it exists only to make the selection deterministic and reproducible across reconciles (the previous behavior picked whichever owner reference happened to be first, which is not guaranteed stable by Kubernetes).

Traced vSphere's own `EnvironmentBrowser.QueryConfigOptionEx` (vpxd source) to see how it resolves the analogous problem across multiple hosts in one cluster: it does **not** aggregate — it deterministically defers to the host running the **lowest product/ESXi version** among hosts supporting the requested key, on the theory that hardware-version capability is monotonic with product version. We don't yet have a comparable per-cluster signal (ESXi/product version) to replicate that "most conservative" selection at the cluster level — `ConfigTarget.status` doesn't expose it today, and this design deliberately avoids per-host granularity (the retired `HostSystem` CRD). Note `ConfigTargetStatus.MaxHardwareVersion` (vmop-3797) is **not** a valid stand-in for this: it's each cluster's own ceiling, not its product/patch vintage, and since co-owners by definition already support the shared key being queried, picking by lowest `MaxHardwareVersion` can select the *newer*-patched cluster rather than the more conservative one -- the opposite of what we want. A real fix needs an actual product-version-like signal, which doesn't exist yet.

Confirmed via ESXi release notes and the vSphere API's guest-OS-identifier changelog that this isn't a hypothetical edge case: at least two documented instances in the 8.x line alone (API 8.0.0.1 added 10 new guest OS identifiers to unchanged hardware version vmx-20; API 8.0.3.0 added another to unchanged vmx-21) show new guest OS support landing on an *already-shipped* hardware version via a patch, independent of any hardware-version bump. Multi-cluster Supervisor environments commonly patch clusters on staggered schedules, so two co-owning clusters disagreeing on `QueryConfigOptionEx`'s answer for the same key is a real, recurring condition, not a corner case.

Severity is currently limited to discovery/informational surfaces: `spec.md`'s admission-enforcement acceptance criteria (US5/S9) only reference `ConfigTarget.status.maxHardwareVersion`, never `VirtualMachineConfigOptions`/`VirtualMachineGuestOptions`, so today's arbitrary-cluster selection can't produce an incorrect admission decision — only a potentially inconsistent read for `kubectl get`/partner tooling. That should be re-escalated if any future story starts reading these objects for enforcement rather than discovery.

Filed a follow-up comment on vmop-3797 flagging that its `MaxHardwareVersion` addition is not sufficient to fix this by itself, and that revisiting this selection logic depends on a real per-cluster product-version signal that doesn't exist yet.

Fixes #

- This PR depends on PR #1671 (vmop-3762) for the `VirtualMachineConfigOptions` admission webhook; that PR is not required to merge first, but review of RBAC/scheme overlap between the two may be easier once both are visible.
- The `test/builder/fake.go` change registers `vimv1` types (`ConfigTarget`, `VirtualMachineGuestOptions`) into the unit-test scheme, in addition to `VirtualMachineConfigOptions` already registered by #1671, which is needed for the controller's fake-client unit tests.
- `make generate-manifests` fails locally because the `hack/tools` module requires Go 1.26.4 but the runner has 1.26.2; the existing `controller-gen` (`v0.19.0`) and `setup-envtest`-provisioned `etcd`/`kube-apiserver` binaries were used instead to fully validate the build, lint, and envtest suite locally.
- New E2E assertions carry the `experimental` label per `e2e-sync-with-changes.md` until validated on a real WCP cluster.
- See the "Known limitation" section above for the temporary multi-cluster co-ownership tie-break; tracked for follow-up in vmop-3797.

```release-note
Add VirtualMachineConfigOptions controller
```

**E2E Test Results**
```
Ran 7 of 138 Specs in 227.930 seconds
SUCCESS! -- 7 Passed | 0 Failed | 5 Pending | 126 Skipped
PASS
```

### DIFF
--- a/.sdd/memory/operator-best-practices.md
+++ b/.sdd/memory/operator-best-practices.md
@@ -181,6 +181,47 @@ func relatedToMyResourceMapper(c client.Client) func(context.Context, client.Obj
 }
 ```
 
+## Fan-out to Child Objects (patch vs. CreateOrUpdate)
+
+`controllerutil.CreateOrPatch` uses a JSON merge patch: map fields merge
+key-by-key, but **list fields are replaced wholesale** and there is **no
+`resourceVersion` precondition**. Fine for disjoint fields or a shared map
+keyed per writer; unsafe for a shared list.
+
+When multiple parents fan-in write to the same *list*-typed field (owner
+references, or a status list each parent upserts an entry into), a plain merge
+patch can silently drop a concurrent writer's entry with no error to trigger a
+retry. Patch with an optimistic lock, and skip the write when nothing changed:
+
+```go
+base := obj.DeepCopy()
+// mutate obj toward desired ...
+if !apiequality.Semantic.DeepEqual(base.Spec, obj.Spec) ||
+    !apiequality.Semantic.DeepEqual(base.OwnerReferences, obj.OwnerReferences) {
+    // MergeFromWithOptimisticLock sends base's resourceVersion, so a racing
+    // writer's patch fails with a conflict instead of overwriting. Reconcile
+    // is idempotent, so just return the error and let the requeue retry.
+    if err := c.Patch(ctx, obj, client.MergeFromWithOptions(base, client.MergeFromWithOptimisticLock{})); err != nil {
+        return err
+    }
+}
+```
+
+- **Compare with `apiequality.Semantic.DeepEqual`** (`k8s.io/apimachinery/pkg/api/equality`),
+  not `reflect.DeepEqual` — plain reflection false-diffs on types like
+  `resource.Quantity` that carry an unexported lazy-formatted-string cache, so a
+  freshly-built value and its round-tripped-through-the-API-server equivalent
+  compare unequal even when identical. That silently defeats the skip-if-unchanged
+  guard below.
+- **Skip unchanged writes** so an idle reconcile does not bump
+  `resourceVersion` and re-trigger watchers. (`CreateOrUpdate`'s `Update` gets
+  the same conflict guarantee but writes the whole object; prefer the patch.)
+- **Status is a subresource:** persist it separately via `c.Status().Patch(...)`
+  (same optimistic lock, same skip guard) after the object exists.
+- **In tests**, register the type with `WithStatusSubresource` (all VM Operator
+  types already are — `test/builder/fake.go` `KnownObjectTypes`) so the fake
+  client enforces the spec/status split and catches a missing status write.
+
 ## VM Update Reconcile Order
 
 The `updateVirtualMachine` method in the vSphere provider follows a strict, intentional ordering. Each step depends on the state established by prior steps. Do NOT reorder these without understanding the dependencies:
@@ -206,6 +247,50 @@ Key ordering rationale:
 - **Snapshot create last**: Captures the final converged state after all changes.
 
 Each step uses `getReconcileErr` to accumulate errors while continuing through subsequent steps where possible. A `NoRequeueError` from any step short-circuits the remaining steps via `errOrReconcileErr`.
+
+## VC Operation IDs (WithVCOpID)
+
+Any code path that calls into vCenter — govmomi, the property collector, or a
+package-level helper in `pkg/providers/vsphere/*.go` that takes a
+`*vim25.Client` directly — MUST wrap the context with `pkgctx.WithVCOpID`
+before making that call, so the operation can be correlated with VC-side
+logs.
+
+**Set it inside the provider method, not in the calling controller.** When a
+controller reaches vCenter through the `VMProvider` interface (the normal
+case), the wrapping belongs at the top of the concrete
+`pkg/providers/vsphere/*.go` method that actually talks to VC — the
+controller just calls the interface method with a plain `ctx`:
+
+```go
+// pkg/providers/vsphere/vmprovider.go
+func (vs *vSphereVMProvider) QueryConfigOptionEx(
+    ctx context.Context,
+    clusterMoID string,
+    hardwareVersion string) (*vimtypes.VirtualMachineConfigOption, error) {
+
+    ctx = pkgctx.WithVCOpID(ctx, nil, "queryConfigOptionEx")
+
+    client, err := vs.getVcClient(ctx)
+    ...
+}
+```
+
+- The object parameter is nil-safe — pass the Kubernetes object being
+  reconciled when the provider method has one in scope (e.g.
+  `pkgctx.WithVCOpID(ctx, vm, "properties")`), or `nil` when the method only
+  takes primitive args (e.g. a `clusterMoID` string) and has no object to
+  fold into the op ID.
+- `operation` is a short, camelCase label naming the specific VC call/step
+  (e.g. `"properties"`, `"createOrUpdateVM"`, `"queryConfigOptionEx"`), not a
+  generic term like `"reconcile"`. The resulting string is
+  `vmoperator-<objName>-<operation>-<reconcileID>`.
+- Usually one op ID per provider method is enough: set it once near the top of
+  the `pkg/providers/vsphere/*.go` method that reaches vCenter rather than
+  labeling every individual govmomi/property-collector call it makes. This is a
+  default to steer toward when nothing else dictates otherwise, not a hard rule
+  — if a method does genuinely distinct operations worth telling apart in
+  VC-side logs, giving them their own op IDs is fine.
 
 ## Requeue and Error Semantics (pkgerr)
 
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -15,6 +15,7 @@ import (
 	"github.com/vmware-tanzu/vm-operator/controllers/storage"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachine"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachineclass"
+	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachineconfigoptions"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinegroup"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachinegrouppublishrequest"
 	"github.com/vmware-tanzu/vm-operator/controllers/virtualmachineimagecache"
@@ -93,8 +94,13 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 	}
 
 	if pkgcfg.FromContext(ctx).Features.VirtualMachineConfigPolicy {
+		// ConfigTarget fans out VirtualMachineConfigOptions, so it's
+		// registered first.
 		if err := configtarget.AddToManager(ctx, mgr); err != nil {
 			return fmt.Errorf("failed to initialize ConfigTarget controller: %w", err)
+		}
+		if err := virtualmachineconfigoptions.AddToManager(ctx, mgr); err != nil {
+			return fmt.Errorf("failed to initialize VirtualMachineConfigOptions controller: %w", err)
 		}
 	}
 
--- a/controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
+++ b/controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
@@ -1,0 +1,449 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package virtualmachineconfigoptions
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"regexp"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	vimv1 "github.com/vmware-tanzu/vm-operator/external/vim/api/v1alpha1"
+	pkgcond "github.com/vmware-tanzu/vm-operator/pkg/conditions"
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
+	"github.com/vmware-tanzu/vm-operator/pkg/patch"
+	"github.com/vmware-tanzu/vm-operator/pkg/providers"
+	"github.com/vmware-tanzu/vm-operator/pkg/record"
+)
+
+const (
+	// Finalizer is the finalizer placed on VirtualMachineConfigOptions objects
+	// by VM Operator.
+	Finalizer = "vmoperator.vmware.com/virtualmachineconfigoptions"
+
+	// ConfigTargetKind is the owner-reference Kind used to identify a
+	// ConfigTarget among a VirtualMachineConfigOptions' owner references.
+	ConfigTargetKind = "ConfigTarget"
+
+	// ConfigTargetNotFoundReason is the Ready condition reason used while a
+	// VirtualMachineConfigOptions has no ConfigTarget owner reference yet.
+	ConfigTargetNotFoundReason = "ConfigTargetNotFound"
+
+	// QueryFailedReason is the Ready condition reason used when resolving the
+	// owning cluster or querying QueryConfigOptionEx fails.
+	QueryFailedReason = "QueryFailed"
+
+	// NotFoundReason is the Ready condition reason used when
+	// QueryConfigOptionEx reports no config option for the requested
+	// hardware version.
+	NotFoundReason = "NotFound"
+)
+
+// SkipNameValidation is used for testing to allow multiple controllers with the
+// same name since Controller-Runtime has a global singleton registry to
+// prevent controllers with the same name, even if attached to different
+// managers.
+var (
+	SkipNameValidation *bool
+
+	// nonAlnumHyphenRE matches any character that is not a lowercase letter,
+	// digit, or hyphen.
+	nonAlnumHyphenRE = regexp.MustCompile(`[^a-z0-9-]+`)
+)
+
+var _ = SkipNameValidation
+
+// AddToManager adds this package's controller to the provided manager.
+func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) error {
+	var (
+		controlledType     = &vimv1.VirtualMachineConfigOptions{}
+		controlledTypeName = reflect.TypeOf(controlledType).Elem().Name()
+
+		controllerNameShort = fmt.Sprintf("%s-controller", strings.ToLower(controlledTypeName))
+	)
+
+	r := NewReconciler(
+		ctx,
+		mgr.GetClient(),
+		ctrl.Log.WithName("controllers").WithName(controlledTypeName),
+		record.New(mgr.GetEventRecorder(controllerNameShort)),
+		ctx.VMProvider,
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(controlledType).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: ctx.GetMaxConcurrentReconciles(controllerNameShort, 1),
+			SkipNameValidation:      SkipNameValidation,
+			LogConstructor:          pkglog.ControllerLogConstructor(controllerNameShort, controlledType, mgr.GetScheme()),
+		}).
+		Complete(r)
+}
+
+// NewReconciler returns a new Reconciler for VirtualMachineConfigOptions.
+func NewReconciler(
+	ctx context.Context,
+	c client.Client,
+	logger logr.Logger,
+	recorder record.Recorder,
+	vmProvider providers.VirtualMachineProviderInterface) *Reconciler {
+
+	return &Reconciler{
+		Context:    ctx,
+		Client:     c,
+		Logger:     logger,
+		Recorder:   recorder,
+		VMProvider: vmProvider,
+	}
+}
+
+// Reconciler reconciles a VirtualMachineConfigOptions object.
+type Reconciler struct {
+	client.Client
+	Context    context.Context
+	Logger     logr.Logger
+	Recorder   record.Recorder
+	VMProvider providers.VirtualMachineProviderInterface
+}
+
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=virtualmachineconfigoptions,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=virtualmachineconfigoptions/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=configtargets,verbs=get;list;watch
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=virtualmachineguestoptions,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=virtualmachineguestoptions/status,verbs=get;update;patch
+
+func (r *Reconciler) Reconcile(
+	ctx context.Context,
+	req ctrl.Request) (_ ctrl.Result, reterr error) {
+
+	ctx = pkgcfg.JoinContext(ctx, r.Context)
+
+	var obj vimv1.VirtualMachineConfigOptions
+	if err := r.Get(ctx, req.NamespacedName, &obj); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	patchHelper, err := patch.NewHelper(&obj, r.Client)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to init patch helper for %s: %w", req, err)
+	}
+	defer func() {
+		if err := patchHelper.Patch(ctx, &obj); err != nil {
+			if reterr == nil {
+				reterr = err
+			} else {
+				reterr = fmt.Errorf("%w,%w", err, reterr)
+			}
+		}
+	}()
+
+	if !obj.DeletionTimestamp.IsZero() {
+		return r.ReconcileDelete(ctx, &obj)
+	}
+	return r.ReconcileNormal(ctx, &obj)
+}
+
+// ReconcileDelete handles deletion of a VirtualMachineConfigOptions object.
+func (r *Reconciler) ReconcileDelete(
+	ctx context.Context,
+	obj *vimv1.VirtualMachineConfigOptions) (ctrl.Result, error) {
+
+	controllerutil.RemoveFinalizer(obj, Finalizer)
+	return ctrl.Result{}, nil
+}
+
+// ReconcileNormal reconciles a VirtualMachineConfigOptions object during normal
+// (non-delete) operations.
+func (r *Reconciler) ReconcileNormal(
+	ctx context.Context,
+	obj *vimv1.VirtualMachineConfigOptions) (ctrl.Result, error) {
+
+	if controllerutil.AddFinalizer(obj, Finalizer) {
+		// Return immediately to persist the finalizer before proceeding.
+		return ctrl.Result{}, nil
+	}
+
+	clusterMoID, err := r.findClusterMoID(ctx, obj)
+	if err != nil {
+		pkgcond.MarkError(obj, vimv1.ReadyConditionType, QueryFailedReason, err)
+		return ctrl.Result{}, err
+	}
+	if clusterMoID == "" {
+		// No ConfigTarget owner found yet; surface it on the Ready condition
+		// and requeue after a short delay.
+		pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, ConfigTargetNotFoundReason,
+			"no %s owner reference found yet", ConfigTargetKind)
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	configOption, err := r.VMProvider.QueryConfigOptionEx(ctx, clusterMoID, obj.Spec.HardwareVersion)
+	if err != nil {
+		pkgcond.MarkError(obj, vimv1.ReadyConditionType, QueryFailedReason, err)
+		return ctrl.Result{}, fmt.Errorf("failed to query config option ex: %w", err)
+	}
+
+	if configOption == nil {
+		pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, NotFoundReason,
+			"config option not found for hardware version %s", obj.Spec.HardwareVersion)
+		obj.Status.ObservedGeneration = obj.Generation
+		return ctrl.Result{}, nil
+	}
+
+	applyStatus(obj, configOption)
+
+	if err := r.fanOutGuestOptions(ctx, obj.Spec.HardwareVersion, configOption.GuestOSDescriptor); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	pkgcond.MarkTrue(obj, vimv1.ReadyConditionType)
+	obj.Status.ObservedGeneration = obj.Generation
+	return ctrl.Result{}, nil
+}
+
+// findClusterMoID looks up a cluster managed object ID from the owner
+// references of the given VirtualMachineConfigOptions. It returns an empty
+// string (with no error) when no ConfigTarget owner is found.
+//
+// A hardware-version-keyed VirtualMachineConfigOptions may legitimately be
+// co-owned by more than one ConfigTarget (see
+// controllers/configtarget's reconcileConfigOptions), since multiple
+// clusters can report supporting the same hardware version. vSphere's own
+// EnvironmentBrowser.QueryConfigOptionEx does not aggregate across hosts in
+// this situation either: it resolves to exactly one host's real, internally
+// consistent ConfigOption, chosen deterministically as the host running the
+// lowest product version among those supporting the requested key. We have
+// no comparable per-cluster signal to replicate that "most conservative"
+// selection today (that would need something like each ConfigTarget's ESXi
+// version, which this design deliberately does not expose -- see the
+// retired HostSystem CRD). So, among co-owners, this picks the
+// lexicographically smallest cluster MoID: an arbitrary but stable and
+// reproducible tie-break, not a proxy for "most conservative".
+//
+// TODO(vmop-3964): Replace this tie-break with a real conservative-selection
+// signal (e.g. minimum ESXi build across co-owning ConfigTargets) and record
+// the selected cluster in VirtualMachineConfigOptions.Status for debugging.
+func (r *Reconciler) findClusterMoID(
+	ctx context.Context,
+	obj *vimv1.VirtualMachineConfigOptions) (string, error) {
+
+	clusterMoIDs := make([]string, 0, len(obj.OwnerReferences))
+	for _, ref := range obj.OwnerReferences {
+		if ref.Kind != ConfigTargetKind {
+			continue
+		}
+		var ct vimv1.ConfigTarget
+		if err := r.Client.Get(ctx, client.ObjectKey{Name: ref.Name}, &ct); err != nil {
+			return "", fmt.Errorf("failed to get ConfigTarget %s: %w", ref.Name, err)
+		}
+		clusterMoIDs = append(clusterMoIDs, ct.Spec.ID.ID)
+	}
+
+	if len(clusterMoIDs) == 0 {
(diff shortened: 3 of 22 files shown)


controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+	obj *vimv1.VirtualMachineConfigOptions) (ctrl.Result, error) {
+
+	if controllerutil.AddFinalizer(obj, Finalizer) {
+		// Return immediately to persist the finalizer before proceeding.
+		return ctrl.Result{}, nil
+	}
+
+	clusterMoID, err := r.findClusterMoID(ctx, obj)
+	if err != nil {
+		pkgcond.Set(obj, pkgcond.FalseCondition(
```
faisalabujabal: we have a helper method that simplifies this for errors. this applies to the entire PR

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+			vimv1.ReadyConditionType,
+			"QueryFailed",
+			"%v", err,
+		))
+		return ctrl.Result{}, err
+	}
+	if clusterMoID == "" {
+		// No ConfigTarget owner found yet; requeue after a short delay.
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
faisalabujabal, line 196: should we have a condition for this? event? check if that is mentioned in the spec and other patterns in this repo

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+
+// findClusterMoID looks up the cluster managed object ID from the owner
+// references of the given VirtualMachineConfigOptions. It returns an empty
+// string (with no error) when no ConfigTarget owner is found.
+func (r *Reconciler) findClusterMoID(
+	ctx context.Context,
+	obj *vimv1.VirtualMachineConfigOptions) (string, error) {
+
+	for _, ref := range obj.OwnerReferences {
+		if ref.Kind == "ConfigTarget" {
```
faisalabujabal: move to a const

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+			var ct vimv1.ConfigTarget
+			if err := r.Client.Get(ctx, client.ObjectKey{Name: ref.Name}, &ct); err != nil {
+				return "", fmt.Errorf("failed to get ConfigTarget %s: %w", ref.Name, err)
+			}
+			return ct.Spec.ID.ID, nil
```
faisalabujabal: check the confgtarget controller but it does support multiple configtargets owning a configoption and your logic here means that we always select the first. we need to think about whether we should have distinct configoptions per configtarget or we should dedup. this will also translate to different guestoptions then right? 

lets research if the configoptions or guestoptions output might differ for the same hardware or guest across different clusters

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+		}
+	}
+	return "", nil
+}
+
+// applyStatus maps configOption fields onto obj.Status and fans out child
```
faisalabujabal: should we bundle the status and fan out in one method?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+
+	name := toGuestOptionsName(desc.Id)
+
+	desired := &vimv1.VirtualMachineGuestOptions{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+	}
+
+	if _, err := controllerutil.CreateOrPatch(ctx, r.Client, desired, func() error {
```
faisalabujabal: instead of calling patch twice can we not have one patch call with a patch method that does the logic of getting and updating the hardware version?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+	if err := r.Client.Status().Patch(ctx, updated, client.MergeFrom(current)); err != nil {
+		return fmt.Errorf("failed to patch status for VirtualMachineGuestOptions %s: %w", name, err)
+	}
+
+	return nil
+}
+
+// buildHardwareVersionStatus builds a
+// VirtualMachineGuestOptionsHardwareVersionStatus from a GuestOsDescriptor.
+func buildHardwareVersionStatus(
```
faisalabujabal, line 398: check the convention in the repo do we usually put such logic in the controller? also check if configcontroller followed the right convention

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+		SupportedNumDisks:           desc.SupportedNumDisks,
+		RecommendedDiskController:   vimv1.VirtualControllerType(desc.RecommendedDiskController),
+		RecommendedSCSIController:   vimv1.VirtualControllerType(desc.RecommendedSCSIController),
+		RecommendedCdromController:  vimv1.VirtualControllerType(desc.RecommendedCdromController),
+		RecommendedEthernetCard:     vimv1.EthernetCardType(desc.RecommendedEthernetCard),
+		SupportsWakeOnLan:           desc.SupportsWakeOnLan,
+		SupportLevel:                vimv1.SupportLevel(desc.SupportLevel),
+	}
+
+	s.SmcRequired = desc.SmcRequired
```
faisalabujabal: why are those values being set here instead of line 311 like the others?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,363 @@
...
+		q := resource.NewMilliQuantity(int64(desc.RecommendedMemMB)*1024*1024, resource.BinarySI)
+		s.RecommendedMem = q
+	}
+
+	return s
+}
+
+// nonAlnumHyphenRE matches any character that is not a lowercase letter,
+// digit, or hyphen.
+var nonAlnumHyphenRE = regexp.MustCompile(`[^a-z0-9-]+`)
```
faisalabujabal: i believe the convention is vars go on top

controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			},
+		}
+
+		configOptions = &vimv1.VirtualMachineConfigOptions{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: objName,
+				OwnerReferences: []metav1.OwnerReference{
+					{
+						APIVersion: vimv1.GroupVersion.String(),
+						Kind:       "ConfigTarget",
```
faisalabujabal: when we move it a const (earlier comment) you can reference it here and anywhere else you are using the literal string

controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			var current vimv1.VirtualMachineConfigOptions
+			Expect(vcsimCtx.Client.Get(vcsimCtx, client.ObjectKeyFromObject(configOptions), &current)).To(Succeed())
+			Expect(pkgcond.IsTrue(&current, vimv1.ReadyConditionType)).To(BeTrue())
+			Expect(current.Status.GuestOSIdentifiers).ToNot(BeEmpty())
+
+			var guestOptionsList vimv1.VirtualMachineGuestOptionsList
+			Expect(vcsimCtx.Client.List(vcsimCtx, &guestOptionsList)).To(Succeed())
+			Expect(guestOptionsList.Items).ToNot(BeEmpty())
```
faisalabujabal, line 658: we should add more assertions here. what about the status? what about the fan out?

controllers/controllers.go
```diff
@@ -77,6 +78,12 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		}
 	}
 
+	if pkgcfg.FromContext(ctx).Features.VirtualMachineConfigPolicy {
```
faisalabujabal: why not register all the controllers related to this feature flag together?

pkg/providers/vsphere/environment_browser.go
```diff
@@ -0,0 +1,52 @@
+// © Broadcom. All Rights Reserved.
```
faisalabujabal, line 1: this file doesn't have tests

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -167,6 +167,62 @@ func Spec(ctx context.Context, inputGetter func() SpecInput) {
...
+
+					Eventually(func(g Gomega) {
+						vco := &vimv1.VirtualMachineConfigOptions{}
+						g.Expect(svClusterClient.Get(ctx, ctrlclient.ObjectKey{Name: name}, vco)).To(Succeed())
+
+						cond := apimeta.FindStatusCondition(vco.Status.Conditions, vimv1.ReadyConditionType)
+						g.Expect(cond).ToNot(BeNil(), "VirtualMachineConfigOptions %q should have a Ready condition", name)
+						g.Expect(cond.Status).To(Equal(metav1.ConditionTrue), "VirtualMachineConfigOptions %q should be Ready", name)
+
+						g.Expect(vco.Status.GuestOSIdentifiers).ToNot(BeEmpty(),
```
faisalabujabal, line 193: should we identify what are the guest os in the testbeds that we run against for additional validation. it can be guarded with the hardware version check right?

faisalabujabal: can we do more assertions on the status?

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -167,6 +167,62 @@ func Spec(ctx context.Context, inputGetter func() SpecInput) {
...
+
+						for _, guestID := range vco.Status.GuestOSIdentifiers {
+							found := false
+							for j := range vgoList.Items {
+								if vgoList.Items[j].Spec.ID == guestID {
+									found = true
+									break
+								}
+							}
+							g.Expect(found).To(BeTrue(),
```
faisalabujabal: can we do more assertions?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Fixed in 3f7ad9a0 -- replaced with pkgcond.MarkError/MarkFalse/MarkTrue, matching the ConfigTarget controller's convention.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Added a condition (new ConfigTargetNotFoundReason) in 3f7ad9a0. No sibling controller in this repo emits events for a similar transient wait state, so I kept this to a condition only rather than adding a new pattern.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Extracted to an exported const (ConfigTargetKind) in 3f7ad9a0, along with the condition reason strings (ConfigTargetNotFoundReason, QueryFailedReason, NotFoundReason).

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Collapsed into a single controllerutil.CreateOrPatch call in 3f7ad9a0 -- its mutate function now sets both spec and status, and CreateOrPatch itself issues the spec patch and the status patch once it detects each differs, so the extra manual Get + Status().Patch was redundant. True single-round-trip isn't possible since VirtualMachineGuestOptions has +kubebuilder:subresource:status, but this removes the wasted extra fetch.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Checked ConfigTarget's convention: its own per-field mapping/upsert helpers (populateStatus, reconcileConfigOptions, garbageCollectConfigOptions) stay in the controller package as pure functions rather than moving to pkg/. Followed the same pattern in 3f7ad9a0 -- extracted the listMap upsert into its own pure, independently testable function (upsertHardwareVersionStatus) alongside the existing buildHardwareVersionStatus, rather than introducing a new pkg/ subpackage for one small function.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Moved into the struct literal in 3f7ad9a0 -- no functional reason they were split out, just inconsistent with the rest of the fields.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Moved nonAlnumHyphenRE up next to the other package-level var declarations in 3f7ad9a0.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
faisalabujabal: Test now references vmconfigoptions.ConfigTargetKind (and the other new exported consts) instead of the literal, in 3f7ad9a0.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
faisalabujabal: Added status and fan-out assertions to the vcsim test in 3f7ad9a0 -- checks ObservedGeneration, GuestOSIdentifiers, and for each guest ID, that a matching VirtualMachineGuestOptions exists with the expected Family and a hardwareVersions entry for this object's hardware version.

controllers/controllers.go
faisalabujabal: Turned out the rebase onto main had actually landed two separate 'if Features.VirtualMachineConfigPolicy' blocks (ConfigTarget's registration came in from main's history). Merged them into one in 3f7ad9a0, with ConfigTarget registered first since it's what fans out VirtualMachineConfigOptions.

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
faisalabujabal: Added an assertion for the universal 'otherGuest' fallback identifier in 3f7ad9a0, gated on it being a virtually version-independent baseline (present since the earliest VI API guest ID sets) rather than a testbed-specific list, so it stays meaningful without being brittle to whichever real cluster CI runs against.

faisalabujabal: Added more status assertions in 3f7ad9a0: ObservedGeneration vs Generation, Description non-empty, and GuestOSDefaultIndex validated as an in-bounds index into GuestOSIdentifiers.

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
faisalabujabal: Strengthened in 3f7ad9a0 -- now also asserts the matched VirtualMachineGuestOptions' status.fullName/status.family are populated and that it has a hardwareVersions entry for the specific VirtualMachineConfigOptions' hardware version, not just that a same-ID object exists.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Researched this -- long thread in the PR conversation. Short version: the co-ownership/dedup model is intentional, already-merged design from ConfigTarget's side (metadata.name = hardware-version key, no per-cluster disambiguation), not something for this PR to redesign. Confirmed via vpxd source (bora/vpx/vpxd/envBrowserMo.cpp) that vSphere's own QueryConfigOptionEx has the same problem and solves it by deterministically picking one host (the lowest-product-version host among those supporting the key) rather than aggregating -- so aggregating/unioning across clusters ourselves would be inconsistent with that precedent and risks producing an internally-inconsistent object no real cluster would return. Fixed the immediate bug (arbitrary non-deterministic 'first owner' selection) in 9ae567d7 by sorting to the lexicographically lowest cluster MoID -- documented in the code as an arbitrary-but-stable tie-break, not a proxy for 'most conservative', since we don't have a comparable per-cluster signal (e.g. ESXi version) to replicate vpxd's actual criterion. Confirmed via ESXi release notes/API changelog that same-hardware-version-different-guestOS is a real, recurring pattern (2 documented instances in the 8.x line so far) but currently only affects discovery/informational status, not admission enforcement (S9's acceptance criteria only reference ConfigTarget.status.maxHardwareVersion). Recommend a research.md clarification note as a follow-up rather than solving the full cross-cluster consistency problem in this PR.

pkg/providers/vsphere/environment_browser.go
faisalabujabal: Added a vcsim-backed test in ef9ddeba that exercises the real QueryConfigOptionEx call path (PropertyCollector RetrieveOne for the EnvironmentBrowser MoRef + the actual govmomi RPC) against a simulated vSphere backend, rather than only through the fully-mocked fake provider.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Looked at this again -- applyStatus already is the single entry point that does both: it maps the ConfigOption fields onto status inline, then loops guest OS descriptors and delegates the fan-out to reconcileGuestOptions. I kept the fan-out in its own function since it's a non-trivial CreateOrPatch dance (see the separate thread on that), but there's one call site or the other, not two independent callers to coordinate. Let me know if you meant something more specific here and I can adjust.

faisalabujabal: i meant should we have a different method that loops over the guest os descriptors and does the fan out instead of hiding it in applyStatus which should only update status?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: did you check conventions outside of the entire config policy work?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Fair, you're right -- applyStatus was doing two unrelated things (pure status mapping + driving the fan-out side effect). Split in eb45b4db: applyStatus is now pure (obj.Status only, no ctx/error return), and a new fanOutGuestOptions method owns the loop that creates/patches VirtualMachineGuestOptions. ReconcileNormal calls both explicitly now.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Fair push -- checking only ConfigTarget wasn't independent evidence since it's the same initiative. Went and checked older, unrelated controllers (virtualmachinesnapshot, storagepolicy, virtualmachine/virtualmachine, virtualmachinegroup, virtualmachineimagecache) plus the provider interface itself. Found two coexisting patterns, split by provider-method contract shape, not by age:

- When a VMProvider method's signature already returns a CRD-shaped type (e.g. GetStoragePolicyStatus returns infrav1.StoragePolicyStatus directly), the controller does zero mapping -- storagepolicy and virtualmachinesnapshot both just assign the provider's return value straight onto status.
- When a VMProvider method returns a raw govmomi type by contract (e.g. GetVirtualMachineConfigTarget returns *vimtypes.ConfigTarget, and our QueryConfigOptionEx returns *vimtypes.VirtualMachineConfigOption), the controller is what maps vSphere fields onto the CRD status -- this is exactly what ConfigTarget's populateStatus does (unit conversions, resource.Quantity construction, the works), and it stays in the controller package, not pkg/providers.
- Separately, virtualmachinegroup (findMemberStatus) and virtualmachineimagecache (reconcileLocations) do listMap-style upserts over CRD/k8s-sourced data entirely in-package too, with no pkg/ delegation -- independent confirmation for the upsert half of this specifically.

QueryConfigOptionEx's signature already returns the raw govmomi type (matches ConfigTarget's shape, not storagepolicy's), so buildHardwareVersionStatus/upsertHardwareVersionStatus staying in-package is consistent with the established convention for that provider-contract shape, not just copying a same-initiative sibling. Changing QueryConfigOptionEx to return an already-CRD-shaped type instead would be the alternative that matches the storagepolicy pattern, but that's a larger interface change (touches the fake provider and every call site) for what's currently a single small helper -- happy to revisit if you think that's worth doing here.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,399 @@
...
+
+	s := vimv1.VirtualMachineGuestOptionsHardwareVersionStatus{
+		HardwareVersion:             hardwareVersion,
+		SupportedMaxCPUs:            desc.SupportedMaxCPUs,
+		NumSupportedPhysicalSockets: desc.NumSupportedPhysicalSockets,
+		NumSupportedCoresPerSocket:  desc.NumSupportedCoresPerSocket,
+		RecommendedColorDepth:       desc.RecommendedColorDepth,
+		SupportedNumDisks:           desc.SupportedNumDisks,
+		RecommendedDiskController:   vimv1.VirtualControllerType(desc.RecommendedDiskController),
+		RecommendedSCSIController:   vimv1.VirtualControllerType(desc.RecommendedSCSIController),
```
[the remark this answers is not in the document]
faisalabujabal: Good catch — `NewMilliQuantity` treats the argument as milli-units, so this was producing a quantity 1000x too small. Fixed in 397594f4 by switching to `NewQuantity`.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,399 @@
...
+		RecommendedCdromController:  vimv1.VirtualControllerType(desc.RecommendedCdromController),
+		RecommendedEthernetCard:     vimv1.EthernetCardType(desc.RecommendedEthernetCard),
+		SupportsWakeOnLan:           desc.SupportsWakeOnLan,
+		SmcRequired:                 desc.SmcRequired,
```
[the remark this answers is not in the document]
faisalabujabal: Same fix as the sibling comment on SupportedMaxMem — switched to `NewQuantity` in 397594f4.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,399 @@
...
+		SmcRecommended:              desc.SmcRecommended,
+		UsbRecommended:              desc.UsbRecommended,
+		SupportsVMI:                 desc.SupportsVMI,
+		SupportedForCreate:          desc.SupportedForCreate,
```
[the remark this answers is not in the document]
faisalabujabal: Same fix as the sibling comments — switched to `NewQuantity` in 397594f4.

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -164,7 +164,99 @@ func Spec(ctx context.Context, inputGetter func() SpecInput) {
...
+				Eventually(func(g Gomega) {
+					var vgoList vimv1.VirtualMachineGuestOptionsList
+					g.Expect(svClusterClient.List(ctx, &vgoList)).To(Succeed())
+					g.Expect(vgoList.Items).ToNot(BeEmpty(),
+						"expected at least one VirtualMachineGuestOptions fanned out from a VirtualMachineConfigOptions")
+
+					for i := range vcoList.Items {
+						vco := &vcoList.Items[i]
+
+						for _, guestID := range vco.Status.GuestOSIdentifiers {
```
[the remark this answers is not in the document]
faisalabujabal: Agreed — moved the `vcoList` List call and a non-empty-guestOSIdentifiers assertion inside the Eventually block in 5a0c7e03, so the fan-out check re-evaluates against fresh data each poll instead of possibly running once against an empty list.

pkg/providers/vsphere/environment_browser.go
```diff
@@ -0,0 +1,52 @@
...
+
+// QueryConfigOptionEx retrieves the VirtualMachineConfigOption for the given
+// hardware version from the EnvironmentBrowser of the cluster identified by
+// clusterMoID.
+func QueryConfigOptionEx(
+	ctx context.Context,
+	vimClient *vim25.Client,
+	clusterMoID string,
+	hardwareVersion string) (*vimtypes.VirtualMachineConfigOption, error) {
+
```
bryanv, line 26: Use `WithVCOpID` to add that to the context so these reconciles can easily be correlated with VC logs

faisalabujabal: Good catch — added `pkgctx.WithVCOpID(ctx, obj, "queryConfigOptionEx")` at the call site in `vmconfigoptions_controller.go` so this reconcile can be correlated with VC logs. Also added a note to `.sdd/memory/operator-best-practices.md` calling out that new VC-calling code should set the op ID in the same change, so this doesn't get missed again.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,399 @@
...
+	name := toGuestOptionsName(desc.Id)
+	hwStatus := buildHardwareVersionStatus(hardwareVersion, desc)
+
+	obj := &vimv1.VirtualMachineGuestOptions{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+	}
+
+	if _, err := controllerutil.CreateOrPatch(ctx, r.Client, obj, func() error {
```
bryanv: IIRC `CreateOrPatch` doesn't update the status when it has to create the object

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,399 @@
...
+// upsertHardwareVersionStatus returns versions with hwStatus inserted,
+// replacing any existing entry for the same hardware version.
+func upsertHardwareVersionStatus(
+	versions []vimv1.VirtualMachineGuestOptionsHardwareVersionStatus,
+	hwStatus vimv1.VirtualMachineGuestOptionsHardwareVersionStatus,
+) []vimv1.VirtualMachineGuestOptionsHardwareVersionStatus {
+
+	for i := range versions {
+		if versions[i].HardwareVersion == hwStatus.HardwareVersion {
+			versions[i] = hwStatus
```
bryanv, line 389: Is there a concern that this could thrash?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: good catch! will patch if createorpatch creates a new object

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: another great catch. I'm adding a method that would retry on conflict which should resolve the issue. it also turns out we had the same issue in the configtarget controller which i'm fixing here as well

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,399 @@
...
+	if err := r.fanOutGuestOptions(ctx, obj.Spec.HardwareVersion, configOption.GuestOSDescriptor); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	pkgcond.MarkTrue(obj, vimv1.ReadyConditionType)
+	obj.Status.ObservedGeneration = obj.Generation
+	return ctrl.Result{}, nil
+}
+
+// findClusterMoID looks up a cluster managed object ID from the owner
```
hpannem, line 223: should we pick the cluster which is lowest common denominator for a vmx version? Probably, the cluster with lowest host version that supports this vmx version?

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,399 @@
...
+// it fetched, so there is no need for a separate Get plus Status().Patch.
+func (r *Reconciler) reconcileGuestOptions(
+	ctx context.Context,
+	hardwareVersion string,
+	desc vimtypes.GuestOsDescriptor) error {
+
+	name := toGuestOptionsName(desc.Id)
+	hwStatus := buildHardwareVersionStatus(hardwareVersion, desc)
+
+	obj := &vimv1.VirtualMachineGuestOptions{
```
hpannem, line 331: when is this GC'd?

test/builder/fake.go
```diff
@@ -90,6 +90,8 @@ func KnownObjectTypes() []client.Object {
 		&vimv1.ConfigTarget{},
 		&vimv1.VirtualMachineConfigOptions{},
 		&vimv1.VirtualMachineConfigPolicy{},
+		&vimv1.VirtualMachineConfigOptions{},
```
hpannem: nit: repeated

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: i'll address this as part of vmop-3744

test/builder/fake.go
faisalabujabal: fixed

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: I do address this in the description. I have an open thread with couple of people discussing how we can better do this. the correct way is to query all the hosts and get the min ESXi build number because it is also possible (and has happened) where vmware shipped the same hardware version but with new guest ids, so just relying on hardware version is not enough. I won't update this until we get to a resolution after the discussions

hpannem: Fair enough. can you add a TODO in the code? Also, do you want to add the sourceClusterId in the status where we fetched from? May be useful for debugging.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
bryanv: We shouldn't be maintaining our CreateOrPatch lite thing - just check the op result and patch the status if needed

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: i won't add the sourceclusterid yet. I will resolve the the issue with determining the correct version first then decide how we might have to update the api

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: reverted to using createorpatch + status subresource update

.sdd/memory/operator-best-practices.md
```diff
@@ -181,6 +181,47 @@ func relatedToMyResourceMapper(c client.Client) func(context.Context, client.Obj
...
 ```
 
+## Fan-out to Child Objects (CreateOrUpdate vs CreateOrPatch)
+
+Default to `controllerutil.CreateOrPatch`. Its patch is a JSON merge patch:
+map/object fields merge key-by-key, but **list fields are replaced
+wholesale**. So disjoint fields, or a shared map keyed per writer, are safe
+under `CreateOrPatch` even with multiple parents fanning into one child.
+
+If instead multiple parents fan-in write to the same *list*-typed field
```
bryanv: I think the best practice is to still use patch but with opportunistic locking

What this is describing isn't very common either - if we're responsible for both creating the object, and updating its status, we'll have a watch on the type which will redrive the reconciliation.

.sdd/memory/operator-best-practices.md
```diff
@@ -181,6 +181,47 @@ func relatedToMyResourceMapper(c client.Client) func(context.Context, client.Obj
...
+    return err
+}
+obj.Status.Bar = desired
+if err := c.Status().Update(ctx, obj); err != nil {
+    return err // conflict or otherwise -- bubble up
+}
+```
+
+Fake-client tests don't catch a missing status follow-up:
+`client/fake` doesn't strip `.status` on `Create`/`Update` the way a real
```
bryanv: The type needs to be registered with the fake client has having a status subresource.

.sdd/memory/operator-best-practices.md
```diff
@@ -207,6 +248,47 @@ Key ordering rationale:
...
+- The object parameter is nil-safe — pass the Kubernetes object being
+  reconciled when the provider method has one in scope (e.g.
+  `pkgctx.WithVCOpID(ctx, vm, "properties")`), or `nil` when the method only
+  takes primitive args (e.g. a `clusterMoID` string) and has no object to
+  fold into the op ID.
+- `operation` is a short, camelCase label naming the specific VC call/step
+  (e.g. `"properties"`, `"createOrUpdateVM"`, `"queryConfigOptionEx"`), not a
+  generic term like `"reconcile"`. The resulting string is
+  `vmoperator-<objName>-<operation>-<reconcileID>`.
+- Set it as close as possible to the VC call it labels — a single provider
```
bryanv: This isn't what we're doing, and something this fine grained is probably more annoying than helpful

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -0,0 +1,423 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+	}
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, obj, func() error {
```
bryanv: Why not CreateOrPatch? You only need to handle the Status with that when it created the obj. We should avoid doing noop Update()'s

faisalabujabal: we discussed more offline. I updated both spec and status updates to use createOrPatch with optimistic lock with a deep equality check before the patch

.sdd/memory/operator-best-practices.md
faisalabujabal: discussed offline I clarified the text more to imply that the intent was to push the generation of the opid down instead of doing it in the controller

bryanv (approved):

PR #1673, opened 2026-06-18
:bug: Wait for disks to be registered before snapshot creation

aruneshpa:

If we invoke VM snapshot API after the CnsRegisterVolume resource for an unmanaged disk has been created, but before CSI has had a chance to register the disk, then CRV will fail in a terminal way because CreateVolume call will refuse to reject a disk that is not the running point.

This change fixes this behavior by waiting for all disks to be registered before a VM snapshot is triggered.  A potential side effect is that a VM snapshot can not be taken if disk registration is broken. But that would have been the same behavior today anyway.

Testing Done:
- Unit tests
- Ran the failing end to end tests from our reco job a few times to make sure they pass

```release-note
 Wait for disks to be registered before snapshot creation
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vmsnapshot.go
+++ b/pkg/providers/vsphere/vmprovider_vmsnapshot.go
@@ -200,6 +200,33 @@ func markSnapshotFailed(vmCtx pkgctx.VirtualMachineContext,
 	return nil
 }
 
+// markSnapshotWaitingForDiskRegistration sets a condition on a snapshot to
+// indicate it is blocked waiting for unmanaged disk CnsRegisterVolume objects
+// to complete registration before the snapshot can be triggered.
+func markSnapshotWaitingForDiskRegistration(
+	vmCtx pkgctx.VirtualMachineContext,
+	k8sClient ctrlclient.Client,
+	vmSnapshot *vmopv1.VirtualMachineSnapshot) error {
+
+	if pkgcond.GetReason(vmSnapshot, vmopv1.VirtualMachineSnapshotCreatedCondition) ==
+		vmopv1.VirtualMachineSnapshotWaitingForDiskRegistrationReason {
+		return nil
+	}
+
+	patch := ctrlclient.MergeFrom(vmSnapshot.DeepCopy())
+	pkgcond.MarkFalse(
+		vmSnapshot,
+		vmopv1.VirtualMachineSnapshotCreatedCondition,
+		vmopv1.VirtualMachineSnapshotWaitingForDiskRegistrationReason,
+		"Snapshot cannot be created until all unmanaged disk registrations complete")
+	if err := k8sClient.Status().Patch(vmCtx, vmSnapshot, patch); err != nil {
+		return fmt.Errorf("failed to patch snapshot status: %w", err)
+	}
+	vmCtx.Logger.V(4).Info("Marked snapshot as waiting for disk registration",
+		"snapshotName", vmSnapshot.Name)
+	return nil
+}
+
 // getVirtualMachineSnapshotsForVM finds all VirtualMachineSnapshot objects that reference this VM.
 func getVirtualMachineSnapshotsForVM(
 	vmCtx pkgctx.VirtualMachineContext,
@@ -799,6 +826,32 @@ func synthesizeVMSpecForSnapshot(
 	}
 }
 
+// ReconcileSnapshotWaitForCRVCondition sets the WaitingForDiskRegistration
+// condition on every pending snapshot for the VM. It is called from the VM
+// reconcile loop when reconcileConfig exits early due to pending
+// CnsRegisterVolume objects, so that snapshots do not appear stuck with empty
+// conditions.
+func ReconcileSnapshotWaitForCRVCondition(
+	vmCtx pkgctx.VirtualMachineContext,
+	k8sClient ctrlclient.Client) error {
+
+	vmSnapshots, err := getVirtualMachineSnapshotsForVM(vmCtx, k8sClient)
+	if err != nil {
+		return err
+	}
+
+	for i := range vmSnapshots {
+		snapshot := &vmSnapshots[i]
+		if pkgcond.IsTrue(snapshot, vmopv1.VirtualMachineSnapshotCreatedCondition) {
+			continue
+		}
+		if err := markSnapshotWaitingForDiskRegistration(vmCtx, k8sClient, snapshot); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // ReconcileCurrentSnapshot reconciles the current snapshot owned by the VM.
 func ReconcileCurrentSnapshot(
 	vmCtx pkgctx.VirtualMachineContext,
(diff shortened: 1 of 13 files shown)


pkg/providers/vsphere/vmprovider_vmsnapshot.go
```diff
@@ -200,30 +200,70 @@ func markSnapshotFailed(vmCtx pkgctx.VirtualMachineContext,
...
+// hasUnmanagedVolumes returns true if the VM still has any volume with
+// VolumeTypeClassic in its status. Classic entries are written by
+// update_status.go for every non-FCD disk that lacks a managed PVC, and are
+// only removed by cleanupVolumeStatus after the volumes controller has added a
+// VolumeTypeManaged entry for the same disk UUID. This covers the full
+// registration lifecycle — Phase 1 (PVC created, no CRV yet), Phase 2 (CRV
+// created), CNS processing, and error cases — without requiring an API call.
+func hasUnmanagedVolumes(vm *vmopv1.VirtualMachine) bool {
+	for i := range vm.Status.Volumes {
+		if vm.Status.Volumes[i].Type == vmopv1.VolumeTypeClassic {
```
aakashchan: What happens if we `vm.spec.promoteDiskMode` is Disabled?

pkg/providers/vsphere/vmprovider_vmsnapshot.go
```diff
@@ -200,30 +200,70 @@ func markSnapshotFailed(vmCtx pkgctx.VirtualMachineContext,
...
+}
+
+// hasUnmanagedVolumes returns true if the VM still has any volume with
+// VolumeTypeClassic in its status. Classic entries are written by
+// update_status.go for every non-FCD disk that lacks a managed PVC, and are
+// only removed by cleanupVolumeStatus after the volumes controller has added a
+// VolumeTypeManaged entry for the same disk UUID. This covers the full
+// registration lifecycle — Phase 1 (PVC created, no CRV yet), Phase 2 (CRV
+// created), CNS processing, and error cases — without requiring an API call.
+func hasUnmanagedVolumes(vm *vmopv1.VirtualMachine) bool {
```
faisalabujabal: nit: I’m surprised we don’t have a public util that does this

pkg/providers/vsphere/vmprovider_vmsnapshot.go
```diff
@@ -929,6 +989,38 @@ func ReconcileCurrentSnapshot(
...
+		waitForRegistration := !pkgcond.IsTrue(vmCtx.VM, vmconfunmanagedvolsreg.Condition)
+
+		if !waitForRegistration {
+			// Defense in depth: even when the condition says True, verify
+			// directly against vm.Status.Volumes. The condition can briefly
+			// lag reality (e.g. after a snapshot converts the disk backing to
+			// a delta, FilterOutLinkedClones drops the disk from info.Disks in
+			// the register reconciler, which may not have corrected the
+			// condition yet on this reconcile).
+			waitForRegistration = hasUnmanagedVolumes(vmCtx.VM)
```
faisalabujabal: wouldn't this cause an issue if we revert FCDs for image disks? or atleast will require a change here too. it just means that the definition of disk promotion is done is in multiple controllers, which i'm not a fan off. what is the issue if the condition lags behind, doesn't this mean it will do the right thing on the next reconcile?

faisalabujabal (approved):

pkg/providers/vsphere/vmprovider_vmsnapshot.go
```diff
@@ -929,6 +973,36 @@ func ReconcileCurrentSnapshot(
 		return nil
 	}
 
+	// When AllDisksArePVCs is enabled, wait for all unmanaged disk registrations
```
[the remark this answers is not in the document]
aruneshpa: Not relevant since our Go version has fixed this issue.

pkg/providers/vsphere/vmprovider_vmsnapshot.go
```diff
@@ -200,30 +200,54 @@ func markSnapshotFailed(vmCtx pkgctx.VirtualMachineContext,
...
 	k8sClient ctrlclient.Client) ([]vmopv1.VirtualMachineSnapshot, error) {
 
-	// List all VirtualMachineSnapshot objects in the VM's namespace
 	var snapshotList vmopv1.VirtualMachineSnapshotList
-	if err := k8sClient.List(vmCtx, &snapshotList, ctrlclient.InNamespace(vmCtx.VM.Namespace)); err != nil {
+	if err := k8sClient.List(
+		vmCtx,
+		&snapshotList,
+		ctrlclient.InNamespace(vmCtx.VM.Namespace),
+		ctrlclient.MatchingLabels{vmopv1.VMNameForSnapshotLabel: vmCtx.VM.Name},
```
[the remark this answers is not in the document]
aruneshpa: Our webhooks always apply this label.  However, there are still a couple circumstances in which the labels can be missing.  Addressed this.

pkg/providers/vsphere/vmprovider_vmsnapshot.go
```diff
@@ -929,6 +973,36 @@ func ReconcileCurrentSnapshot(
...
+	// vmodl.fault.InvalidArgument and leaving the snapshot permanently stuck.
+	//
+	// We rely solely on the condition here. The register reconciler (step 10)
+	// always runs before this function (step 12) in the same reconcile cycle,
+	// so the condition on vmCtx.VM is already up-to-date. The hasPendingCRVs
+	// check in the register reconciler keeps the condition False when a CRV
+	// exists but its disk has been filtered out by FilterOutLinkedClones (e.g.
+	// after a snapshot converted the backing to a delta chain). Checking
+	// VolumeTypeClassic entries directly would block snapshots permanently on
+	// VMs where promoteDiskMode is disabled, because those disks are always
```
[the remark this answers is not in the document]
aruneshpa: I don't want to change the model from only reconciling one (the oldest one) snapshot custom resource.   When we go to the next snapshot, the condition will be marked on that snapshot.

pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
```diff
@@ -211,6 +211,23 @@ func (r reconciler) Reconcile(
...
+	// of info.Disks after a snapshot converted its backing to a delta chain),
+	// the PVC-based pending check above would have returned false while the CRV
+	// is still waiting for CNS to process it. Hold the condition False so that
+	// operations that gate on it (e.g. snapshot creation) continue to wait.
+	// We do NOT return ErrPendingRegister here — that would block the entire VM
+	// reconcile (including power state). The condition being False is sufficient
+	// to guard individual operations that care about registration state.
+	if hasPendingCRVs {
+		pkgcond.MarkFalse(
+			vm,
```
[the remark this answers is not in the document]
aruneshpa: If a CRV is pending, we won't necessarily have information about the progress.  Also, very difficult to collate a bunch of errors and dump them in a single Condition.

bryanv (approved):

pkg/providers/vsphere/vmprovider_vmsnapshot.go
aruneshpa: No longer applicable.  So, resolving this.

pkg/providers/vsphere/vmprovider_vmsnapshot.go
aruneshpa: No longer applicable.  So, resolving this.

PR #1675, opened 2026-06-19
🐛  Relax error when finding VM without InstanceUUID or MoID

bryanv:

For new VMs the Spec.InstanceUUID will be mutated in during create. For created brownfield VMs, the Spec.InstanceUUID is backfilled from the actual VC VM. Relax the condition for when the VM does not have an InstanceUUID or MoID to cover the rare gap of the VM not getting created in the first place prior to upgrade.

Fixes #

```release-note
NONE
```


hpannem (commented):
Looks good. Can we mock this behavior? create a CR with deletion timestamp and run the reconciler?

faisalabujabal (approved):

PR #1677, opened 2026-06-22
🌱 e2e: Add RegisterVM restore-to-new test & Fix Incremental Restore Resiliency

aakashchan:

1. Adds a new E2E test for the RegisterVM *restore-to-new* scenario: a vCenter VM whose Kubernetes CR has been deleted (simulating a post-disaster state) is re-registered via the RegisterVM API, creating a fresh CR from the backup YAML stored in ExtraConfig. 
2. There is a flakiness in our e2e test when we don't wait on backfill to complete. So OS disk is not present in spec at the time of backup and but is expected after InvokeRegisterVM restore.

Fixes #

Added new tests for more coverage.
Waits on Backfill condition to be true.

Incremental Restore - Ran the test multiple times to ensure there is no flakiness. 
Restore To new - Ran VDS, Stretched, didn't see the failure. 

https://uts.lvn.broadcom.net/dashboard/pipeline/detail/6a58759e7221200142432012/

```
./e2e-tests -e2e.e2e-config=test/e2e/vmservice/config/wcp.yaml --ginkgo.focus="Boot disk PVC lifecycle operations should succeed|VI-ADMIN-RegisterVM" --ginkgo.v

STDOUT: [SynchronizedAfterSuite] PASSED [102.027 seconds]

Ran 9 of 149 Specs in 1664.114 seconds
SUCCESS! -- 9 Passed | 0 Failed | 4 Pending | 136 Skipped
PASS
```

```release-note

```

### DIFF
--- a/test/e2e/vmservice/lib/vmoperator/vmoperator.go
+++ b/test/e2e/vmservice/lib/vmoperator/vmoperator.go
@@ -318,6 +318,51 @@ func WaitOnVirtualMachineCondition(
 	}, config.GetIntervals("default", "wait-virtual-machine-creation")...).Should(Succeed(), "Timed out waiting for Condition: %+v on VirtualMachine: %s", expectedCondition, vmName)
 }
 
+// WaitForBootDiskPVC waits for the VirtualMachineUnmanagedVolumesBackfilled and
+// VirtualMachineUnmanagedVolumesRegistered conditions to become true, then polls
+// spec.volumes until the boot disk (ControllerBusNumber 0, UnitNumber 0) has a
+// non-empty PersistentVolumeClaim.ClaimName. Returns the boot-disk volume name.
+// Only call this when the AllDisksArePVCs capability is enabled.
+func WaitForBootDiskPVC(
+	ctx context.Context,
+	config *config.E2EConfig,
+	client ctrlclient.Client,
+	ns, vmName string,
+) string {
+	By("Waiting on virtual machine conditions to become true")
+	for _, condition := range []metav1.Condition{
+		{Type: "VirtualMachineUnmanagedVolumesBackfilled", Status: metav1.ConditionTrue},
+		{Type: "VirtualMachineUnmanagedVolumesRegistered", Status: metav1.ConditionTrue},
+	} {
+		WaitOnVirtualMachineCondition(ctx, config, client, ns, vmName, condition)
+	}
+
+	By("Waiting for the boot disk to be promoted to a PVC")
+	var bootDiskVolName string
+	Eventually(func(g Gomega) bool {
+		vm, err := utils.GetVirtualMachineA5(ctx, client, ns, vmName)
+		if err != nil {
+			e2eframework.Logf("retry due to: %v", err)
+			return false
+		}
+		for _, vol := range vm.Spec.Volumes {
+			if vol.ControllerBusNumber != nil && *vol.ControllerBusNumber == 0 &&
+				vol.UnitNumber != nil && *vol.UnitNumber == 0 {
+				g.Expect(vol.PersistentVolumeClaim).ToNot(BeNil(),
+					"Expected boot disk to have a PersistentVolumeClaim")
+				g.Expect(vol.PersistentVolumeClaim.ClaimName).ToNot(BeEmpty(),
+					"Expected boot disk PVC to have a claim name")
+				bootDiskVolName = vol.Name
+				return true
+			}
+		}
+		return false
+	}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).
+		Should(BeTrue(), "Timed out waiting for boot disk to be found in spec.volumes")
+
+	return bootDiskVolName
+}
+
 // Utility function to ensure that a VirtualMachine is deleted.
 func WaitForVirtualMachineToBeDeleted(ctx context.Context, config *config.E2EConfig, client ctrlclient.Client, ns, vmName string) {
 	Eventually(func(g Gomega) {
--- a/test/e2e/vmservice/lib/vmoperator/vmoperator.go
+++ b/test/e2e/vmservice/lib/vmoperator/vmoperator.go
@@ -328,19 +328,24 @@ func WaitForBootDiskPVC(
 	config *config.E2EConfig,
 	client ctrlclient.Client,
 	ns, vmName string,
-) string {
+) (string, *vmopv1.VirtualMachine) {
 	By("Waiting on virtual machine conditions to become true")
 	for _, condition := range []metav1.Condition{
-		{Type: "VirtualMachineUnmanagedVolumesBackfilled", Status: metav1.ConditionTrue},
-		{Type: "VirtualMachineUnmanagedVolumesRegistered", Status: metav1.ConditionTrue},
+		{Type: consts.VMUnmanagedVolumesBackfilledCondition, Status: metav1.ConditionTrue},
+		{Type: consts.VMUnmanagedVolumesRegisteredCondition, Status: metav1.ConditionTrue},
 	} {
 		WaitOnVirtualMachineCondition(ctx, config, client, ns, vmName, condition)
 	}
 
 	By("Waiting for the boot disk to be promoted to a PVC")
-	var bootDiskVolName string
+	var (
+		bootDiskVolName string
+		vm              *vmopv1.VirtualMachine
+	)
+
 	Eventually(func(g Gomega) bool {
-		vm, err := utils.GetVirtualMachineA5(ctx, client, ns, vmName)
+		var err error
+		vm, err = utils.GetVirtualMachine(ctx, client, ns, vmName)
 		if err != nil {
 			e2eframework.Logf("retry due to: %v", err)
 			return false
@@ -360,7 +365,7 @@ func WaitForBootDiskPVC(
 	}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).
 		Should(BeTrue(), "Timed out waiting for boot disk to be found in spec.volumes")
 
-	return bootDiskVolName
+	return bootDiskVolName, vm
 }
 
 // Utility function to ensure that a VirtualMachine is deleted.
--- a/test/e2e/vmservice/lib/vmoperator/vmoperator.go
+++ b/test/e2e/vmservice/lib/vmoperator/vmoperator.go
@@ -318,6 +318,21 @@ func WaitOnVirtualMachineCondition(
 	}, config.GetIntervals("default", "wait-virtual-machine-creation")...).Should(Succeed(), "Timed out waiting for Condition: %+v on VirtualMachine: %s", expectedCondition, vmName)
 }
 
+// findBootDiskVolume returns the volume at ControllerBusNumber 0, UnitNumber 0
+// (the boot disk), or nil if spec.volumes has no such entry.
+// Note: This assumes there is only a single boot disk. Reconsider if/when
+// multiple boot disks are supported.
+func findBootDiskVolume(volumes []vmopv1.VirtualMachineVolume) *vmopv1.VirtualMachineVolume {
+	for i := range volumes {
+		vol := &volumes[i]
+		if vol.ControllerBusNumber != nil && *vol.ControllerBusNumber == 0 &&
+			vol.UnitNumber != nil && *vol.UnitNumber == 0 {
+			return vol
+		}
+	}
+	return nil
+}
+
 // WaitForBootDiskPVC waits for the VirtualMachineUnmanagedVolumesBackfilled and
 // VirtualMachineUnmanagedVolumesRegistered conditions to become true, then polls
 // spec.volumes until the boot disk (ControllerBusNumber 0, UnitNumber 0) has a
@@ -330,10 +345,18 @@ func WaitForBootDiskPVC(
 	ns, vmName string,
 ) (string, *vmopv1.VirtualMachine) {
 	By("Waiting on virtual machine conditions to become true")
-	for _, condition := range []metav1.Condition{
-		{Type: consts.VMUnmanagedVolumesBackfilledCondition, Status: metav1.ConditionTrue},
-		{Type: consts.VMUnmanagedVolumesRegisteredCondition, Status: metav1.ConditionTrue},
-	} {
+
+	conditions := []metav1.Condition{
+		{
+			Type:   consts.VMUnmanagedVolumesBackfilledCondition,
+			Status: metav1.ConditionTrue,
+		},
+		{
+			Type:   consts.VMUnmanagedVolumesRegisteredCondition,
+			Status: metav1.ConditionTrue,
+		},
+	}
+	for _, condition := range conditions {
 		WaitOnVirtualMachineCondition(ctx, config, client, ns, vmName, condition)
 	}
 
@@ -342,28 +365,20 @@ func WaitForBootDiskPVC(
 		bootDiskVolName string
 		vm              *vmopv1.VirtualMachine
 	)
-
-	Eventually(func(g Gomega) bool {
+	Eventually(func(g Gomega) {
 		var err error
 		vm, err = utils.GetVirtualMachine(ctx, client, ns, vmName)
-		if err != nil {
-			e2eframework.Logf("retry due to: %v", err)
-			return false
-		}
-		for _, vol := range vm.Spec.Volumes {
-			if vol.ControllerBusNumber != nil && *vol.ControllerBusNumber == 0 &&
-				vol.UnitNumber != nil && *vol.UnitNumber == 0 {
-				g.Expect(vol.PersistentVolumeClaim).ToNot(BeNil(),
-					"Expected boot disk to have a PersistentVolumeClaim")
-				g.Expect(vol.PersistentVolumeClaim.ClaimName).ToNot(BeEmpty(),
-					"Expected boot disk PVC to have a claim name")
-				bootDiskVolName = vol.Name
-				return true
-			}
-		}
-		return false
+		g.Expect(err).NotTo(HaveOccurred(), "failed to get VirtualMachine %s/%s", ns, vmName)
+
+		bootDiskVol := findBootDiskVolume(vm.Spec.Volumes)
+		g.Expect(bootDiskVol).ToNot(BeNil(), "boot disk volume not found in spec.volumes")
+		g.Expect(bootDiskVol.PersistentVolumeClaim).ToNot(BeNil(),
+			"Expected boot disk to have a PersistentVolumeClaim")
+		g.Expect(bootDiskVol.PersistentVolumeClaim.ClaimName).ToNot(BeEmpty(),
+			"Expected boot disk PVC to have a claim name")
+		bootDiskVolName = bootDiskVol.Name
 	}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).
-		Should(BeTrue(), "Timed out waiting for boot disk to be found in spec.volumes")
+		Should(Succeed(), "Timed out waiting for boot disk to be found in spec.volumes")
 
 	return bootDiskVolName, vm
 }
--- a/test/e2e/vmservice/vmservice/util.go
+++ b/test/e2e/vmservice/vmservice/util.go
@@ -836,10 +836,23 @@ func waitForBackupToComplete(
 ) {
 	By("Waiting for backup to complete for all PVCs")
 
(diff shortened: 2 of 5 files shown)


test/e2e/vmservice/vmservice/util.go
```diff
@@ -836,6 +836,44 @@ func waitForBackupToComplete(
...
+	volumeNames := make([]string, 0)
+	Eventually(func(g Gomega) bool {
+		vm, err := utils.GetVirtualMachineA5(ctx, clusterProxy.GetClient(), vm.Namespace, vm.Name)
+		if err != nil {
+			framework.Logf("retry due to: %v", err)
+			return false
+		}
+
+		for _, vol := range vm.Spec.Volumes {
+			volumeNames = append(volumeNames, vol.Name)
```
zhangheliu: There'\''s a variable-shadowing bug here that undermines the fix this PR is trying to make:

```go
Eventually(func(g Gomega) bool {
    vm, err := utils.GetVirtualMachineA5(ctx, clusterProxy.GetClient(), vm.Namespace, vm.Name)
    ...
```

`vm, err := ...` shadows the outer `vm` parameter — the refreshed object is scoped to the closure and discarded once `Eventually` returns. The code right after (line 880, `for _, vol := range vm.Spec.Volumes`) still reads the *original*, pre-wait `vm`. So we poll until the boot disk shows up, then throw that result away and build `expectedPVCNames` from stale data anyway — which can still hit the "No PVCs found... skipping backup wait" early return this PR is meant to avoid. Suggest `vm, err = ...` (drop `:=`) so the refreshed copy actually flows into the rest of the function.

Separately: this whole block (conditions wait + boot-disk `Eventually`) is a near line-for-line copy of the existing one in `vm_hardware.go:1826-1871`, down to the `volumeNames` slice and the `"wait-virtual-machine-condition-update"` interval key. The `vm_hardware.go` version actually uses `volumeNames` and additionally captures `bootDiskVolName` afterward — this copy drops both, leaving `volumeNames` (line 853/862) as dead code that'\''s appended to on every retry but never read.

Given this logic now exists in two places and one already has a bug the other didn'\''t, could we extract it into a shared helper (e.g. `WaitForBootDiskPVC` in `test/e2e/vmservice/lib/vmoperator/`) and have both `vm_hardware.go` and `waitForBackupToComplete` call it? That fixes the bug once instead of twice and keeps them from drifting further apart.

test/e2e/vmservice/vmservice/util.go
```diff
@@ -836,6 +836,44 @@ func waitForBackupToComplete(
...
+			Status: metav1.ConditionTrue,
+		},
+		{
+			Type:   "VirtualMachineUnmanagedVolumesRegistered",
+			Status: metav1.ConditionTrue,
+		},
+	}
+	for _, condition := range conditions {
+		vmoperator.WaitOnVirtualMachineCondition(ctx, config, clusterProxy.GetClient(), vm.Namespace, vm.Name, condition)
+	}
```
zhangheliu: This new wait couples `waitForBackupToComplete` to a capability that isn'\''t guaranteed by the FSS gate every caller of this function actually checks.

```go
conditions := []metav1.Condition{
    {Type: "VirtualMachineUnmanagedVolumesBackfilled", Status: metav1.ConditionTrue},
    {Type: "VirtualMachineUnmanagedVolumesRegistered", Status: metav1.ConditionTrue},
}
for _, condition := range conditions {
    vmoperator.WaitOnVirtualMachineCondition(ctx, config, clusterProxy.GetClient(), vm.Namespace, vm.Name, condition)
}
```

Both condition types are only ever set by controllers gated behind the `AllDisksArePVCs` feature (`pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go:44`, gated at `pkg/providers/vsphere/session/session_vm_update.go:1446`; `pkg/vmconfig/volumes/unmanaged/backfill/unmanagedvolumes_backfill.go:29`, gated at `pkg/providers/vsphere/vmprovider_vmsnapshot.go:892-995`) — which is driven by its own capability key `supports_vm_service_all_disks_are_pvcs` (`pkg/config/capabilities/capabilities.go:91-93`), independent of the `WCP_VMService_BackupRestore` FSS.

Every existing caller of `waitForBackupToComplete`/`WaitForBackupToComplete` only checks `vmServiceBackupRestoreEnabled`/`incrementalRestoreEnabled` before calling it (e.g. `registervm.go:148-150`). None of them check `AllDisksArePVCs`. On a testbed where `WCP_VMService_BackupRestore` is on but `AllDisksArePVCs` is off, these conditions will never be set and this wait will poll until timeout — failing every existing call site, not just the new test.

aakashchan: Thanks. AllDisksArePVCs check is absolutely must. I've fixed it now.

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -350,6 +350,51 @@ func WaitOnVirtualMachineCondition(
 	}, config.GetIntervals("default", "wait-virtual-machine-creation")...).Should(Succeed(), "Timed out waiting for Condition: %+v on VirtualMachine: %s", expectedCondition, vmName)
 }
 
+// WaitForBootDiskPVC waits for the VirtualMachineUnmanagedVolumesBackfilled and
+// VirtualMachineUnmanagedVolumesRegistered conditions to become true, then polls
+// spec.volumes until the boot disk (ControllerBusNumber 0, UnitNumber 0) has a
```
aruneshpa, line 338: This is okay for now but only because the images that we have only contain a single disk.  If there was ever an OVF that had two disks, you would return prematurely.
Again, I don't see a huge chance of this blowing up because if one disk is registered, most likely the other one is as well, still.

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -1823,52 +1823,14 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
-				conditions := []metav1.Condition{
-					{
-						Type:   "VirtualMachineUnmanagedVolumesBackfilled",
-						Status: metav1.ConditionTrue,
-					},
-					{
-						Type:   "VirtualMachineUnmanagedVolumesRegistered",
-						Status: metav1.ConditionTrue,
-					},
+				currentVM, err := utils.GetVirtualMachineA5(ctx, svClusterClient, vmSvcNamespace, vmName)
```
aruneshpa: So, I am all for having helpers, but one huge drawback is that this pattern is leading to A LOT of calls to the API server.  For example, even in the best case, the WaitForBootDiskPVC would do about 3 GET API calls for the same VM.  And right after that method returns, we would do another GET call.  Can you see if there's a way to combine these to make it more efficient?

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -1823,52 +1823,14 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
 
 				waitForVMAndBatchAttach(ctx, config, svClusterClient, vmSvcNamespace, vmName, []string{})
 
-				By("Waiting on virtual machine conditions to become true")
+				bootDiskVolName := vmoperator.WaitForBootDiskPVC(ctx, config, svClusterClient, vmSvcNamespace, vmName)
```
aruneshpa: Out of curiosity, I am trying to understand if your change to move this to a helper method is purely cosmetic, or is it required.  The existing code seems to be doing the same thing?

test/e2e/vmservice/vmservice/util.go
```diff
@@ -815,27 +815,26 @@ func decodeGzipBase64(encoded string) (string, error) {
...
+func WaitForBackupToComplete(
 	ctx context.Context,
 	vm *vmopv1a2.VirtualMachine,
 	clusterProxy *common.VMServiceClusterProxy,
 	config *config.E2EConfig,
 ) {
 	By("Waiting for backup to complete for all PVCs")
 
+	// If AllDisksArePVC is not enabled, then we end up waiting forever.
+	asyncSVFSSEnabled, err := utils.CheckSupervisorCapabilitiesCRDSupport(ctx, clusterProxy.GetClient())
```
aruneshpa: This is most likely a copy paste, but we don't need to check this.  e2e tests in this repo only work on 9.x onwards where async SV was always enabled.

test/e2e/vmservice/vmservice/util.go
```diff
@@ -815,27 +815,26 @@ func decodeGzipBase64(encoded string) (string, error) {
...
-func waitForBackupToComplete(
+func WaitForBackupToComplete(
 	ctx context.Context,
 	vm *vmopv1a2.VirtualMachine,
 	clusterProxy *common.VMServiceClusterProxy,
 	config *config.E2EConfig,
 ) {
 	By("Waiting for backup to complete for all PVCs")
 
+	// If AllDisksArePVC is not enabled, then we end up waiting forever.
```
aruneshpa: Curious again, how did you hit this?  AFAICT, this code only runs on 9.1 onwards where the cap has always been enabled.
I am fine keeping the check btw.  Just curious.a

test/e2e/vmservice/vmservice/viadmin/registervm.go
```diff
@@ -1075,4 +1075,68 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
...
+			vmoperator.WaitForVirtualMachineCreation(ctx, config, svClusterClient, input.WCPNamespaceName, vmName)
+			vmoperator.WaitForVirtualMachineMOID(ctx, config, svClusterClient, input.WCPNamespaceName, vmName)
+			vmoperator.WaitOnVirtualMachineCondition(ctx, config, svClusterClient, input.WCPNamespaceName, vmName,
+				metav1.Condition{
+					Type:   "VirtualMachineUnmanagedVolumesRegistered",
+					Status: metav1.ConditionTrue,
+				})
+
+			existingVM, err := utils.GetVirtualMachine(ctx, svClusterClient, input.WCPNamespaceName, vmName)
+			expectedVolCount := len(existingVM.Spec.Volumes)
```
aruneshpa: Hmm.  Do we really need to wait for the unmanged volumes to be registered?  The spec.volumes of the VM is populated immediately during the backfill.  The condition is set to true when in the end when the registration finishes.
If all we care about is the number of volumes, then we can simply wait for the backfill and take the count from there?

I am trying to see if that can shave a few seconds off of the pipeline run.

test/e2e/vmservice/vmservice/viadmin/registervm.go
```diff
@@ -50,7 +45,12 @@ import (
 	"github.com/vmware-tanzu/vm-operator/test/e2e/vmservice/skipper"
 	"github.com/vmware-tanzu/vm-operator/test/e2e/vmservice/vmservice"
 	"github.com/vmware-tanzu/vm-operator/test/e2e/wcpframework"
+	corev1 "k8s.io/api/core/v1"
```
aruneshpa: The imports were ordered on purpose.  This may have been an accidental change?

test/e2e/vmservice/vmservice/util.go
```diff
@@ -815,27 +815,26 @@ func decodeGzipBase64(encoded string) (string, error) {
...
+	asyncSVFSSEnabled, err := utils.CheckSupervisorCapabilitiesCRDSupport(ctx, clusterProxy.GetClient())
+	Expect(err).ToNot(HaveOccurred())
+	if utils.IsSupervisorCapabilityEnabled(ctx, clusterProxy.GetClientSet(), clusterProxy.GetDynamicClient(),
+		consts.AllDisksArePVCapabilityName, asyncSVFSSEnabled) {
+		vmoperator.WaitForBootDiskPVC(ctx, config, clusterProxy.GetClient(), vm.Namespace, vm.Name)
+		Expect(clusterProxy.GetClient().Get(ctx,
+			ctrlclient.ObjectKey{Namespace: vm.Namespace, Name: vm.Name}, vm)).
```
aruneshpa: Hmm.  You are now modifying the input pointer that you received in the argument.  Is that safe? What if a caller wasn't relying on this (e.g., they do a patch and now that fails because the RV is outdated).

test/e2e/vmservice/vmservice/util.go
aakashchan: https://github.com/vmware-tanzu/vm-operator/pull/1677#discussion_r3528575406
I also pointed claude to review my code, which also gave similar suggestions. This is very defensive test writing to avoid that. Shall I remove it?

test/e2e/vmservice/vmservice/util.go
aakashchan: This was part of the hardware tests earlier. Can I remove it now?

test/e2e/vmservice/vmservice/viadmin/registervm.go
aakashchan: Fixed.

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
aakashchan: No, Hardware tests also seem to have similar code. Its purely cosmetic and to address duplicate code maintainability.

test/e2e/vmservice/vmservice/util.go
aakashchan: I will remove all the feature flag checks which we set always like AllDisksArePVC and AsyncSV dependencies later in separate PR.

test/e2e/vmservice/vmservice/util.go
aakashchan: Sure @aruneshpa . You are right, I will handle its part of multiple test. It will take it up in a separate PR for easy to track the intent.

test/e2e/vmservice/vmservice/viadmin/registervm.go
aakashchan: Done. I removed the GetVirtualMachine calls which can be avoided by returning the VM after backfilled or registered state.

test/e2e/vmservice/vmservice/util.go
aakashchan: Thank you fixed.

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
aakashchan: Totally agree with you. Infact, we can do it via Interface type which should ensure we return VM object after state management.

test/e2e/vmservice/vmservice/util.go
```diff
@@ -1012,11 +1006,12 @@ func DeleteVMResource(
 	svClusterClient ctrlclient.Client) string {
 	By("Get VM before powering off")
```
bryanv: Now stale

test/e2e/vmservice/vmservice/util.go
bryanv: Use `func(g Gomega) {...` here, and also scope `err` to just here

bryanv: We should probably just capture the MoID before the VM is deleted or anything to make sure it is still set

aakashchan: Thanks a lot. `vm` could be stale, leading to nil pointer dereference. Fixed it.

aakashchan: Thanks updated.

test/e2e/vmservice/vmservice/util.go
aakashchan: Gotcha. Thanks.

bryanv (approved):

test/e2e/vmservice/lib/vmoperator/vmoperator.go
aruneshpa: Before resolving, I would expect at least a comment on the code so readers are aware of this assumption.

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -318,6 +318,56 @@ func WaitOnVirtualMachineCondition(
...
+// non-empty PersistentVolumeClaim.ClaimName. Returns the boot-disk volume name.
+// Only call this when the AllDisksArePVCs capability is enabled.
+func WaitForBootDiskPVC(
+	ctx context.Context,
+	config *config.E2EConfig,
+	client ctrlclient.Client,
+	ns, vmName string,
+) (string, *vmopv1.VirtualMachine) {
+	By("Waiting on virtual machine conditions to become true")
+	for _, condition := range []metav1.Condition{
```
aruneshpa: Nit: more readable to declare the list of conditions as a separate variable than an anonymous slice in the range expression.

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -318,6 +318,56 @@ func WaitOnVirtualMachineCondition(
...
+		bootDiskVolName string
+		vm              *vmopv1.VirtualMachine
+	)
+
+	Eventually(func(g Gomega) bool {
+		var err error
+		vm, err = utils.GetVirtualMachine(ctx, client, ns, vmName)
+		if err != nil {
+			e2eframework.Logf("retry due to: %v", err)
+			return false
```
aruneshpa: Change this to `g.Expect(err).NotTo(HaveOccurred())` so we know what failed if GetVM is what causes the timeout.

test/e2e/vmservice/vmservice/viadmin/registervm.go
```diff
@@ -1066,4 +1053,67 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
...
+					CloudInit: &manifestbuilders.CloudInit{
+						RawCloudConfig: &manifestbuilders.KeySelector{
+							Key:  "user-data",
+							Name: secretName,
+						},
+					},
+				},
+				PowerState: "PoweredOn",
+			}
+			vmYaml := manifestbuilders.GetVirtualMachineYamlA2(vmParameters)
```
aruneshpa, line 1139: Are you using A2 to keep the API version the min supported one so we can validate b/r works for this?

aruneshpa (approved):
LGTM with a few comments.

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -318,6 +318,56 @@ func WaitOnVirtualMachineCondition(
...
+				bootDiskVolName = vol.Name
+				return true
+			}
+		}
+		return false
+	}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).
+		Should(BeTrue(), "Timed out waiting for boot disk to be found in spec.volumes")
+
+	return bootDiskVolName, vm
+}
```
aruneshpa: Maybe consider rewriting without the bool  return in Eventually.  It is always a little bit confusing to use a return value to do what Eventually supports natively.
```suggestion
Eventually(func(g Gomega) {
    vm, err := utils.GetVirtualMachine(ctx, client, ns, vmName)
    g.Expect(err).NotTo(HaveOccurred(), "failed to get VirtualMachine %s/%s", ns, vmName)

    for _, vol := range vm.Spec.Volumes {
        if vol.ControllerBusNumber != nil && *vol.ControllerBusNumber == 0 &&
            vol.UnitNumber != nil && *vol.UnitNumber == 0 {
            g.Expect(vol.PersistentVolumeClaim).ToNot(BeNil(),
                "Expected boot disk to have a PersistentVolumeClaim")
            g.Expect(vol.PersistentVolumeClaim.ClaimName).ToNot(BeEmpty(),
                "Expected boot disk PVC to have a claim name")
            bootDiskVolName = vol.Name
            return
        }
    }
    g.Fail("boot disk volume not found in spec.volumes")
}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).Should(Succeed())
```

aakashchan: Thank you. I don't Fail() for the Gomega interface, I get the intent. Rewrote it now.

test/e2e/vmservice/vmservice/viadmin/registervm.go
aakashchan: Yes @aruneshpa , Thats correct. I gathered that from the rest of the test which also uses v1a2.

PR #1685, opened 2026-06-24
🐛 Fix imageNames empty strings race in VMGPR

brito-rafa:

The VirtualMachineGroupPublishRequest (VMGPR) controller was triggered to reconcile by the first of the child
VirtualMachinePublishRequest's (VMPR) patch operations, which writes status.conditions with Ready=true.
Because the parent reads child objects via apiReader, it could observe the child after that first patch but before the child controller had committed the the ImageName on a subsequent patch operation.

Fixes # vmop-3627179

```release-note
Addresses race condition on Virtual Machine Group Publish Request can read child objects with empty strings on imageNames
```


brito-rafa:
Close in favor of #1888

PR #1694, opened 2026-06-29
🌱 Dockerfile.e2e speed ups

bryanv:

A few changes to make this faster. Mostly matters for building locally.

- Reorg some layers to make caching more likely
- Build e2e binary and packer in separate stages
- Use cache mount for go mod

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -3,8 +3,9 @@
 
 ARG BASE_IMAGE=mirror.gcr.io/library/photon:5.0
 
-# Build stage - compile tests
-FROM ${BASE_IMAGE} AS builder
+# builder-base: Go toolchain + downloaded dependencies
+# Shared foundation for builder-e2e and builder-packer.
+FROM ${BASE_IMAGE} AS builder-base
 
 # Go module proxy settings. Override at build time to use an internal mirror
 # when proxy.golang.org is not reachable (e.g. network-restricted build systems).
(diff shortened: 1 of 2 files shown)


Dockerfile.e2e
```diff
@@ -14,34 +15,67 @@ ARG GONOSUMDB=
...
+# ../../external/*
+# ../../pkg/{backup/api,constants/testlabels}
+#
+# Go requires the go.mod of each replace target to be present before go mod download
+# can succeed. These directories are stable so keeping them here preserves the module
+# cache across routine source edits.
 COPY go.mod go.sum ./
 COPY api/ ./api/
 COPY external/ ./external/
+COPY pkg/backup/api/ ./pkg/backup/api/
```
faisalabujabal, line 39: why did we add this?

faisalabujabal: same with testlabels

Dockerfile.e2e
```diff
@@ -14,34 +15,67 @@ ARG GONOSUMDB=
...
+COPY pkg/constants/testlabels/ ./pkg/constants/testlabels/
+COPY test/e2e/go.mod test/e2e/go.sum ./test/e2e/
```
faisalabujabal, line 41: why are we copying those? i thought we won't need this

faisalabujabal (approved):
approving assuming that we built this atleast locally and ran the binary from within the image as a sanity check

Dockerfile.e2e
bryanv: It is explained in the paragraph right above this.

Note that despite it being in e2e's go.mod, we don't actually use testlabels in the e2e tests. I debated about removing that I could see us wanting to get rid of the strings we use here.

Dockerfile.e2e
bryanv: WDYM? We need that to pull down the dependencies at the next step. This lets us separate the less likely mod version change vs code change, which I viewed as a win for local dev but can go back to copying all of pkg here.

PR #1695, opened 2026-06-29
vmop-3740: Zone controller fans out ConfigTarget and VirtualMachineConfigPolicy

faisalabujabal:

Extends the Zone controller to implement Story S3 of the VirtualMachineConfigPolicy pipeline (vmop-3740). When the `VirtualMachineConfigPolicy` feature gate is enabled, reconciling a `Zone` now:

- Reads the `ClusterComputeResource` MoIDs directly from `zone.spec.managedVMs.clusterMoIDs` (no vCenter calls required).
- `CreateOrPatch`es a cluster-scoped `ConfigTarget` named after each cluster MoID, setting `spec.id.id` on creation only (idempotent).
- `CreateOrPatch`es a namespace-scoped `VirtualMachineConfigPolicy` whose `spec.zone` matches the zone name, defaulting `spec.syncMode = ConfigTarget` on creation only (preserves user overrides).
- Does **not** delete `ConfigTarget` or `VirtualMachineConfigPolicy` when pools are removed (deletion is out of scope here, tracked separately).

The vendored `external/tanzu-topology/api/v1alpha1` package (the Zone/AvailabilityZone/ZoneResourceUtilization CRDs) is fully re-synced with the upstream source, which added the `clusterMoIDs` field directly onto `Zone.spec.managedVMs` along with several other fields/types that had drifted out of sync. Reading this field directly replaces an earlier approach that derived cluster MoIDs from pool MoIDs via a live vCenter `PropertyCollector` RPC per pool — that RPC is no longer needed at all, and the Zone controller no longer talks to vSphere/VMProvider in any way.

Fixes # vmop-3740

- The zone `Reconciler` no longer has a `VMProvider` field — the earlier PropertyCollector-based approach required one, but the direct `ClusterMoIDs` read does not.
- `external/tanzu-topology/api/v1alpha1` was synced end-to-end against the upstream source rather than adding just the one field this PR needs, per reviewer request, so the vendored copy doesn't keep drifting out of sync piecemeal.
- The vcsim integration tests (`test/builder/vcsim_test_context.go`) now populate `Zone.Spec.ManagedVMs.ClusterMoIDs` and `Zone.Spec.Zone` (the `AvailabilityZoneReference`) to mirror what a real Supervisor cluster populates on the Zone object.
- Pre-existing `modernize`/`wsl_v5`/`embeddedstructfieldcheck` linter violations in touched files are fixed as required (the linter runs on the full file body of any touched file).

```release-note
NONE
```

**Testing**
E2E test runs
```
Ran 3 of 135 Specs in 195.587 seconds
SUCCESS! -- 3 Passed | 0 Failed | 5 Pending | 127 Skipped
PASS
```


controllers/infra/zone/zone_controller.go
```diff
@@ -151,26 +163,141 @@ func (r *Reconciler) ReconcileDelete(
...
+		if poolMoID == "" {
+			continue
+		}
+
+		moRef, err := vcenter.GetResourcePoolOwnerMoRef(ctx, vimClient, poolMoID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get cluster for resource pool %s: %w", poolMoID, err)
+		}
+
+		if moRef.Type != "ClusterComputeResource" {
```
faisalabujabal: this should be moved to a const

test/e2e/vmservice/vmservice/configpolicy/zone_fanout.go
```diff
@@ -0,0 +1,172 @@
...
+	WCPNamespaceName string
+}
+
+// ZoneFanOutSpec verifies that the zone controller fans out ConfigTarget and
+// VirtualMachineConfigPolicy objects when a Zone is reconciled with the
+// VirtualMachineConfigPolicy capability active.
+func ZoneFanOutSpec(ctx context.Context, inputGetter func() ZoneFanOutSpecInput) {
+	const (
+		specName      = "zone-config-target-fan-out"
+		capabilityKey = "supports_vm_service_vm_config_policy"
```
faisalabujabal: should be in a const like the other capabilities

test/e2e/vmservice/vmservice/configpolicy/zone_fanout.go
```diff
@@ -0,0 +1,172 @@
...
+
+		svClusterProxy = input.ClusterProxy.(*common.VMServiceClusterProxy)
+		svClusterClient = input.ClusterProxy.GetClient()
+
+		skipper.SkipUnlessSupervisorCapabilityEnabled(ctx, svClusterProxy, capabilityKey)
+	})
+
+	Context("When a workload namespace has zones", func() {
+		It("Should have at least one ConfigTarget derived from the zone's pool MoIDs",
+			Label("extended-functional"),
```
faisalabujabal: new tests or significant changes to existing e2e tests should be experimental to begin with. add this to the .sdd/memory

controllers/infra/zone/zone_controller.go
```diff
@@ -64,52 +69,62 @@ func NewReconciler(
...
 
 func (r *Reconciler) Reconcile(
 	ctx context.Context,
 	req ctrl.Request) (_ ctrl.Result, reterr error) {
-
 	ctx = pkgcfg.JoinContext(ctx, r.Context)
 	ctx = watcher.JoinContext(ctx, r.Context)
 
 	var obj topologyv1.Zone
-	if err := r.Get(ctx, req.NamespacedName, &obj); err != nil {
```
bryanv, line 99: Why the churn?

controllers/infra/zone/zone_controller.go
```diff
@@ -151,26 +165,141 @@ func (r *Reconciler) ReconcileDelete(
...
+			return nil, fmt.Errorf("failed to get cluster for resource pool %s: %w", poolMoID, err)
+		}
+
+		if moRef.Type != clusterComputeResourceType {
+			continue
+		}
+
+		if _, ok := seen[moRef.Value]; !ok {
+			seen[moRef.Value] = struct{}{}
+			clusterMoIDs = append(clusterMoIDs, moRef.Value)
```
bryanv: If keep the old Zone CR, then if the order doesn't matter just use a `set.Set[string]` and return the list

controllers/infra/zone/zone_controller.go
```diff
@@ -151,26 +165,141 @@ func (r *Reconciler) ReconcileDelete(
...
+		return nil, fmt.Errorf("failed to get vSphere client: %w", err)
+	}
+
+	vimClient := vsphereClient.VimClient()
+
+	seen := make(map[string]struct{})
+
+	var clusterMoIDs []string
+
+	for _, poolMoID := range zone.Spec.ManagedVMs.PoolMoIDs {
```
bryanv: Our vendor of the Zone CRD is quite old - there is a ClusterMoIDs field so we don't have to do this lookup.

controllers/infra/zone/zone_controller.go
faisalabujabal: golinter mentions no inline error so claude fixed the files it touched

faisalabujabal: Confirmed — that blank line removal was required by `wsl_v5` (unnecessary whitespace immediately after a multi-line function signature), not unrelated churn. I briefly tried reverting it in 60186bf3's predecessor and `golangci-lint` flagged it again, so it has to stay.

controllers/infra/zone/zone_controller.go
faisalabujabal: Took this further than `sets.Set[string]` — turned out our vendored Zone CRD really was missing `spec.managedVMs.clusterMoIDs` (confirmed against vcf/tera and a live WCP cluster). Synced the vendor in cadd2abe and now read `zone.Spec.ManagedVMs.ClusterMoIDs` directly off the Zone object, so the manual dedup/seen-map this comment was about doesn't exist anymore — the upstream field is already deduplicated.

controllers/infra/zone/zone_controller_suite_test.go
```diff
@@ -20,5 +19,5 @@ func TestZoneController(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	zone.SkipNameValidation = ptr.To(true)
+	zone.SkipNameValidation = new(true)
```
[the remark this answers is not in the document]
faisalabujabal: This repo is pinned to Go 1.26.4 (see `go.mod`), which added `new(expr)` as sugar for "pointer to a new variable initialized to expr" — equivalent to `ptr.To(true)`. It compiles cleanly and `test-controllers`/`lint-go` both passed in CI, so this isn't actually a compile error. Leaving as-is.

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -0,0 +1,162 @@
...
+						Kind:    configTargetGVK.Kind + "List",
+					})
+					Expect(svClusterClient.List(ctx, &ctList)).To(Succeed())
+					Expect(ctList.Items).ToNot(BeEmpty(),
+						"expected at least one ConfigTarget for zone %q", z.Name)
+
+					for j := range ctList.Items {
+						ct := &ctList.Items[j]
+						id, _, _ := unstructured.NestedString(ct.Object, "spec", "id", "id")
+						Expect(id).To(Equal(ct.GetName()),
```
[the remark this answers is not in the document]
faisalabujabal: Not applicable here — these are exact Kubernetes resource names (DNS-subdomain, lowercase-enforced by validation), not free-typed identifiers a user might enter with inconsistent casing. `EqualFold` would mask a real mismatch rather than catch one, so I'm leaving the exact comparison.

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -0,0 +1,162 @@
...
+
+		It("Should have a VirtualMachineConfigPolicy for each zone in the namespace",
+			Label("extended-functional", "experimental"),
+			func() {
+				zoneList, err := utils.ListZonesByNamespace(ctx, svClusterClient, input.WCPNamespaceName)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(zoneList.Items).ToNot(BeEmpty(),
+					"expected at least one Zone in namespace %q", input.WCPNamespaceName)
+
+				for i := range zoneList.Items {
```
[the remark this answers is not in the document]
faisalabujabal: Good catch, this was a real bug — `Zone` is cluster-scoped so `z.Namespace` is always empty. Fixed in 313737ae to use `input.WCPNamespaceName`.

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -0,0 +1,162 @@
...
+
+					var ctList unstructured.UnstructuredList
+					ctList.SetGroupVersionKind(schema.GroupVersionKind{
+						Group:   configTargetGVK.Group,
+						Version: configTargetGVK.Version,
+						Kind:    configTargetGVK.Kind + "List",
+					})
+					Expect(svClusterClient.List(ctx, &ctList)).To(Succeed())
+					Expect(ctList.Items).ToNot(BeEmpty(),
+						"expected at least one ConfigTarget for zone %q", z.Name)
```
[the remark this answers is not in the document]
faisalabujabal: Fixed in 313737ae — hoisted the `List` call above the per-zone loop so it only runs once.

controllers/infra/zone/zone_controller.go
faisalabujabal: Confirmed and fixed in cadd2abe. Checked our vendored `external/tanzu-topology` against the real source (vcf/tera bora/vim/supervisor/controller-managers/apis/public/tanzu-topology/v1alpha1) and a live WCP cluster (`kubectl get zone -o yaml`) — `spec.managedVMs.clusterMoIDs` is populated directly on the Zone object. Synced the full vendor (also picked up CPU/Memory/VirtualMachineReservations/AllowedClusterComputeResourceMoIDs/SystemInfo and the previously-unvendored ZoneResourceUtilization CRD while there), regenerated deepcopy + CRD manifests, and simplified `reconcileConfigTargets` to read `zone.Spec.ManagedVMs.ClusterMoIDs` directly — no AvailabilityZone Get, no vSphere round trip at all.

controllers/infra/zone/zone_controller.go
```diff
@@ -172,5 +176,72 @@ func (r *Reconciler) ReconcileNormal(
 		}
 	}
 
+	if pkgcfg.FromContext(ctx).Features.VirtualMachineConfigPolicy {
+		if err := r.reconcileConfigTargets(ctx, obj); err != nil {
+			return ctrl.Result{}, err
```
bryanv: Maybe for later, but we should probably make this controller accumulate errors since it is now doing several unrelated things.

controllers/infra/zone/zone_controller.go
```diff
@@ -172,5 +176,72 @@ func (r *Reconciler) ReconcileNormal(
...
+	zone *topologyv1.Zone) error {
+	for _, clusterMoID := range zone.Spec.ManagedVMs.ClusterMoIDs {
+		ct := &vimv1.ConfigTarget{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: clusterMoID,
+			},
+		}
+
+		if _, err := controllerutil.CreateOrPatch(ctx, r.Client, ct, func() error {
+			if ct.UID == "" {
```
bryanv: Why have this check?

external/tanzu-topology/api/v1alpha1/zoneresourceutilization.go
```diff
@@ -0,0 +1,263 @@
+// © Broadcom. All Rights Reserved.
```
bryanv: It probably isn't worth us vendoring this

test/builder/vcsim_test_context.go
bryanv, line 462: Why not update clusterMoIDs in this loop?

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -0,0 +1,164 @@
...
+				zoneList, err := utils.ListZonesByNamespace(ctx, svClusterClient, input.WCPNamespaceName)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(zoneList.Items).ToNot(BeEmpty(),
+					"expected at least one Zone in namespace %q", input.WCPNamespaceName)
+
+				// ConfigTargets are cluster-scoped and named after the cluster
+				// MoID derived from each zone's pool MoIDs. List them once and
+				// verify that for each zone at least one ConfigTarget exists and
+				// has a matching spec.id.
+				var ctList unstructured.UnstructuredList
```
bryanv: Why is this using unstructured?

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -0,0 +1,164 @@
...
+						"VirtualMachineConfigPolicy %q/%q should exist", input.WCPNamespaceName, z.Name)
+
+					zoneField, _, _ := unstructured.NestedString(policy.Object, "spec", "zone")
+					Expect(zoneField).To(Equal(z.Name),
+						"VirtualMachineConfigPolicy %q/%q should reference zone %q",
+						input.WCPNamespaceName, z.Name, z.Name)
+				}
+			})
+
+		It("Should not delete a ConfigTarget when the zone is reconciled again",
```
bryanv: I don't understand what this test is really adding. What is guaranteeing that a reconcile occurs during this period? I think this more effective as a unit test, but it doesn't look like the zone controller has unit just integration tests the version of this test in there is I good enough.

bryanv (approved):
Some of the added/remove empty lines seems to be against the general convention that we use

external/tanzu-topology/api/v1alpha1/zoneresourceutilization.go
faisalabujabal: yeah I was debating whether to pull the entire api or just what we need.

test/builder/vcsim_test_context.go
faisalabujabal: done

controllers/infra/zone/zone_controller.go
faisalabujabal: good catch. did it now

controllers/infra/zone/zone_controller.go
faisalabujabal: you are right it is unnecessary

external/tanzu-topology/api/v1alpha1/zoneresourceutilization.go
faisalabujabal: removed

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
faisalabujabal: fixed

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
faisalabujabal: you are right. i removed it

PR #1698, opened 2026-06-30
🌱 Update VM provider workflow for multiple network providers

bryanv:

After a namespace's networking is migrated, the original network provider becomes the legacy provider, and must still be supported. Update the VM reconcile path to support multiple network interface CR types.

Fixes #

```release-note
NONE
```


pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -409,9 +410,10 @@ func (s *Session) poweredOnReconfigure(
 
 	configSpec := &vimtypes.VirtualMachineConfigSpec{}
 
-	if pkgcfg.FromContext(vmCtx).NetworkProviderType == pkgcfg.NetworkProviderTypeVPC {
-		err := s.handleRestoredVPCInterfaces(vmCtx, vcVM, networkResults)
-		if err != nil {
+	if networkType, err := netsetutil.GetProviderType(vmCtx, s.K8sClient, vmCtx.VM.Namespace); err != nil {
```
bryanv, line 414: A later refactor will get & store this upfront so we don't need to query it multiple times during a reconcile.

pkg/providers/vsphere/network/network.go
```diff
@@ -49,7 +49,13 @@ type NetworkInterfaceResults struct { //nolint:revive
 }
 
 type NetworkInterfaceResult struct { //nolint:revive
+	// ObjectName is the name of the network interface CR backing this result.
+	// Empty for named (testing-only) networks which have no CR.
 	ObjectName string
+	// ObjectProviderType is the network provider type of the CR. Combined with
+	// ObjectName it uniquely identifies the CR across all network providers.
+	ObjectProviderType pkgcfg.NetworkProviderType
```
akutz, line 58: Why not just `ProviderType`? Is this to be consistent with `ObjectName`? If so, should we refactor them both to be `IfaceObjectName` and `IfaceProviderType` to ensure there is no confusion over what is meant by "Object"?

pkg/providers/vsphere/network/network.go
```diff
@@ -49,7 +49,13 @@ type NetworkInterfaceResults struct { //nolint:revive
 }
 
 type NetworkInterfaceResult struct { //nolint:revive
```
akutz, line 52: I wonder if this should not inherit from the `VirtualMachineNetworkInterfaceStatus` type?

pkg/providers/vsphere/network/network.go
```diff
@@ -140,17 +147,19 @@ func CreateNetworkDevices(
 ) ([]Device, error) {
 
 	networkSpec := vm.Spec.Network
-	if networkSpec == nil {
+	if networkSpec == nil || networkSpec.Disabled {
 		return nil, nil
```
akutz, line 149: Should we log the reason we are returning early? I suppose it could be pretty noisy, but maybe at level 4?

pkg/providers/vsphere/network/network.go
```diff
@@ -140,17 +147,19 @@ func CreateNetworkDevices(
...
-	case pkgcfg.NetworkProviderTypeNamed:
-		// Named network is a testing only hack that does not have a CR so
-		// handle it separately.
-		return createNetworkDevicesForNamedNetwork(ctx, vm, finder)
-	case "":
-		return nil, fmt.Errorf("no network provider set")
+	if !pkgcfg.FromContext(ctx).Features.PerNamespaceNetworkProvider {
+		switch pkgcfg.FromContext(ctx).NetworkProviderType {
+		case pkgcfg.NetworkProviderTypeNamed:
+			// Named network is a testing only hack that does not have a CR so
```
akutz, line 155: Is any testing still actually using this? If it's not ripped out in this PR, we should rip it out soon in a separate PR.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -409,9 +410,10 @@ func (s *Session) poweredOnReconfigure(
...
+		return err
```
akutz: Should this be

```golang
return fmt.Errorf(
		"failed to get network provider type while doing powered on reconfig: %w",
		err)
```

?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -504,7 +506,11 @@ func (s *Session) poweredOffReconfigure(
 	}
 
 	if reconfigErr == nil || errors.Is(reconfigErr, ErrReconfigure) {
-		if pkgcfg.FromContext(vmCtx).NetworkProviderType == pkgcfg.NetworkProviderTypeVPC {
+		networkType, err := netsetutil.GetProviderType(vmCtx, s.K8sClient, vmCtx.VM.Namespace)
+		if err != nil {
+			return err
```
akutz: Should this be

```golang
return fmt.Errorf(
		"failed to get network provider type while doing powered off reconfig: %w",
		err)
```

?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1013,7 +1019,11 @@ func (s *Session) resizeVMWhenPoweredStateOff(
 		return err
 	}
 
-	if pkgcfg.FromContext(vmCtx).NetworkProviderType == pkgcfg.NetworkProviderTypeVPC {
+	networkType, err := netsetutil.GetProviderType(vmCtx, s.K8sClient, vmCtx.VM.Namespace)
+	if err != nil {
+		return err
```
akutz: Should this be

```golang
return fmt.Errorf(
		"failed to get network provider type while doing powered off resize: %w",
		err)
```

?

akutz (approved):
Thanks @bryanv! Approved with some feedback.

pkg/providers/vsphere/network/network.go
bryanv: My longer term goal is for the whole Results guy to go away. Internally within this package I've split this apart into Device & Bootstrap, and moved building the bootstrap fields from the device but haven't gotten to updating th consumers.

pkg/providers/vsphere/network/network.go
bryanv: Yah this is to match the existing naming convention. Like I mentioned, I want to remove this result struct, and the corresponding fields in the Device are already

```
type Device struct {
	ProviderType pkgcfg.NetworkProviderType
	InterfaceObj ctrlclient.Object
...
```

I don't think the churn of renaming the existing field is worth when this whole thing will go away.

pkg/providers/vsphere/network/network.go
bryanv: Sure

pkg/providers/vsphere/network/network.go
bryanv: I had peeked at CAPV the other week and didn't see anything that would indicate they were setting this up. I'll confirm with the team and remove it after this commit since we'd be in a decent spot to revert that removal if it is actually later needed.

PR #1699, opened 2026-06-30
:sparkles: Power off ordering for VirtualMachineGroup

aruneshpa:

Implement reverse-order, delay-based power-off sequencing for VirtualMachineGroup members, mirroring the existing power-on ordering.

Controller (reconcileMembers):
- Rename applyPowerOnTime -> applyPowerStateTime and parse the ApplyPowerStateTimeAnnotation for PowerStateOff as well as PowerStateOn, so nested groups inherit the correct base timestamp from their parent.
- Before iterating boot order groups, reverse the slice when the desired power state is Off (reverseBootOrder helper). This gives last-group-first ordering during shutdown.
- Switch the per-group delay accumulation from an if-on-PowerOn to a switch on PowerState, adding the PowerOffDelay branch for PowerStateOff.
- Rename applyPowerStateTime throughout reconcileMember and updateMemberPowerState to match.

updateMemberPowerState:
- Stamp ApplyPowerStateTimeAnnotation for PowerStateOff in addition to PowerStateOn on both VM and VMGroup members.

vmprovider_vm.go (reconcilePowerState):
- Extend the feature-gated delay gate from PowerStateOn-only to also fire for PowerStateOff, so VMs respect the annotation when powering off as well.

Tests:
- Add "power off the group with delay" integration test in the VMG controller suite: sets PowerOffDelay on both boot order groups, patches the group to PowerStateOff, and asserts that vmgroup-2 (index 1) receives bootOrder2Delay and vm-1 (index 0) receives the cumulative bootOrder2Delay + bootOrder1Delay, verifying the reverse-order semantics end-to-end.
- Add unit tests for the power-off delay annotation path in vmprovider_vm_power_test.go, covering both future-time (requeue) and past-time (proceed) branches.

**Testing Done**
```
make test-e2e TEST_FOCUS="Should power off group members in reverse boot order with the expected delay"
Ran 1 of 137 Specs in 282.978 seconds
SUCCESS! -- 1 Passed | 0 Failed | 5 Pending | 131 Skipped
PASS
```

```release-note
Power off ordering for VirtualMachineGroup
```

### DIFF
--- a/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
+++ b/controllers/virtualmachinegroup/virtualmachinegroup_controller.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 	"strings"
 	"time"
 
@@ -268,11 +269,12 @@ func (r *Reconciler) reconcileMembers(
 	}
 
 	// Get the group's apply power state change time that may be set from its
-	// parent group for power-on delay.
-	var applyPowerOnTime time.Time
-	if updatePowerState && ctx.VMGroup.Spec.PowerState == vmopv1.VirtualMachinePowerStateOn {
+	// parent group for power-on or power-off delay.
+	var applyPowerStateTime time.Time
+	if updatePowerState && (ctx.VMGroup.Spec.PowerState == vmopv1.VirtualMachinePowerStateOn ||
+		ctx.VMGroup.Spec.PowerState == vmopv1.VirtualMachinePowerStateOff) {
 		if v := ctx.VMGroup.Annotations[constants.ApplyPowerStateTimeAnnotation]; v != "" {
-			applyPowerOnTime, err = time.Parse(time.RFC3339Nano, v)
+			applyPowerStateTime, err = time.Parse(time.RFC3339Nano, v)
 			if err != nil {
 				ctx.Logger.Error(err, "Failed to parse time from annotation",
 					"annotationKey", constants.ApplyPowerStateTimeAnnotation,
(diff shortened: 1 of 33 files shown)


controllers/virtualmachinegroup/virtualmachinegroup_controller.go
```diff
@@ -808,12 +826,24 @@ func shouldUpdatePowerState(
 	return true, lastUpdateAnnotation, nil
 }
 
+// reverseBootOrder returns a new slice with boot order groups in reverse order.
+func reverseBootOrder(
+	bootOrders []vmopv1.VirtualMachineGroupBootOrderGroup,
+) []vmopv1.VirtualMachineGroupBootOrderGroup {
+
+	reversed := make([]vmopv1.VirtualMachineGroupBootOrderGroup, len(bootOrders))
```
hpannem: nit: slices.clone and slices.reverse?

hpannem (commented):
This looks ok. Consistent with power on. However, what happens if shutdown is stuck?
we have 5 mins max shutdown time. if the poweroff delay is less than that, there could be parallel shutdowns across groups. I believe vCD does force poweroff after the delay and proceeds to the next vm. (My inference from the docs)

also, if a boot order group poweroff successfully before the poweroff delay expires, we still wait right? this is bit of optimization though.

hpannem (approved):
Approving as comments will be taken care in the followup PR.

test/e2e/vmservice/vmservice/virtualmachine/vm_group.go
```diff
@@ -1657,3 +1811,29 @@ func VMGroupSpec(ctx context.Context, inputGetter func() VMGroupSpecInput) {
...
+// be poweredOff. There is no vCenter timestamp equivalent to runtime.bootTime
+// for power-off, so this is used in place of one to verify power-off
+// ordering/delay behavior.
+func waitForVMPoweredOffTime(
+	ctx context.Context,
+	propCollector *property.Collector,
+	moRef types.ManagedObjectReference) time.Time {
+
+	var offTime time.Time
+	Eventually(func() bool {
```
bryanv: Prefer to use `func(g Gomega)` pattern here so failures are clearer

bryanv (approved):

PR #1702, opened 2026-07-01
[x-port] [9.1.1] Block snapshot when unmanaged disk registration is pending

aruneshpa:

This is an x-port of #1673 from main branch.

    If we invoke VM snapshot API after the CnsRegisterVolume resource for
    an unmanaged disk has been created, but before CSI has had a chance to
    register the disk, then CRV will fail in a terminal way because
    CreateVolume call will refuse to reject a disk that is not the running
    point.

    Changes:
    - Add VirtualMachineSnapshotWaitingForDiskRegistrationReason to v1alpha5/v6
    - Fix VirtualMachineUnmanagedVolumesRegistered condition being incorrectly
      set True when CRVs exist with Registered=false
    - Return nil (not ErrPendingRegister) from the hasPendingCRVs path so that
      power state and other operations are not gated on CRV completion
    - Replace hasPendingCRVsForVM K8s List call with hasUnmanagedVolumes, an
      in-memory scan of vm.Status.Volumes for Classic / Unmanaged volumes
    - Fix getVirtualMachineSnapshotsForVM to use server-side VMNameForSnapshotLabel
      filtering instead of listing all snapshots in the namespace

    Test changes:
    - Add VMNameForSnapshotLabel to DummyVirtualMachineSnapshot (webhook sets
      this in production but tests bypass it)
    - Add CnsRegisterVolume to KnownObjectTypes(): it has +kubebuilder:subresource:status
      but was omitted, causing fake-client Status().Update() to be silently ignored

```release-note
Block snapshot when unmanaged disk registration is pending
```

### DIFF
--- a/api/v1alpha5/virtualmachinesnapshot_types.go
+++ b/api/v1alpha5/virtualmachinesnapshot_types.go
@@ -102,6 +102,11 @@ const (
 	// VirtualMachineSnapshotCreationFailedReason documents that the virtual machine
 	// snapshot creation has failed.
 	VirtualMachineSnapshotCreationFailedReason = "VirtualMachineSnapshotCreationFailed"
+
+	// VirtualMachineSnapshotWaitingForDiskRegistrationReason documents that the
+	// virtual machine snapshot is waiting for all unmanaged disk registrations
+	// to complete before the snapshot can be triggered.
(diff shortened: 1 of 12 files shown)


brito-rafa (approved):
LGTM

faisalabujabal (approved):

brito-rafa (approved):
LGTM

PR #1704, opened 2026-07-01
✨ Add ConfigTarget controller cluster-scope reconcile path

faisalabujabal:

Implements the cluster-scope reconcile path of the new `ConfigTarget` controller, per `.sdd/specs/001-class-policy-resize/` (Story S5.b, tasks T063–T065):

- On seeing a cluster-scoped `ConfigTarget`, resolves the vSphere cluster from `metadata.name` and calls `EnvironmentBrowser.QueryConfigTarget` + `QueryConfigOptionDescriptor` via a new narrow provider method `GetVirtualMachineConfigTarget`.
- Populates `ConfigTarget.status` numeric limits, memory quantities, and security flags from the `QueryConfigTarget` result.
- Fans out one `VirtualMachineConfigOptions` object per hardware-version key returned by `QueryConfigOptionDescriptor`.
- Garbage-collects hardware-version keys that are no longer reported. `VirtualMachineConfigOptions` objects use a plain (non-controller) owner reference so the same object can be safely co-owned by multiple `ConfigTarget`s (different clusters sharing a hardware version); GC only removes this `ConfigTarget`'s owner reference and deletes the object once it has no remaining owners, per the multi-owner design documented in `plan.md`.
- Sets `status.observedGeneration` and a `Ready` condition (`ClusterNotFound` / `QueryConfigTargetFailed` / `QueryConfigOptionDescriptorFailed` reasons on failure).

**Out of scope for this PR** (tracked separately, called out in `tasks.md`):
- Per-host iteration (max hardware version aggregation, SR-IOV/DVX discovery) — Story S5.c / vmop-3759.
- Webhook validation for `ConfigTarget` / `VirtualMachineConfigOptions`, controller registration in `controllers/controllers.go`, and RBAC `ClusterRole` manifest regeneration — Story S5.a.

No CRD/API field changes; `ConfigTargetSpec`, `ConfigTargetStatus`, and `VirtualMachineConfigOptionsSpec` are unmodified.

Fixes #

NA

```release-note
Add ConfigTarget controller cluster-scope reconcile path
```

**E2E Test Result**
```
Ran 5 of 135 Specs in 187.006 seconds
SUCCESS! -- 5 Passed | 0 Failed | 5 Pending | 125 Skipped
PASS
```


controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	ctx = pkgcfg.JoinContext(ctx, r.Context)
+
+	logger := pkglog.FromContextOrDefault(ctx)
+	logger = logger.WithName(req.Name)
+	ctx = logr.NewContext(ctx, logger)
+
+	var obj vimv1.ConfigTarget
+
+	err := r.Get(ctx, req.NamespacedName, &obj)
+	if err != nil {
```
faisalabujabal: the repo's convention is inline error handling. this comment extends to the entire diff.

controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,352 @@
...
+// ReconcileNormal reconciles the cluster-scope path of a ConfigTarget:
+// QueryConfigTarget populates status, QueryConfigOptionDescriptor fans out
+// VirtualMachineConfigOptions objects, and any hardware version no longer
+// reported is garbage-collected.
+func (r *Reconciler) ReconcileNormal(
+	ctx context.Context,
+	obj *vimv1.ConfigTarget) error {
+	logger := pkglog.FromContextOrDefault(ctx)
+
+	// The cluster's managed object ID is derived from metadata.name, not
```
faisalabujabal: do we need this or having this comment in the spec is sufficient

controllers/configtarget/configtarget_controller_test.go
```diff
@@ -0,0 +1,433 @@
...
+				Expect(got.Status.SMCPresent).To(BeFalse())
+				Expect(got.Status.SEVSupported).To(BeTrue())
+				Expect(got.Status.SEVSNPSupported).To(BeFalse())
+				Expect(got.Status.TDXSupported).To(BeFalse())
+				Expect(got.Status.MaxMemOptimalPerf).ToNot(BeNil())
+				Expect(got.Status.MaxMemOptimalPerf.Value()).To(Equal(int64(4096) * 1024 * 1024))
+				Expect(got.Status.SupportedMaxMem).ToNot(BeNil())
+				Expect(got.Status.SupportedMaxMem.Value()).To(Equal(int64(8192) * 1024 * 1024))
+
+				// Per this sub-task's scope, devices (including SR-IOV) are left unset.
```
faisalabujabal: if this is temporary then lets add a TODO with the ticket tracking the gap that will be resolved

controllers/configtarget/configtarget_controller_test.go
```diff
@@ -0,0 +1,433 @@
...
+			var vmcoList vimv1.VirtualMachineConfigOptionsList
+			Expect(vcsimCtx.Client.List(vcsimCtx, &vmcoList)).To(Succeed())
+			Expect(vmcoList.Items).ToNot(BeEmpty())
+
+			// SKIP: a GC-via-vcsim-mutation scenario (dropping a previously
+			// reported hardware version key at runtime) is not exercised here.
+			// vcsim's EnvironmentBrowser reports a fixed, version-derived set
+			// of ConfigOptionDescriptor keys that cannot be shrunk without
+			// restarting the simulator, so this behavior is only covered by
+			// the fake-provider-backed unit test above.
```
faisalabujabal, line 405: are we not able to restart the simulator?

controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,352 @@
...
+
+	logger.V(4).Info("Reconciled ConfigTarget", "clusterMoID", clusterMoID)
+
+	return nil
+}
+
+// isClusterNotFoundErr returns true if err indicates the cluster named by
+// the ConfigTarget's metadata.name does not resolve in vSphere.
+func isClusterNotFoundErr(err error) bool {
+	return strings.Contains(err.Error(), "not found")
```
faisalabujabal: this is a weak check. if we are bubbling up the vim fault here isn't there a method that checks the stack of the vim faults and can be used here to determine if the cause of the error was a vim not found fault?

controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,352 @@
...
+
+// removeOwnerRefAndDeleteIfOrphaned removes obj's owner reference from vmco
+// and, if that leaves vmco with no remaining owner references, deletes it.
+// If other ConfigTarget objects still own vmco, only the owner reference is
+// removed and vmco itself is left in place.
+func (r *Reconciler) removeOwnerRefAndDeleteIfOrphaned(
+	ctx context.Context,
+	obj *vimv1.ConfigTarget,
+	vmco *vimv1.VirtualMachineConfigOptions) error {
+	if err := controllerutil.RemoveOwnerReference(obj, vmco, r.Scheme()); err != nil {
```
faisalabujabal: isn't it more optimal to determine if this is the last owner ref that is being removed. either you update the object by removing the owner ref or delete the entire obj?

controllers/configtarget/configtarget_controller.go
faisalabujabal: Fixed in 55ba866b — switched to `if err := X; err != nil` at every site in this file where the non-error return value wasn't needed outside the block. Two spots (`patchHelper, err := patch.NewHelper(...)` and the `_, err := controllerutil.CreateOrPatch(...)` inside the fan-out loop) are left as plain assignment since `patchHelper` is used later in the deferred patch, and the `CreateOrPatch` call has a multi-line closure body where inlining would hurt readability more than it'd help.

Note for the record: I initially cross-checked this against a locally-installed golangci-lint (v2.11.4) and it flagged inline error handling as an actual lint violation (`noinlineerr`). That linter doesn't exist in the repo's pinned version (`hack/tools/go.mod` pins v2.1.6) — built and ran that exact binary and got 0 issues either way. So there's no enforced rule either direction; this is purely the human-reviewed style convention you're describing, and I applied it as requested.

controllers/configtarget/configtarget_controller.go
faisalabujabal: Removed the comment in 55ba866b — you're right that this is otherwise only informative background dueling with the spec doc, and the metadata.name-vs-spec.id decision is now recorded in .sdd/specs/001-class-policy-resize/tasks.md's implementation note instead of duplicated in code.

controllers/configtarget/configtarget_controller_test.go
faisalabujabal: Added a TODO with the ticket reference in 55ba866b: `// TODO(vmop-3759): devices (including SR-IOV) are left unset until per-host discovery lands.` This mirrors the existing ticket reference already on populateStatus's doc comment.

controllers/configtarget/configtarget_controller_test.go
faisalabujabal: No hook exists today to reconfigure/restart the vcsim model mid-test (TestContextForVCSim builds the simulator.Model once in setupVCSim and only tears it down in AfterEach), and vcsim's EnvironmentBrowser derives its descriptor set from the simulated hardware/ESXi version, which isn't something we can shrink for the same MoID without a full model rebuild. Doing that just to re-exercise GC felt disproportionate given the fake-provider unit tests above already cover the GC path (drop-a-version and co-ownership-retention scenarios) exhaustively. Left the SKIP comment as an explicit call-out rather than silently omitting the scenario — happy to revisit if there's a lighter-weight way to reconfigure vcsim mid-test that I'm missing.

controllers/configtarget/configtarget_controller.go
faisalabujabal: Good catch — fixed in 55ba866b. `isClusterNotFoundErr` now uses `fault.As(err, &f)` against `*vimtypes.ManagedObjectNotFound`, matching the pattern in pkg/providers/vsphere/vcenter/getvm.go and vmprovider.go, instead of substring-matching on the error message. Also updated the fake-provider unit test to construct a real `soap.WrapVimFault(&vimtypes.ManagedObjectNotFound{...})`-wrapped error so it actually exercises the fault-tree walk rather than a string match; reran the vcsim-backed 'cluster does not exist' spec too and it still passes end-to-end against the real fault type vcsim returns.

controllers/configtarget/configtarget_controller.go
faisalabujabal: This is already handled — removeOwnerRefAndDeleteIfOrphaned removes the owner ref first, then checks `len(vmco.OwnerReferences) > 0` to decide: if other owners remain, Update; otherwise Delete the whole object. No change needed here, but let me know if you were seeing something different.

controllers/configtarget/configtarget_controller.go
faisalabujabal: Fair point — inlined it in f68c710b. It was only called from the one spot in ReconcileNormal, so the named helper wasn't pulling its weight; the fault.As check is now inline there directly.

controllers/configtarget/configtarget_controller.go
faisalabujabal: Good call, that's a real improvement — restructured in f68c710b. Now checks len(vmco.OwnerReferences) == 1 up front (the caller already confirmed obj is an owner via HasOwnerReference, so a single remaining ref is obj's own) and deletes directly in that case, skipping the RemoveOwnerReference mutation entirely. Otherwise it removes the ref and Updates. Reran all 9 specs (including the co-ownership-retention vcsim/unit tests) — still green.

faisalabujabal: Two follow-ups worth capturing here:

**Patch vs Update:** keeping Update. OwnerReferences is a plain slice with no CRD-level strategic-merge-key handling (that only applies to built-in types compiled into the apiserver; CRDs only get JSON Merge Patch / JSON Patch / Apply). A MergeFrom-based patch would still replace the whole array from a client-side snapshot — same lost-update exposure as Update, just without the conflict error. Since this object can legitimately be mutated concurrently by two different ConfigTarget reconciles (per reconcileConfigOptions's own doc comment), Update's resourceVersion check giving a loud 409-and-requeue is the safer behavior, not a downside.

**Is the len==1 delete safe or repetitive to check:** it was mathematically safe today (the caller already confirms ownership via HasOwnerReference right before calling this), but that made correctness an implicit contract with the one call site rather than something the function verified itself. Tightened it in c2cde591 to also check vmco.OwnerReferences[0].UID == obj.GetUID(), so the delete-outright branch is correct on its own without depending on the caller having pre-checked. No extra API call, still 9/9 green.

controllers/configtarget/configtarget_controller_test.go
faisalabujabal: Landed as an E2E test instead of trying to restart/reconfigure vcsim — added in 11e2a80a to test/e2e/vmservice/vmservice/configpolicy/configpolicy.go. Rather than fighting vcsim's fixed per-model descriptor set, it creates a VirtualMachineConfigOptions with a synthetic hardware-version key owned by a real ConfigTarget, then asserts the controller's next reconcile (triggered automatically by the Owns(..., builder.MatchEveryOwner) watch) garbage-collects it since that key will never appear in the real live descriptor set. That exercises the actual GC code path end-to-end against a live cluster without needing to fake vSphere's capability reporting. Left the vcsim SKIP comment as-is since it's still accurate context for why this scenario isn't covered there.

controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=configtargets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=configtargets/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=virtualmachineconfigoptions,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=virtualmachineconfigoptions/status,verbs=get;update;patch
+
+func (r *Reconciler) Reconcile(
+	ctx context.Context,
+	req ctrl.Request) (_ ctrl.Result, reterr error) {
```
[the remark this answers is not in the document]
faisalabujabal: Not needed here — the whole controller is registered conditionally: `configtarget.AddToManager` is only called from `controllers.go` when `Features.VirtualMachineConfigPolicy` is true, and this repo's feature flags are restart-gated (main.go's SIGUSR2 restart handler), so there's no window where `Reconcile` runs with the flag off. This matches every other feature-gated controller in `controllers.go` (VMSnapshots, VMGroups, FastDeploy, VSpherePolicies, etc.) — none of them re-check the flag inside `Reconcile`; they rely solely on the conditional `AddToManager` call. Happy to add a belt-and-suspenders check if you want one, but it'd be inconsistent with the rest of the codebase.

controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,344 @@
...
+	configTarget, configOptionDescriptors, err := r.VMProvider.GetVirtualMachineConfigTarget(ctx, clusterMoID)
+	if err != nil {
+		reason := QueryConfigTargetFailedReason
+
+		var notFound *vimtypes.ManagedObjectNotFound
+		if _, ok := fault.As(err, &notFound); ok {
+			reason = ClusterNotFoundReason
+		}
+
+		pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, reason, "%v", err)
```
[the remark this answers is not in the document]
faisalabujabal: Fixed in 83f032f2 — wrapped with `fmt.Errorf("failed to get virtual machine config target: %w", err)`.

controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,344 @@
...
+
+		return err
+	}
+
+	populateStatus(obj, configTarget)
+
+	if err = r.reconcileConfigOptions(ctx, obj, configOptionDescriptors); err != nil {
+		pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, QueryConfigOptionDescriptorFailedReason, "%v", err)
+		return err
```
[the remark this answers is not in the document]
faisalabujabal: Fixed in 83f032f2 — wrapped with `fmt.Errorf("failed to reconcile config options: %w", err)`.

controllers/configtarget/configtarget_controller.go
```diff
@@ -0,0 +1,344 @@
...
+// The object itself is deleted only once that removal leaves it with no
+// remaining owner references, so a hardware-version key shared by another
+// live ConfigTarget is left untouched.
+func (r *Reconciler) garbageCollectConfigOptions(
+	ctx context.Context,
+	obj *vimv1.ConfigTarget,
+	liveKeys sets.Set[string]) error {
+	var list vimv1.VirtualMachineConfigOptionsList
+
+	if err := r.List(ctx, &list); err != nil {
```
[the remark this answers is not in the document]
faisalabujabal: Fixed in 83f032f2 — wrapped with `fmt.Errorf("failed to remove owner reference and delete if orphaned: %w", err)`.

controllers/configtarget/configtarget_controller.go
bryanv: We have a `MarkError` to simplify this a bit

controllers/configtarget/configtarget_controller_test.go
```diff
@@ -0,0 +1,438 @@
...
+		vcsimCtx   *builder.TestContextForVCSim
+		reconciler *configtarget.Reconciler
+		obj        *vimv1.ConfigTarget
+		objReq     ctrl.Request
+	)
+
+	BeforeEach(func() {
+		vcsimCtx = suite.NewTestContextForVCSim(builder.VCSimTestConfig{})
+
+		Expect(vimv1.AddToScheme(vcsimCtx.Client.Scheme())).To(Succeed())
```
bryanv: Could add this to the `fake.NewScheme()` instead so it is just there

bryanv (approved):

controllers/configtarget/configtarget_controller.go
faisalabujabal: Good call — replaced both `MarkFalse(..., "%v", err)` call sites in `ReconcileNormal` (the `GetVirtualMachineConfigTarget` and `reconcileConfigOptions` error paths) with `pkgcond.MarkError(obj, vimv1.ReadyConditionType, reason, err)`. Fixed in bb6ad047.

controllers/configtarget/configtarget_controller_test.go
faisalabujabal: You're right, vimv1 is already registered in `test/builder.NewScheme()` (used by `NewTestContextForVCSim` under the hood), so this manual `AddToScheme` call was redundant. Removed in bb6ad047.

PR #1705, opened 2026-07-02
Reconcile VM's location when VM is moved outside of expected folder (…

anwithaun-1:

- validateVMFolder: A new method added to assess if the VM's active folder is the designated namespace folder or a valid child folder associated with a VirtualMachineSetResourcePolicy.
- validateVMResourcePool: Refactored to tightly restrict valid locations to explicit VKS child resource pools derived from active namespace resource policies.
- Integrated an API helper (vcenter.GetParentFolder) to navigate and verify the structural vSphere inventory folder paths.
- Refactored the reconcileLocation workflow in vmprovider_vm.go to evaluate both the resource pool (isVMInValidRP) and the inventory folder (isVMInValidFolder).
- Unit Tests and e2e tests are added

Fixes # vmop-3844

successfull Uts

The same review was merged  to main and the reverted because of pipeline failure
https://github.com/vmware-tanzu/vm-operator/pull/1630
https://github.com/vmware-tanzu/vm-operator/pull/1690

In this PR the e2e test is updated and ensured that it passes. Rest of the core changes remains same.

```release-note

```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -944,6 +944,7 @@ var VMUpdatePropertiesSelector = []string{
 	"runtime",
 	"snapshot",
 	"summary",
+	"parent",
 }
 
 func getReconcileErr(msg string, reconcileErr, err error) error {
@@ -964,11 +965,11 @@ func errOrReconcileErr(reconcileErr, err error) error {
 // updateVirtualMachine performs the following operations in the stated order:
 //
 //  1. Fetch properties
-//  2. Reconcile location
-//  3. Fetch recent tasks
-//  4. Fetch attached tags
-//  5. Fetch volume info
-//  6. Reconcile status
+//  2. Fetch recent tasks
+//  3. Fetch attached tags
+//  4. Fetch volume info
+//  5. Reconcile status
+//  6. Reconcile location
 //  7. Reconcile schema upgrade
 //  8. Reconcile backup state
 //  9. Reconcile snapshot revert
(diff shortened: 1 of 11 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1158,37 +1165,105 @@ func (vs *vSphereVMProvider) reconcileLocation(vmCtx pkgctx.VirtualMachineContex
...
+	isVMInValidRP, err := validateVMResourcePool(vmCtx, expectedRootRPMoID, vcClient)
 	if err != nil {
-		logger.Error(err, "failed to validate VM resource pool")
-		return err
+		return fmt.Errorf("failed to validate VM resource pool: %w", err)
 	}
-
-	// Handle Mismatch
-	if !isValid {
+	isVMInValidFolder, err := validateVMFolder(vmCtx, expectedFolder, vcClient)
```
aruneshpa, line 1172: So, both of these methods are doing their own prop collector calls to get the respective props of the VM.  We can club these in one call and fetch both the pool and the parent?

aruneshpa (approved):
LGTM.  Please make sure to have a good run of tests before submitting.  Thanks!

pkg/providers/vsphere/vmprovider_vm.go
anwithaun-1: you're right that each of these makes its own call to vCenter. It's technically possible to combine them into one call, but I'd like to leave them as two separate functions for now since it keeps the code simpler to read and test, and the extra call is small and only happens when the VM isn't in the default resource pool/folder (e.g. VKS worker nodes). Can revisit later if this ever shows up as a real performance issue.

PR #1707, opened 2026-07-06
Volume Status After Detaching All Volumes

anwithaun-1:

- Adds E2E coverage in test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go verifying that after detaching all non-boot volumes from a VirtualMachine, status.volumes correctly settles to contain only the boot disk entry, with no stale or duplicate entries left behind for the detached PVCs.
- The new test creates a VM with 3 PVC-backed volumes, waits for batch attachment, then patches spec.volumes to remove all of them. It then waits for the VirtualMachineHardwareVolumesVerified condition and asserts that status.volumes:
- Contains no entries for the detached PVC volume names.
- Has no duplicate entries for any volume name.
- Ends up with exactly the expected boot-disk entry/entries — accounting for both the "all disks are PVCs" capability enabled (backfilled boot volumes) and disabled (single classic boot disk) cases.

Fixes # vmop-3616

This is a test-only change (no pkg/api/controllers modifications) per e2e-sync-with-changes.md, adding regression coverage for status.volumes reconciliation after a full volume detach.

```release-note

```

### DIFF
--- a/test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go
@@ -472,7 +472,7 @@ func VMEncryptionSpec(ctx context.Context, inputGetter func() VMEncryptionInput)
 		Expect(cryptoStatus.Encrypted).To(ContainElement(vmopv1.VirtualMachineEncryptionTypeDisks))
 	})
 
-	It("Encrypt PVC using encryption class annotation on the PVC", Label("experimental"), func() {
+	It("Encrypt PVC using encryption class annotation on the PVC", Label("experimental"), FlakeAttempts(3), func() {
 		if !byokFSSEnabled {
 			Skip("BYOK FSS is not enabled")
 		}
--- a/test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go
@@ -472,7 +472,7 @@ func VMEncryptionSpec(ctx context.Context, inputGetter func() VMEncryptionInput)
 		Expect(cryptoStatus.Encrypted).To(ContainElement(vmopv1.VirtualMachineEncryptionTypeDisks))
 	})
 
-	It("Encrypt PVC using encryption class annotation on the PVC", Label("experimental"), FlakeAttempts(3), func() {
+	It("Encrypt PVC using encryption class annotation on the PVC", Label("experimental"), func() {
 		if !byokFSSEnabled {
 			Skip("BYOK FSS is not enabled")
 		}
--- a/test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
@@ -1323,7 +1323,7 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
 					volumeNames)
 			})
 
-			It("Detaching all non-boot volumes should leave only the boot disk in status.volumes", Label("extended-functional","experimental"), func() {
+			It("Detaching all non-boot volumes should leave only the boot disk(s) in status.volumes", Label("extended-functional", "experimental"), func() {
 				pvcs := createPvcsFromSpec(input, vmName, manifestbuilders.PVC{
 					StorageClassName: clusterResources.StorageClassName,
 				}, 3)
@@ -1340,25 +1340,52 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
 
 				Expect(clusterProxy.ApplyWithArgs(ctx, vmYaml)).To(Succeed(), "failed to apply virtualmachine")
 
-				pvcVolumeNames := make(map[string]bool, len(pvcs))
-				volumeNames := make([]string, len(pvcs))
-				for i, pvc := range pvcs {
-					volumeNames[i] = pvc.VolumeName
-					pvcVolumeNames[pvc.VolumeName] = true
+				// Names of the PVC-backed volumes we attach and later detach.
+				pvcVolumeNames := make(map[string]struct{}, len(pvcs))
+				for _, pvc := range pvcs {
+					pvcVolumeNames[pvc.VolumeName] = struct{}{}
 				}
-				// Backfilled volumes (e.g. the boot disk, when the "all disks are PVCs"
-				// capability is enabled) must remain attached; only the PVCs created
-				// above are detached below.
-				backfilledVolumes := getBackfilledVolumes(ctx, config, svClusterClient, vmSvcNamespace, vmName, allDisksArePVCapabilityEnabled)
-				volumeNames = append(volumeNames, backfilledVolumes...)
-				vm := waitForVMAndBatchAttach(ctx, config, svClusterClient, vmSvcNamespace, vmName, volumeNames)
+
+				By("Waiting for all disks to be synced via the VirtualMachineHardwareVolumesVerified condition")
+				vmoperator.WaitForVirtualMachineToExist(ctx, config, svClusterClient, vmSvcNamespace, vmName)
+				vmoperator.WaitOnVirtualMachineCondition(ctx, config, svClusterClient, vmSvcNamespace, vmName,
+					metav1.Condition{Type: "VirtualMachineHardwareVolumesVerified", Status: metav1.ConditionTrue})
+
+				// Once the volumes are synced, snapshot the boot-disk entries the VM
+				// reports: every status.volumes entry that is not one of the PVCs we
+				// attached. This keeps the test agnostic to how many boot disks the
+				// image has and how they are identified (no reliance on the
+				// "removable" flag or a single-boot-disk assumption), and confirms
+				// all added PVCs actually attached so the detach below is meaningful.
+				var vm *vmopv1.VirtualMachine
+				var bootDiskNames []string
+
+				By("Confirming the added PVCs are present in status.volumes before detaching")
+				Eventually(func(g Gomega) {
+					var err error
+					vm, err = utils.GetVirtualMachine(ctx, svClusterClient, vmSvcNamespace, vmName)
+					g.Expect(err).ToNot(HaveOccurred(), "failed to get VirtualMachine")
+
+					attachedPVCs := make(map[string]struct{}, len(pvcs))
+					bootDiskNames = nil
+					for _, vol := range vm.Status.Volumes {
+						if _, ok := pvcVolumeNames[vol.Name]; ok {
+							attachedPVCs[vol.Name] = struct{}{}
+						} else {
+							bootDiskNames = append(bootDiskNames, vol.Name)
+						}
+					}
+					g.Expect(attachedPVCs).To(HaveLen(len(pvcs)),
+						"expected all added PVCs to be present in status.volumes before detaching")
+				}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).
+					Should(Succeed(), "Timed out waiting for the added PVCs to appear in status.volumes")
 
 				By("Detaching all non-boot volumes from the Virtual Machine")
 
 				vmPatch := vm.DeepCopy()
 				vmPatch.Spec.Volumes = nil
 				for _, vol := range vm.Spec.Volumes {
-					if !pvcVolumeNames[vol.Name] {
+					if _, ok := pvcVolumeNames[vol.Name]; !ok {
 						vmPatch.Spec.Volumes = append(vmPatch.Spec.Volumes, vol)
 					}
 				}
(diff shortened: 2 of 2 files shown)


test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go
```diff
@@ -473,7 +473,7 @@ func VMEncryptionSpec(ctx context.Context, inputGetter func() VMEncryptionInput)
 		Expect(cryptoStatus.Encrypted).To(ContainElement(vmopv1a3.VirtualMachineEncryptionTypeDisks))
 	})
 
-	It("Encrypt PVC using encryption class annotation on the PVC", Label("experimental"), func() {
+	It("Encrypt PVC using encryption class annotation on the PVC", Label("experimental"), FlakeAttempts(3), func() {
```
aakashchan: Why is this randomly added? Can you add some documentation or comment regarding this?

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -1325,6 +1325,85 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+				By("Waiting on virtual machine conditions to become true after detach")
+				vmoperator.WaitOnVirtualMachineCondition(ctx, config, svClusterClient, vmSvcNamespace, vmName,
+					metav1.Condition{Type: "VirtualMachineHardwareVolumesVerified", Status: metav1.ConditionTrue})
+
+				By("Verifying status.volumes only contains the boot disk entry")
+				Eventually(func(g Gomega) {
+					vm, err := utils.GetVirtualMachineA5(ctx, svClusterClient, vmSvcNamespace, vmName)
+					g.Expect(err).ToNot(HaveOccurred(), "failed to get VirtualMachine")
+
+					seenCount := make(map[string]int, len(vm.Status.Volumes))
```
aakashchan:
```go
seenCount := make(map[string]int, len(vm.Status.Volumes))
for _, vol := range vm.Status.Volumes {
	seenCount[vol.Name]++
	g.Expect(pvcVolumeNames[vol.Name]).To(BeFalse(),
		"status.volumes has a stale entry for detached volume %q", vol.Name)
}
```
This looks redundant if the below HaveLen conditions are true. Equal length plus no-stale-entries implies no duplicates. 
```go
if allDisksArePVCapabilityEnabled {
  	g.Expect(vm.Status.Volumes).To(HaveLen(len(backfilledVolumes)),
  		"expected status.volumes to only contain the backfilled boot disk entries")
  	for _, name := range backfilledVolumes {
  		g.Expect(seenCount).To(HaveKey(name),
  			"expected backfilled boot volume %q in status.volumes", name)
  	}
  } else {
  	g.Expect(vm.Status.Volumes).To(HaveLen(1),
  		"expected status.volumes to only contain the boot disk entry")
  	g.Expect(vm.Status.Volumes[0].Type).To(Equal(vmopv1a5.VolumeTypeClassic),
  		"expected the remaining status.volumes entry to be the classic boot disk")
  }```

File: test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go (hunk 1 above)
Comment (anwithaun-1) [reply test/e2e/vmservice/vmservice/virtualmachine/vm_encryption.go]: This change is not related to the PR, It was added to avoid the vm_encryption flaky test failure observed while  running uts precheckin against the PR changes. I'll remove it from this PR.

File: test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go (hunk 2 above)
Comment (anwithaun-1) [reply test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go]: Fixed.

File: test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go [hunk 3]
```diff
@@ -1323,6 +1323,81 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
 					volumeNames)
 			})
 
+			It("Detaching all non-boot volumes should leave only the boot disk in status.volumes", Label("extended-functional","experimental"), func() {
+				pvcs := createPvcsFromSpec(input, vmName, manifestbuilders.PVC{
```
aruneshpa, line 1327: I am curious. Why 3 PVCs?  Just having 1 would validate what we want, no?
Worried that having 3 PVCs being bound and then attached to the VM may consume a few seconds and these start to add up slowly over time.

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -1323,6 +1323,81 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+					ImageName:        linuxVMIName,
+					VMClassName:      clusterResources.VMClassName,
+					StorageClassName: clusterResources.StorageClassName,
+					PVCs:             pvcs,
+				})
+				vmYamls = append(vmYamls, vmYaml)
+
+				Expect(clusterProxy.ApplyWithArgs(ctx, vmYaml)).To(Succeed(), "failed to apply virtualmachine")
+
+				pvcVolumeNames := make(map[string]bool, len(pvcs))
```
aruneshpa: Use `map[string]struct{}` as that is memory (and speed) optimized because doesn't allocate storage for the values since all you care about is "is key X present in the map".

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -1323,6 +1323,81 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+					if allDisksArePVCapabilityEnabled {
+						// Only the backfilled boot disk(s) should remain. ConsistOf is
+						// an exact set match, so it also enforces no stale entries for
+						// the detached PVCs and no duplicate entries.
+						g.Expect(statusVolNames).To(ConsistOf(backfilledVolumes),
+							"expected status.volumes to contain exactly the backfilled boot disk entries")
+					} else {
+						// Only the classic boot disk should remain. HaveLen(1) enforces
+						// no stale entries for the detached PVCs and no duplicates.
+						g.Expect(vm.Status.Volumes).To(HaveLen(1),
```
aruneshpa: Doesn't this assume that the image only contains a single boot disk?  What if there are more than one?

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -1323,6 +1323,81 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+				pvcVolumeNames := make(map[string]bool, len(pvcs))
+				volumeNames := make([]string, len(pvcs))
+				for i, pvc := range pvcs {
+					volumeNames[i] = pvc.VolumeName
+					pvcVolumeNames[pvc.VolumeName] = true
+				}
+				// Backfilled volumes (e.g. the boot disk, when the "all disks are PVCs"
+				// capability is enabled) must remain attached; only the PVCs created
+				// above are detached below.
+				backfilledVolumes := getBackfilledVolumes(ctx, config, svClusterClient, vmSvcNamespace, vmName, allDisksArePVCapabilityEnabled)
```
aruneshpa: Looking at that method, it seems like it is identifying a backfilled volume as "removable: true".  That's not correct.  Anyone can create a PVC with removable = false though today that's only backfilled disks.

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -1323,6 +1323,81 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+				Expect(clusterProxy.GetClient().Patch(ctx, vmPatch, ctrlclient.MergeFrom(vm))).
+					To(Succeed(), "failed to patch virtualmachine to detach all non-boot volumes")
+
+				By("Waiting on virtual machine conditions to become true after detach")
+				vmoperator.WaitOnVirtualMachineCondition(ctx, config, svClusterClient, vmSvcNamespace, vmName,
+					metav1.Condition{Type: "VirtualMachineHardwareVolumesVerified", Status: metav1.ConditionTrue})
+
+				By("Verifying status.volumes only contains the boot disk entry")
+				Eventually(func(g Gomega) {
+					vm, err := utils.GetVirtualMachineA5(ctx, svClusterClient, vmSvcNamespace, vmName)
```
aruneshpa: I think you need to rebase.  IIRC, this method has been dropped after BV's recent commit.

aruneshpa (commented):
My general comment is it seems like this test is making too many assumptions -- some of them are not your change but because you are using existing utility methods from our code.
For example: the fact that classic disks will have removable set to true.  That if AllDisksPVC is not enabled, boot disks will only be 1 and so on.

Can we simplify this test to:
- We created a VM with 1 additional PVC (on top of whatever came with the image)
- We waited for all disks to be ready / synced using the HardwareVolumesVerified condition
- We detached the PVC that we added
- We wait for the HardwareVolumesVerified condition
- We confirm that the status doesn't contain our PVC

Doesn't this test what we want?  And don't we already have a test for this?

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
anwithaun-1: Jira says detach all non-boot volumes and verify no duplicate entries. That's only meaningful with multiple volumes, so I kept 3 (consistant with above tests). One PVC would not exercise the "no duplicates / detach all" intent.

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
anwithaun-1: You're right. Removed the single-boot-disk assumption — I now snapshot the actual non-PVC (boot/backfilled) volumes from status.volumes before the detach and assert ConsistOf(bootDiskNames) afterward, so it works regardless of how many boot disks the image has while still catching stale/duplicate entries.

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
anwithaun-1: Updated the code to no longer classify disks by removable; we snapshot the actual non-PVC entries from status.volumes and assert ConsistOf(bootDiskNames). getBackfilledVolumes is now  removed

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
anwithaun-1: Done

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
anwithaun-1: Fixed

anwithaun-1:
> My general comment is it seems like this test is making too many assumptions -- some of them are not your change but because you are using existing utility methods from our code. For example: the fact that classic disks will have removable set to true. That if AllDisksPVC is not enabled, boot disks will only be 1 and so on.
> 
> Can we simplify this test to:
> 
> * We created a VM with 1 additional PVC (on top of whatever came with the image)
> * We waited for all disks to be ready / synced using the HardwareVolumesVerified condition
> * We detached the PVC that we added
> * We wait for the HardwareVolumesVerified condition
> * We confirm that the status doesn't contain our PVC
> 
> Doesn't this test what we want? And don't we already have a test for this?

- Agreed on the assumptions — removed them in the latest revision. No more removable or single-boot-disk reliance: I now snapshot the non-PVC entries from status.volumes before the detach and match against them, and I dropped the batch-attach helper in favor of the HardwareVolumesVerified condition (per your flow).
- On the 1-PVC version: it only covers the no-stale-entry check. Jira also wants no duplicates and only boot disks remaining after detaching all non-boot volumes — neither is possible with a single volume. So the only difference from your proposal is keeping multiple PVCs.
- On "do we already have this?" — the existing "Hot attach and detaching PVs" test detaches one PVC and re-attaches, asserting against the CnsNodeVMBatchAttachment CR, not status.volumes. 
- Switching to 1 PVC narrows the Jira's three checks (no-stale, no-duplicate, only-boot-disks-remain) down to just the first one. 
 
Happy to switch to the 1-PVC version if it's ok narrowing the Jira coverage to just confirming the detached PVC is gone from status.volumes — what do you prefer?
@aruneshpa

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
aruneshpa: I meant one _additional_ PVC, not total 1 PVCs.

anwithaun-1:
Closing this PR as the changes in this PR is not aligned with the updated jira requirement

PR #1711, opened 2026-07-06
✨ ConfigTarget: validation webhook + manifest wiring

faisalabujabal:

Adds the validating webhook for the cluster-scoped `ConfigTarget` CRD
(`vim.vmware.com/v1alpha1`), gated by the `VirtualMachineConfigPolicy`
capability, consistent with how `pkg/manager` already gates the
`ConfigTarget` scheme registration.

The webhook enforces:
* `spec.id` is immutable after create.
* `metadata.name` must be a valid vSphere cluster managed object ID
  (e.g. `domain-c21`).
* `spec.id` must not be empty.

New package: `webhooks/configtarget/validation`, following the same
shape as `webhooks/virtualmachinesetresourcepolicy` (webhook +
validator + unit/integration/suite tests). Wired into
`webhooks/webhooks.go` behind `Features.VirtualMachineConfigPolicy`.
Ran `make generate-manifests` to regenerate
`config/webhook/manifests.yaml` with the new webhook entry.

RBAC: `configtargets` already has `get`/`list` (added earlier by the
Zone controller for `create`/`patch`/`update`/`watch`), so
`config/rbac/role.yaml` did not change after regenerating manifests
with this webhook's `get;list` marker.

**Scope note:** the parent story's "wire the ClusterRole for
configtargets, virtualmachineconfigoptions, and hostsystems" and
"register the ConfigTarget controller in controllers/controllers.go"
items are intentionally left out of this PR. The `ConfigTarget`
reconciler and the `HostSystem` CRD (a stated dependency, S4.a) don't
exist on this branch yet and are tracked separately under
vmop-3758/vmop-3759/vmop-3760/vmop-3761. Adding write RBAC for
resources with no owning controller yet, or RBAC for a CRD that
doesn't exist, would be premature; that wiring belongs with the
controller's own PR.

Fixes vmop-3757

- Unit and integration tests both pass locally (`go test ./webhooks/configtarget/...`), gating the ConfigTarget scheme registration on `Features.VirtualMachineConfigPolicy` the same way `webhooks/virtualmachinegroup` gates on `Features.VMGroups`.
- `ConfigTarget` is cluster-scoped, so integration test object names are made unique per spec via an atomic counter to avoid collisions on the shared envtest API server.

```release-note
Add a validating webhook for the ConfigTarget resource.
```

### DIFF
--- a/webhooks/configtarget/validation/configtarget_validator.go
+++ b/webhooks/configtarget/validation/configtarget_validator.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
-	"regexp"
 
 	"k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -30,10 +29,6 @@ import (
 const (
 	webHookName = "default"
 )
-
-// clusterMoIDPattern matches the managed object ID vCenter assigns to a
-// ClusterComputeResource, e.g. domain-c21.
-var clusterMoIDPattern = regexp.MustCompile(`^domain-c[0-9]+$`)
 
 // +kubebuilder:webhook:verbs=create;update,path=/default-validate-vim-vmware-com-v1alpha1-configtarget,mutating=false,failurePolicy=fail,groups=vim.vmware.com,resources=configtargets,versions=v1alpha1,name=default.validating.configtarget.v1alpha1.vim.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
 // +kubebuilder:rbac:groups=vim.vmware.com,resources=configtargets,verbs=get;list
@@ -71,7 +66,7 @@ func (v validator) ValidateCreate(ctx *pkgctx.WebhookRequestContext) admission.R
 		return webhook.Errored(http.StatusBadRequest, err)
 	}
 
-	fieldErrs := append(v.validateName(configTarget), v.validateSpec(configTarget)...)
+	fieldErrs := v.validateSpec(configTarget)
 
 	validationErrs := make([]string, 0, len(fieldErrs))
 	for _, fieldErr := range fieldErrs {
@@ -106,22 +101,9 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
 	return common.BuildValidationResponse(ctx, nil, validationErrs, nil)
 }
 
-// validateName returns an error if metadata.name is not a valid vSphere
-// cluster managed object ID.
-func (v validator) validateName(configTarget *vimv1.ConfigTarget) field.ErrorList {
-	if clusterMoIDPattern.MatchString(configTarget.Name) {
-		return nil
-	}
-
-	return field.ErrorList{
-		field.Invalid(
-			field.NewPath("metadata", "name"),
-			configTarget.Name,
-			"must be a valid vSphere cluster managed object ID, e.g. domain-c21"),
-	}
-}
-
-// validateSpec returns an error if spec.id is empty.
+// validateSpec returns an error if spec.id is empty. The CRD's +required
+// marker on this field only enforces the JSON key's presence, not a
+// non-empty value, since ID has no omitempty tag.
 func (v validator) validateSpec(configTarget *vimv1.ConfigTarget) field.ErrorList {
 	if configTarget.Spec.ID.ID != "" {
 		return nil
--- a/webhooks/configtarget/validation/configtarget_validator.go
+++ b/webhooks/configtarget/validation/configtarget_validator.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"regexp"
 
 	"k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -29,6 +30,15 @@ import (
 const (
 	webHookName = "default"
 )
+
+// clusterMoIDPattern matches the conventional form of the managed object ID
+// vCenter assigns to a ClusterComputeResource, e.g. domain-c21. The vSphere
+// API documents ManagedObjectReference values as opaque strings, but this
+// prefix has been stable across vSphere releases, is how the Zone
+// controller derives ConfigTarget's name (controllers/infra/zone), and is
+// required by this webhook's task spec
+// (.sdd/specs/001-class-policy-resize/tasks.md, T060).
+var clusterMoIDPattern = regexp.MustCompile(`^domain-c[0-9]+$`)
 
 // +kubebuilder:webhook:verbs=create;update,path=/default-validate-vim-vmware-com-v1alpha1-configtarget,mutating=false,failurePolicy=fail,groups=vim.vmware.com,resources=configtargets,versions=v1alpha1,name=default.validating.configtarget.v1alpha1.vim.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
 // +kubebuilder:rbac:groups=vim.vmware.com,resources=configtargets,verbs=get;list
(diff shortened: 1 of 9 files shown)


webhooks/configtarget/validation/configtarget_validator.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/vmware-tanzu/vm-operator/webhooks/common"
+)
+
+const (
+	webHookName = "default"
+)
+
+// clusterMoIDPattern matches the managed object ID vCenter assigns to a
+// ClusterComputeResource, e.g. domain-c21.
+var clusterMoIDPattern = regexp.MustCompile(`^domain-c[0-9]+$`)
```
faisalabujabal, line 41: where did you get this pattern from?

webhooks/configtarget/validation/configtarget_validator.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+}
+
+// validateSpec returns an error if spec.id is empty.
+func (v validator) validateSpec(configTarget *vimv1.ConfigTarget) field.ErrorList {
+	if configTarget.Spec.ID.ID != "" {
+		return nil
+	}
+
+	return field.ErrorList{
```
faisalabujabal, line 137: is this even required if we already have `+required` annotation in the types file?

webhooks/configtarget/validation/configtarget_validator_intg_test.go
```diff
@@ -0,0 +1,167 @@
...
+		err = ctx.Client.Update(suite, ctx.configTarget)
+	})
+	AfterEach(func() {
+		Expect(ctrlclient.IgnoreNotFound(ctx.Client.Delete(ctx, ctx.configTarget))).To(Succeed())
+
+		err = nil
+		ctx = nil
+	})
+
+	When("update is performed with a changed spec.id", func() {
```
faisalabujabal, line 151: we need a testcase for a valid update

webhooks/configtarget/validation/configtarget_validator_unit_test.go
```diff
@@ -0,0 +1,235 @@
...
+		oldConfigTarget:                     oldConfigTarget,
+	}
+}
+
+func unitTestsValidateCreate() {
+	var (
+		ctx *unitValidatingWebhookContext
+	)
+
+	type createArgs struct {
```
faisalabujabal, line 112: instead of createargs taking booleans it should take the configtarget name, id, etc.. and the cases should pass those

webhooks/configtarget/validation/configtarget_validator_unit_test.go
```diff
@@ -0,0 +1,235 @@
...
+	)
+}
+
+func unitTestsValidateUpdate() {
+	var (
+		ctx *unitValidatingWebhookContext
+	)
+
+	type updateArgs struct {
+		changeID bool
```
faisalabujabal: instead of bools have it take the actual values it makes the describe table easier to read

webhooks/configtarget/validation/configtarget_validator.go
faisalabujabal: Fair callout — this pattern wasn't sourced from any documented vSphere/vCenter contract, just an assumption based on the `domain-c<N>` convention I saw in the Zone controller. Removed it; the webhook no longer validates `metadata.name` format.

webhooks/configtarget/validation/configtarget_validator.go
faisalabujabal: Checked the generated CRD: `+required` only adds `id`/`spec.id` to the OpenAPI `required:` list, which enforces the JSON key's *presence*, not a non-empty value (`ID string` has no `omitempty`, so it's always serialized). Confirmed this empirically — removing this check let an intg test create a ConfigTarget with an empty `spec.id.id` past the apiserver. Kept the check.

webhooks/configtarget/validation/configtarget_validator_intg_test.go
faisalabujabal: Added a valid (no-op) update test case.

webhooks/configtarget/validation/configtarget_validator_unit_test.go
faisalabujabal: Changed `createArgs` to hold a `*string` for the actual ID value under test instead of a boolean flag.

webhooks/configtarget/validation/configtarget_validator_unit_test.go
faisalabujabal: Changed `updateArgs` to hold the actual `vimv1.ManagedObjectID` value instead of a boolean flag.

webhooks/configtarget/validation/configtarget_validator.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.Response {
+	configTarget, err := v.configTargetFromUnstructured(ctx.Obj)
+	if err != nil {
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
+
+	oldConfigTarget, err := v.configTargetFromUnstructured(ctx.OldObj)
+	if err != nil {
```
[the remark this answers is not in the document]
faisalabujabal: Good catch. Fixed: `ValidateUpdate` now runs `validateSpec` alongside `validateAllowedChanges` and accumulates both. This wasn't exploitable today (clearing `spec.id` to empty on update was already rejected by the immutability check regardless of the new value), but the gap was real for defense-in-depth and future fields. Added a unit test asserting both messages appear together when `spec.id` is cleared on update.

webhooks/configtarget/validation/configtarget_validator.go
```diff
@@ -0,0 +1,134 @@
...
+	return common.BuildValidationResponse(ctx, nil, validationErrs, nil)
+}
+
+// validateSpec returns an error if spec.id is empty. The CRD's +required
+// marker on this field only enforces the JSON key's presence, not a
+// non-empty value, since ID has no omitempty tag.
+func (v validator) validateSpec(configTarget *vimv1.ConfigTarget) field.ErrorList {
+	if configTarget.Spec.ID.ID != "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
faisalabujabal: Left as-is. The early-return-nil pattern inside a single helper is fine here because accumulation across validators happens at the call site (`append(v.validateName(...), v.validateSpec(...)...)`), not inside each helper — that's the same pattern used elsewhere in this codebase (e.g. virtualmachinesnapshot_validator.go). Restructuring `validateSpec` internally to build a slice would just reintroduce the pattern the `prealloc` linter already flagged and I had to fix.

webhooks/configtarget/validation/configtarget_validator.go
faisalabujabal: Correction to my last reply: I dug further and this check *is* required — `.sdd/specs/001-class-policy-resize/tasks.md` (T060) explicitly scopes this webhook to validate "cluster MoID format for metadata.name", and `domain-c<N>` is the real, long-standing vSphere convention for a ClusterComputeResource's MoID (used elsewhere in this repo's docs, e.g. `docs/concepts/workloads/vm-group.md`, and hardcoded in govmomi's own type-prefix mapping). Note the vSphere Web Services SDK docs do call `ManagedObjectReference.value` an "opaque string" with no formal format guarantee, so I reworded the code comment to reflect that it's a stable convention rather than a hard vendor contract. Restored the check with that caveat.

webhooks/configtarget/validation/configtarget_validator_unit_test.go
```diff
@@ -0,0 +1,242 @@
...
+	var (
+		ctx *unitValidatingWebhookContext
+	)
+
+	type createArgs struct {
+		name *string
+		id   *string
+	}
+
+	validateCreate := func(args createArgs, expectedAllowed bool, expectedReason string) {
```
bryanv, line 117: If this type is going to get more complicated then prefer to move away from the args based tests since that quickly gets messy & unwieldy IMO

bryanv (approved):

webhooks/configtarget/validation/configtarget_validator_unit_test.go
faisalabujabal: it seems like the adopted pattern in the repo, so I'll keep it for now. if this grows we can evaluate the best path then

bryanv: Experience has shown that it becomes a mess

PR #1716, opened 2026-07-08
🐛 Remove VM ownerref from detached PVCs on delete

akutz:

Today, VM Operator sets an OwnerRef on every PVC it registers for an unmanaged disk, but never clears it. If a disk is detached from a VM before the VM is deleted, Kubernetes garbage collection still cascade-deletes that PVC along with the VM -- unlike a traditional vSphere VM, which only removes disks still attached at delete time.

Add pkg/util/kube/pvc.go's RemoveStaleVMOwnerRefFromPVCs, called from the VirtualMachine controller's ReconcileDelete immediately before the finalizer is removed. It lists PVCs owned by the VM via a new field index on PVC OwnerReferences, and for each PVC no longer listed in spec.volumes, surgically removes just that VM's OwnerReference entry -- leaving any other OwnerReferences and the rest of the PVC untouched.

A PVC can opt out of this cleanup via the new, presence-based vmoperator.vmware.com/keep-owner-ref annotation, for callers that still want a detached PVC garbage-collected along with its VM.

Add unit coverage in pkg/util/kube/pvc_test.go and the VirtualMachine controller's existing unit/envtest suites, plus an experimental E2E scenario in vm_hardware.go that hot-detaches unmanaged disks via govmomi and verifies both the default removal and the annotation opt-out. The E2E scenario is labeled "experimental" and has not yet been validated against real infrastructure.

Also add the spec (spec.md) and implementation plan (plan.md) under .sdd/specs/002-rm-pvc-owner-ref/, and a new persistentvolumeclaims RBAC marker the VirtualMachine controller previously lacked.

Fixes `NA`

```release-note
Remove VM owner ref on detached PVCs by default.
```

----
📚 Documentation preview 📚: https://vm-operator--1716.org.readthedocs.build/en/1716/

### DIFF
--- a/pkg/util/kube/pvc.go
+++ b/pkg/util/kube/pvc.go
@@ -1,0 +1,180 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package kube
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apierrorsutil "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/util/retry"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
+	pkgconst "github.com/vmware-tanzu/vm-operator/pkg/constants"
+)
+
+const (
+	// PVCOwnerRefVMUIDIndexKey is the field-index key used to efficiently
+	// list PersistentVolumeClaims by the UID of any VirtualMachine
+	// OwnerReference they carry. See IndexPVCByVMOwnerRef.
+	PVCOwnerRefVMUIDIndexKey = "metadata.ownerReferences.vmUID"
+
+	vmKind = "VirtualMachine"
+)
+
+// apiGroupNamePrefix is used when matching just the API group of a VM Op
+// object.
+const apiGroupNamePrefix = vmopv1.GroupName + "/"
+
+// PVCOwnerRefVMUIDIndexerFunc extracts the UIDs of any VirtualMachine
+// OwnerReferences from a PersistentVolumeClaim. It is used to populate the
+// field index registered by IndexPVCByVMOwnerRef.
+func PVCOwnerRefVMUIDIndexerFunc(obj ctrlclient.Object) []string {
+	pvc, ok := obj.(*corev1.PersistentVolumeClaim)
+	if !ok {
+		return nil
+	}
+
+	var uids []string
+	for _, ref := range pvc.OwnerReferences {
+
+		if ref.Kind == vmKind &&
+			strings.HasPrefix(ref.APIVersion, apiGroupNamePrefix) {
+
+			uids = append(uids, string(ref.UID))
+		}
+	}
+
+	return uids
+}
+
+// IndexPVCByVMOwnerRef registers a field index on PersistentVolumeClaim,
+// keyed by PVCOwnerRefVMUIDIndexKey, so RemoveStaleVMOwnerRefFromPVCs can
+// efficiently list the PVCs owned by a given VirtualMachine instead of
+// scanning every PVC in the namespace.
+func IndexPVCByVMOwnerRef(ctx context.Context, mgr manager.Manager) error {
+	return mgr.GetFieldIndexer().IndexField(
+		ctx,
+		&corev1.PersistentVolumeClaim{},
+		PVCOwnerRefVMUIDIndexKey,
+		PVCOwnerRefVMUIDIndexerFunc)
+}
+
+// RemoveStaleVMOwnerRefFromPVCs finds every PersistentVolumeClaim in vm's
+// namespace whose OwnerReferences includes vm, and removes that specific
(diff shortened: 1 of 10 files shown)


pkg/util/kube/pvc.go
```diff
@@ -0,0 +1,172 @@
...
+// field index registered by IndexPVCByVMOwnerRef.
+func PVCOwnerRefVMUIDIndexerFunc(obj ctrlclient.Object) []string {
+	pvc, ok := obj.(*corev1.PersistentVolumeClaim)
+	if !ok {
+		return nil
+	}
+
+	var uids []string
+	for _, ref := range pvc.OwnerReferences {
+		if ref.Kind == vmKind && ref.APIVersion == vmopv1.GroupVersion.String() {
```
bryanv: Instead of the entire APIVersion should this only match the group?

pkg/util/kube/pvc.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		if _, ok := pvc.Annotations[pkgconst.KeepOwnerRefAnnotationKey]; ok {
+			return nil
+		}
+
+		filteredRefs := make([]metav1.OwnerReference, 0, len(pvc.OwnerReferences))
+		changed := false
+		for _, ref := range pvc.OwnerReferences {
+			if ref.APIVersion == expOwnerRef.APIVersion &&
```
bryanv: Similar group comment here

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -2235,6 +2237,266 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+
+				keptPVCPatch := keptPVC.DeepCopy()
+				if keptPVCPatch.Annotations == nil {
+					keptPVCPatch.Annotations = map[string]string{}
+				}
+				keptPVCPatch.Annotations[pkgconst.KeepOwnerRefAnnotationKey] = ""
+				Expect(svClusterClient.Patch(ctx, keptPVCPatch, ctrlclient.MergeFrom(keptPVC))).
+					To(Succeed(), "Failed to annotate PVC %s", keptPVCName)
+
+				By("Removing both disks from the vCenter VM, keeping the underlying files (simulating a user detach)")
```
bryanv: Why from the VC VM?

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -2235,6 +2237,266 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+					got := getPVC(removedPVCName)
+					g.Expect(hasVMOwnerRef(got)).To(BeFalse(), "Expected VM OwnerRef to be removed from PVC %s", removedPVCName)
+				}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).
+					Should(Succeed())
+
+				By("Verifying the annotated detached PVC keeps its OwnerRef")
+
+				Consistently(func(g Gomega) {
+					got := getPVC(keptPVCName)
+					g.Expect(hasVMOwnerRef(got)).To(BeTrue(), "Expected VM OwnerRef to remain on annotated PVC %s", keptPVCName)
```
bryanv: With the k8s VM deleted, why would we not expect this to be GC?

akutz: Oh yeah, this is e2e, not envtest, there is a GC in play :)

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
akutz: Because I was being dumb.

akutz:
@bryanv I responded to your PR feedback.

pkg/util/kube/pvc.go
```diff
@@ -0,0 +1,180 @@
...
+	vm *vmopv1.VirtualMachine) error {
+
+	expOwnerRef := metav1.OwnerReference{
+		APIVersion: vmopv1.GroupVersion.String(),
+		Kind:       vmKind,
+		Name:       vm.Name,
+		UID:        vm.UID,
+	}
+
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
bryanv, line 180: Should we expect the patch to fail with a conflict if not patching with MergeFromWithOptimisticLock?

test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go
```diff
@@ -2235,6 +2237,263 @@ func VMHardwareSpec(ctx context.Context, inputGetter func() VMHardwareSpecInput)
...
+					err := svClusterClient.Get(ctx, ctrlclient.ObjectKey{
+						Namespace: vmSvcNamespace,
+						Name:      keptPVCName,
+					}, got)
+					g.Expect(apierrors.IsNotFound(err)).To(BeTrue(),
+						"Expected annotated PVC %s to be garbage collected once its owning VM was deleted", keptPVCName)
+				}, config.GetIntervals("default", "wait-virtual-machine-condition-update")...).
+					Should(Succeed())
+
+				By("Cleaning up the orphaned PVC")
```
bryanv: Ginkgo has a `DeferCleanup` for stuff like this

bryanv (approved):
Just a few comments

pkg/util/kube/pvc.go
akutz: Which patch?

akutz: I see which one you mean, but I am confused. A patch can fail for any number of reasons. Why would we not check the error?

akutz:
Merging after it passed the internal pre-check pipeline with the new, experimental test `pipeline/detail/6a985788eea6fea6b6b45bf1?workload_status=All`

PR #1724, opened 2026-07-10
Fix VC client session logout race and stale VM dir on retry

zhangheliu:
## What does this PR do, and why is it needed?

A `VirtualMachine` create could get stuck permanently retrying
`failed to create vm dir: ... already exists` after a routine rotation
of the VC service-account credentials Secret raced with an in-flight
create.

**Root cause:** `vSphereVMProvider.UpdateVcCreds`/`UpdateVcPNID` swapped
the cached vSphere client and immediately called `Logout()` on the old
one, even though reconciles that had already acquired that client via
`getVcClient` could still be using it mid-transaction. When a
credentials rotation landed between two calls on the same client (a VM
directory create followed by `CreateVM`), the session was killed out
from under the in-flight create. The subsequent failure-cleanup then
reused that same now-dead client and failed too, leaving the VM
directory orphaned. Because the directory path is deterministic
(derived from the VM's own object UID), every retry thereafter collided
with the same leftover directory forever.

**Commit 1** fixes the root cause: the client is now tracked as a
generation paired with a `WaitGroup` of outstanding callers.
`getVcClient` hands out a release func callers must call once done.
`UpdateVcCreds`/`UpdateVcPNID` swap in a new generation immediately for
future callers, but wait for the old generation's outstanding callers
to finish before logging it out. The async VM-create path hands its
release func to the background goroutine, since that goroutine (not
the spawning call) is what actually finishes using the client.
`VSphereClient`'s public signature changes to propagate the same
protection to its two callers (the VM watcher service and the image
cache controller), since both were also observed hitting this exact
race in the original incident.

**Commit 2** adds defense in depth, independent of the above: on a
`FileAlreadyExists` fault while creating a VM's own directory, clean up
the stale leftover and let the normal reconcile-retry path proceed,
instead of colliding with it indefinitely. This covers other causes of
a partial create that commit 1 doesn't address (e.g. a pod restart
mid-create, or an independent failure in the two-step namespace
cleanup).

Fixes #

- This is a draft — tests have intentionally not been added/updated yet.
- One existing test, `vmprovider_update_creds_test.go`'s "VC Client is
  logged out" case, asserts the old client's session is `nil`
  *immediately* after `UpdateVcCreds` returns. That's asserting the
  exact synchronous-logout behavior this PR removes, so it will need
  its assertions updated (not just the compile-fix already applied) in
  a follow-up test pass.
- In commit 2, I couldn't verify against a live vCenter or a raw fault
  payload exactly what format `FileAlreadyExists.File` comes back in
  (a fully-qualified `"[datastore] path"` string vs. a bare name).
  vcsim's `DatastoreNamespaceManager.CreateDirectory` simulator doesn't
  reproduce the random-UUID-remapping behavior observed in the real
  incident, so it couldn't settle this either. Handled both possible
  formats defensively rather than assuming one; flagging for review.

```release-note
Fix a race where a vSphere credentials/PNID update could log out a
vCenter session still in use by an in-flight VirtualMachine create,
which could permanently orphan the VM's directory and get it stuck
retrying "already exists" indefinitely.
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_vm.go
+++ b/pkg/providers/vsphere/vmprovider_vm.go
@@ -191,10 +191,20 @@ func (vs *vSphereVMProvider) createOrUpdateVirtualMachine(
 	)
 	ctx = vmCtx.Context
 
-	client, err := vs.getVcClient(ctx)
+	client, release, err := vs.getVcClient(ctx)
 	if err != nil {
 		return nil, err
 	}
+	// releaseClient tracks whether this function must release the vc client
+	// itself on return, or whether that responsibility has been handed off
+	// to the async create goroutine below (which releases it when the
+	// goroutine, and thus its use of client, is actually done).
+	releaseClient := true
+	defer func() {
+		if releaseClient {
+			release()
+		}
+	}()
 
 	// Set the VC UUID annotation on the VM before attempting creation or
 	// update. Among other things, the annotation facilitates differential
@@ -302,16 +312,20 @@ func (vs *vSphereVMProvider) createOrUpdateVirtualMachine(
 	copyOfCtx := vmCtx
 	copyOfCtx.VM = vmCtx.VM.DeepCopy()
 
-	// Start a goroutine to create the VM in the background.
+	// Start a goroutine to create the VM in the background. The goroutine
+	// takes over responsibility for releasing the vc client, since it, not
+	// this function, is what actually finishes using client.
 	chanErr := make(chan error)
 	go vs.createVirtualMachineAsync(
 		copyOfCtx,
 		client,
 		createArgs,
 		chanErr,
-		cleanupFn)
+		cleanupFn,
+		release)
 
 	asyncCreateStarted = true
+	releaseClient = false
 
 	// Return with the error channel. The VM will be re-enqueued once the create
 	// completes with success or failure.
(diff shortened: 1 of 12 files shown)


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2012,7 +2036,7 @@ func (vs *vSphereVMProvider) vmCreateGetSourceFilePaths(
 	}
 
 	var (
-		datacenterID = vs.vcClient.Datacenter().Reference().Value
+		datacenterID = vcClient.Datacenter().Reference().Value
```
zhangheliu, line 2039: Not just a mechanical rename to satisfy the compiler (`vs.vcClient` no longer exists as a field after this PR; it's now `vs.vcClientGen`, only touched under `vs.vcClientLock`).

This function already takes `vcClient *vcclient.Client` as a parameter — the exact client instance the caller obtained via `getVcClient()` for this create attempt, whose `release()` it's holding. The old code bypassed that parameter for this one line and read `vs.vcClient` directly instead, which was a latent bug even before this PR: if a creds/PNID update swapped the provider's client between the caller's `getVcClient()` call and this line running, it could read a different client instance than the one used everywhere else in this function. Using the `vcClient` parameter here makes this line consistent with the rest of the function.

bryanv (changes requested):
We need to fix the create and delete to be idempotent

zhangheliu:
> We need to fix the create and delete to be idempotent

will address it via https://github.com/vmware-tanzu/vm-operator/pull/1734

PR #1728, opened 2026-07-10
Fail open in collect-support-bundle.sh when testbed data is inaccessible

faisalabujabal:

`hack/e2e/collect-support-bundle.sh` previously exited 1 (hard failure) whenever the testbed blob URL/file was empty, unreachable, or the fetched `testbedInfo.json` did not contain usable vCenter credentials.

In the downstream UTS CI pipeline, this script runs after a testbed-provisioning workload whose completion gateway fires for *any* settled result, including `NOT_RUN` (e.g. when an upstream build dependency failed and provisioning itself never ran). In that case there is nothing to collect, but the script still hard-failed, marking support-bundle collection as a spurious failure on top of the real upstream failure.

This PR treats all three "there is nothing to collect" cases as a graceful no-op (`exit 0` with a warning) instead of a hard failure:
- an explicitly passed but empty `--testbed-blob-url`/`--testbed-info-json`
- a testbed blob URL or file that cannot be fetched/found
- a `testbedInfo.json` that is missing usable VC IP/credentials

A genuine invocation mistake (neither location flag passed at all, or an unknown argument) remains a hard usage error.

Fixes #

No behavior change for the happy path (valid, reachable testbed data) — only the three "testbed not accessible" failure paths change from `exit 1` to `exit 0` with a logged warning.

```release-note
`collect-support-bundle.sh` now exits successfully (instead of failing) when the testbed it would collect a support bundle from was never provisioned or is otherwise inaccessible.
```

### DIFF
--- a/hack/e2e/collect-support-bundle.sh
+++ b/hack/e2e/collect-support-bundle.sh
@@ -23,6 +23,7 @@ _err()  { echo "[${SCRIPT_NAME}] ERROR: $*" >&2; }
 TESTBED_INFO_JSON=""
 TESTBED_BLOB_URL=""
 OUTPUT_DIR="${RESULTSDIR:-/tmp/support-bundles}"
+_LOCATION_FLAG_GIVEN=false
 
 usage() {
     cat >&2 <<EOF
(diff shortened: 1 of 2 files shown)


hack/e2e/collect-support-bundle.sh
```diff
@@ -42,14 +43,24 @@ EOF
 
 while [[ $# -gt 0 ]]; do
     case "$1" in
-        --testbed-info-json) TESTBED_INFO_JSON="$2"; shift 2 ;;
-        --testbed-blob-url)  TESTBED_BLOB_URL="$2";  shift 2 ;;
+        --testbed-info-json) TESTBED_INFO_JSON="$2"; _LOCATION_FLAG_GIVEN=true; shift 2 ;;
```
bryanv, line 46: To double check this script is invoked with an explicit "" param and not that shift needs to change

bryanv (approved):

hack/e2e/collect-support-bundle.sh
faisalabujabal: Confirmed `_LOCATION_FLAG_GIVEN`/`shift 2` is safe here — every UTS invocation always passes the flag with an explicit (possibly empty) value, never omits it. E.g. `support-bundle.json` always calls `.../collect-support-bundle.sh --testbed-blob-url "${TESTBED_DATA_URL}"` with a quoted arg, so the value token is always present even when `TESTBED_DATA_URL` is empty. The only remaining hard-usage error is when neither flag is passed at all.

faisalabujabal:
Splitting the per-subtest syndrome truncation fix out into its own PR (#1757) since it's unrelated to this script's fail-open behavior change. Closing this PR — if the fail-open change is still wanted, happy to reopen or resubmit.

PR #1731, opened 2026-07-10
✨ Use spec.VMName for indexed snapshot lookup for a VM

aruneshpa:

`getVirtualMachineSnapshotsForVM` listed every VirtualMachineSnapshot in a VM's namespace and filtered in memory by Spec.VMName, an O(n) scan that grows with namespace-wide snapshot count. Register a field index on spec.vmName and query it with MatchingFields instead, so the lookup scales with matching snapshots rather than every snapshot in the namespace. The index key and extractor func live in pkg/util/kube so the controller's registration and the fake test client's index can't drift apart.

The `VMNameForSnapshotLabel` label is also kept in sync with `spec.vmName` end to end:

- The controller unconditionally re-asserts the label every
  reconcile, correcting a missing or stale value instead of only
  backfilling an absent one.
- The mutation webhook now backfills the label on update as well
  as create.
- The validator's lock-once-set check is replaced with a "label
  must match spec.vmName if present" rule on both create and
  update, since spec.vmName is already immutable. This allows
  correcting a bad label instead of freezing it forever, at the
  cost of also allowing the label to be removed (the controller
  re-adds it on the next reconcile).

vmop-3927.

```release-note
Use VMNameForSnapshotLabel for indexed snapshot lookup
```

### DIFF
--- a/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
+++ b/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -54,6 +55,16 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 
 		controllerNameShort = fmt.Sprintf("%s-controller", strings.ToLower(controlledTypeName))
 	)
+
+	// Index by Spec.VMName so per-VM snapshot lookups scale with matches,
+	// not with the total snapshots in a namespace.
+	if err := mgr.GetFieldIndexer().IndexField(
+		ctx,
+		&vmopv1.VirtualMachineSnapshot{},
+		kubeutil.VMSnapshotVMNameFieldIndex,
+		kubeutil.VMSnapshotVMNameIndexerFunc); err != nil {
+		return err
+	}
 
 	r := NewReconciler(
 		ctx,
@@ -164,11 +175,14 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) 
 	}
 
 	vmSnapshot := ctx.VirtualMachineSnapshot
-	ctx.Logger.Info("Fetching VirtualMachine from snapshot object")
 
 	if vmSnapshot.Spec.VMName == "" {
 		return ctrl.Result{}, errors.New("vmName is required")
 	}
+
+	ensureVMNameLabel(vmSnapshot)
+
+	ctx.Logger.Info("Fetching VirtualMachine from snapshot object")
 
 	vm := &vmopv1.VirtualMachine{}
 	objKey := client.ObjectKey{Name: vmSnapshot.Spec.VMName, Namespace: vmSnapshot.Namespace}
@@ -179,7 +193,7 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext) 
 	ctx.VM = vm
 
 	// The snapshot must be owned by a VM.  Set an owner reference to the VM.
-	if err := controllerutil.SetOwnerReference(ctx.VM, ctx.VirtualMachineSnapshot, r.Scheme()); err != nil {
+	if err := controllerutil.SetOwnerReference(ctx.VM, vmSnapshot, r.Scheme()); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to set owner reference to snapshot: %w", err)
 	}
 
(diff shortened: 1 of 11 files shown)


test/builder/fake.go
```diff
@@ -48,6 +48,16 @@ func NewFakeClientWithInterceptors(
 		WithInterceptorFuncs(funcs).
 		WithObjects(objs...).
 		WithStatusSubresource(KnownObjectTypes()...).
+		WithIndex(
```
bryanv, line 52: I'd prefer we don't add this to everything here, at least as it is now with duplicating the ID and mapping function. I know most of the existing spots don't do this but can we single shared const & impl in like kube/util? 

For an issue I have I need to clean all this up, and this one is a little out of the way and I think I might miss it.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -129,25 +130,37 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
+// validateVMNameLabelOnUpdate keeps the label consistent with spec.vmName.
+// Once set, it is locked, since it must always identify the same VM. Before
+// that, it may still be filled in, but only to the correct value.
+func (v validator) validateVMNameLabelOnUpdate(
 	vmSnapshot, oldVMSnapshot *vmopv1.VirtualMachineSnapshot) field.ErrorList {
 
-	var allErrs field.ErrorList
-	vmNameLabelPath := field.NewPath("metadata", "labels").Key(vmopv1.VMNameForSnapshotLabel)
+	oldVMNameLabel, oldOK := oldVMSnapshot.Labels[vmopv1.VMNameForSnapshotLabel]
+	if oldOK && oldVMNameLabel != "" {
```
bryanv: Could simplify a bit since `oldOK` isn't really needed

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
```diff
@@ -129,25 +130,37 @@ func (v validator) ValidateUpdate(ctx *pkgctx.WebhookRequestContext) admission.R
...
 
-	// Since this gets called only during an Update, the mutation
-	// webhook is always guaranteed to have set the Labels.
-	newVMNameLabel := vmSnapshot.Labels[vmopv1.VMNameForSnapshotLabel]
-	oldVMNameLabel := oldVMSnapshot.Labels[vmopv1.VMNameForSnapshotLabel]
+// vmNameLabelMismatchErr rejects a VMNameForSnapshotLabel that disagrees
+// with spec.vmName. Returns nil if the label is absent or matches.
+func vmNameLabelMismatchErr(vmSnapshot *vmopv1.VirtualMachineSnapshot) field.ErrorList {
+	label, ok := vmSnapshot.Labels[vmopv1.VMNameForSnapshotLabel]
+	if !ok || label == vmSnapshot.Spec.VMName {
```
bryanv, line 139: For `!ok` doesn't the mutation webhook set that?

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -164,12 +181,20 @@ func (r *Reconciler) ReconcileNormal(ctx *pkgctx.VirtualMachineSnapshotContext)
 	}
 
 	vmSnapshot := ctx.VirtualMachineSnapshot
```
bryanv, line 177: Later code still uses it via the `ctx` so it kind of confusing

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
```diff
@@ -302,6 +327,17 @@ func (r *Reconciler) syncVMSSnapshotTreeStatus(ctx *pkgctx.VirtualMachineSnapsho
 	return r.Client.Status().Patch(ctx, ctx.VM, mergePatch)
 }
 
+// ensureVMNameLabel backfills the VM name label from Spec.VMName if absent.
+// Spec.VMName is immutable once set, so this can never introduce a value
+// that later needs to change.
+func ensureVMNameLabel(vmSnapshot *vmopv1.VirtualMachineSnapshot) bool {
```
bryanv: Could an existing one ever have this set but it is invalid? I'm not sure why not just always set it, and is there really a need to return early since it is going to get patched anyways if different.

bryanv (approved):

controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go
aruneshpa: The label is important because it is used by VCFA UI to figure out snapshots that correspond to a VM - so if the object is missing the label somehow or if the label is incorrect, it makes sense to immediately patch the object.
Snapshot is a critical workflow and I would expect people to be impatient if their snaps don't show up on the UI.

webhooks/virtualmachinesnapshot/validation/virtualmachinesnapshot_validator.go
aruneshpa: It does.  I was just coding defensively for the scenario where the object somehow was created without the webhooks.

test/builder/fake.go
aruneshpa: Deduped the ID and mapping func.  We can also take this a step further by having a helper for the entire indexfield call, but I am not doing that in this PR.

bryanv (approved):

PR #1732, opened 2026-07-13
✨ Populate ConfigTarget device categories and MaxHardwareVersion

faisalabujabal:

`ConfigTarget.status` left every non-SR-IOV device category (CDROM, USB, PCI passthrough, vGPU, SCSI/IDE disks, VFlash, etc.) unmapped even though `QueryConfigTarget` already returns the data, and `status.maxHardwareVersion` was never set. This PR:

- Maps the 19 non-SR-IOV `ConfigTargetDevices` categories directly from the cluster-scope `QueryConfigTarget` result the controller already fetches (`controllers/configtarget/convert.go`).
- Computes `status.maxHardwareVersion` as the max hardware-version key among `QueryConfigOptionDescriptor` entries with `CreateSupported == true` -- data the controller already fetches for the `VirtualMachineConfigOptions` fan-out.

**Design note / correction to vmop-3759 & vmop-3797:** `HostConfigInfo` has no per-host "default hardware version" property -- no such field exists in govmomi's vendored types or public vSphere API docs. This replaces the per-host `PropertyCollector` design described in those tickets with an approach confirmed against vcsim's own `EnvironmentBrowser` simulator, which derives `maxHardwareVersion` the same way (max hardware version across hosts, surfaced via `QueryConfigOptionDescriptor.CreateSupported`). No host enumeration, no new provider method, and no bounded-concurrency helper are needed.

SR-IOV is intentionally excluded from this PR: both `ct.Sriov` and any `VirtualMachineSriovInfo` found inside the `PciPassthrough` union are skipped, since SR-IOV requires per-host attribution that is tracked separately in a new story, vmop-3926. Cross-cluster aggregation of `MaxHardwareVersion` is also out of scope here; that decision is deferred to vmop-3746.

Also updates the `001-class-policy-resize` spec/plan/tasks to reflect the revised design and record the previously-untracked device-mapping work (new tasks T066b/T066c).

Fixes vmop-3759, fixes vmop-3797 (see design-correction note above). Both tickets' non-SR-IOV scope is fully addressed by this PR; their remaining SR-IOV scope has moved to a new story, vmop-3926.

- Two device sub-fields are intentionally left unmapped, with inline comments explaining why: `VendorDeviceGroupInfo.ComponentDeviceInfo[].Device` (a `VirtualDevice` template) and `DVXClassInfo.ConfigParams` (nested `OptionDef`s) -- nothing in vm-operator reads either field today, so converting govmomi's full polymorphic device/option-type hierarchy for them is deferred until a real consumer needs it.
- No CRD/deepcopy regeneration needed -- all target fields already existed; this is pure controller-side population.
- E2E assertions were added (`test/e2e/vmservice/vmservice/configpolicy/configpolicy.go`) but not run against a real Supervisor cluster as part of this change; they're labeled `experimental` per `e2e-sync-with-changes.md` until verified on hardware.

```release-note
ConfigTarget.status now reports non-SR-IOV device inventory (CD-ROM, USB, PCI passthrough, vGPU, SCSI/IDE disks, VFlash, etc.) and maxHardwareVersion, computed from data the controller already queries.
```

**Testing Done**
Config policy E2E test suite
```
Ran 6 of 137 Specs in 194.916 seconds
SUCCESS! -- 6 Passed | 0 Failed | 5 Pending | 126 Skipped
PASS
```

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


.sdd/specs/001-class-policy-resize/plan.md
```diff
@@ -181,18 +181,21 @@ The only API change for this consolidation is additive on `ConfigTarget`:
 **Modify** `external/vim/api/v1alpha1/config_target_types.go` — add a new field on `ConfigTargetStatus`:
```
faisalabujabal, line 183: you are treating the plan and tasks documents as a communication file where you are putting in your thoughts and reasoning. while sometimes that might be correct for future agents to have the context the language and tone should match what the file entails. a tasks or a plan file is the document of record for the target state, so the tone of your wording should not be regarding why was something done for this PR but rather why was something done for the entire initiative.

controllers/configtarget/configtarget_controller.go
```diff
@@ -141,7 +141,8 @@ func (r *Reconciler) Reconcile(
 	}
 
 	defer func() {
-		if err := patchHelper.Patch(ctx, &obj); err != nil {
+		err := patchHelper.Patch(ctx, &obj)
```
faisalabujabal: this change is unnecessary this applies to all changes from an inline err that you have made in this diff. noinlineerr lint rule only applies to test/e2e

controllers/configtarget/configtarget_controller.go
```diff
@@ -198,15 +200,18 @@ func (r *Reconciler) ReconcileNormal(
...
 // populateStatus maps the vSphere QueryConfigTarget result onto the
-// ConfigTarget's status. Per this sub-task's scope, only numeric limits,
-// memory quantities, and security flags are populated; all
-// ConfigTargetDevices fields (including SR-IOV) and the per-host-derived
-// hardware version fields are left unset — see VMSVC-3759.
+// ConfigTarget's status: numeric limits, memory quantities, security flags,
+// and every ConfigTargetDevices category except SR-IOV. SR-IOV requires
+// per-host attribution (host MoID, DVX capabilities) that this cluster-scope
+// result does not carry; that per-host enrichment remains future work — see
+// VMSVC-3759/VMSVC-3797.
```
faisalabujabal: those should be `vmop-xxx` tickets

controllers/configtarget/configtarget_controller_test.go
```diff
@@ -152,11 +152,15 @@ func unitTests() {
 				Expect(got.Status.SupportedMaxMem).ToNot(BeNil())
 				Expect(got.Status.SupportedMaxMem.Value()).To(Equal(int64(8192) * 1024 * 1024))
 
-				// TODO(vmop-3759): devices (including SR-IOV) are left unset
-				// until per-host discovery lands.
+				// SR-IOV requires per-host attribution and is out of scope
```
faisalabujabal: we should have a task in the initiative for the remaining SR-IOV work that is linked here

controllers/configtarget/convert.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+}
+
+// convertPciPassthroughUnion type-switches the BaseVirtualMachinePciPassthroughInfo
+// union returned by QueryConfigTarget. VirtualMachineDynamicPassthroughInfo
+// does not implement this interface (it does not embed
+// VirtualMachinePciPassthroughInfo), so the only concrete types that can
+// appear here are *VirtualMachinePciPassthroughInfo and
+// *VirtualMachineSriovInfo (which embeds VirtualMachinePciPassthroughInfo).
+// The latter is dropped -- SR-IOV is out of scope here and is sourced
```
faisalabujabal: add the future task

controllers/configtarget/convert.go
```diff
@@ -0,0 +1,546 @@
...
+// separately via a future per-host path.
+func convertPciPassthroughUnion(
+	items []vimtypes.BaseVirtualMachinePciPassthroughInfo) []vimv1.VirtualMachinePCIPassthroughInfo {
+	var pciPassthrough []vimv1.VirtualMachinePCIPassthroughInfo
+
+	for _, item := range items {
+		switch v := item.(type) {
+		case *vimtypes.VirtualMachineSriovInfo:
+			// SR-IOV is out of scope; skip.
```
faisalabujabal: instead say is not supported yet

.sdd/specs/001-class-policy-resize/plan.md
faisalabujabal: (Response from Claude:) Fixed. Reworded the ComponentDeviceInfo.Device/DvxClassInfo.ConfigParams comments and the plan.md/tasks.md sections to state the target design and remaining work directly, without narrating what changed in this PR or why. Pushed in 7e6b0a41 and 95d3bb7f.

controllers/configtarget/configtarget_controller.go
faisalabujabal: (Response from Claude:) Confirmed and reverted -- noinlineerr doesn't fire under this repo's actual CI lint invocation (Makefile's lint-go uses --fast-only=true, which excludes noinlineerr; I verified 0 issues locally with that flag). Reverted every inline-err split I'd introduced back to the original `if err := ...; err != nil` form. Pushed in 7e6b0a41.

controllers/configtarget/configtarget_controller.go
faisalabujabal: (Response from Claude:) Fixed -- replaced VMSVC-xxx with vmop-xxx throughout the touched code comments, tasks.md, and plan.md. Pushed in 7e6b0a41.

controllers/configtarget/convert.go
faisalabujabal: (Response from Claude:) Added -- the future-task reference now points at vmop-3926 (the new SR-IOV story) and T121 (its controller-enrichment task), both here and on the sibling comment in the skip-case below.

controllers/configtarget/convert.go
faisalabujabal: (Response from Claude:) Changed to "is not supported yet" in both places (the doc comment and the skip case).

test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -146,6 +148,38 @@ func Spec(ctx context.Context, inputGetter func() SpecInput) {
...
+						g.Expect(svClusterClient.Get(ctx, ctrlclient.ObjectKey{Name: name}, ct)).To(Succeed())
+
+						g.Expect(ct.Status.MaxHardwareVersion).ToNot(BeEmpty(),
+							"ConfigTarget %q status.maxHardwareVersion should be populated", name)
+						_, err := vimtypes.ParseHardwareVersion(ct.Status.MaxHardwareVersion)
+						g.Expect(err).ToNot(HaveOccurred(),
+							"ConfigTarget %q status.maxHardwareVersion %q should be a valid hardware version",
+							name, ct.Status.MaxHardwareVersion)
+
+						// CDROM is asserted here because every ESX host reports
```
hpannem: would it be good to add these tests and skip if the device types are not there based on the cluster inventory?
Should we assert the devices baed on the clusters real inventory?
aka get cluster target devices via vc.
and assert with that here?

faisalabujabal: thanks for commenting I wanted to do that and I forgot before pushing

faisalabujabal: our testbeds are virtualized and don't have real hardware and env browser is returning nothing. i tried to use mock vgpu but wasn't able to make it work using the implemented methods needs more investigation. i created a task (vmop-3965) to try and make mock vgpu work and/or run the test against real hardware. 

the test now itself uses env browser to get the hardware from the testbed to assert against it

pkg/util/vsphere/configtarget/convert.go
```diff
@@ -0,0 +1,545 @@
+// © Broadcom. All Rights Reserved.
+// The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package configtarget
```
bryanv, line 5: If this isn't actually used in the provider, util/vsphere is probably a better location for it.

bryanv (approved):

pkg/util/vsphere/configtarget/convert.go
faisalabujabal: (Response from Claude:) Confirmed -- this package is only imported by controllers/configtarget, nothing under pkg/providers/vsphere uses it. Moved to pkg/util/vsphere/configtarget, alongside the other pure conversion/helper packages there (client, datastore, library, vm). Pushed in 431a3019.

PR #1741, opened 2026-07-16
🐛 Fix image name resolution in boot disk storage policy test

aakashchan:
The test "Should verify boot disk storage policy matches VM spec.storageClass after power-on" was using linuxImageDisplayName (the human-readable display name, e.g. "photon-5.0") instead of linuxVMIName (the resolved Kubernetes CR name, e.g. "vmi-abc123") when building the VM YAML.

This caused the mutating webhook's ResolveImageName function to search by status.name across both namespace-scoped and cluster-scoped VirtualMachineImages, finding one match in each scope. The webhook correctly rejects this ambiguous reference with "multiple VM images exist for 'photon-5.0' in namespace and cluster scope".

By using the resolved CR name (linuxVMIName), the webhook takes the direct lookup path (HasPrefix "vmi-") and skips the scope-search entirely, avoiding the ambiguity. This aligns with every other It block in the same test file which all correctly use linuxVMIName.

Changes:
- Use linuxVMIName instead of linuxImageDisplayName in VM YAML builder

Test changes:
- None (this fixes the broken test without requiring test infrastructure changes)

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
@@ -475,7 +475,7 @@ func VMSpec(ctx context.Context, inputGetter func() VMSpecInput) {
 		vmParameters := manifestbuilders.VirtualMachineYaml{
 			Namespace:        input.WCPNamespaceName,
 			Name:             vmName,
-			ImageName:        linuxImageDisplayName,
+			ImageName:        linuxVMIName,
 			VMClassName:      clusterResources.VMClassName,
 			StorageClassName: clusterResources.StorageClassName,
 			ResourcePolicy:   clusterResources.VMResourcePolicyName,


test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
```diff
@@ -475,7 +475,7 @@ func VMSpec(ctx context.Context, inputGetter func() VMSpecInput) {
 		vmParameters := manifestbuilders.VirtualMachineYaml{
 			Namespace:        input.WCPNamespaceName,
 			Name:             vmName,
-			ImageName:        linuxImageDisplayName,
+			ImageName:        linuxVMIName,
```
bryanv, line 478: This looks to be the only spot that deviated from the correct usage here. Still a little curious what is creating the VMCI for the same name in your env but this is correct regardless.

bryanv (approved):

PR #1745, opened 2026-07-17
🐛 Watch PVCs in VM controller

bryanv:

The VM controller should watch PVCs and reconcile the VM when a PVC that is referenced in the VM Spec.Volumes is changed. This is needed for more timely reconciliation of VM creation it a PVC is not bound.

The volume batch controller already had this watch, so reorganize a bit to better share that. While here, remove an old TODO in the batch controller to stop enqueuing based on the PVC OwnerRef the mapper covers what we need (and outside of PVCs created for CNS registration, PVCs generally won't have the VM as a owner anyways).

Fixes # vmop-3933

smoke, vm-hardware passed

```release-note
VirtualMachine controller now watches PVCs and will reconcile a VM when one of the PVCs referenced in its VM Spec.Volume changes.
```

### DIFF
--- a/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
+++ b/controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -75,6 +76,12 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 
 		controllerNameShort = fmt.Sprintf("%s-controller", strings.ToLower(controlledTypeName))
 	)
+
+	// Set up field index for VirtualMachine by ClaimName to efficiently query VMs
+	// referencing a PVC. This index is also used by the batch controller.
+	if err := kubeutil.IndexVMSpecVolumesPVCs(ctx, mgr); err != nil {
+		return err
+	}
 
 	proberManager, err := prober.AddToManager(ctx, mgr, ctx.VMProvider)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -164,27 +145,7 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
...
-				&vmopv1.VirtualMachine{},
-			),
-		)); err != nil {
-			return fmt.Errorf(
-				"failed to start VirtualMachine watch "+
-					"for PersistentVolumeClaim: %w", err)
-		}
-
-		// Watch for changes for PersistentVolumeClaim, and enqueue
 		// VirtualMachine that reference the PVC in their Spec.Volumes.
```
anwithaun-1, line 149: Nit- This comment line(No.184) looks truncated by the removal of the previous commented line (line no.183)

faisalabujabal (approved):

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -189,38 +195,43 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
...
+		builder = builder.Watches(
+			&vpcv1alpha1.SubnetPort{},
+			handler.EnqueueRequestForOwner(
+				mgr.GetScheme(),
+				mgr.GetRESTMapper(),
+				&vmopv1.VirtualMachine{}),
+		)
 	}
 
+	builder = builder.WatchesRawSource(source.Kind(
```
aruneshpa, line 231: I am curious, any reason for not using the "Watches"?

Sounds like they have the same effect?
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/builder/controller.go#L154-L158

aruneshpa (approved):

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
bryanv: They use different controller runtime methods to setup the controller: `NewControllerManagedBy()` vs `controller.New()`, which have different interfaces. I think the later is newer, and I think you can accomplish the same controller config with either.

PR #1688, opened 2026-06-26
 ✨ Add E2E tests for VM ExtraConfig reconciler

hpannem:
Adds test coverage for the `VirtualMachineAdvancedSpec` ExtraConfig reconciler introduced for the Telco VM Service API. The reconciler applies first-class VMX fields (`numa.vcpu.preferHT`, `sched.mem.lpage.enable1GPage`, `sched.swap.vmxSwapEnabled`, etc.) and user-defined bag keys to vSphere ExtraConfig, with deferred application for keys that require a power cycle or power off. Without this coverage, regressions in the reconciler or its condition-reporting logic would go undetected.

Changes:

- **`pkg/vmconfig/extraconfig/reconciler_test.go`** — Adds unit tests for the two previously uncovered paths in `Reconcile` and `OnResult`:
  - `PNUMANodeAffinity` change on a powered-on VM emits `vmx.reboot.powerCycle` and sets `PowerCyclePending`.
  - Clearing a PowerOff-mode key (`HugePages1GEnabled`) on a powered-on VM defers the change and sets `PowerOffRequired`.
  - `OnResult` correctly prioritises `PowerOffRequired` over `PowerCyclePending` when both are pending simultaneously.
  - `OnResult` leaves `status.ExtraConfig` unchanged when `moVM.Config` is nil.
  
 **Refactor to the code.**
  Calculation of the extraconfig status is moved to the reconcile_status() logic. So, refactored code so that both the reconciler and status checker use the same helpers to compare. 

- **`test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go`** — New E2E spec (`VMExtraConfigSpec`) gated on the `supports_telco_vm_service_api` supervisor capability. Nine `It` blocks cover:
  1. First-class fields and bag key CRUD — syncs immediately, status reflects changes.
  2. `PowerCyclePending` on a running VM — cleared after operator-driven power cycle.
  3. `PowerOffRequired` deferred key — absent from status while VM is on, applied after power-off.
  4. Kubernetes delete + recreate — all keys re-applied from spec on the new VM.
  5. Out-of-band vSphere power-off resolves `PowerCyclePending` and restores `PoweredOn`.
  6. Out-of-band vSphere destroy — operator recreates VM, all keys re-applied (detected via `status.UniqueID` change, not `BiosUUID`, since the operator preserves BiosUUID across recreations).
[...]

PR #1746, opened 2026-07-20
Fix vm-location e2e test failure

anwithaun-1:

- Restores the WCPNamespaceName field on the VM-LOCATION spec input in test/e2e/vmservice/vmservice_test.go.
- When the VM-LOCATION E2E spec was added, its input correctly passed WCPNamespaceName: wcpNamespaceName. The merge for [https://github.com/vmware-tanzu/vm-operator/pull/1688](url) mis-reconstructed this block and dropped that line from the VM-LOCATION input while adding the new VM-EXTRACONFIG context, leaving the field unset (defaulting to ""). Since it's an omitted struct field, nothing caught it at compile time.
- As a result, all three VM-LOCATION specs fail in their BeforeEach guard:
`Invalid argument. input.WCPNamespaceName can't be empty when calling`

Fixes # vmop-3844

```release-note

```


hpannem (approved):
Thanks for fixing this.

faisalabujabal (approved):

PR #1747, opened 2026-07-20
🌱 Add E2E coverage for VM suspend/resume across suspend modes

anwithaun-1:

- Adds E2E coverage for the Suspended VM power state, which previously had no test coverage in the VM lifecycle suite (virtualmachinelcm.go).
- Adds a new Context("when suspending a VM", ...) inside VMSpec that:
- Creates a powered-on VM and waits for it to obtain a guest IP (so VMware Tools is available for graceful suspend paths).
- Exercises suspend → verify Suspended → resume → verify PoweredOn via a DescribeTable with one Entry per supported spec.suspendMode (Hard, Soft, TrySoft).

Fixes # vmop-3579

```release-note

```

### DIFF
--- a/test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go
@@ -48,6 +48,7 @@ import (
 const (
 	poweredOnState  = "PoweredOn"
 	poweredOffState = "PoweredOff"
+	suspendedState  = "Suspended"
 )
 
 type VMSpecInput struct {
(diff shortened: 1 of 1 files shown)


aruneshpa:
Can you please update the PR title to more accurately reflect what it is trying to accomplish?

Also, the commit message doesn't have to call out the helper methods you are adding.  That is evident if someone looks at the code.  Just describe the what and the why.  Not the how.

And you *do not* have to call out what is NOT changed in the PR.  No need to call out "no API changes".

Thanks!

aruneshpa (approved):

PR #1755, opened 2026-07-21
e2e: fix flaky RegisterVM incremental-restore volume race

faisalabujabal:
## What this fixes

Two failure modes were observed in UTS CI, from the exact same commit under test, across consecutive runs:

1. `Incremental Restore - Register VM with pre-existing VM CR` — `Should register VM successfully` failing with:
   ```
   admission webhook "default.validating.virtualmachine.v1alpha6.vmoperator.vmware.com" denied the request:
   spec.volumes[0]: Forbidden: cannot remove volume with removable=false
   ```
2. `Incremental Restore - Register VM with pre-existing VM CR and PVCs` — `Should register VM successfully` failing with:
   ```
   Restored VM must have expected number of restored volumes, expected 2, got 1
   ```

**Update:** further investigation traced Fix 1 to a likely defect in
`vcf/tera`'s `RegisterVM` classic-disk handling
(`bora/vpx/wcp/wcpsvc/src/server/pkg/vmservice/registerpvc_helper.go`), not a pure test
gap. `getDisksToRegister` classifies a live disk as "classic" only if its current backing
filename matches what was recorded in the VM's ExtraConfig at backup time; anything that
doesn't match gets re-registered as a brand-new CNS volume (`restored-<random>`), which
silently drops the existing backfilled boot-disk entry from the patch. The validating
webhook is correctly blocking that drop — the bug is upstream, either in how/when the
classic-disk ExtraConfig key is populated relative to `RegisterVM` reading it (a race —
best fit, since both CI failures passed on immediate retry against the same build), or in
the filename-matching itself (tera's own code has a TODO acknowledging a related
mismatch scenario, not yet confirmed to be the same one hit here).

Given that, **Fix 1's `Removable=true` patch is a workaround for a real upstream issue,
not a fix**: no real customer would proactively mark their boot disk removable before an
incremental restore. It unblocks this test in the interim; it should not be read as
"this was a test-only gap." This should be filed as a product bug against tera and
this workaround revisited once that's resolved.

Fix 2 is unrelated to the above and stands on its own as a real test-timing bug.

### Fix 1 — `registervm.go`

The `Incremental Restore - Register VM with pre-existing VM CR` (no-PVC) context never sets `Removable=true` on the auto-backfilled boot disk volume before pausing/registering. When `RegisterVM`'s incremental restore replaces that backfilled volume entry with a freshly registered one, the validating webhook correctly blocks the removal since the existing entry wasn't marked removable. The `...and PVCs` variant of this same test already does this (`Add the pause annotation and set all volumes as removable`) — this change mirrors that patch into the no-PVC variant so both behave the same way pending the upstream fix (see note above).

### Fix 2 — `util.go` (`VerifyPostRegisterVM`)

The restored-volume-count assertion did a single `Get` immediately after the VM reached `PoweredOff`, but the volume-registration reconciler patches `spec.volumes` with the restored PVCs asynchronously after `RegisterVM`'s task reports success. A single read can race and observe a stale, partially-restored volume list. This wraps the check in `Eventually` (reusing the existing `wait-backup-to-complete` interval) instead of asserting once, matching the retry pattern already used elsewhere in this file (e.g. `waitForBackupToComplete`).

## Testing

`go build ./...` and `go vet ./...` pass for `test/e2e`. `gofmt -l` clean on both changed files.

## Proof of testing

Ran VI-ADMIN-RegisterVM

```
Ran 7 of 151 Specs in 1020.701 seconds
SUCCESS! -- 7 Passed | 0 Failed | 4 Pending | 140 Skipped
PASS
```

### DIFF
--- a/test/e2e/vmservice/vmservice/util.go
+++ b/test/e2e/vmservice/vmservice/util.go
@@ -1122,22 +1122,29 @@ func VerifyPostRegisterVM(
 	vmoperator.WaitForVirtualMachineToExist(ctx, config, svClusterClient, vmNamespace, vmName)
 	vmoperator.WaitForVirtualMachinePowerState(ctx, config, svClusterClient, vmNamespace, vmName, string(vmopv1.VirtualMachinePowerStateOff))
 
-	vm, err := utils.GetVirtualMachine(ctx, svClusterClient, vmNamespace, vmName)
-	Expect(err).ToNot(HaveOccurred())
-
 	By("Restored VM must have expected number of restored volumes")
 
+	// The volume-registration reconciler patches spec.volumes with the
+	// restored PVCs asynchronously after RegisterVM's task reports success,
+	// so a single read here can race and observe a stale, partially-restored
+	// volume list. Poll until it settles instead of asserting once.
 	actualRestoredPVCCount := 0
 
-	for _, vol := range vm.Spec.Volumes {
-		// Volume and PVC both contain "restored-" prefix, so validate that.
-		if vol.PersistentVolumeClaim != nil &&
-			strings.HasPrefix(vol.PersistentVolumeClaim.ClaimName, "restored-") {
-			actualRestoredPVCCount++
-		}
-	}
+	Eventually(func() int {
+		vm, err := utils.GetVirtualMachine(ctx, svClusterClient, vmNamespace, vmName)
+		Expect(err).ToNot(HaveOccurred())
 
-	Expect(actualRestoredPVCCount).To(Equal(expectedRestoredPVCCount), "Restored VM must have expected number of restored volumes, expected %d, got %d",
+		actualRestoredPVCCount = 0
+		for _, vol := range vm.Spec.Volumes {
+			// Volume and PVC both contain "restored-" prefix, so validate that.
+			if vol.PersistentVolumeClaim != nil &&
+				strings.HasPrefix(vol.PersistentVolumeClaim.ClaimName, "restored-") {
+				actualRestoredPVCCount++
+			}
+		}
+		return actualRestoredPVCCount
+	}, config.GetIntervals("default", "wait-backup-to-complete")...).Should(Equal(expectedRestoredPVCCount),
+		"Timed out waiting for restored VM to have expected number of restored volumes, expected %d, got %d",
 		expectedRestoredPVCCount, actualRestoredPVCCount)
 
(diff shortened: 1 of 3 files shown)


test/e2e/vmservice/vmservice/util.go
```diff
@@ -1122,22 +1122,29 @@ func VerifyPostRegisterVM(
...
-	for _, vol := range vm.Spec.Volumes {
-		// Volume and PVC both contain "restored-" prefix, so validate that.
-		if vol.PersistentVolumeClaim != nil &&
-			strings.HasPrefix(vol.PersistentVolumeClaim.ClaimName, "restored-") {
-			actualRestoredPVCCount++
-		}
-	}
+	Eventually(func() int {
+		vm, err := utils.GetVirtualMachine(ctx, svClusterClient, vmNamespace, vmName)
+		Expect(err).ToNot(HaveOccurred())
```
bryanv: Use the `func(g Gomega) { ... g.Expect()` pattern otherwise a failure here doesn't retry. And consider moving the assertion in here too.

faisalabujabal: thanks good catch

test/e2e/vmservice/vmservice/viadmin/registervm.go
```diff
@@ -296,12 +296,19 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
...
 
 			vm, err := utils.GetVirtualMachine(ctx, svClusterClient, input.WCPNamespaceName, vmName)
 			Expect(err).ToNot(HaveOccurred())
 			base := vm.DeepCopy()
 			metav1.SetMetaDataAnnotation(&vm.ObjectMeta, vmopv1.PauseAnnotation, trueString)
+			// RegisterVM's incremental restore replaces the auto-backfilled boot
+			// disk volume entry with a freshly registered one; the validating
+			// webhook blocks that swap unless Removable=true on the existing
+			// entry (mirrors the "...and PVCs" variant of this test below).
+			for i := range vm.Spec.Volumes {
```
aruneshpa, line 309: Trying to parse this.  Why is this required _now_?  The removable field has been added for a while and incremental restore has been working okay.

aruneshpa (approved):
Change LGTM.  Just a question on the timing.

PR #1762, opened 2026-07-22
🐛 Reject bus-sharing SCSI controllers/volumes on snapshotted VMs

faisalabujabal:

vSphere never allows a bus-sharing (Physical/Virtual) SCSI controller to be added to a VM that has a snapshot, and the reverse -- taking a snapshot of a VM with such a controller -- is also unsupported (`vim.fault.SharedBusControllerNotSupported`).

Without this check, adding a `MicrosoftWSFC` volume (which requires a Physical-sharing controller) to a VM that already has a snapshot was admitted by the `VirtualMachine` webhook. vm-operator then retried the failing `ReconfigVM_Task` in a tight loop forever, leaving the VM stuck in `PoweredOff`/`poweredOffReconfigure` indefinitely, with only a spammy `UpdateFailure` event and generic hardware-mismatch conditions to show for it.

This PR:
- **`webhooks/virtualmachine`**: rejects adding a bus-sharing controller, or a `MultiWriter` volume on a *dependent-mode* disk, to a VM that already has a snapshot. Independent-mode `MultiWriter` volumes (e.g. `OracleRAC`, which is always `IndependentPersistent`) are excluded from VM snapshots entirely by vSphere and are correctly still allowed.
- **`webhooks/virtualmachinesnapshot`**: the existing controller-sharing-mode check only inspected `spec.hardware.scsiControllers`; extends it to also reject creating a snapshot when a volume itself has a dependent-mode `MultiWriter` sharing mode.
- **`pkg/providers/vsphere/resources`**: classifies `vim.fault.SharedBusControllerNotSupported` using the same `fault.Is` idiom already used for `NotAuthenticated`/`InvalidLogin` in `pkg/util/vsphere/client`, and returns it as a `NoRequeueError` so the controller stops retrying a reconfigure that will fail identically forever. Also builds the full task error message via `taskutil.ErrorMessageFromTaskInfo` for all reconfigure failures, matching the existing pattern already used in `Customize()`.

This intentionally does *not* change how the reconfigure failure is surfaced as a Kubernetes Event/condition beyond stopping the retry storm -- the existing `HardwareControllersMismatch`/`HardwareVolumesMismatch` conditions are generic and don't distinguish a terminal failure from an in-progress reconcile, which is a separate, larger product decision (new condition reason/event semantics) intentionally left out of scope here.

```release-note
Reject admission of a bus-sharing SCSI controller or a dependent-mode MultiWriter volume on a VirtualMachine or VirtualMachineSnapshot that would conflict with an existing snapshot, and stop retrying a ReconfigVM_Task that vCenter permanently rejects for this reason (vim.fault.SharedBusControllerNotSupported).
```

### DIFF
--- a/pkg/providers/vsphere/resources/resources_suite_test.go
+++ b/pkg/providers/vsphere/resources/resources_suite_test.go
@@ -1,0 +1,19 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package resources_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	_ "github.com/vmware-tanzu/vm-operator/test/builder/log"
+)
+
+func TestResources(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "vSphere Resources Suite")
+}
--- a/pkg/providers/vsphere/resources/vm.go
+++ b/pkg/providers/vsphere/resources/vm.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/vmware/govmomi/fault"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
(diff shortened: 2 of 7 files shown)


faisalabujabal:
Closing and recreating from a differently-named branch/description — this one referenced an internal bug-tracker ID that shouldn't be exposed publicly.

PR #1763, opened 2026-07-22
🐛 Reject bus-sharing SCSI controllers/volumes on snapshotted VMs

faisalabujabal:

vSphere never allows a bus-sharing (Physical/Virtual) SCSI controller to be added to a VM that has a snapshot, and the reverse -- taking a snapshot of a VM with such a controller -- is also unsupported (`vim.fault.SharedBusControllerNotSupported`).

Without this check, adding a `MicrosoftWSFC` volume (which requires a Physical-sharing controller) to a VM that already has a snapshot was admitted by the `VirtualMachine` webhook. vm-operator then retried the failing `ReconfigVM_Task` in a tight loop forever, leaving the VM stuck in `PoweredOff`/`poweredOffReconfigure` indefinitely, with only a spammy `UpdateFailure` event and generic hardware-mismatch conditions to show for it.

This PR:
- **`webhooks/virtualmachine`**: rejects adding a bus-sharing controller, or a `MultiWriter` volume on a *dependent-mode* disk, to a VM that already has a snapshot. Independent-mode `MultiWriter` volumes (e.g. `OracleRAC`, which is always `IndependentPersistent`) are excluded from VM snapshots entirely by vSphere and are correctly still allowed.
- **`webhooks/virtualmachinesnapshot`**: the existing controller-sharing-mode check only inspected `spec.hardware.scsiControllers`; extends it to also reject creating a snapshot when a volume itself has a dependent-mode `MultiWriter` sharing mode.
- **`pkg/providers/vsphere/resources`**: classifies `vim.fault.SharedBusControllerNotSupported` using the same `fault.Is` idiom already used for `NotAuthenticated`/`InvalidLogin` in `pkg/util/vsphere/client`, and returns it as a `NoRequeueError` so the controller stops retrying a reconfigure that will fail identically forever. Also builds the full task error message via `taskutil.ErrorMessageFromTaskInfo` for all reconfigure failures, matching the existing pattern already used in `Customize()`.

Fixes #

This intentionally does *not* change how the reconfigure failure is surfaced as a Kubernetes Event/condition beyond stopping the retry storm -- the existing `HardwareControllersMismatch`/`HardwareVolumesMismatch` conditions are generic and don't distinguish a terminal failure from an in-progress reconcile, which is a separate, larger product decision (new condition reason/event semantics) intentionally left out of scope here.

```release-note
Reject admission of a bus-sharing SCSI controller or a dependent-mode MultiWriter volume on a VirtualMachine or VirtualMachineSnapshot that would conflict with an existing snapshot, and stop retrying a ReconfigVM_Task that vCenter permanently rejects for this reason (vim.fault.SharedBusControllerNotSupported).
```

### DIFF
--- a/pkg/providers/vsphere/resources/vm.go
+++ b/pkg/providers/vsphere/resources/vm.go
@@ -116,11 +116,15 @@ func (vm *VirtualMachine) Reconfigure(
 // vim.fault.SharedBusControllerNotSupported. vSphere never allows a
 // bus-sharing (Physical/Virtual) SCSI controller to be added to a VM that
 // has a snapshot, and the reverse -- taking a snapshot of a VM with such a
-// controller -- is also unsupported. This is a VM-wide, disk-mode-agnostic
-// restriction, distinct from the disk-level MultiWriter flag (see
-// vmopv1util.GetControllerSharingMode). Once a VM is in this state,
-// retrying the same reconfigure will fail identically forever, so callers
-// should stop requeuing instead of retrying in a tight loop.
+// controller -- is also unsupported:
+//   - https://knowledge.broadcom.com/external/article/314360
+//   - https://knowledge.broadcom.com/external/article/311074
+//
+// This is a VM-wide, disk-mode-agnostic restriction, distinct from the
+// disk-level MultiWriter flag (see vmopv1util.GetControllerSharingMode).
+// Once a VM is in this state, retrying the same reconfigure will fail
+// identically forever, so callers should stop requeuing instead of
+// retrying in a tight loop.
 func IsSharedBusControllerNotSupportedFault(err error) bool {
 	return fault.Is(err, &vimtypes.SharedBusControllerNotSupported{})
 }
--- a/pkg/providers/vsphere/resources/vm.go
+++ b/pkg/providers/vsphere/resources/vm.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/vmware/govmomi/fault"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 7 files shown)


pkg/providers/vsphere/resources/vm.go
```diff
@@ -91,12 +92,45 @@ func (vm *VirtualMachine) Reconfigure(
...
 	if err != nil {
+		// When a task fails, WaitForResult returns an error that contains the
+		// LocalizedMessage, but taskInfo.Error contains the full fault details.
+		// Extract a comprehensive error message that includes both the
+		// localized message and all fault messages, skipping it when it's
+		// identical to err.Error() (e.g. when the fault has no additional
+		// FaultMessage detail) to avoid duplicating the same text twice.
+		if errMsg := taskutil.ErrorMessageFromTaskInfo(taskInfo); errMsg != "" && errMsg != err.Error() {
+			err = fmt.Errorf("%s: %w", errMsg, err)
+		}
```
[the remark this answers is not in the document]
faisalabujabal: Fixed in e925121b — wrapped in `if taskInfo != nil`, matching `Customize()`. (Note: `taskutil.ErrorMessageFromTaskInfo` already nil-checks internally so this wasn't an actual panic risk, but the explicit guard matches the sibling method's style and is harmless either way.)

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1915,6 +1915,39 @@ func (v validator) validateSharedVolumesOrControllers(
...
+		// ReconfigVM_Task fails with vim.fault.SharedBusControllerNotSupported
+		// and vm-operator has no way to recover other than the snapshot
+		// being removed.
+		if controllerShared {
+			allErrs = append(allErrs, field.Invalid(
+				volPath.Child("controllerBusNumber"),
+				fmt.Sprintf("%s:%d", vol.ControllerType, *vol.ControllerBusNumber),
+				fmt.Sprintf("Controller with sharing mode %s is not supported for a VM that has a snapshot",
+					controllerSharingMode)))
+		}
```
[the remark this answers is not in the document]
faisalabujabal: Good catch, confirmed real: `controllerShared` can be true via `DefaultControllerSharingMode` (e.g. MicrosoftWSFC) before `validateControllerFields` has enforced `ControllerBusNumber` presence. Fixed in e925121b with a nil-safe bus-number string, plus a regression test that exercises this exact nil case (bypassing the test harness's controller auto-fill) to prove it denies instead of panicking.

webhooks/virtualmachine/validation/virtualmachine_validator_unit_test.go
```diff
@@ -10515,6 +10515,152 @@ func commonCreateAndUpdateValidations(
...
+										},
+									},
+								},
+								SharingMode: vmopv1.VolumeSharingModeMultiWriter,
+								DiskMode:    vmopv1.VolumeDiskModePersistent,
+							},
+						}
+					},
+					validate: doValidateWithMsg(
+						"MultiWriter disk sharing is not supported for a dependent-mode volume on a VM that has a snapshot",
```
bryanv: Please use the full message for these - it it nice to seem the field path in here to verify that too

faisalabujabal: Fixed in 92ffd03d — the three affected entries now assert the full field-path-prefixed message (e.g. `spec.volumes[0].sharingMode: Invalid value: "MultiWriter": ...`) instead of just the trailing detail string, so the field path is verified too.

webhooks/virtualmachine/validation/virtualmachine_validator.go
bryanv, line 1914: Maybe while here use `:=` since `controllerShared` isn't referenced until this point

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -1915,6 +1915,44 @@ func (v validator) validateSharedVolumesOrControllers(
...
+		// restriction: vSphere never allows a bus-sharing controller to be
+		// added to a VM that has a snapshot, regardless of disk mode
+		// (https://knowledge.broadcom.com/external/article/314360). The
+		// ReconfigVM_Task fails with vim.fault.SharedBusControllerNotSupported
+		// and vm-operator has no way to recover other than the snapshot
+		// being removed.
+		if controllerShared {
+			var busNumber string
+			if vol.ControllerBusNumber != nil {
+				busNumber = strconv.Itoa(int(*vol.ControllerBusNumber))
```
bryanv, line 1945: So right below this in the `pvcEncrypted` there is similar code that assumes `ControllerBusNumber` is set. The platform just has a complete stupid mess of what & how sharing is supported that we're forced to deal with. I don't like leaving the bus number empty in the message since that is pretty confusing: I think from `getControllerSharingMode()` unless it is a MicrosoftWSFC volume, the bus number had to be set to have `controllerShared` be true so maybe we can use that to make a better invalid error here.

PR #1766, opened 2026-07-23
🐛 Fix order-dependent VM publish duplicate-item test

bryanv:

In the original E2E tests, the "already exists in the content library" tests depended on the prior test to cause the dup. When the labels were added to the imported E2E tests, the prior test might not run so the dupe assertion would fail.

Fold the duplicate-item assertion into the smoke test itself, immediately after its own successful publish. It now publishes a second request against the item it just created and asserts TargetValid=False/TargetItemAlreadyExists on that second request, so the scenario no longer depends on state shared across specs via the content library.

Add some By()'s to the test to help track what it is up to.

Also simplify the CLS Content Library AfterEach to a single keepTargetLocationCLAttached guard, and switch the Inventory Content Library duplicate-publish cleanup to DeferCleanup instead of a Context-level AfterEach variable, matching the pattern used elsewhere in this file.

Fixes #

Locally can run entire publish suite and it passes, so does UTS smoke

```release-note
NONE
```

### DIFF
--- a/test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go
@@ -127,26 +127,28 @@ func VMPublishRequestSpec(ctx context.Context, inputGetter func() VMPublishReque
 			})
 
 			AfterEach(func() {
-				// Detach the content library from the namespace if attached and not keeping it attached.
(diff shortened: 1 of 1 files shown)


faisalabujabal (approved):
would help to add the proof of testing in the description or as a comment. whether a uts run or a log snippet if you ran it locally

PR #1773, opened 2026-07-27
:bug: Only watch BatchAttach CRD if the SharedDisks cap is enabled

aruneshpa:

We are unconditionally watching the batch attachment CRD which can cause issues due to the circular dependency between VM op and CSI for CRD installation.

Closes vmop-3981.

```release-note
Only watch BatchAttach CRD if the SharedDisks cap is enabled
```


brito-rafa (approved):
LGTM

bryanv (approved):

PR #1776, opened 2026-07-27
✨ Add host-local storage support for VM Service

divyenpatel:
A host-local storage policy requires that a volume be placed on a direct-attached VMFS datastore mounted on a single ESXi host, restricting any virtual machine using that volume to run strictly on that specific host. Previously, VM Service lacked awareness of this constraint, allowing VMs to be provisioned on hosts without access to their underlying disks and leading to power-on failures with the error: "Provided backing file is not accessible from host."

To resolve this, placement now passes the constraint directly to DRS rather than having VM Operator compute a host. For each bound PVC, its actual datastore path—resolved from the PV's volume handle via CNS—is attached to a placement-only disk within the ConfigSpec, enabling DRS to identify and return the single accessible host. VM Operator no longer derives, passes, or records host information; no annotations are read from or written to the VirtualMachine object, and PlaceVMForCreate retains its original signature. Host-local placement specifically utilizes PlaceVm, as testing showed PlaceVmsXCluster returns the same host regardless of the disks defined in the ConfigSpec.

Support is also included for WaitForFirstConsumer volumes. Because these unprovisioned volumes lack an initial disk path for DRS, DRS selects the target host first. The update workflow then sets that host's node on the PVC as selected-node, providing the target node for CSI provisioning. To prevent conflicting updates, both volume reconcilers skip applying zone stamps to host-local PVCs, ensuring a single writer for the field. Additionally, any PVC that is already stamped but not yet bound will requeue to avoid race conditions between binding and potentially conflicting host recommendations. VMs with multiple such volumes will have all of them co-located on the same host.

Host-local is detected from the storage policy's SPBM StorageLocality capability, surfaced as StoragePolicyStatus.HostLocal. This keeps policy-capability identification uniform with the existing StoragePolicy
CRD, which already exposes the policy's other capabilities the same way, rather than adding a separate StorageClass-specific mechanism.

This functionality is feature-gated under the HostLocalStorage capability. Note that placement performed via a VirtualMachineGroup does not enforce host constraints and is explicitly unsupported.

Fixes #

Manual testing result - https://gist.github.com/divyenpatel/1f839bbeac938ffb1d2c100e1ce679fe

```release-note
Add host-local storage support for VM Service
```

----
📚 Documentation preview 📚: https://vm-operator--1776.org.readthedocs.build/en/1776/


pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -2006,6 +2013,24 @@ func processPlacementResult(
...
+
+		hostFQDN, err := vcenter.GetESXHostFQDN(vmCtx, vcClient.VimClient(), hostMoID)
+		if err != nil {
+			return err
+		}
+
+		if vmCtx.VM.Annotations == nil {
+			vmCtx.VM.Annotations = map[string]string{}
+		}
+		vmCtx.VM.Annotations[constants.HostLocalSelectedNodeMOIDAnnotationKey] = hostMoID
```
akutz: 1. I am not sure we should be annotating VMs with the managed object ID of an ESX host.
2. This should not occur until after the VM is successfully created since it is possible to re-enter placement multiple times.

pkg/providers/vsphere/virtualmachine/devices.go
```diff
@@ -99,3 +101,31 @@ func CreateInstanceStorageDiskDevices(isVolumes []vmopv1.VirtualMachineVolume) [
...
 	return devices
 }
+
+// CreateHostLocalPlacementDiskDevices returns placement-only phantom disks,
+// one per given PVC, sized to the PVC's requested storage capacity. These
+// disks are never actually created — they exist only so DRS/SPBM's
+// placement math, driven by the disk's Profile, is forced to consider each
+// PVC's host-local storage policy when scoring candidate hosts. Unlike
+// instance storage disks, no magic VDiskId marker is needed.
+func CreateHostLocalPlacementDiskDevices(pvcs []corev1.PersistentVolumeClaim) []vimtypes.BaseVirtualDevice {
```
akutz: This should not be specific to host local disks. We should be doing this for _all_ PVCs that are listed as part of the VM spec during VM placement. Please refactor.

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
```diff
@@ -0,0 +1,144 @@
...
+// so the datastore returned is one the host can access.
+//
+// It is a no-op unless the HostLocalStorage capability is enabled and the VM
+// references at least one host-local StorageClass PVC.
+func resolveHostLocalPlacement(
+	vmCtx pkgctx.VirtualMachineContext,
+	k8sClient ctrlclient.Client,
+	storageData storage.VMStorageData) ([]corev1.PersistentVolumeClaim, error) {
+
+	if !pkgcfg.FromContext(vmCtx).Features.HostLocalStorage {
```
akutz: Feature/cap checks should NOT occur at the top of functions but instead from the place where the function is called. Otherwise it is still possible to cause an NPE when calling a guarded function. This is the pattern in the rest of this project. Please adhere to it.

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, nil
+	}
+
+	hostLocalPVCs := make([]corev1.PersistentVolumeClaim, 0, len(storageData.PVCs))
+	for _, pvc := range storageData.PVCs {
+		if pvc.Spec.StorageClassName == nil {
+			continue
```
akutz: Please refrain from using `continue` statements as they make testing more difficult.

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		hostLocalPVCs = append(hostLocalPVCs, pvc)
+	}
+
+	if len(hostLocalPVCs) == 0 {
+		return nil, nil
+	}
+
+	if vmCtx.VM.Annotations[constants.HostLocalSelectedNodeMOIDAnnotationKey] != "" {
+		// Host has already been resolved.
```
akutz: The host should be resolved each time until placement is complete unless there is only a single host that can satisfy the storage.

Also, why should the host need to be resolved at all? Per my other note regarding questioning if we should be annotating the VM with the host MoID, why not put that annotation on the PVC instead?

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
```diff
@@ -0,0 +1,144 @@
+// © Broadcom. All Rights Reserved.
```
akutz: Please rename the file `vmprovider_vm_hostlocal_storage.go` to make it clear this is about storage.

Also, I'm not sure we *want* to support host-local storage as it breaks mobility. 

cc @srajashekar-vmw

akutz (changes requested):
A *very* cursory review, there will be more.

pkg/providers/vsphere/vmprovider_vm.go
divyenpatel: The VM MoID annotation is removed entirely — VM Operator no longer records anything about the host on the VirtualMachine object, at any point in its lifecycle. The host is derived fresh from the VM's PVCs on every reconcile (resolveHostLocalStorage), and the outcome (once the VM is actually running on a host) is published to the PVC, not the VM, via reconcileHostLocalStorage. A failed create is therefore free to land on a different host on the next attempt, with nothing to unwind.

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
divyenpatel: Addressed. The HostLocalStorage capability check moved to each call site (vmCreateDoPlacement, reconcileHostLocalStorage's caller); the resolver functions themselves no longer guard on it internally.

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
divyenpatel: rewritten as guard conditions

pkg/providers/vsphere/virtualmachine/devices.go
divyenpatel: Done.  AddPVCPlacementDisks now adds a placement-only disk for every PVC on the VM (not just host-local, not just pending)

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
divyenpatel: The host is now resolved from scratch on every reconcile rather than cached anywhere — there is no VM-side state to go stale.  The decision that does need to persist (which host a pending WaitForFirstConsumer volume must provision on) is published to the PVC's own volume.kubernetes.io/selected-node.

We also investigated whether the annotation-based host resolution itself could be replaced by supplying the PVC's already-provisioned disk to placement and letting DRS infer the host from its real datastore backing. 

Verified against real DRS that this doesn't work — PlaceVm returns no host recommendation for a disk with real backing, and PlaceVmsXCluster recommends a host that can't even see that datastore. 

PlacementSpec.Datastores does constrain DRS correctly and may be worth adopting in a follow-up -  Findings recorded in research.md

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
divyenpatel: done. vmprovider_vm_hostlocal.go → vmprovider_vm_hostlocal_storage.go

controllers/virtualmachine/volumebatch/volumebatch_controller.go
```diff
@@ -973,6 +974,18 @@ func (r *Reconciler) handlePVCWithWFFC(
 		return errors.New("PVC with WFFC storage class support is not enabled")
 	}
 
+	if pvc.Annotations == nil {
```
divyenpatel: move back to original place

pkg/providers/vsphere/placement/cluster_placement.go
```diff
@@ -181,16 +181,28 @@ func CloneVMRelocateSpec(
...
+// what allows a VM that is pinned to a specific host - host-local storage - to
+// still obtain the datastore recommendation that the fast deploy feature
+// requires. Note PlaceVmsXCluster cannot express this constraint for a create,
+// which is why the pinned case uses this API instead.
 func PlaceVMForCreate(
 	vmCtx pkgctx.VirtualMachineContext,
 	cluster *object.ClusterComputeResource,
-	configSpec vimtypes.VirtualMachineConfigSpec) (*Recommendation, error) {
+	configSpec vimtypes.VirtualMachineConfigSpec,
+	hostMoRef *vimtypes.ManagedObjectReference) (*Recommendation, error) {
```
divyenpatel: See if we can use configSpec only for PlaceVMForCreate, and remove hostMoRef

pkg/providers/vsphere/vmprovider_vm_hostlocal.go
divyenpatel: updated logic. see latest commit.

pkg/util/kube/storage.go
```diff
@@ -62,20 +62,74 @@ func getPVCAccessibleZones(pvc corev1.PersistentVolumeClaim) sets.Set[string] {
...
+// unreadable StoragePolicy is returned as an error rather than as false,
+// because falsely reporting a host-local policy as ordinary storage would let
+// the VM be placed on a host that cannot reach its disks — the exact failure
+// this is meant to prevent. Callers should propagate the error and retry.
+func IsHostLocalStorageProfile(
+	ctx context.Context,
+	k8sClient ctrlclient.Client,
+	profileID string) (bool, error) {
+
+	if profileID == "" {
```
akutz, line 90: Why/when would `profileID` be empty? And if you are checking argument values, should you do the same for `ctx` and `k8sClient`. Should the incorrect values return a no-op, error, or panic? In other functions in the code base we panic.

pkg/util/kube/storage.go
```diff
@@ -62,20 +62,74 @@ func getPVCAccessibleZones(pvc corev1.PersistentVolumeClaim) sets.Set[string] {
...
+		return false, nil
+	}
+
+	name := GetStoragePolicyObjectName(profileID)
+	if name == "" {
+		return false, fmt.Errorf(
+			"was not able to construct a storage policy name for profile ID %s",
```
akutz: Please change to `failed to construct storage policy name for profile ID %q`.

* We like to start errors with `failed` or something that makes it clear it is a failure
* Use the `%q` to ensure the string is quoted

pkg/util/kube/storage.go
```diff
@@ -62,20 +62,74 @@ func getPVCAccessibleZones(pvc corev1.PersistentVolumeClaim) sets.Set[string] {
...
+	var obj infrav1.StoragePolicy
+	if err := k8sClient.Get(
+		ctx,
+		ctrlclient.ObjectKey{
+			Namespace: pkgcfg.FromContext(ctx).PodNamespace,
+			Name:      name,
+		},
+		&obj); err != nil {
+
+		return false, fmt.Errorf(
```
akutz, line 110: I am remarking on this because it is a PERFECT example of error strings! Great job!

pkg/util/kube/storage.go
```diff
@@ -62,20 +62,74 @@ func getPVCAccessibleZones(pvc corev1.PersistentVolumeClaim) sets.Set[string] {
...
+			"failed to get StoragePolicy %q for profile ID %s: %w",
+			name, profileID, err)
+	}
+
+	return obj.Status.HostLocal, nil
+}
+
+// HasVirtualMachineDataSourceRef returns true if the given PVC's data source is
+// the VirtualMachine itself, meaning the volume is one of the VM's own disks.
```
akutz: This comment is misleading. The function only verifies the DS ref is a VM, not the VM itself (meaning a specific VM).

pkg/util/vsphere/storage/storage_policy.go
```diff
@@ -22,6 +22,73 @@ import (
...
+
+	if subprofiles == nil {
+		return false
+	}
+	for _, subprofile := range subprofiles.SubProfiles {
+		for _, capIns := range subprofile.Capability {
+			if capIns.Id.Namespace != volumeAllocationNamespace ||
+				capIns.Id.Id != storageLocalityID {
+
+				continue
```
akutz: I *really* prefer to avoid the use of `continue` as it makes code coverage much more difficult. Please rewrite without the use of `continue`.

pkg/util/vsphere/storage/storage_policy.go
```diff
@@ -22,6 +22,73 @@ import (
...
+// NOT be treated as host-local.
+// This function does NOT match vSAN Direct or vSAN locality/SNA policies.
+func IsHostLocalStorageCapabilityPolicy(
+	subprofiles *pbmtypes.PbmCapabilitySubProfileConstraints) bool {
+
+	if subprofiles == nil {
+		return false
+	}
+	for _, subprofile := range subprofiles.SubProfiles {
+		for _, capIns := range subprofile.Capability {
```
akutz: Just curious, what is the `Ins` in `capIns`? Why not just `cap`?

pkg/util/vsphere/storage/storage_policy.go
```diff
@@ -22,6 +22,73 @@ import (
...
+		}
+	}
+	return false
+}
+
+// storageLocalityValueIsHostLocal reports whether a StorageLocality property
+// value denotes host-local storage. The value is normally a plain string, but
+// SPBM may express it as a discrete set of allowed values, in which case any
+// member matching is sufficient.
+func storageLocalityValueIsHostLocal(value vimtypes.AnyType) bool {
```
akutz, line 86: This can be simplified:

```golang
func storageLocalityValueIsHostLocal(value vimtypes.AnyType) bool {
	switch v := value.(type) {
	case string:
		return v == storageLocalityHostLocal
	case pbmtypes.PbmCapabilityDiscreteSet:
		for _, item := range v.Values {
			if s, ok := item.(string); ok && s == storageLocalityHostLocal {
				return true
			}
		}
	case *pbmtypes.PbmCapabilityDiscreteSet:
		if v != nil {
			return storageLocalityValueIsHostLocal(*v)
		}
	}
	return false
}
```

pkg/providers/vsphere/vmprovider_vmgroup.go
```diff
@@ -250,6 +250,17 @@ func (vs *vSphereVMProvider) vmGroupGetVMPlacementConfigSpec(
 			return nil, err
 		}
 
+		// The VM's PVCs are included so that the recommendation accounts for
+		// their storage policies. Note that pinning a VM to a single host for
```
akutz: How do we know this is not supported? Did you check with the DRS team? Did we file an internal tracking issue as an RFE for them to add this support in the future?

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
```diff
@@ -0,0 +1,455 @@
...
+	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/constants"
+	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/storage"
+	pkgutil "github.com/vmware-tanzu/vm-operator/pkg/util"
+	kubeutil "github.com/vmware-tanzu/vm-operator/pkg/util/kube"
+	"github.com/vmware-tanzu/vm-operator/pkg/util/ptr"
+)
+
+// hostLocalPVCBindRequeueDelay is how long to wait before re-checking a
+// host-local PVC that has been given a host but is not Bound yet.
+const hostLocalPVCBindRequeueDelay = 10 * time.Second
```
akutz: Is there a reason you did not write these as:

```golang
const (
	// hostLocalPVCBindRequeueDelay is how long to wait before re-checking a
	// host-local PVC that has been given a host but is not Bound yet.
	hostLocalPVCBindRequeueDelay = 10 * time.Second

	// pvcPlacementStartDeviceKey is the starting key for the placement-only PVC
	// disks. A negative range is traditionally used, and this is kept clear of the
	// ranges used for PCI passthrough and instance storage devices.
	pvcPlacementStartDeviceKey = int32(-1000)

	// esxHostMoIDNodeAnnotationKey is the annotation a Supervisor Node object
	// carries naming the MOID of the ESXi host it corresponds to.
	esxHostMoIDNodeAnnotationKey = "vmware-system-esxi-node-moid"
)
```

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
```diff
@@ -0,0 +1,455 @@
...
+
+// pvcPlacementStartDeviceKey is the starting key for the placement-only PVC
+// disks. A negative range is traditionally used, and this is kept clear of the
+// ranges used for PCI passthrough and instance storage devices.
+const pvcPlacementStartDeviceKey = int32(-1000)
```
akutz: I am not sure you should count on `-1000` being safe as devices are added elsewhere. You may want to find the smallest number and add `1000` to that and then start counting.

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
```diff
@@ -0,0 +1,455 @@
...
+		if scName := pvc.Spec.StorageClassName; scName != nil {
+			policyID = storageData.StorageClassToPolicyID[*scName]
+		}
+
+		backing := &vimtypes.VirtualDiskFlatVer2BackingInfo{
+			ThinProvisioned: ptr.To(false),
+		}
+
+		// A provisioned volume is named by its path and is not created.
+		fileOp := vimtypes.VirtualDeviceConfigSpecFileOperationCreate
```
akutz: The `fileOp` should not be `Create`. This is incorrect for disks that already exist. You can see that at https://github.com/vmware-tanzu/vm-operator/blob/a3cd2ac4b9ef7d4bbdf34ac540c81be676eef0c2/pkg/providers/vsphere/vmlifecycle/create_fastdeploy.go#L488-L501.

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
```diff
@@ -0,0 +1,455 @@
...
+		if p := diskPaths[pvc.Name]; p != "" {
```
akutz: Ah, nevermind. You handle that here. I am wondering then why you default to `Create`? Is this for WffC volumes?

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
```diff
@@ -0,0 +1,455 @@
+// © Broadcom. All Rights Reserved.
```
akutz, line 1: I think this filename should be renamed to `vmprovider_vm_pvc_placement.go` as the primary function, `AddPVCPlacementDisks`, is not specific to host-local storage.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1791,6 +1830,21 @@ func (vs *vSphereVMProvider) vmCreateDoPlacement(
...
+	// Naming each provisioned volume's disk by its real datastore path is what
+	// lets DRS work out the host, so no host or datastore has to be passed to
+	// placement.
+	diskPaths, err := pvcDiskPaths(
+		vmCtx, vcClient.VimClient(), vs.k8sClient, createArgs.Storage.PVCs)
+	if err != nil {
+		return err
+	}
+
+	if err := AddPVCPlacementDisks(
```
akutz: Is there a reason we are not calling this from inside of https://github.com/vmware-tanzu/vm-operator/blob/a3cd2ac4b9ef7d4bbdf34ac540c81be676eef0c2/pkg/providers/vsphere/virtualmachine/configspec.go#L169-L336? Because that function already ends with the `EnsureDisksHaveControllers` call.

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
```diff
@@ -0,0 +1,455 @@
...
+
+// reconcileHostLocalStorage publishes the ESXi host that the VM actually runs
+// on to any of its host-local PVCs that are not yet provisioned, so that CNS
+// provisions their volumes on that same host and the VM can reach them.
+//
+// This is deliberately driven by where the VM really is rather than by the
+// host placement recommended, so that nothing is committed to before the VM
+// exists, and so that a failed attempt is simply retried on the next
+// reconcile.
+func (vs *vSphereVMProvider) reconcileHostLocalStorage(
```
akutz, line 110: Is there not going to be a race that means the VM could be migrated from the initial host prior to the PVC being provisioned on host-local storage, which would effectively tie the VM to said host?

akutz (changes requested):

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
bryanv: PVCs that reference a VM are watched so a delay/retry should not be needed

pkg/util/kube/storage.go
divyenpatel: Agreed, now panics on nil ctx, nil k8sClient, and empty profileID.

pkg/util/kube/storage.go
divyenpatel: Done: `failed to construct storage policy name for profile ID %q.` Applied `%q` to the profile ID in the neighbouring error too.

pkg/util/kube/storage.go
divyenpatel: updated comment to say

```
// HasVirtualMachineDataSourceRef returns true if the given PVC's data source is
// a VirtualMachine, i.e. the volume is one of a VM's own disks. Such disks are
// already in the placement ConfigSpec and must not be counted twice.
//
// Only the API group and kind are compared, not the name.
```

pkg/util/vsphere/storage/storage_policy.go
divyenpatel: Renamed to capability. capIns was mirroring PbmCapabilityInstance.

pkg/util/vsphere/storage/storage_policy.go
divyenpatel: Took your version

pkg/util/vsphere/storage/storage_policy.go
divyenpatel: removed usage of continue

pkg/providers/vsphere/vmprovider_vmgroup.go
divyenpatel: I have verified calling PlaceVmsXCluster against live setup. Filed Jira - CRM-4964  for DRS team

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
divyenpatel: done

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
divyenpatel: done. addressed.

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
divyenpatel:
```golang
fileOp := vimtypes.VirtualDeviceConfigSpecFileOperationCreate
if p := diskPaths[pvc.Name]; p != "" {
    backing.FileName = p
    fileOp = ""
}
```
If diskPaths has a path for the PVC (i.e. the volume already exists), we set backing.FileName and clear fileOp to "". 

It's only left as Create when no path is known yet (WaitForFirstConsumer volumes, or a Bound volume CNS didn't return a path for), which mirrors the fastDeployDirect pattern you linked exactly — Create when FileName == "", "" otherwise.

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
divyenpatel: done.

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
divyenpatel: Yes — mainly WaitForFirstConsumer volumes, since those aren't provisioned yet at ConfigSpec-build time so there's no path to give them.

pkg/providers/vsphere/vmprovider_vm.go
divyenpatel: Done — moved addPVCPlacementDisks into CreateConfigSpecForPlacement in configspec.go

pkg/providers/vsphere/vmprovider_vm_hostlocal_storage.go
divyenpatel: Race exists, but not for vm provisioned with host-local policy. 

- VM provisioned with a host-local storage class — its home and disks are on the host-local datastore, so it can't move without a storage vMotion, which DRS load balancing doesn't do. Not affected.

- VM provisioned with a shared zonal policy that also requests a host-local WaitForFirstConsumer volume — this is the case the race applies to. Nothing pins the VM to the host until its host-local disk is actually attached, since its home isn't on that datastore. While the PVC is still unprovisioned this self-corrects, because reconcileHostLocalStorage re-publishes the selected node from the VM's current host on every reconcile. The gap is narrow: only a DRS migration that lands after CNS has already committed the volume causes a problem — the VM ends up unable to reach it, with no recovery short of deleting the PVC.

To be upfront about test coverage here: what's verified on a real cluster is the happy path (two real VMs, shared zonal + WFFC host-local PVCs, both reached PoweredOn and provisioned correctly). The race itself — a migration actually landing in that window — hasn't been reproduced or tested; the gap is documented from reasoning about the mechanism, not from an observed failure.

PR #1778, opened 2026-07-29
🌱 Tighten E2E poll interval for VM wait helpers

bryanv:

Analysis of an e2e-smoke run showed ~359s (20.7%) of total spec runtime sitting inside Eventually-based waits for VM power-state and creation, with event-timeline deltas landing on multiples of 10s - the signature of the condition becoming true mid-interval but only being checked every 10s.

Tighten the poll interval from 10s to 3s for the default/wait-virtual-machine-* family in wcp.yaml and kind.yaml (creation, deletion, condition-update, powerstate, resize, moid, restart-mode-update, vmip), keeping the existing timeout ceilings unchanged. Each poll is a single cheap k8s GET, so this is low risk and should reclaim a meaningful fraction of the polling-boundary waste.

Basic smoke test passes

Fixes # vmop-3998

```release-note
NONE
```

### DIFF
--- a/test/e2e/vmservice/config/kind.yaml
+++ b/test/e2e/vmservice/config/kind.yaml
@@ -37,13 +37,13 @@ variables:
 intervals:
   # Please keep the wait-virtual-machine values in sync between the different
   # configuration files.
-  default/wait-virtual-machine-creation: ["5m", "10s"]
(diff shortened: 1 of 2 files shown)


faisalabujabal (approved):

faisalabujabal (approved):
please add proof of testing

PR #1780, opened 2026-07-29
🌱 Return NoRequeueError for more PVC errors

bryanv:

After 5e54618 in which the VM controller watches PVCs and will reconcile VMs when a PVC it references is updated, we can return NoRequeueErrors for when the PVC doesn't exist or is not bound, and the watch will take care of reconciling the VM.

ResultFromError() uses errors.As() to determine the wrapped error so we cannot use k8s Aggregate since it does not implement As(), so switch to use errors.Join().

Remove a stale TODO comment about not doing placement: this is called during VM create, and with other changes we basically always do some form of placement.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/storage/storage.go
+++ b/pkg/providers/vsphere/storage/storage.go
@@ -9,10 +9,12 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
 	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	pkgerr "github.com/vmware-tanzu/vm-operator/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/providers/vsphere/storage/storage.go
```diff
@@ -92,8 +94,9 @@ func getVMStorageClassNamesAndPVCs(
 			pvc := &corev1.PersistentVolumeClaim{}
 			pvcKey := ctrlclient.ObjectKey{Name: claim.ClaimName, Namespace: vm.Namespace}
 			if err := client.Get(ctx, pvcKey, pvc); err != nil {
-				// TODO: If later we won't actually do placement, IMO this shouldn't be fatal. Need
-				// to rework the code a bit so this isn't as clunky.
+				if apierrors.IsNotFound(err) {
+					err = pkgerr.NoRequeueErrorf("%s", err.Error())
```
bryanv, line 98: It doesn't matter a ton for our purposes here but it isn't ideal what we lose the underlying error type. This error's message is used to for a condition reason, so we don't want the default NoRequeueError error string in there. I've thought about having a error instead of a message string in the NoRequeueError, and then delegate the NoRequeueError Error() to that don't want to make that churn now.

akutz (approved):

PR #1781, opened 2026-07-29
vmop-3767: Add vcsim tests for GuestOptions HW-version merge

faisalabujabal:

Adds vcsim-backed integration tests for the `VirtualMachineConfigOptions` controller's fan-out to `VirtualMachineGuestOptions`, covering the `status.hardwareVersions` listMap merge (keyed by `hardwareVersion`) across multiple hardware versions:

- Two hardware versions (`vmx-21`, `vmx-22`) reconciled in either order fan in to a single `VirtualMachineGuestOptions` per guest ID with two distinct `status.hardwareVersions` entries.
- Re-reconciling an already-converged hardware version updates its existing entry in place instead of appending a duplicate.
- Deleting the `VirtualMachineConfigOptions` that contributed an entry does not prune it from the shared `VirtualMachineGuestOptions` — deletion semantics for stale hardware versions are deferred (not yet implemented), and this test documents/locks in that current behavior.

These run against the real govmomi vcsim-backed provider (`testlabels.VCSim`), exercising the actual `QueryConfigOptionEx` client and `client.MergeFromWithOptimisticLock` merge-patch codepaths end to end, complementing the existing fake-provider unit test that only covers the first scenario.

Addresses vmop-3767.

No production code changes — test-only PR, consistent with `vmconfigoptions_controller.go` fan-out logic already merged in #1672. The existing single-hardware-version vcsim test (`vmx-19`) is left as-is; this adds a sibling `When` block for the multi-hardware-version scenarios.

```release-note
NONE
```

### DIFF
--- a/controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
+++ b/controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
@@ -13,6 +13,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	vimtypes "github.com/vmware/govmomi/vim25/types"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
```diff
@@ -684,4 +685,145 @@ func vcsimTestsReconcile() {
...
+			_, err := reconciler.Reconcile(vcsimCtx,
+				ctrl.Request{NamespacedName: types.NamespacedName{Name: vmx21.Name}})
+			Expect(err).ToNot(HaveOccurred())
+
+			var deleted vimv1.VirtualMachineConfigOptions
+			err = vcsimCtx.Client.Get(vcsimCtx, client.ObjectKeyFromObject(vmx21), &deleted)
+			Expect(apierrors.IsNotFound(err)).To(BeTrue())
+
+			vgo := guestOptionsFor(guestID)
+			Expect(vgo.Status.HardwareVersions).To(HaveLen(2),
```
hpannem, line 823: will it ever gc'd?

hpannem (approved):

controllers/virtualmachineconfigoptions/vmconfigoptions_controller_test.go
hpannem: I see other PR for this.

faisalabujabal:
## Claude Code review — head `a881d9bd`

Test-only PR, and the vcsim specs are well built: the reconcile-order-independence case, the fan-in via `sharedGuestID()` rather than a hardcoded guest ID (so it doesn't depend on vcsim's descriptor set), and the update-in-place-not-append assertion all test the right things.

One finding, and it's a merge-order collision with another approved PR.

### 🟠 The third spec asserts behavior that #1782 deliberately reverses

`"does not prune the hardwareVersions entry when its VirtualMachineConfigOptions is deleted"` deletes `vmx-21`, reconciles, and asserts `HaveLen(2)` — the `vmx-21` entry survives on the shared `VirtualMachineGuestOptions`. Its comment says "Deletion semantics for stale hardware versions are deferred," which was true when this was written on 07-29.

#1782's commit `ebe74fc5` ("Prune guest options on VirtualMachineConfigOptions deletion", 07-30) makes `ReconcileDelete` call `garbageCollectGuestOptions(ctx, obj.Spec.HardwareVersion, sets.New[string]())` — an empty live set, so the deleted hardware version's entry *is* pruned before the finalizer is removed. That is exactly the call this spec's `reconciler.Reconcile` on the deleted object will execute. This spec asserts `HaveLen(2)` right after asserting the object is gone (which requires `ReconcileDelete` to have run), so under #1782 the entry is removed and the assertion fails.

Both PRs are approved, neither is merged, and whichever lands second breaks `test-controllers`. The fix is a one-liner in whichever goes second — invert the spec to assert the entry *is* pruned and the sole-owner object deleted — but it should be decided now rather than discovered by CI. #1789 changes the gate again (owner-reference count instead of status-entry count), so it's worth settling the intended end state across all three.

Nothing else. `guestOptionsFor` doing a full `List` and matching on `Spec.ID` rather than computing `pkgutil.VimGuestOptionsName(guestID)` is slightly indirect but arguably more robust, since it doesn't re-derive the name transform the code under test uses.

aruneshpa (approved):

PR #1782, opened 2026-07-29
Prune unused VirtualMachineGuestOptions entries

faisalabujabal:

The `VirtualMachineConfigOptions` controller fans out a `VirtualMachineGuestOptions` object per guest OS descriptor and upserts a `status.hardwareVersions` entry into it per hardware version, but never garbage-collected that entry once a hardware version's `QueryConfigOptionEx` result stopped reporting the guest OS. This mirrors a gap already called out as deferred in `.sdd/specs/001-class-policy-resize/spec.md` and `research.md`.

This PR adds `garbageCollectGuestOptions`, run at the end of every `VirtualMachineConfigOptions` reconcile: it removes the current hardware version's `status.hardwareVersions` entry from any `VirtualMachineGuestOptions` no longer reporting that guest OS, and deletes the object outright once no hardware-version entries remain — the same shape as the existing `ConfigTarget` -> `VirtualMachineConfigOptions` garbage collection.

The live set used for comparison is keyed by the sanitized object name (via `toGuestOptionsName`), not the raw guest OS ID, so two distinct guest OS IDs that collide onto the same truncated, DNS-safe object name cannot cause a spurious deletion.

Also updates `spec.md`, `research.md`, and `tasks.md` under `.sdd/specs/001-class-policy-resize/` to reflect that this previously-deferred follow-up is now implemented.

vmop-3932

- No `test/e2e` coverage was added. vcsim's `EnvironmentBrowser` reports a fixed descriptor set per model, so "a guest OS is dropped from a cluster's real descriptor list" can't be triggered against a live cluster -- the same constraint already noted for the sibling `ConfigTarget` GC feature (see `tasks.md` T071a). Coverage instead includes a unit test asserting the real removal/deletion behavior plus a vcsim/envtest-backed test that verifies the JSON merge-patch mechanics against a real API server (not just the controller-runtime fake client).

```release-note
Fixed a resource leak where `VirtualMachineGuestOptions` objects and their per-hardware-version status entries were never cleaned up after a hardware version stopped supporting a guest OS.
```


faisalabujabal:
Pushed a follow-up commit: `QueryConfigOptionEx` returning `nil, nil` (no error) is vSphere's positive "no config option for this hardware version" answer, not a transient signal — that comes back as a non-nil error instead, handled separately above. `ReconcileNormal` was returning before `fanOutGuestOptions` ran in that branch, so a hardware version that stopped being supported entirely left its previously fanned-out `VirtualMachineGuestOptions` entries orphaned forever, even though the owning `VirtualMachineConfigOptions` kept reconciling. Now routes through `fanOutGuestOptions` with an empty descriptor list, same as the already-covered non-nil-but-empty case. Added a unit test for the transition.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -377,6 +392,77 @@ func (r *Reconciler) reconcileGuestOptions(
...
+	if idx == -1 {
+		return nil
+	}
+
+	if len(obj.Status.HardwareVersions) == 1 {
+		if err := r.Delete(ctx, obj); err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete orphaned VirtualMachineGuestOptions %q: %w", obj.Name, err)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
faisalabujabal: Good catch, fixed in 3e13e0e2 — gated the `Delete` on `client.Preconditions{ResourceVersion: &obj.ResourceVersion}` so a racing writer's addition surfaces as a conflict (retried on the next reconcile) instead of being silently discarded.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
hpannem, line 157: what happens when the CR is deleted? will it not GC guest options?

This begs a question why we are not using ownerreferences?

faisalabujabal: good catch and I fixed it

hpannem (approved):
This looks good. Ownerrefs might give some observability as well.

faisalabujabal:
## Claude Code review — head `ebe74fc5`

Reviewed the full diff. The GC design holds up: `MergeFromWithOptimisticLock` on a fan-in-written shared object, `client.Preconditions{ResourceVersion}` on the orphan delete so a racing upsert conflicts instead of being deleted along with the object, status patched via the subresource, and the "nil-but-no-error from `QueryConfigOptionEx` is a positive answer, not an unknown" distinction — all consistent with `.sdd/memory/operator-best-practices.md`. Approved already; nothing here blocks.

### 🟠 Merge-order conflict with #1781

`ebe74fc5` makes `ReconcileDelete` prune the deleted hardware version's entry. #1781 (also approved, not yet merged, single commit from 07-29) contains a spec asserting the opposite — `"does not prune the hardwareVersions entry when its VirtualMachineConfigOptions is deleted"`, expecting `HaveLen(2)` after deleting one contributing `VirtualMachineConfigOptions`. Whichever of these merges second fails `test-controllers`. Details in my comment on #1781; flagging here so it isn't missed from whichever side lands first.

### 🔵 The write path can hot-loop on a name collision the GC path defends against

`fanOutGuestOptions`'s comment is precise about the hazard: "distinct IDs can collide onto the same truncated, DNS-safe name, and comparing garbage collection against the wrong key could delete an object still live under its colliding sibling ID." The GC side is keyed on the sanitized name for exactly that reason. The *write* side has no equivalent guard: if two descriptors in one fan-out collide onto the same name, `reconcileGuestOptions` patches `spec.ID` to descriptor A's value, then to descriptor B's, in a single pass — and does it again on every subsequent reconcile, bumping `resourceVersion` each time. With the controller watching the objects it fans out, that's a self-sustaining reconcile loop rather than an error anyone would notice.

In practice vSphere guest IDs are camelCase-alphanumeric, so `VimGuestOptionsName` (#1779) only lowercases them and a collision would need two IDs differing solely in case — which vSphere doesn't produce. So this is latent, not live. But since the code already documents the collision hazard on one side, it's worth either stating the assumption ("vSphere guest IDs are unique under the transform") next to the helper in #1779, or making the transform collision-resistant (suffix a short hash when the transform changes the input).

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -377,6 +414,84 @@ func (r *Reconciler) reconcileGuestOptions(
...
+// hardwareVersion, this is a no-op -- it was never reported by that hardware
+// version, so there is nothing to reconcile away.
+func (r *Reconciler) removeHardwareVersionAndDeleteIfOrphaned(
+	ctx context.Context,
+	obj *vimv1.VirtualMachineGuestOptions,
+	hardwareVersion string) error {
+
+	idx := -1
+	for i := range obj.Status.HardwareVersions {
+		if obj.Status.HardwareVersions[i].HardwareVersion == hardwareVersion {
```
aruneshpa: Not your change, but this stutters.  We should have called it "version" or "id" so we have `status.hardwareversions.version`. :(

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
```diff
@@ -377,6 +414,84 @@ func (r *Reconciler) reconcileGuestOptions(
...
+// hardware-version entries, deletes obj. If obj has no entry for
+// hardwareVersion, this is a no-op -- it was never reported by that hardware
+// version, so there is nothing to reconcile away.
+func (r *Reconciler) removeHardwareVersionAndDeleteIfOrphaned(
+	ctx context.Context,
+	obj *vimv1.VirtualMachineGuestOptions,
+	hardwareVersion string) error {
+
+	idx := -1
+	for i := range obj.Status.HardwareVersions {
```
aruneshpa: Did you consider using `slices.DeleteFunc` here with a matcher func that can be on the hardware version?  If the returned slice is empty, then you need to delete the object, otherwise, it's a patch?

aruneshpa (approved):
LGTM.  One exploratory comment.

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: i'd prefer to do this in a separate PR since it is a contract update that I don't want to sneak here

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: Rebased onto main to pick up #1779's `pkgutil.VimGuestOptionsName` refactor and #1781's new vcsim tests. #1781 landed a test asserting the opposite of what `ebe74fc5` does ("does not prune the hardwareVersions entry when its VirtualMachineConfigOptions is deleted"); rewrote it to assert the current, correct behavior instead of just carrying it forward stale.

Also adopted `slices.DeleteFunc` here per your suggestion (`ae3a3b19`) — the delete-vs-patch branch now reads directly off the resulting slice length instead of a separate found-index check.

— Faisal + Claude

controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go
faisalabujabal: https://github.com/vmware-tanzu/vm-operator/pull/1793

PR #1783, opened 2026-07-29
Add VirtualMachineConfigPolicy validation webhook

faisalabujabal:

Adds an admission validating webhook for `VirtualMachineConfigPolicy` (`vim.vmware.com`), part of vmop-3745 (Story S8). It validates that:

- `spec.zone` references an existing Zone in the policy's namespace (on create and update).
- Every `spec.extraConfig` `allowed`/`denied` entry has a non-empty key.

The Zone-existence check is skipped for VM Operator's own service account (`ctx.IsVMOperatorAccount`): the Zone controller creates/patches this object in the same reconcile that just created its Zone, using a cache-backed client, so a cache lag could otherwise cause a spurious rejection of VM Operator's own write. This mirrors the same bypass pattern already used by the `virtualmachine` and `persistentvolumeclaim` validators for cross-object checks on VM Operator's own writes.

Ships with unit and integration (envtest) tests.

Fixes vmop-3769

- The originally-scoped defaulting webhook was skipped: every field it would default (`syncMode`, `createMode`, `updateMode`, `powerOnMode`, `vmClassMode`) already carries a `+kubebuilder:default=` marker on the CRD, so API-server structural defaulting already covers it. This mirrors the `configtarget` and `virtualmachineconfigoptions` webhooks, which are validate-only for the same reason.
- This PR is functionally independent of #1784 (the sync controller) — either can merge first. If they land out of order, `config/webhook/manifests.yaml` and `config/rbac/role.yaml` may need a manifest regeneration (`make generate-manifests`) rather than a hand-resolved merge conflict.

```release-note
Add a validating admission webhook for VirtualMachineConfigPolicy that rejects a spec.zone referencing a non-existent Zone and extraConfig entries with an empty key.
```

### DIFF
--- a/external/vim/api/v1alpha1/virtualmachine_config_policy_types.go
+++ b/external/vim/api/v1alpha1/virtualmachine_config_policy_types.go
@@ -116,6 +116,7 @@ const (
 // VirtualMachineConfigPolicy.
 type VirtualMachineConfigPolicySpec struct {
 	// +required
+	// +kubebuilder:validation:MinLength=1
 
 	// Zone is the name of the zones.topology.tanzu.vmware.com resource to which
 	// this policy applies.
--- a/external/vim/api/v1alpha1/virtualmachine_config_policy_types.go
+++ b/external/vim/api/v1alpha1/virtualmachine_config_policy_types.go
@@ -31,6 +31,7 @@ type VirtualMachineConfigPolicyExtraConfigKey struct {
 	Type MatchType `json:"type"`
 
 	// +required
+	// +kubebuilder:validation:MinLength=1
 
 	// Key is the extra config key to match.
 	Key string `json:"key"`
--- a/webhooks/virtualmachineconfigpolicy/validation/virtualmachineconfigpolicy_validator.go
+++ b/webhooks/virtualmachineconfigpolicy/validation/virtualmachineconfigpolicy_validator.go
@@ -1,0 +1,187 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package validation
+
+import (
+	"fmt"
+	"net/http"
+	"reflect"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	vimv1 "github.com/vmware-tanzu/vm-operator/external/vim/api/v1alpha1"
+
+	"github.com/vmware-tanzu/vm-operator/pkg/builder"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+	"github.com/vmware-tanzu/vm-operator/pkg/topology"
+	"github.com/vmware-tanzu/vm-operator/webhooks/common"
+)
+
+const (
+	webHookName = "default"
+)
+
+// +kubebuilder:webhook:verbs=create;update,path=/default-validate-vim-vmware-com-v1alpha1-virtualmachineconfigpolicy,mutating=false,failurePolicy=fail,groups=vim.vmware.com,resources=virtualmachineconfigpolicies,versions=v1alpha1,name=default.validating.virtualmachineconfigpolicy.v1alpha1.vim.vmware.com,sideEffects=None,admissionReviewVersions=v1;v1beta1
+// +kubebuilder:rbac:groups=vim.vmware.com,resources=virtualmachineconfigpolicies,verbs=get;list
+
+// AddToManager adds the webhook to the provided manager.
+func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr ctrlmgr.Manager) error {
+	hook, err := builder.NewValidatingWebhook(ctx, mgr, webHookName, NewValidator(mgr.GetClient()))
+	if err != nil {
+		return fmt.Errorf("failed to create VirtualMachineConfigPolicy validation webhook: %w", err)
+	}
+
+	mgr.GetWebhookServer().Register(hook.Path, hook)
+
+	return nil
+}
+
+// NewValidator returns the package's Validator.
+func NewValidator(client client.Client) builder.Validator {
+	return validator{
+		client:    client,
+		converter: runtime.DefaultUnstructuredConverter,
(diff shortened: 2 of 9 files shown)


webhooks/virtualmachineconfigpolicy/validation/virtualmachineconfigpolicy_validator.go
```diff
@@ -0,0 +1,197 @@
...
+			field.InternalError(f, err),
+		}
+	}
+
+	return nil
+}
+
+// validateExtraConfig returns an error for any extraConfig allowed/denied
+// entry whose key is empty. The Type enum is already enforced by the CRD's
+// OpenAPI schema, so the only remaining rule to enforce here is a non-empty
```
faisalabujabal: can that not be expressed with kubebuilder annotations?

faisalabujabal: Good catch — moved this to the CRD schema instead. Added `+kubebuilder:validation:MinLength=1` on `VirtualMachineConfigPolicyExtraConfigKey.Key` and regenerated the vim CRD manifest, then removed the Go-side `validateExtraConfig`/`validateExtraConfigKeys` check entirely. Strengthened the existing envtest integration test to assert the rejection is a schema validation error (`apierrors.IsInvalid`) rather than a webhook response, proving the CRD enforces it on its own.

webhooks/virtualmachineconfigpolicy/validation/virtualmachineconfigpolicy_validator.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// Re-run the same rules validated on create: spec.zone is not immutable
+	// (a policy may be repointed at a different zone), so an update must
+	// validate the new zone reference exists just as create does.
+	fieldErrs := v.validateSpec(ctx, policy)
+
+	validationErrs := make([]string, 0, len(fieldErrs))
+	for _, fieldErr := range fieldErrs {
+		validationErrs = append(validationErrs, fieldErr.Error())
+	}
```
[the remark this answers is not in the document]
faisalabujabal: Fixed — extracted a shared `validate()` method used by both `ValidateCreate` and `ValidateUpdate`.

webhooks/virtualmachineconfigpolicy/validation/virtualmachineconfigpolicy_validator_test.go
```diff
@@ -0,0 +1,375 @@
...
+		err error
+		ctx *intgValidatingWebhookContext
+	)
+
+	BeforeEach(func() {
+		ctx = newIntgValidatingWebhookContext()
+		Expect(ctx.Client.Create(ctx, ctx.zone)).To(Succeed())
+		Expect(ctx.Client.Create(ctx, ctx.configPolicy)).To(Succeed())
+	})
+	JustBeforeEach(func() {
```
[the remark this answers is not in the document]
faisalabujabal: Fixed — this and the Delete test below both now use `ctx` instead of the package-level `suite`.

webhooks/virtualmachineconfigpolicy/validation/virtualmachineconfigpolicy_validator_test.go
```diff
@@ -0,0 +1,375 @@
...
+		})
+	})
+}
+
+func intgTestsValidateDelete() {
+	var (
+		err error
+		ctx *intgValidatingWebhookContext
+	)
+
```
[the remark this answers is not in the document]
faisalabujabal: Fixed alongside the Update test above — both now use `ctx`.

external/vim/api/v1alpha1/virtualmachine_config_policy_types.go
hpannem, line 123: nit: minlength = 1 avoids webhook validation If zone is empty string.

hpannem (approved):

external/vim/api/v1alpha1/virtualmachine_config_policy_types.go
faisalabujabal: Good catch, thanks — added `+kubebuilder:validation:MinLength=1` on `spec.zone`, regenerated the CRD manifest, and added an envtest integration test confirming an empty zone is rejected by the schema before it ever reaches the webhook's Zone lookup.

faisalabujabal:
## Claude Code review — head `06394cca`

Reviewed the full diff. No blocking findings. The validator is minimal in the right way — one live cluster read that genuinely can't be CEL, with the structural rules (`MinLength=1` on `spec.zone` and on the extraConfig key) pushed into the CRD schema after review, matching the direction #1779 amended the constitution toward. `ValidateUpdate` re-running create's rules is correct given `spec.zone` is mutable, and the reasoning is in the doc comment where the next reader will find it.

Two observations, neither a defect.

### 🔵 An `APIReader` would remove the need for the service-account carve-out

`validateZone` skips the check entirely for `ctx.IsVMOperatorAccount`, because the Zone controller creates a Zone and patches a policy in the same reconcile and the cache may lag. That's a real problem with a precedent in this repo (`webhooks/virtualmachine/validation/virtualmachine_validator.go:1527`), so this isn't wrong. But the bypass is broader than the problem: it also lets any write from the VM Operator service account set `spec.zone` to a zone that doesn't exist at all, including a typo'd one from a future controller, with nothing catching it.

The cache lag is the whole reason for the carve-out, and `mgr.GetAPIReader()` doesn't have it — a quorum read is guaranteed to see a Zone the Zone controller just created through the API server. Using the `APIReader` for this one lookup would let you drop `IsVMOperatorAccount` and validate VM Operator's own writes like anyone else's. Cost is one uncached GET per policy admission, and policy admissions are rare (one per zone, plus tenant edits). Worth considering; not worth blocking on.

### 🔵 `webhooks/virtualmachineconfigpolicy/` and the group-placement rule

`.sdd/memory/constitution.md` says "Webhooks for API groups other than `vmoperator.vmware.com` should not be placed directly in the `webhooks/` directory," and this is a `vim.vmware.com` webhook sitting directly under `webhooks/`. The repo already contradicts the rule in several places (`webhooks/configtarget/`, `webhooks/virtualmachineconfigoptions/`, `webhooks/unifiedstoragequota/`, `webhooks/persistentvolumeclaim/`, and `controllers/vspherepolicy/` on the controller side), so this PR is following the de facto layout, not breaking new ground — and #1785 removes two of those. Since this spec's PRs are already amending the constitution where it conflicts with practice (#1779), this rule looks like the next honest candidate: either the `vim.vmware.com` surface moves under `webhooks/vim/` and `controllers/vim/`, or the rule gets narrowed to describe what the repo actually does.

external/vim/api/v1alpha1/virtualmachine_config_policy_types.go
```diff
@@ -115,6 +116,7 @@ const (
 // VirtualMachineConfigPolicy.
 type VirtualMachineConfigPolicySpec struct {
 	// +required
+	// +kubebuilder:validation:MinLength=1
```
aruneshpa, line 119: Do you also want a maxLength (of 253 I think) given that this is the name of a k8s resource?

aruneshpa (approved):
LGTM.  One minor comment.

PR #1738, opened 2026-07-15
✨ Enforce VirtualMachineConfigPolicy in webhook and power-on reconcile

akutz:
**What does this PR do, and why is it needed?**

Add two enforcement points for a namespace zone's
VirtualMachineConfigPolicy, sharing matching logic so they can't drift apart:

- webhooks/virtualmachine/validation: validateConfigPolicy checks a VM's desired spec (advanced.extraConfig, minHardwareVersion) against the policy's CreateMode/UpdateMode/PowerOnMode on create, update, and power-on transitions.
- pkg/providers/vsphere: verifyConfigPolicy checks a VM's actual, live vSphere ConfigInfo immediately before power-on, catching VMs that were compliant when last reconfigured but whose policy has since tightened (e.g. an extraConfig key now denied, or a hardware version no longer supported by the zone's cluster).
- pkg/util/configpolicy: shared ExtraConfig/HardwareVersion matching used by both enforcement points.

Also adds external/vim/api/v1alpha1.HardwareVersion (parsing, validation, and supported-version range for vmx-N strings) and HardwareVersionRange, and extends VirtualMachineConfigPolicySpec with a HardwareVersions field.

**Which issue(s) is/are addressed by this PR?** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes `NA`

**Are there any special notes for your reviewer**:

**Please add a release note if necessary**:

```release-note
Enforce VirtualMachineConfigPolicy in webhook and power-on reconcile
```

PR #1784, opened 2026-07-29
Add VirtualMachineConfigPolicy sync controller

faisalabujabal:

Adds the `VirtualMachineConfigPolicy` sync controller, part of vmop-3745 (Story S8 / vmop-3770, vmop-3771, vmop-3772). When `spec.syncMode=ConfigTarget`, the controller mirrors the Zone's cluster capabilities (from `ConfigTarget.status`) into the policy's `spec` — min-of-maxima for numeric ranges, AND for boolean support flags, and set-intersection for device categories when a zone spans more than one cluster.

The controller watches both `ConfigTarget` and `Zone` (via field indexes on `VirtualMachineConfigPolicy.spec.zone` and `Zone.spec.managedVMs.clusterMoIDs`), so a capability change on either is reflected within one reconcile, per spec.md's requirement.

Ships with unit tests (including a hot-loop regression: a second reconcile against an unchanged ConfigTarget does not bump `resourceVersion`, checked against both the fake client and a real vcsim/envtest API server), vcsim integration tests, and an E2E extension of the existing `configpolicy` suite.

Fixes vmop-3770, Fixes vmop-3771, Fixes vmop-3772

- The ConfigTarget→policy field-mapping logic lives in a new package, `pkg/util/configpolicysync/`, rather than the `pkg/vmconfig/policy/` path in the original task breakdown — that path is already used by an unrelated existing VM-tag/PolicyEvaluation reconciler. The name also deliberately avoids `pkg/util/configpolicy`, reserved by an unrelated, unmerged PR (#1738) for a different purpose (S9 enforcement matching).
- Cluster MoID derivation uses `zone.Spec.ManagedVMs.ClusterMoIDs` directly, matching what the merged Zone controller actually uses to create `ConfigTarget` objects — not the `spec.namespace.poolMoIDs`-based derivation described in `external/vim/doc/controller-workflows.md`, which is stale relative to the merged code.
- Does not depend on PR #1738 (S9 enforcement); built to merge cleanly with or without it.
- Independent of, but complements, #1783 (the validation webhook) — either can merge first. If they land out of order, `config/rbac/role.yaml` may need a manifest regeneration (`make generate-manifests`) rather than a hand-resolved merge conflict.

```release-note
Add a VirtualMachineConfigPolicy sync controller that mirrors vSphere cluster capabilities from ConfigTarget into the policy's spec.
```

### DIFF
--- a/.sdd/specs/001-class-policy-resize/tasks.md
+++ b/.sdd/specs/001-class-policy-resize/tasks.md
@@ -98,11 +98,21 @@ Dependencies: none. All A-tasks may run in parallel `[P]`.
 - [ ] T100 [S8.a] Author `webhooks/virtualmachineconfigpolicy/defaulting_webhook.go` — default syncMode=ConfigTarget, createMode/updateMode/powerOnMode=Allow, vmClassMode=AsPolicy
 - [ ] T101 [S8.a] Author `webhooks/virtualmachineconfigpolicy/validation_webhook.go` — spec.zone references existing Zone; extraConfig allowed/denied entries have non-empty key and valid type enum
 - [ ] T102 [S8.a] Unit tests — `webhooks/virtualmachineconfigpolicy/`
-- [ ] T103 [S8.b] Author `controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go` — syncMode=ConfigTarget: copy ConfigTarget.status → policy spec; syncMode=Disabled: set Ready=True, skip sync; never overwrite extraConfig/latencySensitivityLevels/txRxThreadModels
-- [ ] T104 [S8.b] Author `pkg/vmconfig/policy/policy_reconciler.go` — ConfigTarget→policy sync field mapping logic (separate from controller for unit testability)
-- [ ] T105 [S8.b] Unit tests — `pkg/vmconfig/policy/policy_reconciler_test.go`
-- [ ] T106 [S8.c] [P] Integration tests with vcsim — `test/intg/virtualmachineconfigpolicy/`: syncMode=ConfigTarget happy path; syncMode=Disabled skip; missing ConfigTarget → condition; multi-cluster zone selects correct ConfigTarget
-- [ ] T107 [S8.d] E2E test — policy spec filled from real cluster capabilities after Zone creation
+- [x] T103 [S8.b] [vmop-3745] Author `controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go` — syncMode=ConfigTarget: copy ConfigTarget.status → policy spec; syncMode=Disabled: set Ready=True (reason SyncDisabled), skip sync; never overwrite extraConfig/latencySensitivityLevels/txRxThreadModels. spec.zone not resolving to a live Zone surfaces Ready=False/ZoneNotFound rather than blocking other policies, per spec.md's edge-case note — the complement to the validation webhook's create/update-time check.
+- [x] T104 [S8.b] [vmop-3745] Author `pkg/util/configpolicysync/configpolicysync.go` — ConfigTarget→policy sync field mapping logic (separate from controller for unit testability). Multi-cluster zones merge by intersection (min of numeric maxima, AND of boolean flags, common ConfigTargetDevices entries), per vmop-3746's acceptance criteria.
+- [x] T105 [S8.b] [vmop-3745] Unit tests — `pkg/util/configpolicysync/configpolicysync_test.go`; `controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller_test.go` covers the controller's Disabled/ZoneNotFound/ConfigTargetNotFound/toggle/hot-loop-regression paths with a fake client.
+- [x] T106 [S8.c] [vmop-3745] Integration tests — co-located in `controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller_test.go` as a second `Describe` (Label `EnvTest`+`VCSim`, per `testing-standards.md`'s single-file-per-package convention, not a `test/intg/` tree): syncMode=ConfigTarget happy path against a real vcsim-derived ConfigTarget; missing ConfigTarget → condition.
+- [x] T107 [S8.d] [vmop-3745] E2E test — extended `test/e2e/vmservice/vmservice/configpolicy/configpolicy.go`'s existing `Spec()`: policy spec (numCPUCores, memory) filled from the zone's real ConfigTarget after Zone creation; toggling syncMode=Disabled stops spec changes. Labeled `experimental` pending validation on real hardware, per `e2e-testing.md`.
+
+  > Implementation note: T104's path was changed from `pkg/vmconfig/policy/`
+  > to `pkg/util/configpolicysync/` — `pkg/vmconfig/policy/` already exists
+  > and implements unrelated per-VM tag/PolicyEvaluation reconciliation
+  > (`Reconcile(ctx, k8sClient, vimClient, vm)`), and `pkg/util/configpolicy`
+  > is reserved for the enforcement-side matching logic PR #1738 introduces
+  > for Story S9. T106's vcsim coverage is folded into the controller's own
+  > test file rather than a `test/intg/` directory, matching
+  > `testing-standards.md` and the `configtarget` controller's precedent
+  > (`controllers/configtarget/configtarget_controller_test.go`).
 
 ### Story S9 — VM admission webhook enforcement (vmop-3746)
 
--- a/.sdd/specs/001-class-policy-resize/tasks.md
+++ b/.sdd/specs/001-class-policy-resize/tasks.md
@@ -104,15 +104,7 @@ Dependencies: none. All A-tasks may run in parallel `[P]`.
 - [x] T106 [S8.c] [vmop-3745] Integration tests — co-located in `controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller_test.go` as a second `Describe` (Label `EnvTest`+`VCSim`, per `testing-standards.md`'s single-file-per-package convention, not a `test/intg/` tree): syncMode=ConfigTarget happy path against a real vcsim-derived ConfigTarget; missing ConfigTarget → condition.
 - [x] T107 [S8.d] [vmop-3745] E2E test — extended `test/e2e/vmservice/vmservice/configpolicy/configpolicy.go`'s existing `Spec()`: policy spec (numCPUCores, memory) filled from the zone's real ConfigTarget after Zone creation; toggling syncMode=Disabled stops spec changes. Labeled `experimental` pending validation on real hardware, per `e2e-testing.md`.
 
-  > Implementation note: T104's path was changed from `pkg/vmconfig/policy/`
-  > to `pkg/util/configpolicysync/` — `pkg/vmconfig/policy/` already exists
-  > and implements unrelated per-VM tag/PolicyEvaluation reconciliation
-  > (`Reconcile(ctx, k8sClient, vimClient, vm)`), and `pkg/util/configpolicy`
-  > is reserved for the enforcement-side matching logic PR #1738 introduces
-  > for Story S9. T106's vcsim coverage is folded into the controller's own
-  > test file rather than a `test/intg/` directory, matching
-  > `testing-standards.md` and the `configtarget` controller's precedent
-  > (`controllers/configtarget/configtarget_controller_test.go`).
+  > Implementation note: T104's path was changed from `pkg/vmconfig/policy/` to `pkg/util/configpolicysync/` — `pkg/vmconfig/policy/` already exists and implements unrelated per-VM tag/PolicyEvaluation reconciliation (`Reconcile(ctx, k8sClient, vimClient, vm)`), and `pkg/util/configpolicy` is reserved for the enforcement-side matching logic PR #1738 introduces for Story S9. T106's vcsim coverage is folded into the controller's own test file rather than a `test/intg/` directory, matching `testing-standards.md` and the `configtarget` controller's precedent (`controllers/configtarget/configtarget_controller_test.go`).
 
 ### Story S9 — VM admission webhook enforcement (vmop-3746)
 
--- a/controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
+++ b/controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
@@ -62,8 +62,17 @@ const (
 	ZoneNotFoundReason = "ZoneNotFound"
 
 	// ConfigTargetNotFoundReason is the Ready condition reason used when
-	// none of the zone's cluster MoIDs resolve to an existing ConfigTarget.
+	// the zone has no cluster MoIDs to sync from.
 	ConfigTargetNotFoundReason = "ConfigTargetNotFound"
+
+	// ConfigTargetNotReadyReason is the Ready condition reason used when
+	// one or more of the zone's cluster MoIDs do not resolve to a Ready
+	// ConfigTarget. spec is left untouched in this case: a ConfigTarget
+	// with Ready!=True has not finished populating its status, and merging
+	// its zero-valued fields would publish a false capability denial rather
+	// than "unknown," since a zero maximum on a numeric field means "not
+	// supported" per VirtualMachineConfigPolicySpec's doc comments.
+	ConfigTargetNotReadyReason = "ConfigTargetNotReady"
 )
 
 // SkipNameValidation is used for testing to allow multiple controllers with the
@@ -294,6 +303,7 @@ func (r *Reconciler) ReconcileNormal(
 		if apierrors.IsNotFound(err) {
 			pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, ZoneNotFoundReason,
 				"zone %q not found", obj.Spec.Zone)
+			obj.Status.ObservedGeneration = obj.Generation
 
 			return nil
 		}
@@ -303,15 +313,30 @@ func (r *Reconciler) ReconcileNormal(
 		return fmt.Errorf("failed to get zone %q: %w", obj.Spec.Zone, err)
 	}
 
-	targets, err := r.getConfigTargets(ctx, zone.Spec.ManagedVMs.ClusterMoIDs)
+	if len(zone.Spec.ManagedVMs.ClusterMoIDs) == 0 {
+		pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, ConfigTargetNotFoundReason,
+			"zone %q has no cluster to sync from", zone.Name)
+		obj.Status.ObservedGeneration = obj.Generation
+
+		return nil
+	}
+
+	targets, notReady, err := r.getConfigTargets(ctx, zone.Spec.ManagedVMs.ClusterMoIDs)
 	if err != nil {
-		pkgcond.MarkError(obj, vimv1.ReadyConditionType, ConfigTargetNotFoundReason, err)
+		pkgcond.MarkError(obj, vimv1.ReadyConditionType, ConfigTargetNotReadyReason, err)
 		return fmt.Errorf("failed to get config targets for zone %q: %w", zone.Name, err)
 	}
 
-	if len(targets) == 0 {
-		pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, ConfigTargetNotFoundReason,
-			"no ConfigTarget found for zone %q", zone.Name)
+	if len(notReady) > 0 {
+		// Require every cluster behind the zone to have a Ready
+		// ConfigTarget before syncing: merging a subset would compute an
+		// intersection over fewer clusters than the zone actually spans,
+		// which is a strictly wider (more permissive) result than the true
+		// intersection -- the opposite of the safe direction for data that
+		// feeds capability enforcement.
+		pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, ConfigTargetNotReadyReason,
+			"ConfigTarget(s) not ready for cluster(s) %v", notReady)
+		obj.Status.ObservedGeneration = obj.Generation
 
 		return nil
 	}
@@ -331,31 +356,41 @@ func (r *Reconciler) ReconcileNormal(
 }
 
 // getConfigTargets returns the ConfigTarget status of every clusterMoID that
-// resolves to an existing ConfigTarget. A clusterMoID with no matching
-// ConfigTarget is skipped rather than treated as an error: today a zone
-// maps to a single vSphere cluster in practice, and per
-// external/vim/doc/controller-workflows.md the remaining entries exist for
-// infrastructure mobility / cluster decommissioning, so a stale MoID should
-// not block a sync that the live cluster(s) can otherwise satisfy.
+// resolves to a Ready ConfigTarget, and the subset of clusterMoIDs that do
+// not (missing, or present but Ready!=True). Every clusterMoID behind the
+// zone must be Ready before the caller may sync: a ConfigTarget that has not
+// finished populating its status still reports zero-valued numeric fields,
+// and a zero maximum is a real capability denial per
+// VirtualMachineConfigPolicySpec's field docs, not "no data yet." Merging a
+// not-yet-Ready target's status, or silently dropping a missing one from a
+// multi-cluster zone's intersection, would therefore compute a result no
+// narrower than -- and possibly wider (more permissive) than -- the true
+// intersection across every cluster the zone spans.
 func (r *Reconciler) getConfigTargets(
 	ctx context.Context,
-	clusterMoIDs []string) ([]vimv1.ConfigTargetStatus, error) {
-	targets := make([]vimv1.ConfigTargetStatus, 0, len(clusterMoIDs))
+	clusterMoIDs []string) (targets []vimv1.ConfigTargetStatus, notReady []string, err error) {
+	targets = make([]vimv1.ConfigTargetStatus, 0, len(clusterMoIDs))
 
 	for _, clusterMoID := range clusterMoIDs {
 		var ct vimv1.ConfigTarget
 
-		err := r.Get(ctx, client.ObjectKey{Name: clusterMoID}, &ct)
-		if err != nil {
-			if apierrors.IsNotFound(err) {
+		getErr := r.Get(ctx, client.ObjectKey{Name: clusterMoID}, &ct)
+		if getErr != nil {
+			if apierrors.IsNotFound(getErr) {
+				notReady = append(notReady, clusterMoID)
 				continue
 			}
 
-			return nil, fmt.Errorf("failed to get ConfigTarget %q: %w", clusterMoID, err)
+			return nil, nil, fmt.Errorf("failed to get ConfigTarget %q: %w", clusterMoID, getErr)
+		}
+
+		if !pkgcond.IsTrue(&ct, vimv1.ReadyConditionType) {
+			notReady = append(notReady, clusterMoID)
+			continue
 		}
 
 		targets = append(targets, ct.Status)
 	}
 
-	return targets, nil
+	return targets, notReady, nil
 }
--- a/controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
+++ b/controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
@@ -17,8 +17,11 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	topologyv1 "github.com/vmware-tanzu/vm-operator/external/tanzu-topology/api/v1alpha1"
 	vimv1 "github.com/vmware-tanzu/vm-operator/external/vim/api/v1alpha1"
 	pkgcond "github.com/vmware-tanzu/vm-operator/pkg/conditions"
 	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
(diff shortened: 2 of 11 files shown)


test/e2e/vmservice/vmservice/configpolicy/configpolicy.go
```diff
@@ -370,6 +371,94 @@ func Spec(ctx context.Context, inputGetter func() SpecInput) {
...
+							"VirtualMachineConfigPolicy %q should be Ready", policy.Name)
+
+						g.Expect(policy.Spec.NumCPUCores).ToNot(BeNil(),
+							"VirtualMachineConfigPolicy %q spec.numCPUCores should be populated from ConfigTarget.status",
+							policy.Name)
+						g.Expect(policy.Spec.NumCPUCores.Max).To(Equal(ct.Status.NumCPUCores),
+							"VirtualMachineConfigPolicy %q spec.numCPUCores.max should equal ConfigTarget %q status.numCPUCores",
+							policy.Name, ct.Name)
+
+						g.Expect(policy.Spec.Memory).ToNot(BeNil(),
```
faisalabujabal, line 436: can we assert on more fields? can we get the actual values from VC to assert that they match

faisalabujabal: Also had an Opus review pass on this PR that dug into this exact spec and found it could pass vacuously: `ct` was fetched once outside the `Eventually`, so a not-yet-populated ConfigTarget status (all zeros) would compare 0==0 and pass without proving anything.

Fixed: `ct` is now re-fetched on every poll inside the `Eventually`, gated on the ConfigTarget's own Ready condition first, and the assertions now cover NumNUMANodes, NumSimultaneousThreads, Memory, and the SEV/SEVSNP/TDX/SMCPresent flags in addition to NumCPUCores.

.sdd/specs/001-class-policy-resize/tasks.md
```diff
@@ -98,11 +98,21 @@ Dependencies: none. All A-tasks may run in parallel `[P]`.
...
+- [x] T106 [S8.c] [vmop-3745] Integration tests — co-located in `controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller_test.go` as a second `Describe` (Label `EnvTest`+`VCSim`, per `testing-standards.md`'s single-file-per-package convention, not a `test/intg/` tree): syncMode=ConfigTarget happy path against a real vcsim-derived ConfigTarget; missing ConfigTarget → condition.
+- [x] T107 [S8.d] [vmop-3745] E2E test — extended `test/e2e/vmservice/vmservice/configpolicy/configpolicy.go`'s existing `Spec()`: policy spec (numCPUCores, memory) filled from the zone's real ConfigTarget after Zone creation; toggling syncMode=Disabled stops spec changes. Labeled `experimental` pending validation on real hardware, per `e2e-testing.md`.
+
+  > Implementation note: T104's path was changed from `pkg/vmconfig/policy/`
+  > to `pkg/util/configpolicysync/` — `pkg/vmconfig/policy/` already exists
+  > and implements unrelated per-VM tag/PolicyEvaluation reconciliation
+  > (`Reconcile(ctx, k8sClient, vimClient, vm)`), and `pkg/util/configpolicy`
```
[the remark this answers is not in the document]
faisalabujabal: Fixed — un-wrapped the blockquote to a single line.

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,361 @@
...
+			requests = append(requests, policiesReferencingZone(ctx, c, zone.Namespace, zone.Name)...)
+		}
+
+		return requests
+	}
+}
+
+// policiesReferencingZone lists every VirtualMachineConfigPolicy in
+// namespace whose spec.zone equals zoneName.
+func policiesReferencingZone(
```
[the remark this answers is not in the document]
faisalabujabal: Fixed — added the godoc comment.

pkg/util/configpolicysync/configpolicysync.go
```diff
@@ -0,0 +1,229 @@
...
+
+// mergeIntRangeMax returns a copy of existing with Max set to maxVal, or a
+// new IntRange{Max: maxVal} if existing is nil. Min is left untouched: it is
+// not derived from any ConfigTarget, and is a tenant-managed floor.
+func mergeIntRangeMax(existing *vimv1.IntRange, maxVal int32) *vimv1.IntRange {
+	out := vimv1.IntRange{Max: maxVal}
+	if existing != nil {
+		out.Min = existing.Min
+	}
+
```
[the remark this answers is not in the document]
faisalabujabal: I don't think this is safe to apply, and I want to flag why rather than silently skip it.

A separate review pass on this PR found that `getConfigTargets` now requires every ConfigTarget behind a zone to be Ready before Merge runs at all — a zero here is real reported data (e.g. a zero `NumNumaNodes` explicitly means the cluster doesn't support NUMA alignment, per `VirtualMachineConfigPolicySpec`'s field docs), never "not populated yet."

Given that, collapsing `{Max: 0}` down to `nil` when `existing` is nil would flip a legitimate capability *denial* into a *no restriction specified* signal for whatever consumes this policy's spec downstream (S9 enforcement). That's the exact fail-open direction the other review pass was trying to close, just one level lower. I'm leaving this as `&IntRange{Max: 0}` rather than `nil`.

pkg/util/configpolicysync/configpolicysync.go
```diff
@@ -0,0 +1,229 @@
...
+// mergeResourceQuantityRangeMax returns a copy of existing with Max set to
+// maxVal, or a new ResourceQuantityRange{Max: maxVal} if existing is nil.
+// Min is left untouched; see mergeIntRangeMax.
+func mergeResourceQuantityRangeMax(
+	existing *vimv1.ResourceQuantityRange, maxVal resource.Quantity) *vimv1.ResourceQuantityRange {
+	out := vimv1.ResourceQuantityRange{Max: maxVal}
+	if existing != nil {
+		out.Min = existing.Min
+	}
```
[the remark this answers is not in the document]
faisalabujabal: Same reasoning as the comment on mergeIntRangeMax above — leaving this as-is.

faisalabujabal:
Two findings from a test-design review of the epic (written up in `.sdd/specs/001-class-policy-resize/tds.md` § 15 as CHK-7 and CHK-8). Both are in code that's still in review here, so they're cheaper to address now than after merge.

To be clear up front: the merge logic is **correct**. `mergeIntRangeMax` overwrites `Max` rather than taking `max(existing, new)`, which is exactly right — a policy must narrow when its cluster's capability drops. The findings are about that correctness being unprotected, and about one edge the overwrite creates.

---

### 1. The sync is never driven in the shrinking direction

Every test in `pkg/util/configpolicysync/configpolicysync_test.go` drives `Max` upward or sideways:

- *"copies its capacity limits and security flags into spec, preserving non-derived fields"*
- *"intersects numeric ranges to the minimum of the per-target maxima"*
- *"preserves an existing Min on a range field"*
- *"ANDs boolean support flags"*

None of them starts from a policy whose `Max` is **higher** than what the `ConfigTarget` now reports and asserts that it comes down.

That's the direction with security consequences. A future refactor to `mergeIntRangeMax` returning `max(existing, maxVal)` — which reads like a perfectly reasonable "don't clobber the admin's value" change — would leave a permanently over-permissive policy after a cluster is downgraded or loses hosts, and would pass this entire suite.

**Suggested:** one unit case per range type — existing spec has `Max: 64`, target reports `32`, assert the result is `32`. Plus the controller-level equivalent, since that's where a real cluster shrink arrives.

---

### 2. A shrink can produce an inverted range, and nothing rejects it

`mergeIntRangeMax` overwrites `Max` while deliberately **preserving** an admin-set `Min`:

```go
func mergeIntRangeMax(existing *vimv1.IntRange, maxVal int32) *vimv1.IntRange {
	out := vimv1.IntRange{Max: maxVal}
	if existing != nil {
		out.Min = existing.Min
	}
	return &out
}
```

Preserving `Min` is the right call — it's tenant intent, not `ConfigTarget`-derived data. But if the cluster's capability drops below that `Min`, the sync writes `Min > Max`.

Nothing catches it: `IntRange`, `LongRange`, and `ResourceQuantityRange` in `external/vim/api/v1alpha1/range_types.go` carry no `Min <= Max` validation, and #1783 doesn't add a CEL rule for it either. `mergeResourceQuantityRangeMax` has the same shape.

Concretely: tenant admin sets `numCPUCores.min = 8` on a cluster reporting `max = 64`; the cluster is later downgraded to `max = 4`; the policy silently becomes `{min: 8, max: 4}`. Whatever consumes that range downstream — the vmop-3746 admission gate — gets a range that admits nothing, or admits everything, depending on how it's read.

**Two defensible resolutions, either is fine, but pick one deliberately:**

- **Clamp** — `out.Min = min(existing.Min, maxVal)`, and surface the clamp on a condition so the admin can see their floor was lowered. Keeps the policy usable.
- **Reject** — leave the write, add a `Min <= Max` CEL rule, and let the controller surface `Ready=False` with a reason. Keeps the admin's intent visible and makes the conflict loud.

Clamping silently is the one option I'd avoid — it changes tenant intent with no signal.

---

Everything else in this PR's test set reads well. Two I'd specifically call out as worth keeping exactly as written, because they're the subtle ones:

- *"does not sync from the single Ready target -- that would be wider than the true intersection"* — this is the safety-critical case in a multi-cluster zone and it's easy to "simplify" away later.
- *"intersects a zero value on one target literally, as real reported data"* — the zero-vs-unset distinction is exactly the kind of thing that gets refactored into a `if v != 0` guard by someone being helpful.

faisalabujabal:
Both addressed in 2faf24ab.

**1. Shrinking-direction coverage** — added a unit case in `configpolicysync_test.go` ("narrows Max when the cluster's reported capability has shrunk") covering all four range fields, plus the controller-level equivalent in `vmconfigpolicy_controller_test.go` ("a cluster's reported capability has shrunk since the last sync"). Both assert Max comes down while Min survives untouched.

**2. Inverted range** — went with **reject**, not clamp, per your own call-out that silent clamping changes tenant intent with no signal. `Merge` now returns `(spec, error)`; when a merge would push Max below an existing Min on any range field, it returns the original spec unchanged and a descriptive error instead of publishing `{min: 8, max: 4}`. The controller surfaces that as `Ready=False` / reason `InvalidRange`, consistent with how `ZoneNotFound` and `ConfigTargetNotReady` already work — no new condition-handling pattern introduced.

Didn't add a CEL `min <= max` rule to `IntRange`/`ResourceQuantityRange` in `range_types.go`, since these range types are shared across every consumer of that struct (not just this policy's ConfigTarget-derived fields), and the rejection in `Merge` already makes it impossible for *this* controller to ever write an inverted range — happy to add the CEL rule too as defense-in-depth if you think it's warranted for admin-authored writes that bypass this controller entirely (e.g. an admin directly setting an inverted range on create/update).

Kept your two flagged tests exactly as written — didn't touch them.

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(controlledType).
+		Watches(
+			&topologyv1.Zone{},
+			handler.EnqueueRequestsFromMapFunc(zoneToPolicyMapper(r.Client))).
+		Watches(
+			&vimv1.ConfigTarget{},
+			handler.EnqueueRequestsFromMapFunc(configTargetToPolicyMapper(r.Client))).
```
hpannem, line 153: what about sync disabled cases?

Does it also disable sync for zone changes?

hpannem (commented):
general question: what happens if policy changed and existing vm is not compatible with the policy changes?

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,410 @@
...
+	err := c.List(ctx, &policies,
+		client.InNamespace(namespace),
+		client.MatchingFields{policyByZoneIndex: zoneName})
+	if err != nil {
+		return nil
+	}
+
+	requests := make([]reconcile.Request, len(policies.Items))
+	for i := range policies.Items {
+		requests[i] = reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&policies.Items[i])}
```
hpannem, line 224: we can filter disbaled cases here

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,410 @@
...
+
+			return nil
+		}
+
+		pkgcond.MarkError(obj, vimv1.ReadyConditionType, ZoneNotFoundReason, err)
+
+		return fmt.Errorf("failed to get zone %q: %w", obj.Spec.Zone, err)
+	}
+
+	if len(zone.Spec.ManagedVMs.ClusterMoIDs) == 0 {
```
hpannem, line 336: Not sure what should be the behavior? The policy that is sync'd has values. However, the zone is has no clusters. What happens to policy values? they are stale. Should these be reverted?

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,410 @@
...
+			Reason: SyncDisabledReason,
+		})
+		obj.Status.ObservedGeneration = obj.Generation
+
+		return nil
+	}
+
+	zone, err := topology.GetZone(ctx, r.Client, obj.Spec.Zone, obj.Namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
hpannem, line 323: similar case here... stale values?

faisalabujabal:
## Claude Code review — head `2faf24ab` (delta since the last review pass at `4b0770e3`)

The earlier pass on this PR reviewed through `4b0770e3`; this covers the two commits since — `91e64440` (test determinism) and `2faf24ab` (range inversion) — and then takes a run at @hpannem's four open questions, since they're all answerable from the code as it stands.

`2faf24ab` is the right fix for CHK-7/CHK-8 and the test coverage is genuinely good: the shrink-direction spec asserts all four derived Maxes come *down* while every tenant Min survives, which is exactly the mutation that would otherwise pass a whole suite silently. One real gap in it, below.

### 🟠 `InvalidRange` collides with the "zero Max means not supported" convention

`validateRanges` flags any derived range where `Min > Max`. But `ConfigTargetNotReadyReason`'s own doc comment in this file, and `getConfigTargets`'s, both rest on the premise that **a zero maximum is real reported data meaning "not supported"** — not "no data yet." (You made that argument yourself in this thread when declining the bot's `mergeIntRangeMax` nil suggestion, and it's why `getConfigTargets` requires Ready before merging at all.)

Put those together: a cluster that legitimately reports `NumNumaNodes: 0` — NUMA not supported — against a policy where a tenant has set `numNUMANodes.min: 1` now produces `0 < 1`, which `validateRanges` treats as an inversion. `Merge` returns `spec` untouched, and the controller stops. That doesn't just freeze the offending field: it freezes **the entire sync for that policy** — devices, TDX, SEV, every other cluster capability — indefinitely, until a human notices `Ready=False` and lowers the Min. Before `2faf24ab` the same input published `Min: 1, Max: 0`, which is also wrong, but only for that one field.

Every new spec drives the inversion with a non-zero narrowed Max (`NumCPUCores: 4` against `Min: 8`), so the `Max == 0` case is untested and the distinction is invisible in the suite. Three ways out, in rough order of how well they fit the rest of this design:

1. Apply the merge and zero (or leave stale) only the inverted field, reporting the conflict per-field in the condition message. Keeps every unrelated capability converging.
2. Keep all-or-nothing, but say so in the condition message and name the remedy — an admin reading `numNUMANodes: min (1) exceeds max (0)` has no way to know their vGPU profile list also stopped updating.
3. Treat `Max == 0` as "unsupported" and exempt it from the inversion check, so only a genuine narrowing-below-Min trips it. Still needs a distinct signal, since `Min: 1` on an unsupported field is a real conflict — just a different one.

### 🔵 A CEL rule would keep the Go check honest about what it's for

`validateRanges` runs over the *merged* output, so it also fires when a tenant hand-sets `Min > Max` at create time on a derived field — same permanent wedge, before any cluster ever narrows. A `self.min <= self.max` `XValidation` rule on `IntRange` / `ResourceQuantityRange` would reject that at admission and leave this Go check meaning only what its doc comment says it means (a cluster-side narrowing). That's also the direction #1779 amended the constitution toward and #1785 is applying to the sibling CRDs.

### 🔵 `91e64440` — trade-off is right, note what it gives up

Dropping the real `configtarget` controller and the vSphere provider from the watch test to drive `ConfigTarget` Ready via `Status().Update()` is the correct call: an unbounded vcsim query in the middle of a watch test is testing the wrong thing, and finding the nil `initProvidersFn` panic along the way was worth the trip. What's no longer covered anywhere in this suite is the real chain — `configtarget` controller populates status → policy syncs — which now lives only in T107's E2E. Worth one line in the PR description so the next person doesn't re-add the vcsim dependency looking for it.

---

## On @hpannem's open questions

**"What about sync disabled cases? Does it also disable sync for zone changes?" (:134) / "we can filter disabled cases here" (:212)**

`zoneToPolicyMapper` and `configTargetToPolicyMapper` enqueue every policy whose `spec.zone` matches, regardless of `syncMode`; the Disabled check is the first thing in `ReconcileNormal`. So sync itself is correctly disabled — `spec` is never written — but every Zone or ConfigTarget event still wakes every disabled policy in the namespace **and patches it** (`ObservedGeneration` plus the `SyncDisabled` condition go through the deferred `patchHelper`). That's a write per disabled policy per cluster-capability change.

@hpannem's filter suggestion is worth taking, and it's cheap: fold the predicate into `policyByZoneIndex` so the mapper's `List` is server-side filtered — index a zone name only when `policy.Spec.SyncMode != Disabled`. A policy flipping Disabled→ConfigTarget still reconciles immediately via `For(controlledType)` catching its own update, so nothing is lost. This gets materially more important if S11/T130 lands option 1 (`syncMode=Disabled` as the Zone controller's default), where the *typical* policy is disabled.

**"Stale values?" (:311, :324) — zone with no clusters, or a not-Ready ConfigTarget**

Confirmed: `spec` keeps whatever was last synced, and the only signal is `Ready=False` with `ConfigTargetNotFound` / `ConfigTargetNotReady`. Leaving it is the better half of the choice — reverting to zeros would mean "nothing supported" per the field docs, so a transient ConfigTarget blip would deny every VM operation in the namespace, which is far worse than briefly-stale maxima.

But the asymmetry deserves to be stated rather than left emergent, because it points at a real gap outside this PR: **stale values are more permissive than reality**, and this spec is what S9's admission webhook (T110–T117) will enforce against. A zone stuck at `Ready=False` therefore leaves an over-permissive policy in force, with nothing at the enforcement point noticing. `tds.md` GAP-5 already tracks the fail-closed case for a missing/not-Ready `ConfigTarget` — the answer to @hpannem's question is probably "stale by design, and S9 must gate enforcement on `Ready=True`", which is worth writing into the CRD field docs and GAP-5 now while the reasoning is fresh, rather than rediscovering it during S9.

pkg/util/configpolicysync/configpolicysync.go
```diff
@@ -0,0 +1,268 @@
...
+	targets ...vimv1.ConfigTargetStatus) (vimv1.VirtualMachineConfigPolicySpec, error) {
+	if len(targets) == 0 {
+		return spec, nil
+	}
+
+	agg := aggregate{
+		numCPUCores:            targets[0].NumCPUCores,
+		numNUMANodes:           targets[0].NumNumaNodes,
+		numSimultaneousThreads: targets[0].MaxSimultaneousThreads,
+		memory:                 quantityOrZero(targets[0].SupportedMaxMem),
```
hpannem: if nil, 0 or ignore? if 0, all other cluster values are discarded?

pkg/util/configpolicysync/configpolicysync.go
```diff
@@ -0,0 +1,268 @@
...
+// value-for-value, on every target.
+//
+// Fields the policy spec owns independently of any ConfigTarget --
+// createMode, updateMode, powerOnMode, vmClassMode, extraConfig,
+// latencySensitivityLevels, txRxThreadModels, and the *Supported flags with
+// no ConfigTargetStatus counterpart (cpuLockedToMaxSupported,
+// memoryLockedToMaxSupported, hugePagesSupported, iommuSupported,
+// rssSupported, udpRSSSupported, lroSupported) -- are left untouched by
+// Merge.
+package configpolicysync
```
hpannem, line 22: can't we keep this in the same package as configpolicy?

pkg/util/configpolicysync/configpolicysync.go
faisalabujabal: Confirmed a real bug here, thanks for pushing on this. The seed `quantityOrZero(targets[0].SupportedMaxMem)` turned a first target that simply didn't report `SupportedMaxMem` (`+optional`) into a literal zero — and since nothing after the seed can raise the running minimum back up, that zero silently discarded every other target's real reported value for the rest of the multi-cluster fold, regardless of how many clusters actually reported real data.

Fixed in `3411096c`: `aggregate.memory` is now `*resource.Quantity` so nil (not reported) and a real zero stay distinguishable through the whole fold — matching `minInt32`'s existing zero-is-real-data convention for the sibling fields, which was never the ambiguity here. If *every* target omits `SupportedMaxMem`, `Memory` is now left untouched entirely rather than set to zero. Added regression tests for both cases.

pkg/util/configpolicysync/configpolicysync.go
faisalabujabal: Fair question — the short answer is timing, not a permanent design decision. `pkg/util/configpolicy` is the name PR #1738 introduces for the enforcement/matching side (webhook + power-on reconcile checking a VM against `policy.spec`), and #1738 is still open/unmerged. This package (`configpolicysync`) is the producer side — it only computes what goes into `policy.spec` from `ConfigTarget.status`, with no enforcement logic and no dependency on #1738's types.

I named it separately to avoid a collision (two PRs independently claiming `pkg/util/configpolicy`) rather than because the two are conceptually unrelated — they're tightly coupled (one produces the spec, the other consumes it for enforcement), so once #1738 merges it's worth revisiting whether these should actually live together, e.g. `pkg/util/configpolicy/sync.go` alongside the enforcement code. Flagging as a deliberate follow-up rather than doing it now, since folding this PR's package into #1738's before #1738 has even merged would create the dependency this PR was specifically designed to avoid.

faisalabujabal:
Addressed in `3411096c`.

**🟠 InvalidRange collision** — you're right, and the failure mode was exactly as described: a genuine `Max: 0` (not supported) against an existing `Min > 0` froze the *entire* policy's sync, which is worse than the bug it replaced. Went with your option 1: `Merge` now applies every field independently — a field whose derived range would invert is left unchanged and named in the returned error, but every other field (including other range fields) still converges normally. The controller still applies the merged spec on this path and only uses `Ready=False`/`InvalidRange` to surface the conflict. Added a test proving an unrelated field (Memory) still converges in the same `Merge` call despite a `NumCPUCores` conflict, plus a multi-field-conflict test asserting the joined error names both.

**🔵 CEL rule** — holding off for now. Per-field skip already makes it impossible for *this controller* to ever write an inverted range, so the CEL rule would only catch a tenant hand-setting `min > max` on a derived field at create time — a real but narrower gap. Given `IntRange`/`ResourceQuantityRange` in `range_types.go` are shared beyond this policy's ConfigTarget-derived fields, I'd rather raise it as its own follow-up than fold a schema-wide change into this PR's fix.

**🔵 91e64440 coverage note** — noting here rather than editing the PR description after the fact: the real `configtarget` controller → policy sync chain is no longer covered by any test in this suite; it lives only in T107's E2E spec (`test/e2e/vmservice/vmservice/configpolicy/configpolicy.go`).

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,426 @@
...
+	ConfigTargetNotFoundReason = "ConfigTargetNotFound"
+
+	// ConfigTargetNotReadyReason is the Ready condition reason used when
+	// one or more of the zone's cluster MoIDs do not resolve to a Ready
+	// ConfigTarget. spec is left untouched in this case: a ConfigTarget
+	// with Ready!=True has not finished populating its status, and merging
+	// its zero-valued fields would publish a false capability denial rather
+	// than "unknown," since a zero maximum on a numeric field means "not
+	// supported" per VirtualMachineConfigPolicySpec's doc comments.
+	ConfigTargetNotReadyReason = "ConfigTargetNotReady"
```
aruneshpa, line 75: Typically, condition reasons are part of the API.  Is there a reason why you are deviating from that pattern that we've established?

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,426 @@
...
+	// converges normally, so this reason does not mean spec as a whole is
+	// untouched.
+	InvalidRangeReason = "InvalidRange"
+)
+
+// SkipNameValidation is used for testing to allow multiple controllers with the
+// same name since Controller-Runtime has a global singleton registry to
+// prevent controllers with the same name, even if attached to different
+// managers.
+var SkipNameValidation *bool
```
aruneshpa, line 92: This is almost never a good idea to have a singleton variable because it can lead to unexpected results if the state is changed by one test while another test is consuming it.
Can we think of an alternative way to achieve this?

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,426 @@
...
+		ctx,
+		mgr.GetClient(),
+		ctrl.Log.WithName("controllers").WithName(controlledTypeName),
+		record.New(mgr.GetEventRecorder(controllerNameShort)))
+
+	err := mgr.GetFieldIndexer().IndexField(
+		ctx,
+		&vimv1.VirtualMachineConfigPolicy{},
+		policyByZoneIndex,
+		func(rawObj client.Object) []string {
```
aruneshpa, line 113: Going forward, we want to define the indexer funcs separately and simply specify them here.  Kind of like: https://github.com/vmware-tanzu/vm-operator/blob/main/controllers/virtualmachinesnapshot/virtualmachinesnapshot_controller.go#L61-L67
(possibly even moving the entire IndexField call in the kube package).

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,426 @@
...
+			policy := rawObj.(*vimv1.VirtualMachineConfigPolicy) //nolint:forcetypeassert
```
aruneshpa, line 114: Is this new linter?  We do this in all of the existing indexer functions and I have not seen our linter complain.

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,426 @@
...
+		})
+	if err != nil {
+		return fmt.Errorf("failed to index VirtualMachineConfigPolicy by spec.zone: %w", err)
+	}
+
+	err = mgr.GetFieldIndexer().IndexField(
+		ctx,
+		&topologyv1.Zone{},
+		zoneByClusterMoIDIndex,
+		func(rawObj client.Object) []string {
```
aruneshpa, line 138: Same comment here.  Consider moving to the kube package.

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,426 @@
...
+func configTargetToPolicyMapper(c client.Client) handler.MapFunc {
+	return func(ctx context.Context, o client.Object) []reconcile.Request {
+		ct, ok := o.(*vimv1.ConfigTarget)
+		if !ok {
+			return nil
+		}
+
+		var zones topologyv1.ZoneList
+
+		err := c.List(ctx, &zones, client.MatchingFields{zoneByClusterMoIDIndex: ct.Name})
```
aruneshpa, line 191: Just curious: if there are 100 zones, but only two namespace - and only one of the zones is associated with the namespace, then this would still return all 100 resources, correct?  Is there a way to optimize that or is it intentional?

controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go
```diff
@@ -0,0 +1,426 @@
...
+	var policies vimv1.VirtualMachineConfigPolicyList
+
+	err := c.List(ctx, &policies,
+		client.InNamespace(namespace),
+		client.MatchingFields{policyByZoneIndex: zoneName})
+	if err != nil {
+		return nil
+	}
+
+	requests := make([]reconcile.Request, len(policies.Items))
```
aruneshpa, line 222: I have a subtle preference for using range loops here, I think that's a bit more readable.

```golang
requests := make([]recocnile.Request, 0, len(policies.Items))
for _, policy := range policies.Items {
    requests = append(requests, 
        reconcile.Request{NamespacedName: client.ObjectKeyFromObject(policy),
    )
}

File: controllers/virtualmachineconfigpolicy/vmconfigpolicy_controller.go [hunk 19]
```diff
@@ -0,0 +1,426 @@
...
+
+		return nil
+	}
+
+	zone, err := topology.GetZone(ctx, r.Client, obj.Spec.Zone, obj.Namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			pkgcond.MarkFalse(obj, vimv1.ReadyConditionType, ZoneNotFoundReason,
+				"zone %q not found", obj.Spec.Zone)
+			obj.Status.ObservedGeneration = obj.Generation
```
aruneshpa, line 326: The semantic of when the stauts.observedGen is set vs not set is confusing to me.  Can you explain that please?  In some failure cases, we seem to set this, others we don't.  If we unconditionaly need to do this (it's replicated all over) then let;s move it in the patch helper for this controller?

pkg/util/configpolicysync/configpolicysync.go
```diff
@@ -0,0 +1,278 @@
...
+
+	return agg
+}
+
+// minInt32 returns the smaller of a and b. Callers are expected to only pass
+// status from a Ready ConfigTarget (see the Reconciler's getConfigTargets),
+// so a zero value here is real reported data, not "not populated yet" --
+// e.g. a zero MaxSimultaneousThreads means the cluster does not support
+// HT/SMT, per VirtualMachineConfigPolicySpec's field docs.
+func minInt32(a, b int32) int32 {
```
aruneshpa, line 151: We at least need to call out why the builtin `min` is not sufficient here.

aruneshpa (commented):
Took a first pass.  Looks generally okay.

PR #1785, opened 2026-07-30
vmop-3766: Move ConfigTarget/VirtualMachineConfigOptions validation to CEL

faisalabujabal:

Follow-up to #1779 (vmop-3766, `VirtualMachineGuestOptions` validation webhook), which amended `.sdd/memory/constitution.md` to prefer CEL for plain, unconditional structural validation and reserve Go webhook code for complex/cross-field/vSphere-data-dependent rules. That PR applied the change to `VirtualMachineGuestOptions`'s `spec.id`. This PR applies the same change to its two siblings, `ConfigTarget` and `VirtualMachineConfigOptions`, whose webhooks (merged in #1711 and #1672, respectively) had the exact same shape of checks: plain scalar immutability, non-emptiness, format, and name-equality — nothing that needed Go.

**ConfigTarget** (webhook was in #1711):

- `spec.id` (`ManagedObjectID`) immutability → `self == oldSelf` CEL transition rule. Compares the whole `ManagedObjectID` struct (it has an optional `ServerID` field alongside `ID`), matching exactly what the Go check it replaces already compared — pinned by a test that mutates only `ServerID` and still expects `id is immutable`.
- `spec.id.id` non-emptiness → `size(self.id) > 0` field-level CEL rule.
- `metadata.name` must match `^domain-c[0-9]+$` → root-level CEL rule (`self.metadata.name.matches(...)`). Verified controller-gen correctly emits `x-kubernetes-validations` referencing `self.metadata.name` at the object root before relying on this — Kubernetes CRD CEL rules support `self.metadata.name`/`self.metadata.generateName` specifically at that level.

Those three checks were this webhook's entire validator. With nothing left for it to do, **the webhook is removed outright**: the validator package, its `AddToManager` registration in `webhooks/webhooks.go`, and its entry in `config/webhook/manifests.yaml`. RBAC for `configtargets` is unaffected — it's owned by `controllers/configtarget` and `controllers/infra/zone`, not this webhook.

**VirtualMachineConfigOptions** (webhook was in #1672):

- `spec.hardwareVersion` immutability → `self == oldSelf` CEL transition rule.
- `spec.hardwareVersion` format → `self.matches('^vmx-[0-9]+$')` field-level CEL rule, which subsumes the old non-empty check.
- `metadata.name == spec.hardwareVersion` → root-level CEL rule.

Same outcome: this was its entire validator, so **the webhook is removed outright** the same way. RBAC for `virtualmachineconfigoptions` is unaffected — it's owned by `controllers/virtualmachineconfigoptions`, `controllers/configtarget`, and `controllers/infra/zone`.

Also removed the now-dead `test/builder.DummyVirtualMachineConfigOptions`, whose only caller was the deleted webhook test.

**Test coverage — restored, not just moved:**

Deleting each webhook package also deleted its `*_validator_intg_test.go` file, which took real envtest coverage of these checks along with it — even though the checks themselves are still real, user-visible CRD behavior, just enforced by CEL instead of a webhook now. That gap is closed in this PR, not left open:

- Added a `Describe("CRD validation", ...)` block to each controller's own envtest suite (`controllers/configtarget`, `controllers/virtualmachineconfigoptions`) that exercises the CEL rules directly against a real API server via `suite.NewIntegrationTestContext()` — independent of any controller reconcile or webhook — asserting on the exact CEL rejection messages for immutability, non-emptiness/format, and the name-match rules.
- This surfaced a real, pre-existing gap in `controllers/configtarget`'s suite: it used a plain `builder.NewTestSuite()`, whose default context never enables `Features.VirtualMachineConfigPolicy`. `vimv1.AddToScheme` is gated behind that exact flag in `pkg/manager/manager.go`, so the manager's typed Go client couldn't (de)serialize `ConfigTarget` at all — even though the CRD itself is installed in the real envtest API server regardless of the flag (`CRDDirectoryPaths` installs manifests unconditionally). Fixed with `builder.NewTestSuiteWithContext(ctx)` (the no-controller counterpart of `NewTestSuite()`) and the feature flag enabled in that context, matching the sibling `virtualmachineconfigoptions` suite, which already did this. (The suite's existing vcsim-backed tests never caught this because they run against a fake client, which doesn't enforce CRD schema/CEL validation at all.)
- `controllers/virtualmachineconfigoptions`'s `intgTests()` was previously an empty placeholder — filled in with the same pattern.

**Bug fix surfaced by restoring that coverage — predates this PR, not a CEL consequence:**

Review on this PR (see thread below) caught that `configtarget_controller.go`'s `reconcileConfigOptions` fan-out only skipped descriptors with an *empty* `Key`, so a malformed one (e.g. `"not-a-version"`) would fan out a `VirtualMachineConfigOptions{Name: "not-a-version"}` — which the deleted `#1672` webhook already rejected on CREATE via its `^vmx-\d+$` format check, and which the new CEL rule rejects the same way. This bug predates the CEL retrofit; CEL just makes it easier to notice with real envtest coverage in place. Fixed by extracting the malformed-key check `computeMaxHardwareVersion` already used into an exported `configtarget.IsValidHardwareVersionKey` and applying it in the fan-out loop too, so both call sites agree on what counts as a usable descriptor. A unit test now pins that a malformed key produces neither a `status.maxHardwareVersion` value nor a `VirtualMachineConfigOptions` object.

The same review also caught that `test/e2e/vmservice/vmservice/configpolicy/configpolicy.go`'s garbage-collection spec created its synthetic stale fixture as `vmx-e2e-stale-vmop-3760` — a name the `#1672` webhook already rejected on CREATE for the same reason, so that spec could only ever have run green against a supervisor where the `vim.vmware.com` webhooks weren't actually being enforced. Renamed to `vmx-9999`, which satisfies the format rule while staying far above any real hardware version (max is `vmx-22` today), preserving the test's intent of injecting staleness directly.

**User-visible effects:**
- An immutable/required/format/name-match rejection on either CRD now comes from the API server (CEL) instead of a webhook admission response, with the message text changed to match (e.g. `field is immutable` → `id is immutable` / `hardwareVersion is immutable`).
- The `configtarget` and `virtualmachineconfigoptions` validating webhooks no longer exist. Nothing is lost — every check they made is now enforced by the CRD schema itself; behavior is equivalent, just enforced one layer lower in the stack.
- `ConfigTarget`'s `reconcileConfigOptions` fan-out no longer fails an entire reconcile when a descriptor has a malformed `Key`; it now skips just that descriptor, matching `status.maxHardwareVersion`'s existing behavior.

**Verified, not just written:**
- Regenerated `config/crd/external-crds/*.yaml`, `config/webhook/manifests.yaml`, and `config/rbac/role.yaml` — diffs are exactly the expected CEL rule additions and webhook-entry removals. RBAC regeneration produced no diff, confirming those grants come from the controllers, not the removed webhooks.
- Ran the `configtarget` (20/20) and `virtualmachineconfigoptions` (19/19) envtest suites repeatedly (including after the CRD-validation and bug-fix additions) — clean in isolation. Also ran `infra/zone`'s suite to confirm the controllers that actually create these objects (all gated behind the same `VirtualMachineConfigPolicy` feature flag the webhooks were) already produce CEL-compliant name/id/hardwareVersion values.
- `go build`, `go vet`, and `golangci-lint` clean across the root module, `external/vim/api`, and `test/e2e`.

**Note on the constitution amendment:** `.sdd/memory/constitution.md` in this diff is a stacking artifact, not a second amendment — it's the identical hunk from #1779 (the `spec.id` immutability rewording plus the `Last updated` bump), inherited because this branch is built on top of that commit. It will disappear from this PR's diff once #1779 merges and this branch rebases past it. No new constitutional rule is being introduced or justified here; the rationale for the one amendment lives entirely in #1779's description.

Fixes #

- This PR depends on #1779 having established the CEL-for-simple-structural-rules precedent (see that PR's description for the full rationale and the review comment from @hpannem that started this).
- No E2E changes beyond the one fixture rename above: `ConfigTarget` and `VirtualMachineConfigOptions` materialization is already covered by `test/e2e/vmservice/vmservice/configpolicy/configpolicy.go`; this PR only changes how their existing validation is enforced, not what's validated.
- If you run `./controllers/configtarget`, `./controllers/virtualmachineconfigoptions`, and `./controllers/infra/zone` together in one `hack/test.sh` invocation, you may see `kube-apiserver` fail to start on retry ("timeout waiting for process kube-apiserver to start successfully") — that's local envtest resource contention from stacking three separate control-plane spin-ups back to back, not a real failure. Each suite is clean run individually.
- Open question raised in review, still being run down: whether the `vim.vmware.com` validating webhooks (`#1711`/`#1672`, and by extension `#1779`/`#1783`) were ever actually reaching admission on the supervisor the E2E suite runs against, given the stale-fixture finding above. Will follow up on this thread once confirmed either way; it doesn't block this PR's correctness either way (CEL enforcement doesn't depend on the answer).

```release-note
Remove the ConfigTarget and VirtualMachineConfigOptions validating webhooks; their immutability, format, and name-match checks are now enforced by CRD CEL validation rules instead. Also fixes a pre-existing bug where a malformed hardware-version descriptor from vCenter could fail an entire ConfigTarget reconcile instead of being skipped.
```


faisalabujabal:
Two findings from a test-design review of the epic (written up in `.sdd/specs/001-class-policy-resize/tds.md` § 15 as CHK-1 and CHK-18). Both are consequences of this PR's CEL retrofit that the existing test layers structurally cannot catch, so flagging them here rather than as follow-ups.

**Root cause for both:** the fake client used by the unit and vcsim suites enforces no CRD schema at all, so nothing below envtest can see a CEL violation coming. This PR already handles that correctly for the rules it introduces — the added `Describe("CRD validation", …)` envtest blocks are exactly right. These two cases are the ones outside that net.

---

### 1. The E2E garbage-collection spec will fail once this merges

`test/e2e/vmservice/vmservice/configpolicy/configpolicy.go:356-357` creates its synthetic stale fixture as:

```go
ObjectMeta: metav1.ObjectMeta{Name: "vmx-e2e-stale-vmop-3760"},
Spec:       vimv1.VirtualMachineConfigOptionsSpec{HardwareVersion: "vmx-e2e-stale-vmop-3760"},
```

That does not match the `self.matches('^vmx-[0-9]+$')` rule this PR adds to `spec.hardwareVersion`. The API server rejects the `Create`, and the spec fails before it asserts anything about garbage collection.

The PR description says it verified the controllers produce CEL-compliant values — that part holds, but the E2E suite wasn't in scope of that check.

**No chart bump buys time here.** These CRDs are `//go:embed`ed via `config/crd/crd.go` / `pkg/crd/crd.go` and listed in `config/default/kustomization.yaml`, so the schema ships inside the operator image. The E2E breaks as soon as it runs against an image built from this commit. envtest is exposed the same way — `test/builder/test_suite.go` points `CRDDirectoryPaths` at the same directory.

**Fix:** rename the fixture to something syntactically valid that a real cluster will never report, e.g. `vmx-9999`. The test's intent (inject staleness directly, since vcsim's EnvironmentBrowser reports a fixed descriptor set and a real cluster's supported set can't be shrunk on demand) is unaffected.

I checked for other offenders: the only other non-conforming fixtures are in the two webhook test files this PR already deletes. This is the one survivor.

---

### 2. The fan-out itself can now produce a rejected `Create`

`controllers/configtarget/configtarget_controller.go:219` — `reconcileConfigOptions` skips only *empty* descriptor keys:

```go
for _, d := range descriptors {
    if d.Key == "" {
        continue
    }
    ...
    vmco := &vimv1.VirtualMachineConfigOptions{ObjectMeta: metav1.ObjectMeta{Name: key}}
```

So a descriptor with `Key: "not-a-version"` produces a `Create` of `VirtualMachineConfigOptions{Name: "not-a-version"}`, which the same new CEL rule rejects — failing the entire reconcile, not just that one object.

The unit spec that covers this input (`configtarget_controller_test.go:597`, *"skips the malformed key without erroring"*) asserts only `maxHardwareVersion` and runs against the fake client, so it cannot observe the failure.

A real cluster returning a malformed key is unlikely, but the code explicitly tolerates one today and after this PR it no longer can.

**Fix:** filter the fan-out to well-formed keys, matching what `computeMaxHardwareVersion` in `pkg/util/vsphere/configtarget/status.go:90` already does — the two should agree on what counts as a usable descriptor.

---

While you're in there, one adjacent observation that isn't this PR's to fix but is worth knowing (CHK-5): `computeMaxHardwareVersion` filters on `CreateSupported == true`, but the fan-out doesn't filter on it at all. A cluster can therefore publish a `vmx-21` `VirtualMachineConfigOptions` while `status.maxHardwareVersion` is `vmx-19`. That may be correct — run-supported-but-not-create-supported is a real vSphere state — but it's undocumented and unasserted, and it lands on the vmop-3746 admission gate, where a user would see an object advertising a version admission then refuses.

faisalabujabal:
Both confirmed and fixed in dbf13abf:

1. Renamed the E2E stale fixture from `vmx-e2e-stale-vmop-3760` to `vmx-9999` — satisfies the `^vmx-[0-9]+$` CEL format rule while staying far above any real hardware version (max is `vmx-22` today).
2. Extracted the malformed-key check `computeMaxHardwareVersion` already used into an exported `configtarget.IsValidHardwareVersionKey`, and applied it in `reconcileConfigOptions`'s fan-out loop so a malformed descriptor `Key` is skipped there too instead of producing a `Create` the new CEL rule rejects. Strengthened the existing "malformed key" unit test to assert no `VirtualMachineConfigOptions` is created for it.

Left the CHK-5 observation (fan-out not filtering on `CreateSupported`) as-is per your note — real, but out of scope here.

Re-ran `./controllers/configtarget` (19/19) and `./controllers/virtualmachineconfigoptions` (19/19) envtest suites, plus `go vet`/`golangci-lint` on the touched packages and the `test/e2e` module build.

faisalabujabal:
## Claude Code review — head `dbf13abf`

Reviewed the whole diff (`+277/-1348`, 24 files). The CEL retrofit itself is sound: both `self == oldSelf` transition rules sit on required scalar/struct fields (so `oldSelf` always exists), `size(self.id) > 0` is safe because `ManagedObjectID.id` is itself `required` in the schema (an optional field would make `self.id` a CEL "no such key" runtime error rather than a clean rejection), and the two root-level rules using `self.metadata.name` are at the one level where Kubernetes actually supports that reference. The restored envtest coverage runs against a real API server, which is the only place these rules can be verified — the fake client enforces neither schema nor CEL. No lingering references to either removed webhook anywhere in `config/`, `webhooks/`, or `main.go`. All checks green.

Nothing blocking. Five notes, roughly in descending order of value.

### 🔵 The API-compatibility argument is correct but unstated

The constitution treats even additive changes to a shipped API as unsafe (k8s#111703), and this tightens validation on two CRDs that shipped in 9.2. The reason it's safe here is specific and worth saying out loud in the description: **the removed webhooks already enforced the identical checks on CREATE and UPDATE**, so no conforming object exists that these CEL rules would newly reject. That's a stronger claim than "behavior is equivalent" and it's the one a reviewer worried about upgrade needs.

One residual case the equivalence doesn't cover: an object created *by the old E2E* — `vmx-e2e-stale-vmop-3760`, which the pre-#1785 webhook allowed because `metadata.name` was only checked against `spec.hardwareVersion`, not against `^vmx-[0-9]+$`... (worth double-checking against #1672's validator; if it did enforce the format, this case can't exist and the note is moot). If any such object is sitting on a long-lived dev/test supervisor, it becomes permanently invalid: every patch to it, status or otherwise, is now rejected. The GC path mostly self-heals — with a single owner reference it takes the `Delete` branch, and deletes aren't schema-validated — but with two or more owners it patches owner references, which would fail on every reconcile of the owning `ConfigTarget` and wedge that reconcile with a permanent error. A line in the release note ("delete any pre-existing `VirtualMachineConfigOptions` whose name is not `vmx-<n>`") would close it.

### 🔵 `IsValidHardwareVersionKey` fixes a pre-existing bug, not one this PR introduces

Good catch and the right shape (one predicate shared by `computeMaxHardwareVersion` and the fan-out, with a unit spec pinning that a malformed key is excluded from *both* `maxHardwareVersion` and the `VirtualMachineConfigOptions` list). But the doc comment and description frame it as "spec.hardwareVersion carries a CEL rule ... would otherwise reject the Create" — the deleted webhook enforced the same `^vmx-\d+$` format, so a malformed descriptor `Key` would already have been rejected on CREATE before this PR. That makes this a real bug fix that predates the retrofit, which is more interesting than a refactor consequence and belongs in the release note.

### 🔵 `NewTestSuiteWithContext` is the smaller fix for the scheme-gating bug

Finding the `configtarget` suite couldn't (de)serialize `ConfigTarget` at all because `vimv1.AddToScheme` is gated on `Features.VirtualMachineConfigPolicy` is the most valuable thing in this PR. The fix, though, swaps `builder.NewTestSuite()` for `NewTestSuiteForControllerWithContext(ctx, manager.AddToManagerNoopFn, manager.InitializeProvidersNoopFn)` — a controller suite that registers no controller. `builder.NewTestSuiteWithContext(ctx)` (`test/builder/test_suite.go:154`) is the direct counterpart of `NewTestSuite()` and expresses the same thing without the two noop funcs.

### 🔵 The whole-struct immutability claim isn't pinned by a test

The description makes a specific point that `self == oldSelf` on `spec.id` compares the entire `ManagedObjectID`, including the optional `ServerID` alongside `ID` — matching what the Go check compared. Every new spec mutates `.ID` only. A one-line case mutating `ServerID` and expecting `id is immutable` would pin the claim; without it, a future narrowing of the rule to `self.id == oldSelf.id` would pass CI.

### 🔵 Description says it doesn't touch `constitution.md`, but the diff does

The body states "this PR does not re-touch `.sdd/memory/constitution.md` — that amendment landed in #1779", yet `.sdd/memory/constitution.md` is the first file in this diff, carrying a hunk byte-identical to #1779's (the `Last updated` bump plus the `spec.id` immutability rewrite). It's a stacking artifact, same as #1789's, and harmless on merge — but per the constitution's own rule that a PR bending a constitutional rule must call the amendment out in its description, the description should either claim the amendment or note that the hunk is inherited from #1779 and will vanish on rebase.

faisalabujabal:
Resolving the hedge in note 1 above — I checked the validator this PR deletes, and it changes the conclusion:

`webhooks/virtualmachineconfigoptions/validation/virtualmachineconfigoptions_validator.go` enforced `hardwareVersionPattern = ^vmx-\d+$` against `spec.hardwareVersion` on **CREATE** (`validateHardwareVersionFormat`, called from `ValidateCreate`), plus `metadata.name == spec.hardwareVersion`. So the equivalence is airtight and **the upgrade case I raised cannot exist** — no `VirtualMachineConfigOptions` named `vmx-e2e-stale-vmop-3760` could ever have been admitted on a supervisor running the webhook. Disregard the suggested release-note line; the "no conforming object can be newly rejected" argument stands on its own.

That does surface a different question, though, and it's the more interesting one: if the #1672 webhook already rejected that name on CREATE, then `test/e2e/.../configpolicy.go`'s `adminClient.Create(ctx, stale)` should have been failing on any supervisor where the `vim.vmware.com` webhooks were actually registered. `tasks.md` T206 / `tds.md` § 15 CHK-1 framed the rename as "that E2E spec breaks the moment #1785 merges" — but it should have been broken already. Two possibilities, and they have different consequences:

- The E2E hadn't been run green against a supervisor with those webhooks deployed, in which case the rename is still the right fix and nothing else follows.
- The `vim.vmware.com` validating webhooks were never actually reaching admission on the test supervisor (chart wiring, `failurePolicy`, namespace/object selectors, or the feature gate not registering them). If that's the case, then #1711's and #1672's webhook coverage was never enforced end-to-end in the first place — which is an argument *for* this PR, since CEL on the CRD can't be bypassed that way, but it's also worth knowing before you rely on the remaining vim webhooks (`virtualmachineguestoptions` from #1779, `virtualmachineconfigpolicy` from #1783) being live.

Cheap to settle: `kubectl get validatingwebhookconfiguration validating-webhook-configuration -o yaml | grep -A2 vim.vmware.com` on the test supervisor the E2E runs against.

faisalabujabal:
Addressed notes 2–4 from the last review pass in 052eb3e1:

- **Note 2 (framing):** reworded the `IsValidHardwareVersionKey` doc comments, its call site, and the unit test comment — the malformed-key fan-out bug predates this PR (the deleted `#1672` webhook already enforced `^vmx-\d+$` on CREATE), so it's not a consequence of moving the check to CEL. Updated the PR description and release note to say so.
- **Note 3 (simplification):** swapped in `builder.NewTestSuiteWithContext(ctx)` for the `NewTestSuiteForControllerWithContext(ctx, manager.AddToManagerNoopFn, manager.InitializeProvidersNoopFn)` call — confirmed it's the exact same thing under a name (`test/builder/test_suite.go:154`).
- **Note 4 (test gap):** added a case mutating only `ManagedObjectID.ServerID` (leaving `.ID` unchanged) and expecting `id is immutable`, pinning that `self == oldSelf` compares the whole struct.
- **Note 5 (stale description):** rewrote the constitution note to say what's actually true — the `.sdd/memory/constitution.md` hunk here is inherited from #1779 (stacking artifact), not a second amendment, and will disappear once this branch rebases past #1779 merging.

Left note 1 (whether the `vim.vmware.com` webhooks were ever actually enforced on the E2E test supervisor) for @faisalabujabal to run down directly — flagged as an open, non-blocking question in the updated PR description.

Re-ran `./controllers/configtarget` (20/20, includes the new ServerID case) and confirmed lint/build clean.

pkg/util/vsphere/configtarget/status.go
```diff
@@ -91,19 +91,33 @@ func computeMaxHardwareVersion(descriptors []vimtypes.VirtualMachineConfigOption
...
 	return maxVer.String()
 }
+
+// IsValidHardwareVersionKey reports whether key is a well-formed hardware
+// version (e.g. "vmx-21"), the same criterion computeMaxHardwareVersion uses
+// to skip malformed descriptors. Callers that fan out a
+// VirtualMachineConfigOptions per descriptor Key -- whose spec.hardwareVersion
+// must match that same ^vmx-\d+$ shape or the API server rejects the Create
+// -- must apply this filter too. That constraint predates this function:
+// the now-removed VirtualMachineConfigOptions webhook enforced the identical
```
hpannem, line 113: nit: do we need to say now-removed webhook etc.? We can treat it never existed right?

hpannem (approved):
do we need to update tasks.md with the cel based approach?

akutz (approved):
Thank you for taking care of this!

PR #1789, opened 2026-07-31
Add owner refs from guest options to config options

faisalabujabal:
**Draft — stacked on #1782, do not review until that merges.**

Follow-up to #1782 (vmop-3932). hpannem asked there why
`VirtualMachineGuestOptions` doesn't carry owner references back to the
`VirtualMachineConfigOptions` objects that report it, given
`controllers/configtarget` already uses that pattern one layer up
(`VirtualMachineConfigOptions` <- `ConfigTarget`). The status-based GC
landed in #1782 is correct on its own, but it depends on every
reconcile running to completion. Native Kubernetes garbage collection
gives the same end state for free if a `VirtualMachineConfigOptions` is
force-deleted or otherwise removed outside a normal reconcile.

Each contributing `VirtualMachineConfigOptions` now adds a
(non-controller) owner reference to every `VirtualMachineGuestOptions`
it fans out to, mirroring `configtarget`'s `reconcileConfigOptions`.
`status.hardwareVersions[]` is unchanged in shape and still carries all
the substantive per-hardware-version data (CPU/memory limits,
recommended controllers, etc.) -- an owner reference has no payload, so
it can't replace that field, only add a second, GC-visible signal
alongside it.

What *did* change from #1782: the deletion trigger in
`removeHardwareVersionAndDeleteIfOrphaned` moves from "no
`status.hardwareVersions` entries remain" to "no owner references
remain". In the normal path these two counts always move together (every
write path adds/removes both in the same pass), so this is not an
observable behavior change on its own -- the reason for the switch is
that gating on the owner-reference count makes a reconcile that fails
between the status patch and the owner-reference patch resume correctly
on retry (next pass finds no status entry, a no-op there, but still
finds the dangling owner reference and finishes removing it), instead of
silently leaking a reference forever.

**Reviewing this diff before #1782 merges**: because GitHub requires a
PR's base branch to exist in the base repository, and
`vmop-3932-gc-unused-guestoptions` only exists on my fork, this PR is
opened against `main` and its diff currently includes #1782's full
changeset. In particular, in `.sdd/specs/001-class-policy-resize/tasks.md`,
**T093a and T093b were completed in #1782, not here** -- only **T093c
and T093d** are new to this PR. Once #1782 merges this branch will be
rebased onto `main` and the diff will shrink to just the
owner-reference work (commit `36a972d5`).

Tracked under vmop-3932 (same ticket as #1782).

Hold review until #1782 merges. To review only the net-new work now,
look at commit `36a972d5` directly rather than the full PR diff.

```release-note
None
```


faisalabujabal:
## Claude Code review — head `36a972d5`, net-new commit `36a972d5` only (per the PR notes, #1782's changeset excluded)

Reviewed scope: `controllers/virtualmachineconfigoptions/vmconfigoptions_controller.go` and its `_test.go`. Both CRDs are cluster-scoped (`+kubebuilder:resource:scope=Cluster` on `VirtualMachineGuestOptions` and `VirtualMachineConfigOptions`), so a cluster→cluster non-controller owner reference is valid and won't trip the GC's "cluster-scoped dependent with namespaced owner" path. The fan-in write path is exactly what `.sdd/memory/operator-best-practices.md` prescribes: `MergeFromWithOptimisticLock`, `apiequality.Semantic.DeepEqual` for the skip-if-unchanged guard, status patched separately, `client.Preconditions` on the delete. All 20 checks green.

One real finding.

### 🟠 Switching the delete gate from status entries to owner references is not upgrade-safe

`removeHardwareVersionAndDeleteIfOrphaned` now deletes only when `hasOwnerRef && len(obj.OwnerReferences) == 1`. Every `VirtualMachineGuestOptions` created before this change has **zero** owner references, and there are two ways that bites during the upgrade window:

1. **Permanent leak.** A pre-existing object that is *already* stale at upgrade time (no longer reported by any hardware version) never gets its owner reference backfilled, because backfill happens in `reconcileGuestOptions`, which only runs for live descriptors. GC takes the `idx != -1` branch, removes the status entry, and skips the delete because `hasOwnerRef` is false. On every subsequent pass `idx == -1 && !hasOwnerRef` returns early — the object is leaked forever. Pre-change code (`len(obj.Status.HardwareVersions) == 1`) would have deleted it. These are precisely the objects #1782's GC exists to clean up.

2. **Premature delete of a still-live object.** Mirror image: `obj` has vmco's freshly-added owner reference (so `len(OwnerReferences) == 1`) but also a legacy status entry from a *second* `VirtualMachineConfigOptions` whose reconcile hasn't backfilled its own owner reference yet. The new gate deletes the shared object even though another hardware version still reports that guest OS. It's recreated on the other VMCO's next reconcile, but the delete is observable to consumers and the surviving version's status entry is lost until then. The old status-count gate (2 entries) would not have deleted it.

The retry-resumption argument in the new doc comment is sound — it just doesn't need to be exclusive. Gating on "no contribution of either kind remains" keeps that property and stays correct across the upgrade:

```go
lastOwnerRef := hasOwnerRef && len(obj.OwnerReferences) == 1
lastStatusEntry := idx != -1 && len(obj.Status.HardwareVersions) == 1
if lastOwnerRef && (lastStatusEntry || len(obj.Status.HardwareVersions) == 0) {
    // ... delete
}
if !hasOwnerRef && lastStatusEntry {
    // legacy object, never backfilled: fall back to the pre-#1789 gate
}
```
— or, more simply, keep `len(obj.Status.HardwareVersions) == 1 || (hasOwnerRef && len(obj.OwnerReferences) == 1)` as the delete condition and let the two signals agree in the steady state, which the PR description already argues they do.

### 🔵 Neither legacy branch is covered

The new unit specs construct every `VirtualMachineGuestOptions` through the new code, so both cases above are untested. A spec that seeds a `VirtualMachineGuestOptions` with a `status.hardwareVersions` entry and **no** owner references, then reconciles a VMCO that no longer reports it, would pin finding #1.

### 🔵 E2E: owner references are cluster-observable

Per `.sdd/memory/e2e-sync-with-changes.md`, the existing GuestOptions fan-out spec in `test/e2e/vmservice/vmservice/configpolicy/configpolicy.go` is the natural home for a one-line assertion that each fanned-out `VirtualMachineGuestOptions` carries an owner reference per contributing hardware version. Today that file only *sets* an owner ref on the synthetic stale fixture (line 359); it never asserts the controller sets them.

### 🔵 Description will go stale on rebase

The "Reviewing this diff before #1782 merges" section and the T093a/T093b-vs-T093c/T093d note are correct today but become misleading once this rebases onto `main`. Worth trimming when the draft flips to ready.

faisalabujabal:
needs rebase after #1782

PR #1794, opened 2026-08-03
🐛 Merge CnsUnregisterVolume into api/v1alpha1 pkg

brito-rafa:

This PR merges CnsUnregisterVolume into vsphere-csi-driver/api/v1alpha1.
CnsUnregisterVolume lived in its own package (from the e2e migration) - `api/cnsunregistervolume/v1alpha1` - that set its API group via a Go "const GroupName" rather than the controller-gen "+groupName" marker. 
So, the `make generate` emitted a bogus `config/crd/external-crds/_cnsunregistervolumes.yaml` (empty
group, hence the leading underscore) alongside the correctly named cns.vmware.com_cnsunregistervolumes.yaml.
This patch merges the api and addresses the bogus file.

Fixes # vmop_4004

```release-note

```


aruneshpa (approved):
LGTM.

PR #1795, opened 2026-08-04
✨ Add Dependabot config and post-update codegen workflow

brito-rafa:

Configure Dependabot for gomod updates across the three go.mod directories in this multi-module repo (`/`, `/hack/tools`, `/test/e2e`), checked daily with up to 10 open PRs at a time. Related dependencies are grouped into single PRs (Kubernetes/sigs.k8s.io, golang.org/x, golangci, ginkgo/gomega, observability, go-logr, and a catch-all for everything else) to cut down on PR noise.
Modules that are vmware related (image-registry-operator-api, net-operator-api, nsx-operator, govmomi) must be bumped manually so they are ignored by dependabot.

Dependabot only edits go.mod/go.sum, so a bumped dependency can leave stale generated CRDs, webhooks, RBAC, or tidied modules behind, failing verify-go-modules/verify-codegen in CI. Add a workflow that runs `make modules` and `make generate` on Dependabot PRs and pushes the results back to the PR branch.

Also exclude config/crd/external-crds from the labeler and the testing-needed diff, matching the existing config/crd/bases exclusion, since both are generated output.

Fixes # vmop_3991

This will synchronize the modules version across multiple go.mods in this repo. Checked daily with up to 10 open PRs at a time but eventually we can reduce this number after the first wave of updates.

```release-note

```

### DIFF
--- a/.github/workflows/dependabot-codegen.yml
+++ b/.github/workflows/dependabot-codegen.yml
@@ -1,0 +1,60 @@
+name: dependabot-codegen
+
+# Dependabot only edits go.mod/go.sum. It never runs `make modules` (to
+# re-tidy every module in this multi-module repo) or `make generate` (to
+# regenerate CRDs, webhooks, RBAC, and Go sources), so a bumped dependency
+# can silently leave stale generated files behind, which then fails
+# verify-go-modules/verify-codegen in ci.yml. This workflow runs both right
+# after Dependabot opens or updates a PR and pushes any resulting changes
+# back onto the PR branch, so the PR arrives with generated output already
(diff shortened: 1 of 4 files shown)


.github/workflows/dependabot-codegen.yml
```diff
@@ -0,0 +1,60 @@
+name: dependabot-codegen
```
faisalabujabal, line 1: do we have a pipeline that runs make modules + generate and fails if the tree is dirty? we should have that regardless of the author because it was probably a miss.

faisalabujabal (approved):

PR #1797, opened 2026-08-05
🐛 e2e: Serialize default KMS provider across concurrent jobs

aakashchan:

vCenter's default KMS provider (`CryptoManagerKmip.SetDefaultKmsCluster`
/ `GetDefaultKmsCluster` called with `entity=nil`) is a single, VC-wide
setting — it is not scoped per-namespace, per-session, or per-test. The
`vm_encryption.go` E2E specs save and restore this value in
`BeforeEach`/`AfterEach` as if each test owned it exclusively. When two
E2E jobs run encryption specs concurrently against the same vCenter
appliance, one job's cleanup can clear or overwrite the default the
other job just set and is relying on, producing a spurious
`VirtualMachineEncryptionSynced=NoDefaultKeyProvider` failure and
timeout partway through VM creation.

This was root-caused from two UTS builds (1132205 and 1132208) that
ran encryption specs concurrently against the same shared vCenter: the
first build's `AfterEach` cleared the global default KMS provider 3
seconds before the second build's vTPM spec tried to create its VM,
which then failed because vm-operator correctly saw no default
provider configured.

This PR adds a distributed lock backed by a Kubernetes `Lease` in the
target supervisor cluster's `vmware-system-vmop` namespace, held for
the full read-mutate-restore window (from `BeforeEach`'s read of the
current default through `AfterEach`'s restore, via `DeferCleanup`), so
only one job at a time can mutate the default KMS provider for a given
vCenter appliance.

Changes:
- Add `AcquireDefaultKMSProviderLock` in
  `test/e2e/vmservice/utils/kmslock.go`: an optimistic-concurrency
  create-or-take-over `Lease` lock with a bounded TTL so a crashed
  holder cannot permanently block other jobs.
- Wire the lock into `vm_encryption.go`'s `BeforeEach`, released after
  `AfterEach` restores the original default provider.
- Register `coordination/v1` in the shared e2e scheme
  (`vmservice/common/scheme.go`) so controller-runtime clients can
  read/write `Lease` objects — this was previously missing and caused
  the lock to fail with `no kind is registered for the type v1.Lease`.

Fixes #

UTS precheckin succeeded for all environment. Ran the encryption tests.

This is a test-infrastructure-only change (`test/e2e/**`); no product
(`pkg/`, `api/`) code is touched. Verified with `go build ./...`,
`go vet`, and `gofmt` on the e2e module; also validated live against a
real WCP testbed running the `vm_encryption.go` specs (surfaced and
fixed the missing `coordination/v1` scheme registration in the
process).

```release-note
NONE
```

### DIFF
--- a/test/e2e/vmservice/utils/kmslock.go
+++ b/test/e2e/vmservice/utils/kmslock.go
@@ -24,24 +24,15 @@ import (
 const (
 	// defaultKMSProviderLockName and defaultKMSProviderLockNamespace back a
 	// Lease used to serialize mutation of vCenter's default KMS provider
-	// across concurrently-running E2E jobs that target the same vCenter
-	// appliance.
+	// (CryptoManagerKmip.SetDefaultKmsCluster/GetDefaultKmsCluster called
+	// with entity=nil) across concurrently-running E2E jobs that target the
+	// same vCenter appliance — that setting is VC-wide, not scoped per
+	// namespace, session, or test.
 	//
-	// vCenter's default KMS provider (CryptoManagerKmip.SetDefaultKmsCluster
-	// / GetDefaultKmsCluster called with entity=nil) is a single, VC-wide
-	// setting — it is not scoped per-namespace, per-session, or per-test.
-	// Encryption specs save/restore it around each spec as if it were
-	// exclusively theirs; when two E2E jobs run encryption specs
-	// concurrently against the same vCenter, one job's cleanup can clear or
-	// overwrite the value the other job just set and is relying on,
-	// producing a spurious VirtualMachineEncryptionSynced=NoDefaultKeyProvider
-	// failure.
-	//
-	// The lock is a Lease in the target supervisor cluster rather than
-	// something stored in vCenter itself, because only jobs pointed at the
-	// same supervisor cluster/vCenter share that cluster's API server —
-	// which is exactly the set of jobs capable of racing on that vCenter's
-	// global setting.
+	// The lock lives in the target supervisor cluster rather than in
+	// vCenter itself because only jobs pointed at the same supervisor
+	// cluster/vCenter share that cluster's API server — exactly the set of
+	// jobs capable of racing on that vCenter's global setting.
 	defaultKMSProviderLockName      = "vmservice-e2e-default-kms-provider-lock"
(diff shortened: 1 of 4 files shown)


test/e2e/vmservice/utils/kmslock.go
```diff
@@ -0,0 +1,149 @@
...
+	// cluster/vCenter share that cluster's API server — exactly the set of
+	// jobs capable of racing on that vCenter's global setting.
+	defaultKMSProviderLockName      = "vmservice-e2e-default-kms-provider-lock"
+	defaultKMSProviderLockNamespace = "vmware-system-vmop"
+
+	// defaultKMSProviderLockLeaseDuration bounds how long a lock holder that
+	// crashes or panics without releasing can block other jobs. It must
+	// exceed the slowest encryption spec's full BeforeEach+It+AfterEach
+	// runtime.
+	defaultKMSProviderLockLeaseDuration = 15 * time.Minute
```
faisalabujabal: find the default ginkgo test timeout and lets match it or use the same source of the timeout (minus few min)

faisalabujabal (approved):
post a message on the channel as we discussed that we are going with this solution. lets give the team a chance to object if anyone wants to

test/e2e/vmservice/utils/kmslock.go
```diff
@@ -0,0 +1,149 @@
...
+
+	lease := &coordinationv1.Lease{}
+	if err := c.Get(ctx, key, lease); err != nil {
+		return
+	}
+	if ptr.DerefWithDefault(lease.Spec.HolderIdentity, "") != holder {
+		// Lease already expired and another job reclaimed it.
+		return
+	}
+	if err := c.Delete(ctx, lease); err != nil {
```
bryanv: We could also specify the delete option `client.Preconditions(&metav1.Preconditions{})` with the UID & ResourceVersion of the Lease we just got so no body else updated after our Get.

aakashchan: Yes, makes sense. Fixed.

test/e2e/vmservice/utils/kmslock.go
aakashchan: Thank you. Done.

PR #1799, opened 2026-08-05
✨  Remove network interface result struct

bryanv:

 NetworkInterfaceResult was a flat struct mixing three unrelated concerns — network-interface-CR identity, vSphere ethernet-card backing/state, and guest-customization data — that every caller had to partially populate and partially ignore. This PR deletes it and splits its responsibilities into the two types that already existed for exactly this purpose:                                             
                                                                                                                                                                                                                                                                                                                                                                                                                   
  - Device — vSphere-facing: CR identity, backing, MAC/external ID, and now also InterfaceName, EthCard, and EthCardKey.                                                                                                                                                                                                                                                                                           
  - Bootstrap — guest-customization-facing: DHCP/IPAM, DNS, routes, MTU. Built on demand via the new BuildBootstraps(ctx, vm, devices), which dispatches per-Device on the interface CR's concrete type instead of being assembled eagerly and carried around half-stale.                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                   
  This removes the devAndBootstrapToNetworkInterfaceResult bridge function that existed solely to keep producing the old struct, and lets every consumer (GuestOSCustomization, NetPlanCustomization, ListOrphanedNetworkInterfaces, VPCPostRestoreBackingFixup, ReconcileNetworkInterfaces, vmlifecycle.GetBootstrapArgs) work off the narrower type it actually needs. vmlifecycle.BootstrapArgs drops its       
  NetworkResults field in favor of Bootstraps []network.Bootstrap + UpdatedEthCards bool.                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                   
  Building Bootstrap after MAC-address fixup (rather than before, like the old bridge function did) also makes Bootstrap.MacAddress correct by construction, which surfaced and fixed a couple of latent bugs along the way:                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                   
  - UpdatedEthCards never reached guest customization. VMUpdateArgs/VMResizeArgs.NetworkResults held a copy of the reconcile-scoped networkResults, so mutations made while reconfiguring ethernet cards (UpdatedEthCards, backfilled EthCardKey/MacAddress) never propagated to the value later used to build Bootstraps — meaning the post-reconfigure MAC/eth-card re-sync never fired when cards actually      
  changed. Fixed by making those fields pointers into the same networkResults. (Verified this can't introduce a stale-read race: the vCenter Reconfigure task is awaited synchronously before the re-fetch runs.)                                                                                                                                                                                                  
  - findExistingEthCardForOrphanedCR matched the wrong orphaned CR whenever ≥2 orphaned interface CRs existed and more than one lacked the interface-name label — an off-by-index bug (iterating position instead of the stored index). Fixed and covered by a new regression test.                                                                                                                                
  - VM-create NIC ordering could desync from vm.Spec.Network.Interfaces order when an interface with an explicit Type was mixed with class-provided (untyped) ethernet cards, because reused class cards kept their original ConfigSpec.DeviceChange position while new cards were appended at the end. vmCreateGenConfigSpecZipNetworkInterfaces now rebuilds the ethernet-card device changes from scratch, in   
  interface order, so a class device only ever pairs with its own positional interface (an interface with Type set forfeits its class device rather than reusing it, and that device is simply dropped, not reassigned).                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  Are there any special notes for your reviewer:                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                   
Fixes # vmop-3999

  - vmCreateDoNetworking's NetworkReady condition now derives from len(devices) > 0 rather than checking Spec.Network directly — intentional, but it means an enabled network spec that legitimately resolves to zero interfaces now deletes the condition instead of marking it True, a small change from prior behavior worth a second look.                                                                     
  - CreateVirtualEthernetCard currently always creates a VMXNet3 card regardless of interfaceSpec.Type (see the TODO in devices.go)
  - Most non-network/vmlifecycle/session renames in this diff are mechanical (r → b for loop variables now iterating Bootstrap instead of NetworkInterfaceResult).   

```release-note
NONE
```


pkg/providers/vsphere/network/devices.go
```diff
@@ -26,6 +28,82 @@ import (
...
+	ethCard := ethDev.(vimtypes.BaseVirtualEthernetCard).GetVirtualEthernetCard()
+	ethCard.ExternalId = dev.ExternalID
+	if dev.MacAddress != "" {
+		ethCard.MacAddress = dev.MacAddress
+		ethCard.AddressType = string(vimtypes.VirtualEthernetCardMacTypeManual)
+	} else {
+		ethCard.AddressType = string(vimtypes.VirtualEthernetCardMacTypeGenerated)
+	}
+
+	// TODO: interfaceSpec.VNUMANodeID, VMXNet3. We need to sort out how to deal
```
hpannem, line 66: you are right here. I think we can reorder the nic extra config after vmCreateGenConfigSpecZipNetworkInterfaces. However, for new vms, this is always no-op as there is no device key yet. on the next reconfigure, extra config will be configured.

hack/new-schema-version.py
```diff
@@ -1877,8 +1877,8 @@ def step_update_bootstrap_templatedata(self) -> None:
             1,
         )
         content = content.replace(
-            f"make([]vmopv1.NetworkDeviceStatus, 0, len(bsArgs.NetworkResults.Results))",
-            f"make([]{old_alias}.NetworkDeviceStatus, 0, len(bsArgs.NetworkResults.Results))",
+            f"make([]vmopv1.NetworkDeviceStatus, 0, len(bsArgs.NetBootstraps))",
```
hpannem: heads-up: https://github.com/vmware-tanzu/vm-operator/pull/1804
I have changed this file. If I checkin first, you may need to merge or vice versa.

pkg/providers/vsphere/network/bootstrap_test.go
```diff
@@ -1309,3 +1326,149 @@ var _ = Describe("VPCInterfaceBootstrap",
...
+
+		It("returns an error", func() {
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("unsupported network interface CR type"))
+		})
+	})
+
+	Context("device/interface length mismatch", func() {
+		BeforeEach(func() {
+			devices = []network.Device{
```
hpannem: multi device positive test case?

pkg/providers/vsphere/network/devices.go
```diff
@@ -26,6 +28,82 @@ import (
...
+
+	backing, err := dev.Backing.EthernetCardBackingInfo(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get Ethernet card backing info for network ID %s (%s): %w",
+			dev.NetworkID, dev.Backing.Reference().Value, err)
+	}
+
+	var cardType string
+	// TODO: Honor interfaceSpec.Type if set.
+	switch interfaceSpec.Type { //nolint:gocritic
```
hpannem, line 46: what about other vmxnet types barring sriov?

		&types.VirtualE1000{},
		&types.VirtualE1000e{},
		&types.VirtualVmxnet{},
		&types.VirtualVmxnet2{},
		&types.VirtualVmxnet3{},
		&types.VirtualVmxnet3Vrdma{},
		&types.VirtualPCNet32{},
//		&types.VirtualSriovEthernetCard{},

hpannem:
From Claude: (Ignore if it does not make sense)

 Code review(review · 10 findings)
pkg/providers/vsphere/session/session_vm_update.go
  ●  316 [correctness]    VMUpdateArgs/VMResizeArgs.NetworkResults changed from a value type to a pointer aliasing the same local networkResults, so UpdatedEthCards set inside poweredOffReconfigure/resizeVMWhenPoweredStateOff now leaks into the later reconcileNetworkAndGuestCustomizationState call — the exact propagation the PR's own design doc says must not be reintroduced.
pkg/providers/vsphere/network/network.go
  ●  288 [correctness]    MAC addresses are now unconditionally lowercased in the shared Device-constructing getters (getNetOPNetworkInterfaceDevice, getNCPNetworkInterfaceDevice, getVPCSubnetPortDevice, createAndWaitNamedNetworkInterface), which are also called by the VM-create path, even though the PR's plan states the create path is 'intentionally untouched'.
  ●   64 [design]         The Device struct's doc comment calls InterfaceName/EthCard/EthCardKey fields temporary ('just for the in progress transition'), but they are load-bearing in reconcile.go, nsxt.go, and session_vm_update.go with no remaining migration path off them.
pkg/providers/vsphere/vmprovider_vm.go
  ● 3189 [correctness]    When interfaceSpec.Type is non-empty, the zip logic drops the VM-Class-provided ethernet card and creates a new one via CreateVirtualEthernetCard, but that function has a TODO to honor interfaceSpec.Type and unconditionally hardcodes vmxnet3 — so the class card's config is discarded for a replacement that still ignores the requested type.
  ● 3221 [correctness]    The rewritten vmCreateGenConfigSpecZipNetworkInterfaces always appends all ethernet DeviceChange entries after all non-ethernet entries, whereas the old code preserved each ethernet device's original relative position within ConfigSpec.DeviceChange.
  ● 3171 [correctness]    classEthCards[i] is assumed to correspond to vm.Spec.Network.Interfaces[i]/createArgs.NetworkDevices[i] purely via a code comment, with no type-level guarantee tying the two independently-populated slices together positionally.

pkg/providers/vsphere/network/devices.go
  ●   91 [correctness]    New line ethCard.SubnetId = "" unconditionally clears the VM-Class-provided ethernet card's SubnetId in UpdateVMClassEthCardFromDevice, a behavior absent from the deleted pre-PR equivalent used on the create path, with no test covering this specific clear on create.
  ●   33 [simplification] CreateVirtualEthernetCard and UpdateVMClassEthCardFromDevice duplicate the entire backing-lookup + ExternalId + MAC/AddressType-branch logic almost verbatim instead of sharing a helper.
pkg/providers/vsphere/network/bootstrap.go
  ●  452 [simplification] The three-way CR-type switch (NetworkInterface / VirtualNetworkInterface / SubnetPort) in the new BuildBootstraps duplicates the identical dispatch already present in CreateNetworkDevices (network.go:115) and findExistingEthCardForOrphanedCR (reconcile.go:93).
  ●  122 [reuse]          MAC addresses are lowercased twice: once when Device.MacAddress is constructed in network.go's getters, and again unconditionally in InterfaceBootstrap from the already-lowercased value.

PR #1799 review — 10 findings, most severe first (all correctness issues before cleanup items):

1. session_vm_update.go:316 — NetworkResults changed to a pointer, letting UpdatedEthCards leak between reconfigure phases — the exact bug the PR's design doc says must not be reintroduced.
2. network.go:288 — MACs are now unconditionally lowercased in getters shared with the VM-create path, which the plan says is "intentionally untouched." Confirmed by a test expecting strings.ToLower(macAddress0).
3. vmprovider_vm.go:3189 — setting interfaceSpec.Type discards the VM-Class ethernet card and replaces it with a hardcoded vmxnet3 that still doesn't honor the requested type (regression vs. old positional reuse).
4. devices.go:91 — new unconditional ethCard.SubnetId = "" on the create path, untested.
5. vmprovider_vm.go:3221 — ethernet DeviceChange entries are now always appended last, losing original relative ordering from the VM Class template.
6. vmprovider_vm.go:3171 — classEthCards[i] positional pairing is only guaranteed by a comment, no type-level safety.
7. devices.go:33 — CreateVirtualEthernetCard/UpdateVMClassEthCardFromDevice duplicate backing/MAC logic almost verbatim.
8. bootstrap.go:452 — the CR-type switch is now duplicated in three places across two files.
9. network.go:64 — Device struct fields documented as "temporary transition" are actually permanent/load-bearing.
10. bootstrap.go:122 — MAC is lowercased a second, redundant time.

pkg/providers/vsphere/network/devices.go
bryanv: Yah, I would like to see if we can instead do that offset calc via the unit number (and/or the implied ordering) at create time.

hack/new-schema-version.py
bryanv: Merge your's first

pkg/providers/vsphere/network/devices.go
bryanv: I had that in an earlier version of this branch but dropped to not scope creep too much in this MR: I know that support is coming and I wanted to make sure that what I was changing here would make it easy to implement later.

bryanv:
> From Claude: (Ignore if it does not make sense)
> 

I too have run this through a few rounds of Claude's built in code review skill. Mine raised basically the same issues and either they are actual fixes (like the first one to change it to a pointer), items that I have already updated in the PR, or non issues.

hpannem (approved):
Looks good to me!

PR #1800, opened 2026-08-05
🐛 Do not clear backfilled SubnetPort fields

bryanv:

NSX Operator backfills the InterfaceIPType and StaticIPAllocationType fields, so we should not clear them when the corresponding fields in the VM interface spec are not set.

This means that the fields cannot be cleared in the interface spec to get back the defaults. If we want to be slightly stricter, we could only set these fields when creating the SubnetPort.

Fixes # vmop-4001

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/network/network.go
+++ b/pkg/providers/vsphere/network/network.go
@@ -1125,8 +1125,13 @@ func createVPCNetworkInterface(
 		}
 
 		if pkgcfg.FromContext(ctx).Features.WorkloadIPv6 {
-			subnetPort.Spec.InterfaceIPType = IPAMModesToVPCInterfaceIPType(interfaceSpec.IPAMModes)
-			subnetPort.Spec.StaticIPAllocationType = DeriveStaticIPAllocationType(interfaceSpec)
+			// NSX Operator backfills these fields so don't clear if already set.
+			if ipType := IPAMModesToVPCInterfaceIPType(interfaceSpec.IPAMModes); ipType != "" {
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/network/network.go
```diff
@@ -1125,8 +1125,13 @@ func createVPCNetworkInterface(
 		}
 
 		if pkgcfg.FromContext(ctx).Features.WorkloadIPv6 {
-			subnetPort.Spec.InterfaceIPType = IPAMModesToVPCInterfaceIPType(interfaceSpec.IPAMModes)
-			subnetPort.Spec.StaticIPAllocationType = DeriveStaticIPAllocationType(interfaceSpec)
+			// NSX Operator backfills these fields so don't clear if already set.
+			if ipType := IPAMModesToVPCInterfaceIPType(interfaceSpec.IPAMModes); ipType != "" {
```
hpannem, line 1129: Should this be done at NSX operator level? Suppose if I have ipammodes to IPv4 to start with and later I want to move to platform default, I send ipammodes as [] list. We don't have a way to set that anymore?

bryanv: This is an issue with these fields being backfilled. I'm not sure of the reason was they decided to do it. We could annotation the SubnetPort for the what-we-thought-we-set-here but that does not solve the issue for CRs that are created externally for us, like the issue that cropped up from was an import, since the fields will likely be backfilled by the time we see the CR.

hpannem (approved):

aruneshpa (approved):

PR #1802, opened 2026-08-07
🌱 Better handle Zone decom in VirtualMachineSetResourcePolicy

bryanv:

The VMSetRP itself is immutable but the underlying cluster state could change: specifically here when a zone is removed, we update the status to better indicate that. The child RPs were already handled (note that the actual delete of them is deferred to when the namespace RP is removed so they are just not included in the status) but we should also not include the cluster modules for the removed zone.

Fixes # vmop-4026

Verified manually. Add item for e2e zone decom work

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/vmprovider_resourcepolicy.go
+++ b/pkg/providers/vsphere/vmprovider_resourcepolicy.go
@@ -7,11 +7,14 @@ package vsphere
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	vimtypes "github.com/vmware/govmomi/vim25/types"
 	apierrorsutil "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	vmopv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha6"
+	pkglog "github.com/vmware-tanzu/vm-operator/pkg/log"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/clustermodules"
 	"github.com/vmware-tanzu/vm-operator/pkg/providers/vsphere/vcenter"
 	"github.com/vmware-tanzu/vm-operator/pkg/topology"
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/vmprovider_resourcepolicy.go
```diff
@@ -77,6 +83,24 @@ func (vs *vSphereVMProvider) CreateOrUpdateVirtualMachineSetResourcePolicy(
 		}
 	}
 
+	if len(errs) == 0 {
+		// In the case of a zone being removed (decommissioned) remove the cluster
+		// modules for that CCR. Our status is the only source of the CMs that were
+		// created for this CCR so treat the delete as just best effort.
+		resourcePolicy.Status.ClusterModules = slices.DeleteFunc(resourcePolicy.Status.ClusterModules,
```
aniket-deole, line 90: As VMSetRP is immutable, will this result in the clustermodule being recreated due to: [vmprovider_resourcepolicy.go#L133-L147](https://github.com/vmware-tanzu/vm-operator/blob/2ae4dc13739643cdd3f07e45152f145876aaf05d/pkg/providers/vsphere/vmprovider_resourcepolicy.go#L133-L147)

bryanv: I don't believe so - the cluster MoID for the Zone would exist anymore for the code you linked to, so that is why the added code in this MR removes them.

aniket-deole: Okay. Now I see clusterMorefs being filtered by rpMoIDs for the namespace.
```
	folderMoID, rpMoIDs, err := topology.GetNamespaceFolderAndRPMoIDs(ctx, vs.k8sClient, resourcePolicy.Namespace)
```

aniket-deole (approved):

aruneshpa (approved):
LGTM.

PR #1803, opened 2026-08-07
✨ SAP config e2e test and status.hardware.memory conversion fix

hpannem:

  Adds a composite "SAP HANA profile" E2E scenario to `VMComputeConfigSpec` (`test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go`) that creates a single VM covering every VCFA compute advanced-parameter request from the SAP HANA / VCD-to-VCFA migration requirements that VM Operator supports today, and fixes two bugs uncovered while writing it.

  **New E2E scenario**

  - `numa.vcpu.preferHT` — first-class `spec.advanced.preferHtEnabled`
  - `sched.node{0-3}.affinity` and `smbios.assetTag` — the generic `spec.advanced.extraConfig` bag, since neither has a first-class field
  - `ethernet0.ctxPerDev` / `ethernet0.pnicFeatures` — first-class `spec.network.interfaces[].vmxnet3` fields
  - A real 4-way vNUMA topology (`vnumaNodeCount: 4`, `coresPerSocket: 1`, `size.cpu: 4`, `minHardwareVersion: 20`) so each of the four `sched.nodeX.affinity` keys binds to an actual vNUMA client instead of being an inert ExtraConfig value — verified directly against vSphere via govmomi (`config.numaInfo.coresPerNumaNode`), both powered off and after power-on
  - Full CPU reservation (`spec.resources.requests.cpu`, derived from the host's `CpuMhz`) and a full memory limit (`spec.resources.limits.memory`), plus `memoryAdvanced.reservationLockedToMax` for the memory reservation guarantee
  - `disk.enableUUID` is asserted (not set) since it's a system-reserved key vm-operator sets unconditionally on every VM

  Not covered, by design: SVGA video memory amount has no VM Operator API today — no ExtraConfig key, no first-class field, no VM Class video device knob — so it isn't a test gap closeable with a test-only change.

  **Bug fix: `status.hardware.memory` used decimal instead of binary MB**

  `pkg/providers/vsphere/vmlifecycle/update_status.go` converted vSphere's `MemoryMB` (always true binary mebibytes) into `status.hardware.memory.{total,reservation,limit}` using `*1_000_000` (decimal mega) instead of binary math, under-reporting real memory by ~4.66% on every VM and contradicting every other MemoryMB conversion in this codebase (`pkg/util/vmopv1/compute_overwrite.go`'s `desiredMemoryAllocation`, the `ConfigSpec.MemoryMB` assignment). Added `virtualmachine.MbToBytes` next to its existing inverse `MemoryQuantityToMb` in `pkg/providers/vsphere/virtualmachine/conversion.go` (no such reverse utility existed anywhere), and fixed all three conversions to use it. Updated the tests that were pinned to the old, wrong values: `update_status_test.go` (4 assertions, including recalculating the `math.MaxInt32` boundary test for the new divisor) and `vm_compute_config.go` (a `"3072M"` round-trip assertion now correctly expects `"3Gi"` — an exact match instead of a lossy one). Added two small unit tests for `MbToBytes` in `conversion_test.go`.

  **Bug fix: condition waits racing against their own stale prior state**

  `waitForComputeConfigSynced` and `waitForExtraConfigSynced` support (or, for the latter, now support) an `afterTime` sentinel specifically to guard against this, but several call sites weren't using it: when a condition is already `True` before a patch, waiting for `True` again afterward with no sentinel can pass instantly on the stale, pre-patch value without ever confirming the reconciler processed the new spec. Fixed in the new SAP HANA scenario (two waits, threading `condBaseline`/`condResize.LastTransitionTime` forward) and in the existing `"re-applies extraConfig after the vSphere VM is destroyed out-of-band"` scenario, where `ExtraConfigSynced` sits untouched at its pre-destroy `True` for the entire window the underlying vSphere VM is gone (`reconcileStatusExtraConfig` skips updating it whenever `moVM.Config` is nil), and `status.uniqueID` changing doesn't prove otherwise since it's set by a separate patch in `createVirtualMachineAsync`. Added `afterTime ...metav1.Time` support to `waitForExtraConfigSynced` itself, matching the pattern already used by `waitForComputeConfigSynced`/`waitForNICExtraConfigSynced`.
already used by `waitForComputeConfigSynced`/`waitForNICExtraConfigSynced`.
already used by `waitForComputeConfigSynced`/`waitForNICExtraConfigSynced`.

Fixes #

- New scenario is labeled `extended-functional` and `experimental` (skips when `supports_telco_vm_service_api` is disabled or when the govmomi vCenter client isn't available), per `e2e-testing.md` — has not yet been run against a real Supervisor/vSphere cluster.
- The `status.hardware.memory` fix is a real behavior change for every VM, not test-only — flagging per `e2e-sync-with-changes.md`'s "found a potential bug while writing E2E tests" carve-out rather than scoping it to a separate PR, since the E2E test is what surfaced it and the fix needed the matching test updates in the same change set.

```release-note
Fix status.hardware.memory.{total,reservation,limit} under-reporting VM memory by ~4.66% due to a decimal/binary unit conversion bug.
``

### DIFF
--- a/test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go
@@ -13,8 +13,11 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/property"
 	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/mo"
 	vimtypes "github.com/vmware/govmomi/vim25/types"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	e2eframework "k8s.io/kubernetes/test/e2e/framework"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go [hunk 1]
```diff
@@ -771,6 +774,216 @@ func VMExtraConfigSpec(ctx context.Context, inputGetter func() VMExtraConfigSpec
...
+				CPU: ptr.To(resource.MustParse(fmt.Sprintf("%d", fullCPURes))),
+			}
+			latest.Spec.Resources.Limits = &vmopv1.VirtualMachineResourceQuantity{
+				Memory: memStatus.Total,
+			}
+			Expect(svClusterClient.Patch(ctx, latest, patch)).To(Succeed(),
+				"failed to patch VM %s full compute reservation", vmName)
+
+			By("Waiting for ComputeConfigSynced=True after full reservation/limit patch")
+			waitForComputeConfigSynced(ctx, svClusterClient, config, vmKey, metav1.ConditionTrue, "")
```
bryanv: It might not matter very much for this but isn't there a chance that observes the existing true condition before the change changed VM is even reconciled

test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go
```diff
@@ -771,6 +774,216 @@ func VMExtraConfigSpec(ctx context.Context, inputGetter func() VMExtraConfigSpec
...
+	// vmxnet3 fields), a 4-way vNUMA topology so the four sched.nodeX.affinity
+	// keys each bind to a real vNUMA client, and full CPU/memory reservation
+	// plus a full memory limit (spec.resources / memoryAdvanced). Exercises
+	// all three ExtraConfig-family reconcilers (VM-level, NIC-level,
+	// compute-config) on one VM. disk.enableUUID is asserted rather than set:
+	// it is a system-reserved key vm-operator sets unconditionally on every
+	// VM, not a user-configurable field. SVGA video memory has no VM Operator
+	// API today (no ExtraConfig key, no first-class field, no VM Class device
+	// knob) and is intentionally not covered here.
+	It("creates SAP HANA-profile VM with 4-way NUMA node affinity, asset tag, NIC tuning, and full compute reservation",
```
bryanv, line 803: Does any of this require any special about how the testbed is deployed?

bryanv (approved):

test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go
hpannem: Very valid comment. Should we need to have observedGeneration in vm CR to distinguish reconciles?

Anyways, I added lastUpdatedTime as a check to distinguish old reconcile values.

test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go
hpannem: It does not fail for the numa nodeaffinity. If enough physical NUMA nodes are not there (aka your configuration is wrong), ESXi falls back to without affinity.

I tested this on single socket and dual socket (1 and 2 numa nodes). On both, I am able to power on vm with wrong values. esxtop shows that they are scheduled without affinity.

hpannem:
```
VM-COMPUTE-CONFIG
....
  I0810 17:25:33.580741   64491 util.go:1950] Successfully cleaned up cluster role bindings for e2e tests
[SynchronizedAfterSuite] PASSED [31.075 seconds]
[ReportAfterSuite] Autogenerated ReportAfterSuite for --json-report
autogenerated by Ginkgo
[ReportAfterSuite] PASSED [0.009 seconds]

Ran 17 of 183 Specs in 1020.794 seconds
SUCCESS! -- 17 Passed | 0 Failed | 4 Pending | 162 Skipped
PASS

test/e2e/vmservice/vmservice/virtualmachine/vm_extraconfig.go
bryanv: Yes. When we switched to metav1 Conditions we did gain that field but we need to update the condition helpers and callsites.

The LastTransitionTime is only updated on state transitions so it isn't clear that we can really depend on that.

hpannem: For this test case, the state transition should happen since the intent is changed, it should go back to fase and then true. ObservedGeneration would be the right choice though.

PR #1804, opened 2026-08-07
✨ Add V1alpha6 template functions to differentiate IPv4/IPv6

hpannem:

Adds E2E coverage for the V1alpha6 vApp/Sysprep template functions (vmop-1152, epic vmop-787), and fixes two dual-stack-testbed bugs found while running the suite against a real NSX dual-stack testbed.

- Expands the V1alpha6 vApp-property test in `vm_guestcustomization.go` from covering 6 of the 14 registered `V1alpha6_*` template functions to all 14, split across three VM-creation "rounds" since the `photon-ovf-vapp-properties` OVF image only exposes 4 readable string-typed vApp properties (plus 1 bool/1 int slot) to reuse.
- Each round now builds a `*vmopv1.VirtualMachine` directly as a typed Go struct and creates it via `svClusterClient.Create`, instead of going through the `manifestbuilders`/`.yaml.in` text-template pipeline.
- Rounds A and B request dual-stack IPAM (`spec.network.interfaces[0].ipamModes: [IPv6, IPv4]`) so the IPv6-returning functions exercise a real IPv6 address instead of only ever having IPv4 to work with. Gated on a new `supports_workload_ipv6` Supervisor capability check.
- `hack/e2e/setup-testbed-env.sh`: on a dual-stack testbed, `decryptK8Pwd.py` reports the Supervisor IP as `"<IPv4>,<IPv6>"`. `scp`/`sed`/the `kubectl-vsphere` download URL need a single connectable host, so a new `wcp_ip_primary` (IPv4-only) is derived for those call sites; `WCP_IP`/`SUPERVISOR_CLUSTER_IP` still export the full raw value.
- `test/e2e/infrastructure/vsphere/supervisor/apiserver.go`: `GetControlPlaneVMConnectionDetails`'s regex assumed the decrypt-script's `IP: <ip>` line ends right after the IP, which fails to match on the same dual-stack format above. Made the regex tolerant of an optional `,<ipv6>` suffix, and wrapped the matched IP in `regexp.QuoteMeta` defensively (IPv4 addresses contain literal `.` characters, which are regex metacharacters).

vmop-1152

All three new `It`s are labeled `"experimental"` per `e2e-testing.md` convention — they haven't yet been validated against real dual-stack hardware. `IsUsableIP`/`FormatNameservers`/`FirstIPv6`-family assertions are intentionally lenient where they depend on VM/environment state (an unrendered `"{{ ... }}"` fallback is accepted), since even with dual-stack IPAM requested this test can't guarantee the network hands out an address in time.

```release-note
Add E2E coverage for all V1alpha6 vApp/Sysprep template functions, including dual-stack (IPv4/IPv6) scenarios, and fix Supervisor IP parsing in E2E test setup for dual-stack testbeds.


----
📚 Documentation preview 📚: https://vm-operator--1804.org.readthedocs.build/en/1804/

### CODE REVIEW

**Review by bryanv — COMMENTED** [review]
Have AI update hack/new-schema-version.py too

**Comment by hpannem** [comment]
> Have AI update hack/new-schema-version.py too

From Claude

```
You're right — I missed this. Confirmed: _make_v1aN_template_functions_block (line 353) is a hardcoded Python string template that generates an entire v1a{N}TemplateFunctions function from scratch when scaffolding a new version. It reproduces the old 8-closure pattern verbatim — the exact shape v1a6TemplateFunctions had before this PR's changes. It has none of:

- FirstIPv4/FirstIPv6/FirstIPv4FromNIC/FirstIPv6FromNIC
- IsUsableIP/PrefixLength
- The IPv4→IPv6 fallback in FirstIP/FirstIPFromNIC/IPsFromNIC
- The IPv6-input guards on SubnetMask/IP
- Any awareness of Gateway6/IPv6Addresses

So if someone runs this script today to scaffold v1alpha7 from v1alpha6, the new version would regress to IPv4-only, directly contradicting this PR's premise (v1alpha7 would become the new "latest, most-encouraged" version per the docs, yet lose all the IPv6 work). Good catch — this does need updating, and my earlier answer was wrong because I hadn't looked at this function specifically.

Let me check the companion piece — there's almost certainly a matching hardcoded generator for toTemplateNetworkStatusV1A{N} (the function that would need the Gateway6/IPv6Addresses population logic) — before I scope the fix.

  Read 1 file

Confirmed — and it's worse than just that one function. The whole scaffolding step that inserts a new version's block (step_add_v1aN_hub_functions or similar, around line 1740-1956) is a fragile, string-anchor-based patcher, and at least two of its anchors no longer match the current file at all:

1. insert_after (line 1893-1898) anchors on the literal text "// This is basically identical to v1a2TemplateFunctions.\nfunc v1a3TemplateFunctions(" to know where to insert the new toTemplateNetworkStatusV1A{N} function. That text still exists (v1a3's comment is untouched), so this one still matches — but it inserts the old IPv4-only body (no Gateway6/IPv6Addresses else-branch) verbatim (lines 1899-1926).
2. marker_end_prev (line 1944-1948) anchors on constants.{old_ver_go}FormatIP:   v1alpha{old_num}FormatIP,\n\t}\n}\n — the literal end of the old-style v1aXTemplateFunctions FuncMap literal, to know where to insert the new version's whole function. Since I restructured v1a6TemplateFunctions (now it ends with V1alpha6PrefixLength: v1a6PrefixLength, in a totally different FuncMap, and the closures were extracted into standalone package-level functions), this anchor no longer matches v1a6's actual content at all. str.replace on a non-match is a silent no-op in Python — so if someone runs this script today to scaffold v1alpha7, this step would silently do nothing, and _make_v1aN_template_functions_block's hardcoded old-style output would just... not get inserted anywhere, or the scaffold would end up in some partially-broken intermediate state depending on what else depends on it.

So the real scope is: (a) rewrite _make_v1aN_template_functions_block to emit the new-style content (matching current v1a6TemplateFunctions: standalone helper functions, all 6 new functions, the fallback/guard logic), and (b) fix the anchor strings in the insertion logic to match the new file structure. Given how brittle this string-patching approach already is, do you want me to: (a) do a surgical fix keeping the same brittle-anchor approach, or (b) take this as a cue to make the insertion logic more robust (e.g., anchor on function names/markers less likely to drift), given it's already broken once?

bryanv (approved):

PR #1805, opened 2026-08-10
:sparkles: Functional verification TDS, plan and tasks for VM ReplicaSets

aruneshpa:

This change contains the functional TDS, plan and tasks for the VirtualMachineReplicaSets feature.  The replica sets were implemented a while ago hidden behind a capability that is not yet enabled.  We need to verify the end-to-end functionality before we enable the feature.  This change contains the specs that will be used to drive the tests.

```release-note
NONE
```

### DIFF
--- a/.sdd/specs/003-virtualmachinereplicaset/tasks.md
+++ b/.sdd/specs/003-virtualmachinereplicaset/tasks.md
@@ -1,0 +1,71 @@
+# Tasks: `VirtualMachineReplicaSet` Functional Test Coverage
+
+- **TDS**: [`tds.md`](./tds.md) — the acceptance-criteria source for this task list (in place of a `spec.md`; each task cites the TDS scenario number(s) it implements as `[SCn]`)
+- **Test plan**: [`test-plan.md`](./test-plan.md)
+- **Epic**: vmop-1701
+
+<!--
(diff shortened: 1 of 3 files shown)


hpannem:
Looks good. Just that directory should be changed to new number.  

Minor nits (claude):

2. Undeclared constitutional deviation. The constitution requires spec.md + plan.md + tasks.md; this PR ships tds.md + test-plan.md + tasks.md instead. tasks.md even says "in place of a spec.md" — but the constitution says bending a rule requires either an amendment PR or, per sdd-standards.md anti-pattern #2, an explicit "Complexity tracking" entry in plan.md. There's no plan.md to hold that entry, and the PR description doesn't call out the deviation at all. Given this is a recurring pattern for test-verification specs (worth allowing), the cleaner fix is a small constitution amendment permitting a tds.md/test-plan.md pair as an alternate artifact set for test-verification-only specs — not silently diverging per-PR.
3. Wiki link format. tds.md writes (Confluence page 1007649487) — the required literal form is WIKI page 1007649487.
4. Missing Epic: header. Neither tds.md nor test-plan.md carries the required Epic: vmop-NNN header that spec.md would normally have; the epic only shows up inside tasks.md.

hpannem (approved):

PR #1806, opened 2026-08-11
:bug: Use controller owner refs to identify CnsRegisterVolumes owned by a VM

aruneshpa:

Currently, we use "created-by" label on the CRV resource to identify the resources that was created to register the classic disks of a VM as PVC -- what we call unmanaged disk registration.

This has a flaw.  Because Kubernetes restricts the length of the value of the label to 63 whereas the object names can be up to 254 characters long.  This means, classic disks of any VM that has name exceeding 63 characters can never complete registration thereby blocking a lot of day-2 workflows on the disk (e.g., expanding the disk).

This change fixes this bug by changing the lookup to use ownerRef on the CRV that is set by the controller.  Since the CRV can't be edited by the developer user, it's safe to assume that no one can remove the OwnerRef from the CRV.

While here, also fix a bug where we would silently adopt an existing CRV by name if one existed (from a prior VM).  We now delete those and re-create new ones for the current VM.  This will have a consequence of the FCD being deleted if the CRV mapped a disk, but I would argue that explicitly deleting and re-creating it is the correct contract here.

```release-note
Use controller owner refs to identify CnsRegisterVolumes owned by a VM
```

### DIFF
--- a/pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
+++ b/pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
@@ -707,7 +707,7 @@ func ensurePVCForUnmanagedDisk(
 		if !pkgcfg.FromContext(ctx).Features.VMSharedDisks {
 			volumeMode = corev1.PersistentVolumeFilesystem
 		}
-		obj.Spec.VolumeMode = ptr.To(volumeMode)
+		obj.Spec.VolumeMode = new(volumeMode)
 
 		// Set dataSourceRef to point to this VM.
 		obj.Spec.DataSourceRef = &expDSRef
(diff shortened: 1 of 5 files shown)


pkg/vmconfig/volumes/unmanaged/register/unmanagedvolumes_register.go
```diff
@@ -822,94 +822,106 @@ func ensureCnsRegisterVolumeForDisk(
...
-		case pvc.Spec.VolumeMode != nil:
-			volumeMode = *pvc.Spec.VolumeMode
+	// Get the datastore URL for the provided file name.
+	finder := pkgctx.GetFinder(ctx)
+	if finder == nil {
+		return nil, errors.New("failed to get finder from context")
+	}
+	datastoreURL, err := pkgdatastore.GetDatastoreURLFromDatastorePath(
+		ctx,
+		pkgctx.GetFinder(ctx),
```
bryanv: Update to use `finder` var while here

test/e2e/vmservice/lib/vmoperator/vmoperator.go
bryanv, line 1330: How possible is this to observe all the observed CRVs are completed but the rest CRVs just haven't been created yet?

bryanv (approved):

test/e2e/vmservice/lib/vmoperator/vmoperator.go
aruneshpa: CRV creation happens in one shot (during one reconcile) for all disks.  So I would assume that if one CRV exists, all of them must exist.  Obviously there's a window where we check while they are being created, but that's pretty slim.

Furthermore, if you trace the call stack, this method is only called after backup is successful.  Which is only considered successful when the boot disks have a corresponding PVC.
https://github.com/vmware-tanzu/vm-operator/blob/main/test/e2e/vmservice/vmservice/util.go#L829

PR #1810, opened 2026-08-11
Register a dedicated testbed PyKMIP VM as gce2e-standard KMS

aakashchan:
## Summary
- Some testbed deployers now provision a dedicated PyKMIP server VM as part of testbed provisioning, but nothing consumed it — `setup-testbed-env.sh` had no way to see it, so `gce2e-standard` was only ever backed by pykmip freshly installed on the gateway VM, which is fragile (SSH probing, DNS fixes) and unavailable on testbeds without a reachable gateway.
- `setup-testbed-env.sh`: adds `_parse_pykmip_server`, reading the dedicated PyKMIP VM's ip/username/password directly out of `testbed_info` (`genericVm[]`/`type==PyKMIPServer` for VDS-style testbeds, or a top-level `pykmipserver[]` key for others).
- `kms.sh`: `install()` now prefers the dedicated VM (`PYKMIP_HOST_IP`/`USERNAME`/`PASSWORD`) over the gateway VM as its target — it installs pykmip fresh there with the same self-signed cert used for the gateway-VM path, so `setup()` (which does `govc kms.add`/`govc kms.trust` for `gce2e-standard`) needs no changes at all: it just trusts whichever host `install()` pointed at. Falls back to the existing gateway-VM behavior when no dedicated VM is present.

## Test plan
- [x] `bash -n` on both modified scripts
- [x] Run `setup-testbed-env.sh --e2e` against a real testbed with a dedicated PyKMIP VM and confirm `govc kms.ls gce2e-standard` reports `OverallStatus: green`
- [x] Run an encrypted-PVC e2e test end-to-end against the registered dedicated KMS
- [x] Ran the tests on VDS almost 20 times. No failure related QLC or Key missing
```
STDOUT: [SynchronizedAfterSuite] PASSED [54.604 seconds]

Ran 7 of 182 Specs in 1821.195 seconds
SUCCESS! -- 7 Passed | 0 Failed | 4 Pending | 171 Skipped
PASS
```

### DIFF
--- a/hack/e2e/kms.sh
+++ b/hack/e2e/kms.sh
@@ -15,6 +15,13 @@ GATEWAY_VM_USERNAME="${GATEWAY_VM_USERNAME:-root}"
 # GATEWAY_VM_PASSWORD must be set by the caller (setup-testbed-env.sh passes
 # the discovered password). No default — empty fails fast.
 GATEWAY_VM_PASSWORD="${GATEWAY_VM_PASSWORD:-}"
+# When the testbed already has a dedicated PyKMIP server VM (provisioned by
+# the testbed deployer; see setup-testbed-env.sh:_parse_pykmip_server),
+# pykmip is installed/configured against it directly instead of the gateway
+# VM. No default — empty means "no dedicated VM, use gateway".
+PYKMIP_HOST_IP="${PYKMIP_HOST_IP:-}"
+PYKMIP_HOST_USERNAME="${PYKMIP_HOST_USERNAME:-root}"
+PYKMIP_HOST_PASSWORD="${PYKMIP_HOST_PASSWORD:-}"
 script_dir="$(dirname "$0")"
 
 # Common SSH/SCP options for all connections to the gateway VM.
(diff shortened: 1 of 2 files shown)


hack/e2e/kms.sh
```diff
@@ -59,9 +79,9 @@ install() {
...
     fi
 
-    if [ -n "$2" ]; then
-      target="$1@$2"
-      password=$3
+    if [ -n "$target_ip" ]; then
+      target="$target_user@$target_ip"
+      password="$target_password"
 
       sshpass -p "$password" scp $SSH_OPTS "$crt_dir"/pykmip-*.pem "$script_dir"/install-pykmip.sh "$target":
```
bryanv: Does the config already installed on the dedicated PyKMIP server differ from what that script will do?

aakashchan: It comes with a pre installed pykmip and certs. We just need to fetch and call govc kms.trust and govc kms.add

hack/e2e/setup-testbed-env.sh
bryanv, line 646: I suspect this is only really required if we're going to be installing pykmip on the gateway VM, not for squid (if it is even really needed anymore since it is quite old). Fine to defer this until later but if it not needed it would be nice to not do it just so there is one less thing to fail.

bryanv (approved):

hack/e2e/setup-testbed-env.sh
aakashchan: Sure @bryanv, I will remove this as part of another PR.

PR #1813, opened 2026-08-12
:sparkles: Embed extension compatibility constraints in CreateVM

aruneshpa:

Register the 6 INVARIANT extension compatibility constraints (SERVICE, FOLDER, POOL, VM_STORAGE_POLICY, DISK_STORAGE_POLICY, DEVICE) on every VM's ConfigSpec at create time, gated behind the new ExtensionCompatConstraint capability/feature flag.

This lets vpxd's Safety Rails mechanism reject admin-initiated relocate/reconfigure operations that would break Supervisor's namespace, storage-class, and network-identity invariants, while still allowing host and datastore changes within the same cluster and StorageClass.

Bump govmomi to the pseudo-version that carries the new VirtualMachineExtensionCompatibilityConstraint(Set) types and the SkipExtensionCompatibilityChecks field, matching the commit Mobility Operator already depends on for its own bypass ticket.

```release-note
Embed extension compatibility constraints in CreateVM
```


pkg/providers/vsphere/virtualmachine/configspec.go
```diff
@@ -43,6 +43,10 @@ func CreateConfigSpec(
 		Type:         vmopv1.ManagedByExtensionType,
 	}
 
+	if pkgcfg.FromContext(vmCtx).Features.ExtensionCompatConstraint {
+		configSpec.ExtensionCompatibilityConstraint = extensionCompatibilityConstraintSet()
+	}
```
bryanv, line 50: `} else { // set to nil` ?

bryanv (approved):

PR #1765, opened 2026-07-23
Fix the vm-location e2e test failure wrt zone mismatch in vds-ssv setup

anwithaun-1:
**What does this PR do, and why is it needed?**

- Fixes a flaky failure in the vm-location E2E specs (VM-LOCATION … moved outside the namespace RP hierarchy) on multi-zone Supervisors.
- The test helper getNsRPAndFolder returned the RP/folder of the first Zone in the namespace's ZoneList (always zone-1, since client.List sorts by name) rather than the zone the VM was actually scheduled into. On a multi-zone Supervisor the namespace resource pool is per-zone (each zone has its own resgroup), while the folder is shared across zones. So when the scheduler placed the VM in zone-2/zone-3, the spec "restored" it to zone-1's RP — a location the controller correctly rejects — and VirtualMachineLocationValid never returned to True, timing out after 300s
```
[38;5;9mâ€¢ [FAILED] [398.771 seconds][0m
[0mTesting VM Services [38;5;243mVIRTUAL-MACHINE [0mVM-LOCATION [38;5;243mwhen VM is moved outside the namespace RP hierarchy [38;5;9m[1m[It] sets condition False, then recovers to True when VM is returned to the correct location[0m [38;5;204m[devops, viadmin, virtualmachine, vmservice, core-functional, experimental][0m
[38;5;243m/vm-operator/test/e2e/vmservice/vmservice/virtualmachine/vm_location.go:260[0m

  [38;5;9m[FAILED] Timed out after 300.009s.
  Timed out waiting for Condition: {Type:VirtualMachineLocationValid Status:True ObservedGeneration:0 LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:} on VirtualMachine: vm-location-lwh9
  The function passed to Eventually failed at /vm-operator/test/e2e/vmservice/lib/vmoperator/vmoperator.go:313 with:
  Expected
      <v1.ConditionStatus>: False
  to equal
      <v1.ConditionStatus>: True[0m
```
- This made the spec flaky: it passed only when placement happened to land the VM in zone-1, and failed otherwise. The Folder spec was unaffected because folders are zone-independent.

**Which issue(s) is/are addressed by this PR?** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes #
vmop-3844

**Are there any special notes for your reviewer**:

[...]

PR #1821, opened 2026-08-14
Fix vm-location E2E permission scoping and namespace lookup

anwithaun-1:

- Replaces the global, role-ID-based EnsureRolePrivileges (which permanently mutated the shared VM-Service-VM-Management role that WCP itself manages) with a new GrantExtraPrivileges/swapEntityPermissionRole helper in roles.go. It clones the base role into a test-scoped role, swaps that clone onto only the specific RP/folder entities each spec relocates a VM across, and restores the original permissions via DeferCleanup when the spec ends — instead of leaving a cluster-wide privilege change behind.
- Drops the AvailabilityZone lookup from getNsRPAndFolder/getOtherNamespaceFolder entirely. Both now read Zone.Spec.ManagedVMs — the same fields the controller's topology.GetNamespaceFolderAndRPMoID returns and reconcileLocation validates the VM against — and a missing Zone fails the spec rather than silently resolving a value the controller would never have used. This addresses review feedback on https://github.com/vmware-tanzu/vm-operator/pull/1765#discussion_r3648583378 that the AvailabilityZone CR should no longer be used for this. 
- Un-skips the three vm-location specs (XWhen → When) now that both fixes above make them reliable.

Fixes # vmop-4033

Ran UTS postcheckin against the changes and got a successfull run for these tests

```release-note
NONE
```

### DIFF
--- a/test/e2e/infrastructure/vsphere/vcenter/roles.go
+++ b/test/e2e/infrastructure/vsphere/vcenter/roles.go
@@ -5,9 +5,11 @@ package vcenter
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"net/url"
+	"slices"
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/ssoadmin"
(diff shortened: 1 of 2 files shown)


aruneshpa (approved):
Make sure you get at least a few runs of this (locally) before we merge.

PR #1822, opened 2026-08-14
Fix VM publish test "should preserve vAppConfig properties on a VM deployed from the published image" by skipping  disk promotion

anwithaun-1:
Fixes a flake in the VirtualMachinePublishRequest E2E spec should preserve vAppConfig properties on a VM deployed from the published image, which timed out after  waiting for Complete=True.
```
â€¢ [FAILED] [583.156 seconds]
Testing VM Services VIRTUAL-MACHINE VM-PUBLISH-REQUEST VirtualMachinePublishRequest CLS Content Library [It] should preserve vAppConfig properties on a VM deployed from the published image [devops, viadmin, virtualmachine, vmservice, extended-functional, experimental]
/vm-operator/test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go:305

  [FAILED] Timed out after 360.003s.
  Timed out waiting for Condition: {Type:Complete Status:True ObservedGeneration:0 LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:} on VirtualMachinePublishRequest: vm-publish-ouoa-vapp
  The function passed to Eventually failed at /vm-operator/test/e2e/vmservice/lib/vmoperator/vmoperator.go:695 with:
  Expected
      <v1.ConditionStatus>: False
  to equal
      <v1.ConditionStatus>: True
  In [It] at: /vm-operator/test/e2e/vmservice/lib/vmoperator/vmoperator.go:700 @ 08/13/26 20:37:22.703
```

- The `VirtualMachinePublishRequest` spec `should preserve vAppConfig properties on a VM deployed from the published image` intermittently times out waiting for `Complete=True`
- This is a race with VM Operator's background disk promotion. The spec creates its own source VM as a Fast Deploy linked clone and publishes it ~3 minutes later — while `PromoteDisks_Task` is still running. vCenter serializes the OVF capture behind that task, so `Uploaded` stays `Queued`/`Uploading` and `Complete` never flips inside the spec's 6m budget. The neighbouring smoke publish spec is unaffected because it publishes the long-lived suite VM, whose promotion finished ~50 minutes earlier.

**Fix:** 
- Deploy the source VM with promotedisks=disabled, this skips the disk promotion task and there'll not be any compete task with publishrequest.
- This doesn't weaken the spec — publish has no dependency on disk promotion. The spec asserts that a vApp property set via `bootstrap.vAppConfig` survives publish and reappears on a VM deployed from the published image, and vApp properties live in `config.vAppConfig`, captured into the OVF envelope — disk provenance is irrelevant to it.

Also re-adds `experimental` to the spec's labels, since it is being changed and has not been validated on hardware in this form.

Fixes # vmop-4048

```release-note

```

### DIFF
--- a/test/e2e/vmservice/config/wcp.yaml
+++ b/test/e2e/vmservice/config/wcp.yaml
@@ -52,7 +52,6 @@ intervals:
   default/wait-virtual-machine-vmip: ["5m", "3s"]
   default/wait-virtual-machine-service-endpoints: ["2m", "5s"]
   default/wait-virtual-machine-image-creation: ["10m", "5s"]
-  default/wait-virtual-machine-disk-promotion: ["10m", "10s"]
   default/consistent-virtual-machine-condition: ["30s", "5s"]
   default/wait-virtual-machine-annotation-update: ["30s", "5s"]
   default/wait-virtual-machine-group-deletion: [ "60s", "5s" ]
--- a/test/e2e/vmservice/config/wcp.yaml
+++ b/test/e2e/vmservice/config/wcp.yaml
@@ -52,7 +52,7 @@ intervals:
   default/wait-virtual-machine-vmip: ["5m", "3s"]
   default/wait-virtual-machine-service-endpoints: ["2m", "5s"]
   default/wait-virtual-machine-image-creation: ["10m", "5s"]
-  default/wait-virtual-machine-disk-promotion: ["6m", "5s"]
+  default/wait-virtual-machine-disk-promotion: ["10m", "10s"]
   default/consistent-virtual-machine-condition: ["30s", "5s"]
   default/wait-virtual-machine-annotation-update: ["30s", "5s"]
   default/wait-virtual-machine-group-deletion: [ "60s", "5s" ]
--- a/test/e2e/vmservice/config/wcp.yaml
+++ b/test/e2e/vmservice/config/wcp.yaml
@@ -52,6 +52,7 @@ intervals:
   default/wait-virtual-machine-vmip: ["5m", "3s"]
   default/wait-virtual-machine-service-endpoints: ["2m", "5s"]
   default/wait-virtual-machine-image-creation: ["10m", "5s"]
+  default/wait-virtual-machine-disk-promotion: ["6m", "5s"]
   default/consistent-virtual-machine-condition: ["30s", "5s"]
   default/wait-virtual-machine-annotation-update: ["30s", "5s"]
   default/wait-virtual-machine-group-deletion: [ "60s", "5s" ]
--- a/test/e2e/vmservice/lib/vmoperator/vmoperator.go
+++ b/test/e2e/vmservice/lib/vmoperator/vmoperator.go
@@ -127,32 +127,6 @@ func WaitForVirtualMachineImageCacheReady(ctx context.Context,
 		Should(Succeed(), "Timed out waiting for VirtualMachine %s/%s image cache to be ready", ns, vmName)
 }
 
-// WaitForVirtualMachineDiskPromotionSynced waits for the VM's disk promotion to
-// finish, as indicated by the VirtualMachineDiskPromotionSynced condition.
-//
-// A missing condition means no promotion is in flight (e.g. FastDeploy is
-// disabled, or promoteDisksMode is Disabled), not that one is pending, so it
-// is treated as success rather than something to keep waiting on.
-func WaitForVirtualMachineDiskPromotionSynced(ctx context.Context,
-	config *config.E2EConfig,
-	client ctrlclient.Client, ns, vmName string) {
-	By("Waiting for any VirtualMachine disk promotion to complete")
-
-	Eventually(func(g Gomega) {
-		vm, err := utils.GetVirtualMachine(ctx, client, ns, vmName)
-		g.Expect(err).ToNot(HaveOccurred(), "failed to get VirtualMachine %s/%s", ns, vmName)
-
-		cond := meta.FindStatusCondition(vm.GetConditions(), vmopv1.VirtualMachineDiskPromotionSynced)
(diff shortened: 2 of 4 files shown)


test/e2e/vmservice/config/wcp.yaml
```diff
@@ -82,6 +82,7 @@ intervals:
   default/wait-virtual-machine-vmip: ["5m", "3s"]
   default/wait-virtual-machine-service-endpoints: ["2m", "5s"]
   default/wait-virtual-machine-image-creation: ["10m", "5s"]
+  default/wait-virtual-machine-disk-promotion: ["10m", "10s"]
```
aakashchan: How are these values derived?

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -127,6 +127,32 @@ func WaitForVirtualMachineImageCacheReady(ctx context.Context,
 		Should(Succeed(), "Timed out waiting for VirtualMachine %s/%s image cache to be ready", ns, vmName)
 }
 
+// WaitForVirtualMachineDiskPromotionSynced waits for the VM's disk promotion to
+// finish, as indicated by the VirtualMachineDiskPromotionSynced condition.
+//
+// A missing condition means no promotion is in flight (e.g. FastDeploy is
+// disabled, or promoteDisksMode is Disabled), not that one is pending, so it
+// is treated as success rather than something to keep waiting on.
+func WaitForVirtualMachineDiskPromotionSynced(ctx context.Context,
```
aakashchan: There are already existing util methods. Check `WaitOnVirtualMachineCondition`, does it not help?

aruneshpa: Yep, please consider using the existing helper.

test/e2e/vmservice/config/wcp.yaml
aruneshpa: Can you reduce the polling period of the promotion to 3s?  Disk promotion doesn't take long and we don't want to wait for several seconds because of a large poll interval.

anwithaun-1: Derived from the failing run: the source VM's own promotion task was still running past 6m42s when the publish timed out, and a similar task on the same testbed took roughly 7m12s to finish. So 10m gives some margin above that while matching the other slow vSphere waits already in this file (image-creation, backup-to-complete, etc., all 10m). Went with a 10s poll for the same reason those use it — promotion takes minutes either way, so polling faster just adds load without catching it any sooner.

test/e2e/vmservice/lib/vmoperator/vmoperator.go
anwithaun-1: I checked it first. It's close but doesn't quite work for this case — WaitOnVirtualMachineCondition requires the condition to exist (actualCondition != nil), but here a missing VirtualMachineDiskPromotionSynced is actually fine, not something to fail on. The diskpromo reconciler only runs when FastDeploy is enabled, and it deletes the condition entirely when promoteDisksMode: Disabled. If I used the existing helper, this spec would hard-fail on any testbed without FastDeploy instead of just skipping the wait.

It also defaults to the wait-virtual-machine-creation interval (5m/3s), which is too short — promotion can queue behind guest customization and take longer than that on its own

test/e2e/vmservice/config/wcp.yaml
anwithaun-1: changed the fix. Please review the updated changes.

hpannem:
Looks ok to me. This test is not concerned with promote disks at all.

test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go
```diff
@@ -312,6 +312,12 @@ func VMPublishRequestSpec(ctx context.Context, inputGetter func() VMPublishReque
 					StorageClassName: clusterResources.StorageClassName,
 					ResourcePolicy:   clusterResources.VMResourcePolicyName,
 					PowerState:       "PoweredOn",
+					Annotations: map[string]string{
```
hpannem: you can use promoteDisks: Disabled

anwithaun-1: Made the changes to use this flag, Had to create the VM with typed struct to use this flag and stop using rendered yamls as this flag is not added to virtualmachineyaml's yet

hpannem (approved):

test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go
```diff
@@ -285,7 +286,7 @@ func VMPublishRequestSpec(ctx context.Context, inputGetter func() VMPublishReque
 				vmoperator.DeleteVirtualMachine(ctx, svClusterClient, input.WCPNamespaceName, newVmName)
 			})
 
-			It("should preserve vAppConfig properties on a VM deployed from the published image", Label("extended-functional"), func() {
+			It("should preserve vAppConfig properties on a VM deployed from the published image", Label("extended-functional", "experimental"), func() {
```
faisalabujabal, line 289: if there are no significant changes or refactor to a test (like this PR) I don't think we need to set it as experimental. just please make sure it passes before you merge

faisalabujabal (approved):

test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go
anwithaun-1: It's already merged by Rafa, So I'll montor the pipeline and then remove experimental in the next PR.  However the test passed locally with multiple runs

PR #1824, opened 2026-08-14
:sparkles: Reconcile stale extension compat constraints on VMs

aruneshpa:

Fold a generic, level-triggered check into the existing config reconfigure path: compare a VM's current
config.extensionCompatibilityConstraint set against the desired 6 INVARIANTs (ignoring the cosmetic ConstraintName and array order), and re-send the full desired set whenever they differ.

This is folded into doReconfigure's shared ConfigSpec rather than given its own standalone Reconfigure call, so it batches with other pending config changes into a single vCenter round trip. The trade-off is that reconcileConfig's existing verifyConnectionState check means this fix-up is skipped for VMs with a disconnected host, same as the rest of config reconciliation.

vmop-3779

```release-note
Reconcile stale extension compat constraints on VMs
```

### DIFF
--- a/pkg/providers/vsphere/virtualmachine/extensioncompatconstraint.go
+++ b/pkg/providers/vsphere/virtualmachine/extensioncompatconstraint.go
@@ -1,0 +1,64 @@
+// © Broadcom. All Rights Reserved.
+// The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+
+package virtualmachine
+
+import (
+	vimtypes "github.com/vmware/govmomi/vim25/types"
+)
+
+// extensionCompatConstraintKey identifies an extension compatibility
+// constraint's identity on the wire: the (ConstraintType, ConstraintKind)
+// pair. ConstraintName is a descriptive label only and is deliberately
+// excluded from this key.
+type extensionCompatConstraintKey struct {
+	constraintType string
+	constraintKind string
+}
+
+// extensionCompatibilityConstraintSetMatches reports whether current already
+// contains exactly the desired set of INVARIANT constraints, ignoring
+// ConstraintName and the order of the Constraint slice.
+func extensionCompatibilityConstraintSetMatches(
+	current *vimtypes.VirtualMachineExtensionCompatibilityConstraintSet) bool {
+
+	desired := extensionCompatibilityConstraintSet().Constraint
+
+	if current == nil || len(current.Constraint) != len(desired) {
+		return false
+	}
+
+	currentKeys := make(map[extensionCompatConstraintKey]struct{}, len(current.Constraint))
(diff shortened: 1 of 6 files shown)


yanleizhao-vmware:
I believe it's already taken care of because it's a full set reconfigure, but just want to emphasize in case I'm missing some corner case -- vpxd would reject setting extension compatibility constraints if the vm lacks "managedBy" field, since these constraints would otherwise be stored in vcdb and have no effect when managedBy is missing.

aruneshpa:
> I believe it's already taken care of because it's a full set reconfigure, but just want to emphasize in case I'm missing some corner case -- vpxd would reject setting extension compatibility constraints if the vm lacks "managedBy" field, since these constraints would otherwise be stored in vcdb and have no effect when managedBy is missing.

Yes.  VM Operator ensures that the managedBy is always set on the VMs that we reconcile. 

To save time, I had AI trace the code:

----
- On create: CreateConfigSpec unconditionally sets ManagedBy = {ExtensionKey: vmopv1.ManagedByExtensionKey, Type: vmopv1.ManagedByExtensionType} (pkg/providers/vsphere/virtualmachine/configspec.go:41-44).

- On day-2 drift: every update/resize reconcile path calls overwriteManagedBy (pkg/util/vmopv1/resize_overwrite.go:96-121, invoked from pkg/providers/vsphere/session/session_vm_update.go at lines 164, 417, 660, 990). It reads the VM's live ConfigInfo.ManagedBy, compares against VM Operator's expected value, and forces it back to VM Operator's ExtensionKey/Type whenever it's non-empty — so if someone manually clears or overwrites ManagedBy on day 2, the next reconcile Reconfigure call restores it.
-----

FWIW, we do clear this when the VM is being deleted.

pkg/providers/vsphere/virtualmachine/extensioncompatconstraint.go
```diff
@@ -0,0 +1,64 @@
...
+
+// extensionCompatibilityConstraintSetMatches reports whether current already
+// contains exactly the desired set of INVARIANT constraints, ignoring
+// ConstraintName and the order of the Constraint slice.
+func extensionCompatibilityConstraintSetMatches(
+	current *vimtypes.VirtualMachineExtensionCompatibilityConstraintSet) bool {
+
+	desired := extensionCompatibilityConstraintSet().Constraint
+
+	if current == nil || len(current.Constraint) != len(desired) {
```
bryanv, line 28: If `current` is an expected common'ish case, check that first thing?

bryanv (approved):

pkg/providers/vsphere/virtualmachine/extensioncompatconstraint.go
aruneshpa: I will talk to you async and deal with this in a followup if needed.

PR #1833, opened 2026-08-20
Bound RegisterVM disk-only-restore cleanup, repair FCD first & new fault type

aakashchan:
An assertion failure partway through the disk-only-restore spec (e.g. the FCD backing-file fault-type check) leaves the disk's original FCD catalog entry dangling: its backing file has already been moved/ deleted, but RegisterVM never ran to reconcile or re-register it. The outer DeferCleanup that deletes the VM and its PVCs then hangs for hours, since CNS/CSI retries indefinitely against an FCD stuck in this ambiguous state, and with no timeout on that DeferCleanup node, Ginkgo lets it run until the whole suite's grace period expires, blocking every later spec in the process.

Changes:
- Give the VM/PVC DeferCleanup a 3-minute NodeTimeout so a stuck cleanup fails fast instead of consuming the suite's timeout budget.
- Register a second, LIFO-earlier DeferCleanup right after the disk is detached that best-effort force-deletes the FCD catalog entry directly via fcdManager. This repairs the vSphere-side state before the VM/PVC cleanup runs, so CNS/CSI sees a clean NotFound instead of the ambiguous dangling-catalog state and doesn't get stuck retrying.

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/vmservice/vmservice/viadmin/registervm.go
+++ b/test/e2e/vmservice/vmservice/viadmin/registervm.go
@@ -1052,7 +1052,8 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
 			// Expect to fail w/ orphaned FCD
 			_, err = fcdManager.Retrieve(ctx, ds, disk.VDiskId.Id)
 			Expect(err).To(HaveOccurred())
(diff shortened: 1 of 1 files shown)


aruneshpa (approved):

hpannem:
precheckin passed.

PR #1834, opened 2026-08-20
🐛 Improve the describe diagnostics on failure across the e2e suite

brito-rafa:

Add `vmoperator.DescribeObjectsOnFailure`, an AfterEach-callable helper that lists every vm-operator object e2e tests upon failure commonly create, instead of relying on an outer variable (vmYaml/vmName) that can be stale, reassigned to a different resource, or never set for the specific object that actually failed.

Roll this out as a single AfterEach call across the ~19 spec files that create these resources. Files that only touch cluster-scoped or non-vm-operator resources (VM classes, content libraries, plain namespaces) are left unchanged.

Fixes # vmop_3782245

There will be more PRs to standardize some practices across the AfterEach of the suites.

```release-note

```

### DIFF
--- a/test/e2e/vmservice/lib/vmoperator/vmoperator.go
+++ b/test/e2e/vmservice/lib/vmoperator/vmoperator.go
@@ -588,6 +588,55 @@ func DescribeResourceIfExists(ctx context.Context, client ctrlclient.Client, kub
 	e2eframework.Logf("kubectl describe %s -n %s %s:\n%s", resource, ns, resourceName, stdout)
 }
 
+// DescribeObjectsOnFailure describes every instance of the vm-operator CR kinds tests
+// commonly create in ns, if the current spec failed — without depending on the caller
+// having tracked each resource's name/YAML correctly in an outer variable.
+func DescribeObjectsOnFailure(ctx context.Context, client ctrlclient.Client, kubeconfigPath, ns string) {
+	if !CurrentSpecReport().Failed() {
+		return
+	}
+
+	if ns == "" {
+		e2eframework.Logf("skipping vm-operator failure diagnostics: empty namespace")
+		return
+	}
+
+	// Typed List results don't have TypeMeta/Kind populated on their items (a
+	// well-known controller-runtime/client-go quirk), so each List is paired with
+	// the kubectl "resource" argument we pass to DescribeResourceIfExists below.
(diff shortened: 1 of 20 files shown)


bryanv:
> A VM can also land on a host whose NSX logical-switch port isn't yet reported accessible right after creation (PlaceVmsXCluster "which is not accessible" fault); vm-operator's reconciler already retries and self-resolves this within a couple of minutes. WaitForVirtualMachineIP now widens its wait budget on NSX-backed networking via a new nsx/wait-virtual-machine-vmip interval in wcp.yaml, leaving the default budget for every other topology and caller unchanged.

I don't really follow this. Errors from placement would be prior to the VM even being created, so increasing the wait for a powered on VM to get assigned an IP (apply the bootstrap customization and VM Tools to report the IP) don't really matter.

brito-rafa:
Ok, I can revert the logic of bigger wait budget on IPv4 on NSX deployments. Should we keep the other changes such as the collection of events and the describe of the affected VMs for better troubleshooting in the future?

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -588,6 +588,55 @@ func DescribeResourceIfExists(ctx context.Context, client ctrlclient.Client, kub
...
+	}{
+		{&vmopv1.VirtualMachineList{}, virtualMachineKind},
+		{&vmopv1.VirtualMachineGroupList{}, "VirtualMachineGroup"},
+		{&vmopv1.VirtualMachineGroupPublishRequestList{}, "VirtualMachineGroupPublishRequest"},
+		{&vmopv1.VirtualMachinePublishRequestList{}, "VirtualMachinePublishRequest"},
+		{&vmopv1.VirtualMachineServiceList{}, "VirtualMachineService"},
+		{&vmopv1.VirtualMachineSnapshotList{}, "VirtualMachineSnapshot"},
+		{&vmopv1.VirtualMachineWebConsoleRequestList{}, "VirtualMachineWebConsoleRequest"},
+	} {
+		if err := client.List(ctx, entry.list, ctrlclient.InNamespace(ns)); err != nil {
```
bryanv, line 619: We do precreate 3 VMs at the start of the suite so this is going to include them when they are not relevant so I am worried this could too verbose

PR #1835, opened 2026-08-21
🐛 Update MapEthernetDevicesToSpecIdx for mixed network providers

bryanv:

MapEthernetDevicesToSpecIdx was still using the network provider envvar, which does not support mixed network providers since a namespace may have a mix.Instead like is done for for when creating the network interface CRs, use the Network.APIVersion to lookup the CR for each interface.

Fixes # vmop-4023

UTS smoke passed

```release-note
NONE
```

### DIFF
--- a/pkg/providers/vsphere/network/devices.go
+++ b/pkg/providers/vsphere/network/devices.go
@@ -104,8 +104,15 @@ func UpdateVMClassEthCardFromDevice(
 	return nil
 }
 
-// MapEthernetDevicesToSpecIdx maps the VM's ethernet devices to the corresponding
-// entry in the VM's Spec.
+// MapEthernetDevicesToSpecIdx maps the VM's Ethernet cards to the corresponding entry
+// in the VM's Spec. The result of this is used only to populate the VM Status with the
+// interface spec name. Note that since we populate status very early in the VM reconcile,
+// the VM Spec.Network.Interface may have been changed which means there isn't an Ethernet
+// card added to this VM yet. Therefore, all this is best-effort, and we have to use some
+// additional bit of information from the interface spec CR to match it to the device if
+// there is one.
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/network/devices.go
```diff
@@ -145,17 +152,27 @@ func findMatchingEthCardForInterfaceSpec(
...
-	switch pkgcfg.FromContext(vmCtx).NetworkProviderType {
-	case pkgcfg.NetworkProviderTypeVDS:
-		matchingIdx = findMatchingEthCardVDS(vmCtx, client, interfaceSpec, ethCards)
-	case pkgcfg.NetworkProviderTypeNSXT:
-		matchingIdx = findMatchingEthCardNSXT(vmCtx, client, interfaceSpec, ethCards)
-	case pkgcfg.NetworkProviderTypeVPC:
-		matchingIdx = findMatchingEthCardVPC(vmCtx, client, interfaceSpec, ethCards)
-	case pkgcfg.NetworkProviderTypeNamed:
-		matchingIdx = findMatchingEthCardNamed(vmCtx, client, interfaceSpec, ethCards)
+	// For this interface spec, try to find the Ethernet card that it most likely
```
aruneshpa, line 163: Just double checking: nothing in the below code is dependent on the PerNamespaceNetworkProvider capability?  I don't think so since you are only trying to parse the API version of the netIf resource, but just to double check.

aruneshpa (approved):

pkg/providers/vsphere/network/devices.go
bryanv: Right. What this does is get us is support for multiple network providers for free since we just use what is in the spec, with the what is allowed is enforced up front in the validation webhook.

PR #1839, opened 2026-08-24
🌱  Retry kubectl calls on transient connectivity errors

bryanv:

Create, Delete, and Apply on VMServiceClusterProxy would fail an
entire e2e spec on a single transient blip, e.g. the conversion
webhook being briefly unreachable during a vm-operator pod rollout.
Retry the underlying kubectl call every 15s for up to 2.5 minutes,
but only when the error matches a set of known transient network
substrings (dial tcp, connection refused, context deadline exceeded,
etc.); a real admission/validation error still fails immediately.

Since create/delete are not idempotent, also treat an AlreadyExists
(create) or NotFound (delete) on a retried attempt as success rather
than a hard failure, since the object may have been persisted before
the connection reporting the failure dropped.

CreateRawWithArgs is left un-retried since it is used to assert on an
expected admission error message in a negative test path.

Fixes #

Need to handle the ctrl-runtime client next (and ideally someday replace all the kubectl calls with that as well)

```release-note
NONE
```

### DIFF
--- a/test/e2e/vmservice/common/vmservice_clusterproxy.go
+++ b/test/e2e/vmservice/common/vmservice_clusterproxy.go
@@ -3,12 +3,16 @@ package common
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"net/url"
 	"os"
+	"strings"
+	"time"
 
 	. "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	e2eframework "k8s.io/kubernetes/test/e2e/framework"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


hpannem:
Pipeline run successful.

hpannem (approved):

aakashchan (approved):

test/e2e/vmservice/common/vmservice_clusterproxy.go
```diff
@@ -18,6 +22,143 @@ import (
...
+	}
+
+	return false
+}
+
+// alreadySucceededFunc reports whether a retried write actually landed on a
+// previous attempt before its connection dropped, e.g. kubectl create sees
+// AlreadyExists, kubectl delete sees NotFound.
+//
+// These predicates assume a single-document manifest (no "---" separators),
```
aruneshpa, line 77: Is there a way to detect and throw an error if such a manifest file is ever used?  I fear that this soft rule will get missed during code reviews and inevitably someone will commit a multi document manifest.

aruneshpa (approved):

test/e2e/vmservice/common/vmservice_clusterproxy.go
bryanv: There are already two multidoc - VM and PVC - so generally if one is going to fail they both will due to the existing validation webhooks. It isn't worth investing time to try to fix that since the whole go template yaml crap and the use of kubectl needs to just go away

PR #1844, opened 2026-08-25
🌱 Retriable controller runtime client for transient errors

hpannem:

Wraps every controller-runtime `client.Client` used by the vmservice e2e suite in a retrying decorator, so a single transient connectivity blip — e.g. the vm-operator conversion webhook being briefly unreachable while its pod restarts after losing leader election — no longer fails an entire spec. This is the controller-runtime-client counterpart to the kubectl-based retry in `#1839`: that PR retries `kubectl create/delete/apply`, this one retries the Go client's `Get`, `Create`, `Update`, `Patch`, `Delete`, and `DeleteAllOf`.

`retryingClient` (new, `test/e2e/framework/retry_client.go`) embeds `client.Client` and overrides only the mutating/read methods; `List`, `Status()`, `SubResource()`, `Scheme()`, `RESTMapper()`, etc. pass through unmodified. It retries every 15s for up to 4 minutes, but only on errors classified as transient: a dead conversion webhook surfaces as a typed `apierrors.IsInternalError`, so that check runs first, with a substring fallback (`dial tcp`, `connection refused`, `context deadline exceeded`, ...) for raw transport errors that never reach a typed `apierrors.APIStatus`. A real admission or validation error is returned immediately.

`Create` and `Delete` treat `AlreadyExists`/`NotFound` on a *retried* attempt (not the first) as success, since the write may have landed before the connection reporting failure dropped. `Update` and `Patch` intentionally get no such carve-out: a `Conflict` on retry is ambiguous — it can mean the earlier attempt landed, or that someone else wrote to the object — so it's surfaced as a real error. Call sites that mutate via `Update` already re-fetch and retry at a higher level (`lib/vmoperator`), which self-heals a spurious conflict from this case.

The wrapper is wired in at every client-construction path actually reachable from the suite: `clusterProxy.GetClient()` (the single chokepoint ~20 test files get their client from), `VMServiceClusterProxy.GetAdminClient()` (used by the admin-privileged flows in `util.go`, `registervm.go`, `configpolicy.go`), and the one-off non-admin test-user client in `vm_publishrequest.go`. No existing call site needed to change since `GetClient()`'s return type is unchanged. `workloadClusterProxy.GetClientNoExpect` was left untouched — it has no callers anywhere in the suite.

The retryable-error substring list and retry interval/timeout are exported (`framework.RetryableTransientErrSubstrings`, `framework.IsRetryableTransientErrString`, `framework.TransientRetryInterval`, `framework.TransientRetryTimeout`) so that once `#1839` merges, its kubectl-based classifier and constants can be pointed at these instead of maintaining a second copy.

Fixes #

Depends on nothing from `#1839` to build or pass on its own — the two retry paths are independent today. Once `#1839` merges, a small follow-up here (or there) should collapse `vmservice_clusterproxy.go`'s `retryableKubectlErrSubstrings`/`kubectlRetryInterval`/`kubectlRetryTimeout` down to references to the exported `framework` symbols, dropping the duplicate list and constants.

```release-note
NONE

### DIFF
--- a/test/e2e/framework/cluster_proxy.go
+++ b/test/e2e/framework/cluster_proxy.go
@@ -132,7 +132,12 @@ func (p *clusterProxy) GetScheme() *runtime.Scheme {
 	return p.scheme
 }
 
-// GetClient returns a controller-runtime client for the cluster.
+// GetClient returns a controller-runtime client for the cluster. The
+// returned client retries Get, Create, Update, Patch, Delete, and
+// DeleteAllOf on transient connectivity errors -- e.g. the conversion
+// webhook being briefly unreachable during a vm-operator pod rollout --
+// instead of failing an entire e2e spec on a single blip. See
+// NewRetryableClient.
 func (p *clusterProxy) GetClient() client.Client {
 	config := p.GetRESTConfig()
 
@@ -150,7 +155,7 @@ func (p *clusterProxy) GetClient() client.Client {
 		return c
 	}, 5*time.Minute, 10*time.Second).ShouldNot(BeNil(), "Failed to get controller-runtime client")
 
-	return c
+	return NewRetryableClient(c)
 }
 
 // GetClientSet returns a client-go client for the cluster.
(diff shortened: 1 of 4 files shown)


aruneshpa (approved):
Assuming this is the same thing as #1839 , but for controller-runtime clients, this LGTM.

bryanv (approved):

PR #1846, opened 2026-08-25
:sparkles: Bootstrap KubeVM project to provide generic VM APIs for cross-provider VM management

aruneshpa:

Add a new Go module carrying a first, concrete implementation of the generic, provider-agnostic virtual machine API proposed for the kube-vm.io group, along with CRD generation and the design one-pager.

The goal is to make the proposed API reviewable as something that can be generated, installed, and applied, rather than as a Go snippet in a design document. A VirtualMachine here describes a VM portably and binds to a specific backend through spec.infrastructureRef, which points at a provider-owned object. The generic layer observes that backend only through a duck-typed status contract of well-known field paths, so this module imports no provider code and a provider can evolve independently behind that contract.

Notable choices, and why:

- The API group is kube-vm.io rather than something under x-k8s.io. An x-k8s.io suffix is Kubernetes-project namespace, and adopting one requires SIG and CNCF TAG naming approval; a self-owned domain removes that dependency and the schedule risk with it. It also reads as the project's own rather than as an accepted Kubernetes sub-project, which is the more honest signal for a proposal. Provider groups derive from it, for example infrastructure.vsphere.kube-vm.io.

- Power states are spelled PoweredOn and PoweredOff rather than On and Off. "On" and "Off" are YAML 1.1 boolean literals, so an unquoted "powerState: On" parses as true and is rejected by the API server. These spellings also match VM Operator's existing enum, so the vSphere provider needs no value mapping.

- There is no dedicated accelerator field. Accelerators are requested through instanceType, because on vSphere and EC2 an accelerator is inseparable from the class or instance type, and a standalone field would collapse back into profile selection for most providers. A portable per-VM field waits for a Dynamic Resource Allocation shape that holds across providers.

- Kubernetes dependencies are pinned deliberately low, matching the sibling api module. This is a types module meant to be imported by provider implementations, where a low floor maximizes the range of  Kubernetes versions a consumer can build against.

- The Makefile follows this repo's conventions: the same target names, grouped help, OS/arch-scoped tooling directory, and podman-locally and docker-in-CI runtime selection. It does not replicate the hack/tools sub-module, since two pinned tools do not justify a second go.mod.

This module intentionally contains API types and generation only. There is no controller, the provider contract is not yet expressed in code, there is no conformance suite, and the companion catalog types that bootDisk and interfaces reference are left undefined pending community discussion. Known gaps and open API questions are recorded in the module README.

The generated CRD has been verified against a live API server: it establishes, is discoverable as kube-vm.io/v1alpha1, accepts a valid object, applies its defaults, renders its printer columns, and rejects a missing infrastructureRef, an out-of-enum power state, and duplicate disk names.

The module lives under this repo only so the API can be iterated on with tooling already available here, and is expected to move once a permanent home is approved. Its module path will change on that move, so nothing outside it should import it.

```release-note
Bootstrap KubeVM project to provide generic VM APIs for cross-provider VM management
```


akutz (changes requested):
* The `kubevm` package should be placed in the `external` directory. 
* There should also be an SDD dir describing this.
* The K8s annotations should be placed ahead of the token comments to avoid showing up in GoDocs.

aruneshpa:
> * The `kubevm` package should be placed in the `external` directory.
> * There should also be an SDD dir describing this.
> * The K8s annotations should be placed ahead of the token comments to avoid showing up in GoDocs.

I will move the project to `external` dir.

> * There should also be an SDD dir describing this.
So, as the PR description and the README calls out, this is a temporary placeholder location for the Sandbox application while we figure out a home for this.  So, an .sdd for this is not needed (since no agent will read or implement it).

> * The K8s annotations should be placed ahead of the token comments to avoid showing up in GoDocs.
Will fix.

external/kubevm/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	// PowerStateOn indicates the VM should be powered on.
+	PowerStateOn PowerState = "PoweredOn"
+
+	// PowerStateOff indicates the VM should be powered off.
+	PowerStateOff PowerState = "PoweredOff"
+
+	// PowerStateSuspended indicates the VM should be suspended.
+	PowerStateSuspended PowerState = "Suspended"
+)
```
zjs, line 50: **suggestion**: Consider how this will be generalized to handle the different state semantics from different infrastructure providers.

For example, Azure distinguishes between `Stopped (Allocated)` vs `Stopped (Deallocated)`. Does a simple enum allow the difference between these to be communicated?

Or for even more nuance, GCP allows for some instance types to be suspended, but not all. (If I'm understanding correctly, suspend is not supported on N2, C2, or GPU instances.) How would a user of the generic API understand/discover the way the choice of instance affects viable power states?

In general, it may be useful to have a markdown file considering each dimension of the API from the perspective of a few different providers. (Perhaps starting with the vendors who have the most actively maintained Cluster API provider implementations?)

I think a directory of such documents would be a good companion to `external/kubevm/docs/one-pager-kubevm-generic-api.md`.

external/kubevm/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	// PowerOpModeHard powers off the VM without involving the guest.
+	PowerOpModeHard PowerOpMode = "Hard"
+
+	// PowerOpModeSoft asks the guest to shut down and fails if it does not.
+	PowerOpModeSoft PowerOpMode = "Soft"
+
+	// PowerOpModeTrySoft attempts a guest shutdown, then falls back to hard.
+	PowerOpModeTrySoft PowerOpMode = "TrySoft"
+)
```
zjs, line 80: **question**: Does this generalize across providers?

external/kubevm/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	VirtualMachinePhasePending      VirtualMachinePhase = "Pending"
+	VirtualMachinePhaseProvisioning VirtualMachinePhase = "Provisioning"
+	VirtualMachinePhaseRunning      VirtualMachinePhase = "Running"
+	VirtualMachinePhaseStopping     VirtualMachinePhase = "Stopping"
+	VirtualMachinePhaseStopped      VirtualMachinePhase = "Stopped"
+	VirtualMachinePhaseSuspended    VirtualMachinePhase = "Suspended"
+	VirtualMachinePhaseFailed       VirtualMachinePhase = "Failed"
+	VirtualMachinePhaseDeleting     VirtualMachinePhase = "Deleting"
+)
```
zjs, line 176: **question**: Do we need phases? Kubernetes API conventions have discouraged phase, in favor of conditions.

From [Kubernetes API conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties):
 > Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted [system-design principles](https://git.k8s.io/design-proposals-archive/architecture/principles.md#control-logic) and hampered evolution, since [adding new enum values breaks backward compatibility](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md). Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See [#7856](http://issues.k8s.io/7856) for more details and discussion.

We're not bound by Kubernetes SIG guidance here, but the reasons for deprecating the pattern seem to apply.

external/kubevm/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,161 @@
...
+// ObjectReference references another object in the same namespace, by group,
+// kind, and name. It is the single reference shape used everywhere in this API:
+// for the infrastructure object, the boot image, a snapshot, and a network.
+//
+// Two deliberate omissions.
+//
+// There is no namespace field. A referent is always in the same namespace as
+// the referring VirtualMachine, so a VirtualMachine cannot reach across a
+// namespace boundary to adopt an object or read a catalog entry it was not
+// granted.
```
zjs: **question:** Is this an assumption that's going to generalize? Instance types, image catalogs, network definitions, etc. are often shared across workloads. Limiting the scope to a namespace makes sense in the context of the vSphere Namespace tenancy model, but I'm not sure that other users in other contexts would want that same behavior (e.g., even consider running KubeVM in a VKS cluster — would we want to have to duplicate all referenced objects across all namespaces?).

Perhaps we need a generic mechanism for indicating which cross-namespace references are allowed. This could be viable as a future goal, but current seems to be framed as more of a non-goal.

external/kubevm/api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Type describes the reachability scope of this address.
+	Type VirtualMachineAddressType `json:"type"`
+
+	// +kubebuilder:validation:MinLength=1
+	// +required
+
+	// Address is the IP address or DNS name.
+	Address string `json:"address"`
+}
```
zjs, line 290: **question**: How does this all relate to `NetworkInterfaceSpec`? It seems like there's some confusing asymmetry here between how networking topology is captured in the spec vs. status.

external/kubevm/api/v1alpha1/virtualmachine_types.go
```diff
@@ -0,0 +1,475 @@
...
+	// +optional
+
+	// CPUs is the number of virtual CPUs presented to the guest.
+	CPUs int32 `json:"cpus,omitempty"`
+
+	// +optional
+
+	// Memory is the guest-visible memory, for example "8Gi".
+	Memory *resource.Quantity `json:"memory,omitempty"`
+}
```
zjs, line 55: ([related](https://github.com/vmware-tanzu/vm-operator/pull/1846/changes#r3857448890))
**question**: How will this map to GCP where GPUs are attached dynamically to general-purpose instances? (Seems like that aligns more with a DRA-style resource claim.)

zjs, line 55: **question**: How will this handle GCP custom machine types? If I understand correctly, these require specifying a machine family. Does this somehow happen at the infrastructure provider layer?

external/kubevm/api/v1alpha1/virtualmachine_types.go
```diff
@@ -0,0 +1,475 @@
...
+	// +optional
+
+	// StorageClass is a provider-interpreted storage selector.
+	StorageClass string `json:"storageClass,omitempty"`
+
+	// +optional
+
+	// DeleteOnTermination deletes this disk when the VM is deleted.
+	DeleteOnTermination *bool `json:"deleteOnTermination,omitempty"`
+}
```
zjs, line 262: **question**: Does the distinction between these generalize? (E.g., seems like AWS decouples provisioned IOPS and MB/s throughput from disk size, and allows performance characteristics to be specified for boot disks as well?)

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -0,0 +1,228 @@
...
+| **GCP** | `guestAccelerators[]` (type and count) attached to the instance, or built into `a2`/`g2` machine types | the provider selects the corresponding machine type (or attaches the accelerator) |
+
+A dedicated, portable per-VM accelerator field is deliberately deferred. On vSphere and EC2 the accelerator is inseparable from the class or instance type — a vGPU is a property of the `VirtualMachineClass`, a GPU is intrinsic to an EC2 instance type — so a standalone field would collapse straight back into profile selection for two of the three providers. A separate, orthogonal accelerator request is only coherent once providers share a per-instance *attach* model, and the Kubernetes-wide vehicle for that is Dynamic Resource Allocation (DRA), which is still maturing and has no settled cross-provider shape. Until a DRA strategy holds across providers, accelerators stay on the sizing profile; a portable accelerator field can be introduced once that strategy exists. SR-IOV and latency-sensitive scheduling are deferred on the same grounds — they live on the class or the provider object until two providers agree on a portable shape.
```
zjs: ([related](https://github.com/vmware-tanzu/vm-operator/pull/1846/changes#r3857329467))

**thought**: I think the DRA-style part of this is going to be important. I'm not sure we can defer it and still have something that's useful on, for example, GCP.

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -0,0 +1,228 @@
...
+
+One reconciliation to be explicit about, because it differs from the illustrative examples above. Those examples show a bespoke, capability-only `VSphereVirtualMachine` provider object for clarity. The vSphere realization is expected to differ: the provider object would be **VM Operator's own `vmoperator.vmware.com/VirtualMachine` CRD, reused directly** — a full-featured, and therefore *thick*, object — rather than a slim bespoke type. Reusing the native CRD buys day-one parity with everything VM Operator already does, at the cost of a thick provider object managed through a layered-override model, where the generic object supplies a baseline and the provider object may override it. In other words, "thin provider object" is the aspiration the generic API is built toward and is realistic for greenfield cloud providers; a mature platform's provider may reasonably trade thinness for parity. This is also the strongest argument that the design generalizes: it has to accommodate a provider whose native API is richer than the portable core, not only providers built to fit it.
```
zjs: **thought**: Handling duplicate fields between the generic and provider-specific APIs seems … messy.

external/kubevm/docs/one-pager-kubevm-generic-api.md
zjs, line 1: **style**: It might help to use [semantic line breaks](https://sembr.org/) for this sort of MD file, to make it a little easier to leave targeted comments.

zjs (commented):
As a preliminary draft to spark discussion, I think this works. I've included some comments below — but maybe these are topics that should be a part of the post-merge discussion, rather than a debate to gate the merge.

akutz (approved):

akutz (changes requested):
Is there a reason this cannot live in `support/kubevm` until the other repository is available? I do not want this on `main` since it is not its long-term home. The entire purpose of this is to demonstrate to the CNCF that there is a home for this proposal, but this is not the long-term home, just a staging area since the long-term repo is not yet available. Therefore a support branch in this repository should be fine. Putting this on `main` is strictly misleading since this code will not live in `main` in this repository long-term.

aruneshpa:
Merging this in feature/kube-vm per my discussion with Andrew.

PR #1848, opened 2026-08-26
🐛 Guard status.Volumes writes with optimistic lock

faisalabujabal:

`vm.Status.Volumes` is independently read-modify-written by both the `virtualmachine` controller (`vmlifecycle/update_status.go`, which backfills `ControllerType`/`ControllerBusNumber`/`UnitNumber`/etc. from live vSphere hardware info) and the `volumebatch`/`volume` controllers (which track CNS attachment state). All three patch the object via `pkg/patch.Helper`, which today issues a plain `client.MergeFrom` patch with no conflict detection.

A JSON merge patch replaces a list field wholesale rather than diffing individual elements. So whenever a volume controller's own view of a volume changes at all (e.g. `Attached` flipping to `true`), its patch includes the *entire* `Volumes` list as it understands it — silently dropping any field (like `ControllerType`) the `virtualmachine` controller concurrently wrote. Once both controllers go idle, whichever patch landed last simply sticks, with nothing left to correct it.

Reproduced concretely as: after a `VirtualMachine` detaches and reattaches all of its volumes (a supported day-2 operation — see `test/e2e/vmservice/vmservice/virtualmachine/vm_hardware.go`'s "Detaching and reattaching all disks"), `status.Volumes[].ControllerType` (and the other vSphere-derived fields) can come back permanently empty instead of being backfilled, because a volume controller's very next reconcile races the `virtualmachine` controller's backfill and wins.

**The fix, and why it's shaped this way:** adds a `StatusOptimisticLocker` interface to `pkg/patch`. A type that implements it gets its status patch's `resourceVersion` sent as a precondition, so a concurrent write causes a conflict — and a normal, idempotent requeue — instead of a silent overwrite. `VirtualMachine` implements it.

This is a **type-level opt-in** (mirroring how `conditions.Setter`/`Getter` already scope condition-patching), not a per-call boolean option and not a new default for every `patch.Helper` caller. An earlier version of this change made every status patch optimistically locked unconditionally. That's simpler, but a `resourceVersion` precondition is whole-object, not per-field, so it conflicts with *any* concurrent write to *any* part of the object — not just the shared field it's meant to protect. Running the full repo test suite against that version confirmed the cost is real: it broke an unrelated controller's integration test purely because a completely disjoint annotation write raced its status patch and had to recover via a full reconcile requeue. Traced the actual code path too: that recovery is not a cheap requeue — the deferred `patchHelper.Patch` error becomes a plain returned error (not a `pkg/errors` sentinel), so controller-runtime treats it as a genuine reconcile failure with rate-limited backoff. Scoping the lock to the one type that actually has a multi-controller status hazard avoids that blast radius entirely, while still closing the gap for every current *and future* controller that patches `VirtualMachine.status` — `volume_controller.go`'s legacy path included, at no extra call-site cost.

This isn't a new idiom for the repo: `controllers/configtarget` already upserts owner references onto a shared `VirtualMachineConfigOptions` object with an ad hoc `client.MergeFromWithOptimisticLock`, and `controllers/virtualmachineconfigoptions` does the same for `Status.HardwareVersions` on a shared `VirtualMachineGuestOptions` object. Both fan out from one controller's reconcile to a child object at a single call site, so hand-rolling the lock there is natural. Our case differs structurally — three separate controllers each treat `VirtualMachine` as their own primary reconciled object through the shared `pkg/patch.Helper`, so there's no single call site to hand-edit — which is why this needed a marker interface on the shared helper rather than another inline `MergeFromWithOptimisticLock` block. Documented the split (and cross-referenced the existing examples) in `.sdd/memory/operator-best-practices.md`, with a pointer from the constitution's existing fan-out bullet.

Root-caused via UTS pipeline log analysis (controller-manager logs showing the detach fully clearing `status.Volumes`, followed by a burst of interleaved reconciles from both controllers) and confirmed with a deterministic unit-level reproduction (`controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go`) that drives the exact read/compute/patch interleaving against a fake client — no timing, no flakiness. The existing E2E test above already asserts this exact field round-trips correctly, so no new E2E coverage is needed; it will validate this fix once it runs.

Fixes vmop-4108

- `pkg/patch.StatusOptimisticLocker` is opt-in per-type (zero behavior change for every other type/caller); `VirtualMachine` (`api/v1alpha6`) is the only implementer.
- Validated the scoping decision by running the entire `./controllers/...` and `./pkg/patch/...` test suites both with the (rejected) unconditional-lock version and with this type-scoped version — only the latter is regression-free.
- Audited every `pkg/patch.NewHelper` call site in the repo for other multi-writer status hazards; found none beyond `VirtualMachine.Status.Volumes`. The two other `*VirtualMachine` patch.Helper callers (`pkg/prober/worker` for the readiness condition, `virtualmachinereplicaset` for a label) pick up the new interface for free since it's type-scoped, and neither is actually at risk (one never touches `status`, the other only touches `status.conditions`, which is diffed and patched separately).
- Per `.sdd/memory/sdd-standards.md`'s exemption table, a behavior change observable on a Supervisor cluster nominally calls for `spec.md`/`plan.md`/`tasks.md`. Given the scope (one opt-in interface + a single implementing type) and that it's a defect fix against an already-filed bug rather than new product behavior, we're opening this without SDD artifacts — flagging for reviewer judgment on whether that's the right call here.
- `controllers/virtualmachine/volumebatch/volumebatch_controller_unit_test.go` gained a regression test ("Reattaching volumes while another controller concurrently updates status") that reproduces the race and verifies the fix (conflict + successful retry) end to end. Its assertions target the observable outcome (stale write rejected, value survives, next reconcile converges), not the specific conflict-error string, per review feedback that these tests should read as end-user scenarios rather than descriptions of the locking mechanism.
- Per AI review feedback: `Helper.Patch`'s sub-patch reordering (status before conditions) is now scoped to `h.statusOptimisticLock`. Running `patchStatusConditions` last unconditionally would have meant *any* type whose reconcile changes both a condition and spec/status in the same call pays a self-induced conflict-and-retry on the conditions patch (it has its own hardcoded `MergeFromWithOptimisticLock` in its retry loop), for zero benefit on types that never use `StatusOptimisticLocker`. Every other type keeps the original, pre-PR ordering; only the locked path (`VirtualMachine`) accepts that bounded retry cost, and only when status.Volumes and conditions change together.
- Per human review feedback: reverted an unnecessary `noinlineerr`-motivated rewrite of the deferred `patchHelper.Patch` blocks in `virtualmachine_controller.go` and `volumebatch_controller.go` back to the inline `if err := ...; err != nil` form — verified empirically that `noinlineerr` doesn't flag that specific pattern; both files now diff identically to `upstream/main` on that point.

```release-note
Fix a race between the VirtualMachine and volume controllers that could
permanently drop vSphere-derived volume status fields (e.g. ControllerType)
after a VM's volumes were fully detached and reattached.
```

— Faisal + Claude


api/v1alpha6/virtualmachine_types.go
```diff
@@ -1596,6 +1596,14 @@ func (vm *VirtualMachine) SetConditions(conditions []metav1.Condition) {
 	vm.Status.Conditions = conditions
 }
 
+// OptimisticallyLockStatus implements pkg/patch's StatusOptimisticLocker.
+// Status.Volumes is independently read-modify-written by the virtualmachine
+// controller and by the volumebatch/volume controllers, so a status patch
+// must fail on a concurrent write rather than silently overwrite it.
+func (vm *VirtualMachine) OptimisticallyLockStatus() bool {
```
faisalabujabal, line 1603: should this be true for older versions?

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
```diff
@@ -437,7 +437,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 			ctx,
 			vm.Namespace,
 			vm.Spec.StorageClass)
-		if err := patchHelper.Patch(ctx, vm); err != nil {
+
+		err := patchHelper.Patch(ctx, vm)
```
faisalabujabal: why not keep it inline? you do this every now and then because you incorrectly think the no inline error in the lint rules applies to this module when it only applies to the e2e test module. update this in the local claude file.

api/v1alpha6/virtualmachine_types.go
faisalabujabal: No — `v1alpha1`..`v1alpha5` are conversion-only representations of the hub/storage version. Confirmed `api/v1alpha5/virtualmachine_conversion.go` has `ConvertTo`/`ConvertFrom` to the hub, same pattern for the other older versions. `pkg/patch.Helper` is only ever constructed with `*v1alpha6.VirtualMachine` — I grepped every `patch.NewHelper(...)` call site in the repo (23 total) and the three that touch `VirtualMachine` (`virtualmachine_controller.go`, `volumebatch_controller.go`, `volume_controller.go`) all import the `v1alpha6` alias. An older-version `VirtualMachine` Go value never reaches `patch.Helper.Patch` directly — any write to an older version round-trips through the conversion webhook to `v1alpha6` before the API server persists it. So implementing `OptimisticallyLockStatus` on older versions would be dead code: nothing would ever call it.

— Faisal + Claude

controllers/virtualmachine/virtualmachine/virtualmachine_controller.go
faisalabujabal: You're right, reverted — pushed in d55d0381. I checked why I made this change in the first place: I over-generalized from an actual `noinlineerr` hit elsewhere and assumed it applied here too without testing this specific line. It doesn't — `noinlineerr` is enabled repo-wide (`default: all` in `.golangci.yml`, not in the `disable` list) and is only turned *off* for `test/e2e`, but empirically it doesn't flag this particular block either way (confirmed by reverting to inline and running `golangci-lint run --enable-only noinlineerr` — 0 hits on this line, though it does flag three other genuinely-inline error checks elsewhere in the same file). `--new-from-rev=upstream/main` is 0 issues with the inline form restored. Reverted the same unnecessary change in `volumebatch_controller.go` too — both files now diff identically to `upstream/main` on this point. Will add a note to the local Claude file so I stop assuming this without checking.

— Faisal + Claude

PR #1751, opened 2026-07-21
 ✨ CRD change of supporting VM snapshot disk as the volume source

cc005511:
**What does this PR do, and why is it needed?**
This PR updates the Custom Resource Definitions (CRDs) to support using a virtual machine snapshot disk as a volume source. 
1. Adds a new VirtualMachineSnapshot field of type *VirtualMachineSnapshotDiskSpec to the VirtualMachineVolumeSource struct in the v1alpha6 API.
2. Introduces the VirtualMachineSnapshotDiskSpec struct, which requires two fields:
- Name: The name of the VirtualMachineSnapshot object (must be in the same namespace).
- DiskID: The specific ID of the disk to be mounted (as it appears in the snapshot's status.disks[].id).

To facilitate the backup of VM Service VMs, a dedicated "data-mover" VirtualMachine is used. This data-mover VM needs the ability to directly mount disks from snapshots taken of target VMs. By adding VirtualMachineSnapshot as a valid volume source, users (or the backup controller) can now explicitly instruct a VM to attach a specific disk from an existing snapshot, enabling the data-mover VM to read the snapshot data and back it up to external storage.

ok  	github.com/vmware-tanzu/vm-operator/api/test/v1alpha2	7.234s
ok  	github.com/vmware-tanzu/vm-operator/api/test/v1alpha3	6.565s
ok  	github.com/vmware-tanzu/vm-operator/api/test/v1alpha4	6.590s
ok  	github.com/vmware-tanzu/vm-operator/api/test/v1alpha5	9.058s

**Please add a release note if necessary**:
Add support for using a disk from a VirtualMachineSnapshot as a volume source for a VirtualMachine. This allows a VM to mount specific disks captured in a snapshot by referencing the snapshot name and the disk ID.

PR #1857, opened 2026-08-27
 ✨ Mount vm snapshot disk

cc005511:
VirtualMachineSnapshotDiskSpec is used to define a specific disk from a VirtualMachineSnapshot. VirtualMachineSnapshotDiskSpec field added in VirtualMachineVolumeSource is to support mounting a VirtualMachine snapshot disk as a volume of the VM service VM. 

- Only one member (virtualMachineSnapshot or persistentVolumeClaim) of VirtualMachineVolumeSource could be specified at a time. The webhook validation will reject a VirtualMachineVolume whose source sets both persistentVolumeClaim and virtualMachineSnapshot.
- The name or the diskID of VirtualMachineSnapshotDiskSpec must not be empty.
- If virtualMachineSnapshot is set in VirtualMachineVolumeSource but the VirtualMachineSnapshot is not-found or not-ready, the specific error will be set in status.volumes[].error.
- Once the volume is attached (VirtualMachine.status.volumes[].attached = true), changes on the volume source (the name or the diskID) is not allowed.

PASS
go test -v ./pkg/providers/vsphere/session/...
go test -v ./webhooks/virtualmachine/validation/...
go test -v ./controllers/virtualmachine/volume/...

Expose the disk list in VirtualMachineSnapshot: https://github.com/vmware-tanzu/vm-operator/pull/1744
CRD change of supporting VM snapshot disk as the volume source: https://github.com/vmware-tanzu/vm-operator/pull/1751/

Support mounting a VM snapshot disk as a VM volume. When using the VM Snapshot disk as the volume source, diskMode must be set to "IndependentNonPersistent" and removable must be set to true.


pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -877,6 +878,12 @@ func (s *Session) reconcileVolumes(vmCtx pkgctx.VirtualMachineContext) error {
 			} else {
 				notFoundNames = append(notFoundNames, volSpec.Name)
 			}
+		} else if volSpec.VirtualMachineSnapshot != nil {
+			// For snapshot volumes, we don't block power on for them being unattached here, as they are attached
+			// during the reconcile loop itself (in reconcileSnapshotDisks), unlike PVCs which
+			// are attached by an external controller.
+			// We also don't require them to be in status yet, as they might be added in the same reconcile loop.
+			continue
```
stevenhuang-vcf, line 886: This `else if` is meaningless here as your description. So why not just add comments without the check of the spec type of VM snapshot disk?

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1691,3 +1709,385 @@ func UpdateVMGuestIDReconfiguredCondition(
...
+	case *vimtypes.VirtualDiskFlatVer2BackingInfo:
+		// In vcsim, if we set Parent, it might complain if the parent is not found or already attached.
+		// Actually, let's just set DiskMode and Uuid.
+		return &vimtypes.VirtualDiskFlatVer2BackingInfo{VirtualDeviceFileBackingInfo: vimtypes.VirtualDeviceFileBackingInfo{FileName: backing.FileName}, DiskMode: string(vimtypes.VirtualDiskModeIndependent_nonpersistent), Parent: backing, Uuid: backing.Uuid}
+	case *vimtypes.VirtualDiskSeSparseBackingInfo:
+		return &vimtypes.VirtualDiskSeSparseBackingInfo{VirtualDeviceFileBackingInfo: vimtypes.VirtualDeviceFileBackingInfo{FileName: backing.FileName}, DiskMode: string(vimtypes.VirtualDiskModeIndependent_nonpersistent), Parent: backing, Uuid: backing.Uuid}
+	case *vimtypes.VirtualDiskSparseVer2BackingInfo:
+		return &vimtypes.VirtualDiskSparseVer2BackingInfo{VirtualDeviceFileBackingInfo: vimtypes.VirtualDeviceFileBackingInfo{FileName: backing.FileName}, DiskMode: string(vimtypes.VirtualDiskModeIndependent_nonpersistent), Parent: backing, Uuid: backing.Uuid}
+	default:
+		return targetDisk.Backing
```
stevenhuang-vcf, line 1752: should return the target backing info for a unsupported type?
Comparing the logic in `isSnapshotDisk` only `VirtualDiskFlatVer2BackingInfo`, `VirtualDiskSeSparseBackingInfo` and `VirtualDiskSparseVer2BackingInfo` are supported.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1691,3 +1709,385 @@ func UpdateVMGuestIDReconfiguredCondition(
...
+	if targetDisk == nil && len(moSnap.Config.Hardware.Device) == 0 && moVM.Config != nil {
+		for _, dev := range moVM.Config.Hardware.Device {
+			if disk, ok := dev.(*vimtypes.VirtualDisk); ok {
+				if strings.EqualFold(getVirtualDiskUUID(disk), diskID) {
+					targetDisk = disk
+					break
+				}
+			}
+		}
+	}
```
stevenhuang-vcf, line 1908: Why need to look for a target snapshot disk in a VM's device list?
I think it should be in a VM snapshot's device list. Is there any specific case?

stevenhuang-vcf (changes requested):
You need to add more comments in the code.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1571,6 +1578,17 @@ func doReconfigure(
 		}
 	}
 
+	if err := reconcileSnapshotDisks(
```
stevenhuang-vcf, line 1581: Need to add FSS.

pkg/providers/vsphere/session/session_vm_update.go
```diff
@@ -1691,3 +1709,385 @@ func UpdateVMGuestIDReconfiguredCondition(
...
+		Name:     volName,
+		Type:     vmopv1.VolumeTypeClassic,
+		Attached: true,
+		DiskUUID: diskUUID,
+		Error:    "",
+	})
+	pkglog.FromContextOrDefault(ctx).Info("Appended new volume to status", "volName", volName, "newVolumesLen", len(vm.Status.Volumes))
+}
+
+func isSnapshotDiskAttached(moVM mo.VirtualMachine, configSpec *vimtypes.VirtualMachineConfigSpec, diskID string) bool {
```
stevenhuang-vcf, line 1794: why not as `setSnapshotVolumeAttached` does, the check runs on the status list of volumes?

stevenhuang-vcf (changes requested):

webhooks/virtualmachine/validation/virtualmachine_validator.go
```diff
@@ -3672,3 +3714,74 @@ func isFirstClassNICAdvancedProperty(key string) bool {
 func isNetworkDeviceProperty(key string) bool {
 	return vmopv1util.IsEthernetDeviceKey(key)
 }
+
+func (v validator) validateSnapshotVolume(
```
zhanhank, line 3716: [CRITICAL] Synchronous SubjectAccessReview inside Validation Webhook: The validation webhook (virtualmachine_validator.go) makes three synchronous out-of-band calls to the K8s API server for SubjectAccessReview (checking get, create, delete verbs). Performing external network calls to the API server from within a validating webhook blocks the cluster admission pipeline. Under load, this can cause webhook timeouts, failing VM admission entirely. Furthermore, requiring create and delete permissions on a snapshot to merely mount it seems over-scoped—typically, read or use access should suffice.

pkg/providers/vsphere/vmprovider_vmsnapshot.go
```diff
@@ -849,6 +849,49 @@ func ReconcileSnapshotWaitForCRVCondition(
 	return nil
 }
 
+// getDisksFromSnapshot retrieves the hardware configuration of the snapshot and extracts the disks.
+func getDisksFromSnapshot(vmCtx pkgctx.VirtualMachineContext, vcVM *object.VirtualMachine, snapRef vimtypes.ManagedObjectReference) ([]vmopv1.VirtualMachineSnapshotDiskStatus, error) {
+	logger := pkglog.FromContextOrDefault(vmCtx)
+
+	var moSnap mo.VirtualMachineSnapshot
+	ctx := pkgctx.WithVCOpID(vmCtx, vmCtx.VM, "getDisksFromSnapshot")
+	if err := vcVM.Properties(ctx, snapRef, []string{"config.hardware.device"}, &moSnap); err != nil {
```
zhanhank, line 858: [MAJOR] N+1 vSphere API queries in Reconciler Loop: In session_vm_update.go -> reconcileSnapshotDisks, reconcileSnapshotVolume is called in a loop for each snapshot volume. Inside this function, vcVM.Properties is executed to fetch the snapshot hardware configuration. If a VM mounts multiple volumes from the same snapshot, this triggers redundant vSphere API calls, severely blocking the reconciler worker thread.

webhooks/virtualmachine/validation/virtualmachine_validator.go
zhanhank: It's safe to remove the privilege check since there is no cross-namespace mounting.

cc005511: Thank you. I will remove SubjectAccessReview inside Validation Webhook.

pkg/providers/vsphere/vmprovider_vmsnapshot.go
cc005511: Good point. I will add the cache for it. Thank you.

pkg/providers/vsphere/session/session_vm_update.go
cc005511: MoVM hardware devices and pending configSpec.DeviceChange are checked here to prevent false negatives (re-adding an already-attached disk if status was not yet updated) and false positives (skipping attach if status says attached but disk was detached).

PR #1859, opened 2026-08-28
✨  Reconcile AutomaticVMEvictionPolicy and BestEffortRestartPolicy

hpannem:
**What does this PR do, and why is it needed?**

Wires `AutomaticVMEvictionPolicy` and `BestEffortRestartPolicy` (`vsphere.policy.vmware.com/v1alpha1`) into the `policyevaluation` controller, evaluated the same way `ComputePolicy` already is. Reconciler half of `.sdd/specs/007-vm-eviction-policy/`; CRD types shipped in a prior PR.

- `policyevaluation_controller.go`: one explicit `reconcileMandatory<Kind>Policies`/`add<Kind>PolicyRef`/`Watches(...)` per new kind, mirroring `ComputePolicy` rather than a generic table. Match-evaluation and result-recording converge on a `matchablePolicy` interface (`ctrlclient.Object` plus `GetPolicyEnforcementMode()`/`GetPolicyMatch()`/`GetPolicyTagNames()`), implemented directly by `ComputePolicy`, `AutomaticVMEvictionPolicy`, and `BestEffortRestartPolicy` — no field-copying struct. A generic `toMatchablePolicies[T, PT]` helper converts each kind's `List().Items` into `[]matchablePolicy` by address. `kind` is passed as an explicit parameter to the shared `addPolicyResult`/`resolvePolicyTags`/`addExplicitPolicyMatch`/`processMandatoryPolicies` helpers, since the interface itself carries no kind identity. `ComputePolicy` is unchanged in behavior (its existing tests pass unmodified) but now also implements `matchablePolicy` so all three kinds share the same helpers. Both new kinds gated on `Features.VMEviction`.
- RBAC markers for `automaticvmevictionpolicies`/`besteffortrestartpolicies` (+`/status`); regenerated `config/rbac/role.yaml`.
- `VirtualMachineSpec.Policies` doc comment: now lists `ComputePolicy`, `AutomaticVMEvictionPolicy`, and `BestEffortRestartPolicy` as valid explicit-reference kinds; regenerated affected CRD manifests. `docs/concepts/workloads/vm.md`'s "Policy Types"/"Configuration" sections updated to match.
- `BestEffortRestartPolicy.GetConditions`/`SetConditions` (and the `Status` type's) now have godoc and a pointer receiver on `GetConditions`, matching `AutomaticVMEvictionPolicy`.
[...]

PR #1860, opened 2026-08-28
✨ BestEffortRestartPolicy e2e

hpannem:

Adds the remaining E2E coverage for the `AutomaticVMEvictionPolicy`/`BestEffortRestartPolicy` compute-policy CRDs (product code already merged in #1859), closing out `tasks.md` T021/T022 and adding T040:

- Explicitly referencing a matching Optional `BestEffortRestartPolicy` via `spec.policies` tags the VM.
- Explicitly referencing a non-matching Optional `BestEffortRestartPolicy` surfaces a not-ready `PolicyEvaluation` condition (`Ready=False`, message containing "does not match").
- A VM matching a Mandatory `AutomaticVMEvictionPolicy` and explicitly referencing an Optional `BestEffortRestartPolicy` surfaces both in `status.policies`.
- Both policies Mandatory and matching the same VM simultaneously (no explicit references).
- Both policies Optional and explicitly referenced together via `spec.policies`.

All new `It`s are labeled `experimental` per `e2e-testing.md`, pending validation on real hardware.

Also removes a standalone `ComputePolicy` e2e spec (`computepolicy.go`) and its `Context("COMPUTE-POLICY", ...)` wiring that were added earlier under the mistaken assumption `ComputePolicy` had no e2e coverage — it already has thorough mandatory/optional combination coverage through the real WCP admin API path in `test/e2e/vmservice/vmservice/virtualmachine/virtualmachinelcm.go` (`Context("IaaS Policies", ...)`, `Context("PLACEMENT-POLICY-COMBINATIONS", ...)`), so the parallel direct-CRD-creation spec was redundant. This PR's scope is intentionally limited to `AutomaticVMEvictionPolicy`/`BestEffortRestartPolicy`.

`.sdd/specs/007-vm-eviction-policy/plan.md` and `tasks.md` updated to match.

Fixes #

- Not yet run against a real WCP/vSphere testbed — all new specs carry the `experimental` label and are excluded from CI until validated on hardware.
- `createAutomaticVMEvictionPolicy`/`createBestEffortRestartPolicy` gained an `enforcementMode` parameter (previously hardcoded per kind) so both Mandatory and Optional variants are exercisable from the same helpers; all existing call sites updated accordingly.

```release-note
NONE

### CODE REVIEW

File: external/vsphere-policy/api/v1alpha1/besteffortrestartpolicy_types.go [hunk 1]
```diff
@@ -0,0 +1,129 @@
...
+	Status BestEffortRestartPolicyStatus `json:"status,omitempty"`
+}
+
+func (p BestEffortRestartPolicy) GetConditions() []metav1.Condition {
+	return p.Status.Conditions
+}
+
+func (p *BestEffortRestartPolicy) SetConditions(conditions []metav1.Condition) {
+	p.Status.Conditions = conditions
+}
```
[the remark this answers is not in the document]
hpannem: stale.

external/vsphere-policy/api/v1alpha1/besteffortrestartpolicy_types.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (p BestEffortRestartPolicyStatus) GetConditions() []metav1.Condition {
+	return p.Conditions
+}
+
+func (p *BestEffortRestartPolicyStatus) SetConditions(conditions []metav1.Condition) {
+	p.Conditions = conditions
+}
+
```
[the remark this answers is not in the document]
hpannem: stale.

pkg/providers/vsphere/vmprovider_vm.go
```diff
@@ -1714,9 +1714,13 @@ func (vs *vSphereVMProvider) vmCreateDoPlacement(
 			// processPlacementResult may already have set a more specific
 			// reason (e.g. ZoneMismatch); only fall back to the generic
 			// NotReady reason when the condition is not already False.
+			// Update message (which may contain updated details) only
+			// when its the same "NotReady" reason.
 			if !pkgcond.IsFalse(
 				vmCtx.VM,
-				vmopv1.VirtualMachineConditionPlacementReady) {
+				vmopv1.VirtualMachineConditionPlacementReady) ||
+				pkgcond.GetReason(vmCtx.VM,
```
aruneshpa, line 1722: I may be missing something here. :(  But what you are adding seems redundant? The only way your code executes is when the VirtualMachineConditionPlacementReady is False AND the reason is NotReady -- upon which we set the reason to `NotReady` in the if block?

pkg/vmconfig/policy/policy_reconciler_util.go
```diff
@@ -205,12 +205,18 @@ func getPolicyEvaluationResults(
 			"spec", obj.Spec)
 
 		// The object is still being processed.
+		msg := pkgcond.GetMessage(obj, vspherepolv1.ReadyConditionType)
+		if msg != "" {
```
aruneshpa, line 209: This is confusing.  Instead of creating this weird formatted string by sticking this next to the error, let's either have another key (e.g., "msg", <value of this>).  Or, change the error to include the message somehow.

aruneshpa (approved):
LGTM.  Two minor comments.

pkg/vmconfig/policy/policy_reconciler_util.go
hpannem: Should I append "Reason: <msg>"

pkg/providers/vsphere/vmprovider_vm.go
hpannem: This is to refresh the message.
- reconcile 1.
1. The condition is set to VirtualMachineConditionPlacementReady to false as policy evaluation has not yet reconciled with generic message "saying policies not ready".

- reconcile 2:
1. Policy evaluation updated a actual message (condition not matched etc.) 
2. The condition is already false. We used to skip the  updated message earlier. Now, If the condition is not set to false OR the condition's reason is "NotReady", we will update with refreshed message.

pkg/vmconfig/policy/policy_reconciler_util.go
hpannem: Makes sense. Done

PR #1862, opened 2026-08-28
e2e: Fix flaky VM publish request specs

anwithaun-1:

De-flakes the two VM publish E2E specs that failed together in a recent smoke run (VM-PUBLISH-REQUEST / CLS and VM-GROUP-PUBLISH-REQUEST). And fix another connection issue wrt the VM-PUBLISHREQUEST

Both failures are timeouts on a loaded testbed that correlate with background disk promotion running on the VM under test:
Failure 1
- VM-GROUP-PUBLISH-REQUEST — failed at VerifyVirtualMachineGroupPublishRequestCompleted (vmoperator.go:731): the subset publish never reported Complete within 360 s. The first group publish, issued before promotion begins, completed in ~30 s; the subset publish was issued about a minute after the member VMs were created — inside the promotion window implied by the source VM's timings — and hung.

Failure 2
- VM-PUBLISH-REQUEST / CLS Content Library — failed at WaitForVirtualMachineIP (vmoperator.go:204): the VM deployed from the freshly published image reached PoweredOn but reported no status.network.primaryIP4 for the full 300 s wait-virtual-machine-vmip window. In the same run, the source VM (promoteDisksMode: Online, the API default) flipped VirtualMachineGuestNetworkConfigSynced to True at the same second as VirtualMachineDiskPromotionSynced — i.e. the guest IP only surfaced once online promotion finished. A VM deployed from a just-published image promotes off cold, uncached parent disks, so that window is longer.

Failure 3
- The VM-GROUP-PUBLISH-REQUEST test failed because a transient supervisor apiserver connectivity stall caused the kubectl create for the subset publish request to hang and die with http2: client connection lost, an error signature not yet covered by the e2e transient-retry classifier.

Changes:
- test/e2e/vmservice/vmservice/virtualmachine/vm_publishrequest.go — build the VM deployed from the published image as a typed vmopv1.VirtualMachine with promoteDisksMode: Disabled instead of the rendered v1alpha2 fixture. The fixture cannot express the field: promoteDisksMode does not exist before v1alpha4, so the API server prunes it from a v1alpha2 request. All other fields (class, image, storage class, reserved.resourcePolicyName, powerState: PoweredOn) match what the fixture rendered.
- test/e2e/vmservice/vmservice/virtualmachine/vm_group_publishrequest.go — create the group member VMs directly with promoteDisksMode: Disabled via a new package-level newGroupMemberVM helper, replacing vmservice.DeployVMWithCloudInit. No coverage is lost: that helper sets SecretName, which the v1a5singlevm.yaml.in template never references (it only reads .Bootstrap.*), so the cloud-config secret it created was never attached to the VMs — confirmed by the rendered YAML in the failing run, which has no bootstrap: stanza. The typed VM is field-equivalent plus promoteDisksMode.
-  Add "http2: client connection lost" to the e2e transient-error retry list so a lost HTTP/2 connection during kubectl create (as seen in the VM-GROUP-PUBLISH-REQUEST flake) gets retried instead of failing the test outright.

Fixes #
vmop-4106

On failure # 2, promotion holding back guest property updates is not provable from these logs
- promotion holding back guest property updates is not provable from these logs. What is observable is on the source VM, the only VM whose conditions were captured: across its entire promotion window (VirtualMachineDiskPromotionStarted True at 11:19:41 → VirtualMachineDiskPromotionSynced True at 11:21:05) there is no guest network update at all — VirtualMachineGuestNetworkConfigSynced stays unset and only flips True, with IPv4="172.26.0.114", at 11:21:05 as promotion completes. Whether promotion caused that quiet period or merely coincided with it cannot be told apart from this log, and the VM that actually failed the IP wait had no diagnostics captured, so its promotion state is unknown. Disabling promotion for this spec is therefore a precaution rather than a proven fix: the spec asserts publish and deploy behavior, disk provenance is irrelevant to that, and removing promotion costs no coverage here

```release-note

```

### DIFF
--- a/test/e2e/framework/retry_client.go
+++ b/test/e2e/framework/retry_client.go
@@ -49,6 +49,7 @@ var RetryableTransientErrSubstrings = []string{
 	"unexpected eof",
 	"broken pipe",
 	"http2: server sent goaway",
+	"http2: client connection lost",
 	"context deadline exceeded",
 	"no endpoints available for service",
 	"unable to connect to the server",
--- a/test/e2e/vmservice/vmservice/virtualmachine/vm_group_publishrequest.go
+++ b/test/e2e/vmservice/vmservice/virtualmachine/vm_group_publishrequest.go
@@ -14,6 +14,7 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	capiutil "sigs.k8s.io/cluster-api/util"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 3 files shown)


aruneshpa (approved):
LGTM.

aruneshpa:
Merging this.  @anwithaun-1 mentioned that she was able to get a good run of affected pipelines.

PR #1865, opened 2026-08-31
✨ Tag Controller CRD, Controller & Webhooks

aniket-deole:

Add Tag CRD, controller, and webhook for VM tagging

Add a new Tag CRD (vsphere.policy.vmware.com/v1alpha1) with a
controller and a validating webhook. A Tag stands for one vCenter
tag (a key/value pair) used for VM affinity. It is owned by one or
more VirtualMachines. The controller marks a Tag Ready, copies its
key/value onto labels, and deletes the Tag once all owners are gone.

Fixes # vmop-3882

**Are there any special notes for your reviewer**: NA

```release-note

```


webhooks/webhooks.go
```diff
@@ -66,6 +67,10 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr ctrlmgr.Manager) err
 		return fmt.Errorf("failed to initialize UnifiedStorageQuota webhooks: %w", err)
 	}
 
+	if err := vspherepolicy.AddToManager(ctx, mgr); err != nil {
```
aruneshpa: Add the capability check here so we don't call into the AddToManager at all if the cap is not enabled (as opposed to in vspherepo/webbooks.go).

webhooks/vspherepolicy/webhooks.go
```diff
@@ -0,0 +1,28 @@
...
+	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"github.com/vmware-tanzu/vm-operator/webhooks/vspherepolicy/tag"
+
+	pkgcfg "github.com/vmware-tanzu/vm-operator/pkg/config"
+	pkgctx "github.com/vmware-tanzu/vm-operator/pkg/context"
+)
+
+// AddToManager adds the vSphere Policy webhooks to the provided manager.
+func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr ctrlmgr.Manager) error {
```
aruneshpa: You can nuke this and just call the tag webhook directly if the cap is enabled from the webhooks/webhooks.go.

webhooks/vspherepolicy/tag/validation/tag_validator.go
```diff
@@ -0,0 +1,229 @@
...
+
+// allowedSystemAccountsForTag are the control-plane clients that must be
+// allowed to write a Tag even though they are not privileged accounts.
+// generic-garbage-collector prunes a dangling owner reference on UPDATE and
+// deletes a Tag with zero owners on DELETE; namespace-controller deletes a
+// Tag as part of namespace teardown on DELETE. Without this allow-list,
+// both operations are denied, which wedges owner-reference cleanup and
+// namespace termination respectively.
+var allowedSystemAccountsForTag = map[string]struct{}{
+	"system:serviceaccount:kube-system:generic-garbage-collector": {},
```
aruneshpa, line 45: Is this defined elsewhere in upstream k8s/k8s so we don't have to have strings here?  There's also a worry of these service accounts changing which may break this code.

webhooks/vspherepolicy/tag/validation/tag_validator.go
```diff
@@ -0,0 +1,229 @@
...
+
+package validation
+
+import (
+	"fmt"
+	"net/http"
+	"reflect"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	unversionedvalidation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
```
aruneshpa: Hmm.  Any specific reason for calling this "unversionedalias"?  That somehow indicates that other validations are versioned.
Maybe don't use an alas for that since the other import is already aliased?

webhooks/vspherepolicy/tag/validation/tag_validator.go
```diff
@@ -0,0 +1,229 @@
...
+// validateLabelValue requires spec.value to be a valid Kubernetes label
+// value. An empty value is permitted.
+func validateLabelValue(value string) field.ErrorList {
+	var fieldErrs field.ErrorList
+	for _, msg := range utilvalidation.IsValidLabelValue(value) {
+		fieldErrs = append(fieldErrs, field.Invalid(specValuePath, value, msg))
+	}
+
+	return fieldErrs
+}
```
aruneshpa, line 183: Why not use https://pkg.go.dev/k8s.io/apimachinery@v0.37.0/pkg/apis/meta/v1/validation#ValidateLabels?

That seems to combine both of these?

webhooks/vspherepolicy/tag/validation/tag_validator.go
```diff
@@ -0,0 +1,229 @@
...
+
+// validatePrivileged permits only a privileged account, or one of the
+// allow-listed system service accounts, to create, update or delete a Tag.
+// The allow-list is checked before IsPrivilegedAccount, and neither test
+// admits a DevOps user: the Tag API is internal bookkeeping and is not
+// exposed to namespace users.
+func (v validator) validatePrivileged(
+	ctx *pkgctx.WebhookRequestContext,
+	op admissionv1.Operation) *admission.Response {
```
aruneshpa: Our convention is for helper methods to return `field.ErrorsList`.  Is there a reason why we are returning the response here instead of return the errors and parsing and returning the response from the caller?

webhooks/vspherepolicy/tag/validation/tag_validator.go
```diff
@@ -0,0 +1,229 @@
...
+		return webhook.Errored(http.StatusBadRequest, err)
+	}
+
+	fieldErrs := make(field.ErrorList, 0)
+
+	fieldErrs = append(fieldErrs, validateLabelKey(tag.Spec.Key)...)
+	fieldErrs = append(fieldErrs, validateLabelValue(tag.Spec.Value)...)
+
+	if tag.Spec.Key != oldTag.Spec.Key {
+		fieldErrs = append(fieldErrs, field.Forbidden(specKeyPath, "field is immutable"))
```
aruneshpa:
```suggestion
		fieldErrs = append(fieldErrs,
				validation.ValidateImmutableField(tag.Spec.Key, oldTag.Spec.Key, field.NewPath("spec", "key"))...)
```

webhooks/vspherepolicy/tag/validation/tag_validator.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	if tag.Spec.Value != oldTag.Spec.Value {
+		fieldErrs = append(fieldErrs, field.Forbidden(specValuePath, "field is immutable"))
```
aruneshpa: Same here.  Use the upstream helper for immutable field comparison.

webhooks/vspherepolicy/tag/validation/tag_validator.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	return common.BuildValidationResponse(ctx, nil, fieldErrsToStrings(fieldErrs), nil)
+}
+
+// ValidateDelete admits only privileged requesters.
+func (v validator) ValidateDelete(ctx *pkgctx.WebhookRequestContext) admission.Response {
+	if resp := v.validatePrivileged(ctx, admissionv1.Delete); resp != nil {
```
aruneshpa: Trying to understand what this means.  I have never seen delete be blocked using webhooks.  Can you verify and confirm that doing this actually prevents the API server from slapping the DeletionTimestamp on the object?  IIRC, once that is done, that can't be undone.

pkg/util/hash.go
```diff
@@ -19,6 +22,12 @@ func SHA1Sum17(s string) string {
 	return hex.EncodeToString(h.Sum(nil))[:17]
 }
 
+// XXHash64Hex returns the full 64-bit xxhash.Sum64 digest of the provided
+// string, formatted as 16 hex characters.
+func XXHash64Hex(s string) string {
```
aruneshpa, line 27: I am guessing you are using this as opposed to the SHA1 because SHA 1 tripping up all kinds of security warnings?
I would add a comment explaining that (wherever you are calling this would be more appropriate).

test/e2e/vmservice/consts/consts.go
```diff
@@ -44,4 +44,8 @@ const (
 	VirtualMachineConfigPolicyCapabilityName = "supports_vm_service_vm_config_policy"
 	TelcoVMServiceAPICapabilityName          = "supports_telco_vm_service_api"
 	WorkloadIPv6CapabilityName               = "supports_workload_ipv6"
+
+	// VMHardAffinityDuringExecutionCapabilityName gates the Tag CRD,
+	// controller, and admission webhook (Features.TaggingAPI).
+	VMHardAffinityDuringExecutionCapabilityName = "supports_vm_service_vm_hard_affinity_during_execution"
```
aruneshpa, line 50: So, just for my understanding:
Even though the cap on SV is called "hard_aaf", you ca calling the internal flag as TaggingAPI because it is more aligned to this feature?  And is my understanding correct that you are reusing an already enabled cap (supports_vm_service_vm_hard_affinity_during_execution)?

If yes, we should introduce a new cap at SV level.

pkg/providers/vsphere/virtualmachine/vmtags.go
```diff
@@ -0,0 +1,23 @@
...
+
+// TagResourceName returns the derived name of the Tag resource that
+// represents the given label key/value pair in the given namespace:
+// "tag-" followed by the 16 hex character XXHash64 digest of
+// "<namespace>:<key>:<value>".
+func TagResourceName(namespace, key, value string) string {
+	return "tag-" + pkgutil.XXHash64Hex(namespace+":"+VCenterTagName(key, value))
+}
+
+// VCenterTagName returns the vCenter tag name for the given label
```
aruneshpa: There is nothing specific about vCenter here, right?  Why call that out?  Just mention that this is serialized tag name and value with ":" as a separator.

controllers/vspherepolicy/tag/tag_controller.go
```diff
@@ -0,0 +1,167 @@
...
+		if err != nil {
+			if reterr == nil {
+				reterr = fmt.Errorf("failed to create patch helper: %w", err)
+			} else {
+				reterr = fmt.Errorf("%w,%w", err, reterr)
+			}
+		}
+	}()
+
+	return r.ReconcileNormal(ctx, &obj)
```
aruneshpa, line 122: Hmm.  Shouldn't we handle Tags that are marked for deletion (e.g., a non-zero DeletionTimestamp)?

external/vsphere-policy/api/v1alpha1/tag_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// Type describes the origin/category of this Tag.
+	Type TagType `json:"type,omitempty"`
+}
+
+// TagStatus defines the observed state of Tag.
+type TagStatus struct {
+	// +optional
+
+	// ID is reserved for the resolved vCenter Tag UUID. This field is not
+	// populated by this feature.
```
aruneshpa: Consider removing "by this feature".  API definitions are not supposed to be tied to point-in-time features.

external/vsphere-policy/api/v1alpha1/tag_types.go
```diff
@@ -0,0 +1,131 @@
...
+
+// SetConditions sets the Tag's conditions, allowing Tag to implement the
+// conditions.Setter interface.
+func (t *Tag) SetConditions(conditions []metav1.Condition) {
+	t.Status.Conditions = conditions
+}
+
+// GetConditions returns the TagStatus's conditions, allowing TagStatus to
+// implement the conditions.Getter interface.
+func (t *TagStatus) GetConditions() []metav1.Condition {
```
aruneshpa: Do we need GetConditions on both, Tag and Tag.Status?  Calling it on parent Tag object should be sufficient?

controllers/controllers.go
```diff
@@ -87,7 +87,7 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		}
 	}
 
-	if pkgcfg.FromContext(ctx).Features.VSpherePolicies {
+	if features := pkgcfg.FromContext(ctx).Features; features.VSpherePolicies || features.TaggingAPI {
```
aruneshpa: Can you please split this in two?  Have explicit controller.AddToManagers depending on which cap is enabled so it is more readable.

pkg/util/hash.go
aniket-deole: Correct. Will add comments.

test/e2e/vmservice/consts/consts.go
aniket-deole: The capability is not enabled yet. HardAaf capability drives two features: Tagging API & Actual HardAffinity of VMs.

As cap is not enabled yet, do you still see the need for a separate Capability ? The new capability will be based on the same VC FSS anyway.

webhooks/vspherepolicy/tag/validation/tag_validator.go
aniket-deole: There is separate validation as the spec contains `spec.key` & `spec.value`.

The fieldPath argument accepts a single path.

webhooks/vspherepolicy/tag/validation/tag_validator.go
aniket-deole: We have a bunch of similar strings defined in `webhooks/persistentvolumeclaim/validation/persistentvolumeclaim_validator.go`

webhooks/vspherepolicy/tag/validation/tag_validator.go
aniket-deole: This is a pattern used from `webhooks/virtualmachineservice/validation/virtualmachineservice_validator.go`.
I've changed it to be `metavalidation`.

webhooks/vspherepolicy/tag/validation/tag_validator.go
aniket-deole: I cannot test it in this PR, as the object is garbage collected because of no owners. I'll test it out in the next MR as the Tags will have VM owners.

controllers/vspherepolicy/tag/tag_controller.go
aniket-deole: Tags are garbage collected. Do you think we still need explicit deletion?

PR #1866, opened 2026-08-31
 🐛 Fix RegisterVM restore-disk-only cleanup deadlock

aakashchan:
The "Restore disk only" spec pauses the VM (sets the vmoperator.vmware.com/paused annotation) before invoking RegisterVM, but its DeferCleanup (deleteRegisteredVMAndPVCs) never clears that annotation before deleting the VM. virtualmachine_controller's ReconcileDelete unconditionally skips deletion while the VM is paused, so the finalizer is never removed and the delete blocks until the DeferCleanup's own 3-minute NodeTimeout fires. This happened even on runs where the It body itself passed, turning a correct restore into a spurious [TIMEDOUT] failure.

Also fixed two related issues discovered while root-causing the flakiness of this same spec against live UTS runs:

- vmObj.RemoveDevice's reconfigure task can report done before vSphere actually releases the file lock on the original vmdk, so the immediately-following fileManager.Delete could fail with a FileFault ("...vmdk attached to vm-<id>"). Retry it briefly to absorb that lock-release lag instead of failing the whole spec.
- Boot-disk promotion for this spec is a live Storage vMotion (VM home on nfsDatastore, child disk vSAN-backed) whose duration depends on NFC/migration bandwidth contention with whatever else is running on the same hosts; we measured it from ~27s up to 4m37s+ across different runs, occasionally exceeding the shared "default" wait-virtual-machine-condition-update budget. Threaded a specName parameter through WaitForBootDiskPVC and WaitForBackupToComplete so only this spec's call site can opt into a longer, dedicated interval without loosening the timeout used by every other caller.

Changes:
- deleteRegisteredVMAndPVCs: best-effort clear the pause annotation before deleting the VM, mirroring the pattern already used by DeleteVMResource elsewhere in this file.
- Wrap the post-RemoveDevice fileManager.Delete call in a bounded Eventually (30s/3s) to retry past the file-lock-release race.
- Add a specName parameter to WaitForBootDiskPVC (vmoperator.go) and WaitForBackupToComplete (util.go); all existing callers pass "default" to preserve current behavior.
- Add a register-vm-restore-disk-only/wait-virtual-machine-condition-update interval ("12m", "5s") to wcp.yaml and kind.yaml, used only by the "Restore disk only" spec via the new restoreDiskOnlyIntervalSpec constant.

Fixes: vmop-4097

Fixes #

```release-note

```


test/e2e/vmservice/vmservice/viadmin/registervm.go
```diff
@@ -311,7 +331,7 @@ func VIAdminRegisterVMSpec(ctx context.Context, inputGetter func() VIAdminRegist
 
 			// Wait for backup to complete before reading the backup data.
 			// Use the returned VM to avoid a redundant Get call.
-			existingVM := vmservice.WaitForBackupToComplete(ctx, vmName, input.WCPNamespaceName, clusterProxy, config)
+			existingVM := vmservice.WaitForBackupToComplete(ctx, vmName, input.WCPNamespaceName, clusterProxy, config, "default")
```
aruneshpa: Do you want to fold these in a type and pass that down?  That will allow you to have optional (pointer) fields where you can infer things like unset = "default" timeout?

test/e2e/vmservice/vmservice/util.go
```diff
@@ -816,17 +816,22 @@ func decodeGzipBase64(encoded string) (string, error) {
 // WaitForBackupToComplete waits for the VM backup process to complete by verifying
 // that all PVCs in the VM spec have corresponding entries in the backup data stored
 // in the VM's ExtraConfig.
+//
+// specName is forwarded to vmoperator.WaitForBootDiskPVC to select a
+// spec-specific "wait-virtual-machine-condition-update" interval; pass
+// "default" unless the caller needs a longer boot-disk-promotion budget.
 func WaitForBackupToComplete(
```
aruneshpa, line 823: Same comment here.  Consider using a custom type.

aruneshpa (approved):
One comment, rest looks good.

test/e2e/vmservice/vmservice/viadmin/registervm.go
aakashchan: Thanks for the suggestion. Updated.

test/e2e/vmservice/vmservice/util.go
aakashchan: Done. Thanks

aakashchan:
Test pipelines are passing. Builds are successful. Merging.

PR #1872, opened 2026-09-01
✨ VM Provider VM tests should use fast deploy

bryanv:

Fast deploy is our default so the VM Provider tests should by default run with that feature enabled. Dedupe the repeated test context setup code so we have a single path to enable fast deploy.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/builder/fake.go
+++ b/test/builder/fake.go
@@ -5,7 +5,12 @@
 package builder
 
 import (
+	"context"
+
+	"github.com/google/uuid"
+
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/events"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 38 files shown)


test/builder/fake.go
```diff
@@ -56,6 +70,44 @@ func NewFakeClientWithInterceptors(
...
+// A UID the caller set explicitly is left alone, since a spec that pins one is
+// usually asserting on it.
+func assignUIDOnCreate(next createFunc) createFunc {
+	return func(
+		ctx context.Context,
+		c client.WithWatch,
+		obj client.Object,
+		opts ...client.CreateOption) error {
+
+		ensureUID(obj)
```
akutz, line 89: Should this verify the object type to be a VM?

akutz (approved):
LGTM with one comment.

test/builder/fake.go
bryanv: Offline discussion - this is just for the fake client so this is fine. While wrt to this PR it is just strictly needed on the VM since the UID is use as part of the file path for fast deploy, it should not be an issue to just do it for all types.

PR #1888, opened 2026-09-02
🐛 Atomic status patching and controller cleanups for VM publish operations

bryanv:

Refactors the status patching logic across VirtualMachinePublishRequest and group publish controllers to prevent race conditions and incomplete status updates caused by split patch operations. It also includes minor cleanup for controller loggers, condition validation, and requeue logic.
Key Changes
•	Atomic Status Updates: Added an option to the patch helper to issue a single status patch rather than splitting conditions and other status fields. Updated publish and group publish controllers to patch status in one atomic operation, ensuring fields like image names are set simultaneously with completion conditions.
•	Controller Cleanups:
•	Switched publish controllers to use contextual loggers instead of dedicated logger instances.
•	Corrected invalid condition reasons on invalid group publish specs to comply with field patterns.
•	Simplified condition checks in the publish request requeue logic.
•	Cleaned up inaccurate documentation regarding CompleteTime behavior.

Fixes # vmop-4152

UTS publish & group publish tests pass

```release-note
NONE
```

### DIFF
--- a/controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
+++ b/controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
@@ -37,7 +37,6 @@ import (
 const (
 	finalizerName         = "vmoperator.vmware.com/virtualmachinegrouppublishrequest"
 	undefinedSpecErrorMsg = "spec.%s is undefined"
-	invalidSpecErrorMsg   = "webhooks failed to mutate/validate spec. please delete and create again."
 	delTTLMsg             = "%s vm group publish request due to TTL expired"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/patch/patch_test.go
```diff
@@ -599,6 +599,48 @@ func intgTests() {
...
+					}
+
+					return obj.Status.NodeName == objAfter.Status.NodeName &&
+						conditions.IsTrue(objAfter, vmopv1.ReadyConditionType) &&
+						reflect.DeepEqual(obj.Spec, objAfter.Spec)
+				}, timeout).Should(BeTrue())
+			})
 		})
 
 		/*
```
brito-rafa, line 646: Should we remove this block of commented code since we are not using patch with WithStatusObservedGeneration option?

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
```diff
@@ -359,14 +359,13 @@ func (r *Reconciler) verifySpec(vmGroupPublishReq *vmopv1.VirtualMachineGroupPub
...
-		// set condition to failure
+		// Mutation and validation webhooks should prevent this.
+		err := fmt.Errorf("invalid spec: %w", errors.Join(errs...))
 		conditions.MarkError(vmGroupPublishReq,
 			vmopv1.VirtualMachineGroupPublishRequestConditionComplete,
-			invalidSpecErrorMsg,
-			errors.Join(errs...))
-
-		return pkgerr.NoRequeueError{Message: invalidSpecErrorMsg}
+			"Invalid",
```
brito-rafa, line 366: Should we create a constant for the "Invalid"?

brito-rafa (approved):
LGTM

akutz (approved):

pkg/patch/patch_test.go
bryanv: I think eventually should eventually use this so the generation in the conditions is set

controllers/virtualmachinegrouppublishrequest/virtualmachinegrouppublishrequest_controller.go
bryanv: This is not suppose to be possible, and in the old code if we would have gotten there it the reason would be invalid and we wouldn't be able to actually update the object.

PR #1893, opened 2026-09-02
✨  One pager for KubeVM: A pluggable, provider agnostic APIs for Virtual Machines on Kubernetes

aruneshpa:

KubeVM is a generic set of Kubernetes-native, hypervisor agnostic APIs to deploy and manage the lifecycle of virtual machines from a Kubernetes cluster. This project provides a consistent interface allowing infrastructure providers to expose core platform capabilities, including scheduling, performance, storage, migration and more.   The APIs also allow the platform team to define guardrails and policies around the compute, storage, networking, placement of virtual machines.

N / A

### DIFF
--- a/external/kubevm/docs/one-pager-kubevm-generic-api.md
+++ b/external/kubevm/docs/one-pager-kubevm-generic-api.md
@@ -2,6 +2,398 @@
 
 **Author:** Arunesh Pandey · **Status:** Proposal, open for discussion
 
-This proposal is under review. Please go to
-[#1893](https://github.com/vmware-tanzu/vm-operator/pull/1893) to read it
-and leave comments/feedback.
+This is a design proposal, not an accepted API.
+The types it describes are
+implemented as a strawman in this repository so the shape can be reviewed
+concretely; see the repository README for what is and is not built.
+
+## Business Problem
+
+Kubernetes has become the default control plane for modern infrastructure, yet the ecosystem still lacks a cross-platform, VM-centric API: a single declarative surface through which any hypervisor or cloud can expose both the full lifecycle of a virtual machine and the hardware capabilities that demanding workloads depend on.
+This gap is becoming urgent because a new class of workload is arriving faster than the tooling to run it.
+Agentic workloads — long-running processes that execute model-generated code and orchestrate tools — are increasingly deployed inside virtual machines, both for the strong isolation a VM provides around untrusted code and for direct access to the hardware accelerators, such as GPUs, SR-IOV network functions, and passthrough devices, that hypervisors already virtualize well.
+The community needs a credible way to run these workloads, and today it does not have one.
+
+The CNCF landscape today addresses adjacent needs but not this one.
+Kata Containers provides VM-strength isolation for individual workloads by wrapping a Pod in a lightweight micro-VM — well suited to isolating untrusted code at the granularity of a container.
+KubeVirt takes a different approach, converging the virtual machine into the container model by running a QEMU/KVM process inside a Pod, which is an excellent fit when Kubernetes is the sole infrastructure layer and rich, device-level VM modeling on Kubernetes nodes is the goal.
+Both are strong at their design point.
+What neither sets out to be is a portable, Kubernetes-native front door to a full-blown, hypervisor-native estate — an existing vSphere deployment or a public-cloud VM service — that exposes that platform's own lifecycle and hardware capabilities (GPUs, SR-IOV, passthrough) through one vendor-neutral API.
+That is the gap KubeVM fills, and it is complementary to both.
+
+This document proposes **KubeVM**, a generic and vendor-neutral `VirtualMachine` API served under the `kube-vm.io` group, together with a provider model that lets hypervisors and cloud VM services expose their machines — and, critically, their accelerators — through one portable, Kubernetes-native interface.
+KubeVM is intended to complement KubeVirt, not to replace it: it addresses the hypervisor-native design point that the VM-as-Pod model leaves unaddressed.
+The `VirtualMachine` resource is the starting point rather than the whole proposal.
+Standardizing the machine is what makes a set, a service, a rolling deployment — and eventually a workload model spanning both VMs and Pods — something the ecosystem writes once instead of once per platform.
+That trajectory is described in *Beyond a single machine*, and it is where most of the long-term value of a generic API lies.
+
+## Goals
+
+- As a **DevOps user**, I should be able to declare a virtual machine — its size, image, bootstrap configuration, networking, storage, and accelerators — through a single portable API, regardless of the platform that ultimately runs it.
+- As a **DevOps user**, I should be able to request GPU-backed VMs through the portable sizing profile, with the provider resolving that profile to its own accelerator representation.
+  Note that this is a weaker guarantee than "identically everywhere": where a platform models accelerators as a separate per-instance attachment rather than as a property of the profile, v1 cannot express it.
+  Richer accelerator types (SR-IOV, arbitrary passthrough) follow as a portable shape is established.
+- As a **platform engineer**, I should be able to target different providers — vSphere, EC2, or GCP — with the same portable `VirtualMachine` schema, so that moving a workload means pointing at a different provider object and supplying that platform's specifics, not re-authoring the machine's shape.
+- As a **provider author**, I should be able to ship a thin provider that maps the generic API onto my platform, implementing only the behavior that is genuinely unique to it.
+- As a member of the **CNCF community**, I should have a hypervisor-native, vendor-neutral VM API that is well suited to agentic and accelerated workloads and that fills the gap left open by the VM-as-Pod model.
+
+## Non Goals
+
+- KubeVM does not replace KubeVirt.
+  The VM-as-Pod model remains valid for clusters where Kubernetes is the only infrastructure layer, and KubeVM can even expose a KubeVirt provider; the two are complementary points in the design space.
+- KubeVM is not itself a hypervisor or a virtualization implementation.
+  It is an API and a set of controllers; the hypervisor or cloud is always the provider, and no nested virtualization is introduced.
+- KubeVM does not re-implement hypervisor capabilities such as live migration, high availability, or resource scheduling on top of raw Kubernetes primitives.
+  These are delegated to the provider that already implements them.
+- A dedicated, portable accelerator/GPU field is not part of the initial version.
+  Accelerators are requested through the compute-sizing profile (see *Hardware specification*); a standalone portable field waits for a Dynamic Resource Allocation (DRA) strategy that holds across providers.
+- KubeVM ports the *declaration* of a machine, not a running instance's state or its disk contents.
+  Cross-provider migration of a live VM or its data, and import of pre-existing VMs not created through the API, are out of scope for the initial version.
+- Backup and disaster recovery, marketplace, and billing integrations are outside the scope of the initial version.
+
+## Big Picture
+
+The proposal introduces a new API group, `kube-vm.io`, whose central resource is a `VirtualMachine` supported by a small set of companion types for sizing, images, networking, and snapshots.
+The design deliberately places the common surface — everything that is shared across backends — in the generic API, so that a provider needs to contribute only the settings that are unique to its platform.
+A machine is bound to its backend through a Cluster-API-style `spec.infrastructureRef`, and the generic layer observes the backend exclusively through a duck-typed status contract: a small, fixed set of well-known status fields — provider identifier, readiness, and network addresses — that the contract *requires* each provider to surface at agreed field paths.
+The generic core reads only those paths and imports no provider code, so it needs no per-provider translation and does not change when a provider is added.
+A provider whose native status already carries the same information under its own field names — as VM Operator does today — publishes the contract fields on its own object alongside them.
+Settling the exact contract — the canonical `providerID` form in particular — is part of defining the API.
+A generic controller reconciles the `VirtualMachine` against its provider object, and each provider contributes a controller that translates the resolved intent into calls against its platform.
+To keep the generic API from degenerating into the union of every vendor's feature set, a field is promoted into the portable core only once at least two providers converge on a common shape for it; until then it stays on the provider object where it originated.
+
+```mermaid
+flowchart TB
+  user(["kubectl / GitOps"]) --> gvm
+
+  subgraph kubevm["KubeVM Generic API — kube-vm.io"]
+    gvm["VirtualMachine<br/>sizing · image · bootstrap<br/>network · storage · power"]
+  end
+
+  gvm -->|"spec.infrastructureRef<br/>(duck-typed status contract)"| gc["Generic Controller"]
+
+  gc --> vsp
+  gc --> ec2
+  gc --> gcp
+
+  subgraph providers["Providers"]
+    vsp["vSphere provider<br/>(VM Operator)"]
+    ec2["EC2 provider"]
+    gcp["GCP provider"]
+  end
+
+  vsp --> esxi[("ESXi / vCenter<br/>vGPU · vMotion · HA")]
+  ec2 --> aws[("AWS EC2<br/>GPU instances")]
+  gcp --> gce[("Google Compute<br/>accelerators")]
+```
+
+An example makes the split concrete.
+The user writes one portable `VirtualMachine` that carries the whole intent — power state, sizing (including any accelerator, via the sizing profile), boot image, networking, and bootstrap — and the provider object it points at carries only what cannot be said portably, such as a storage policy, a firewall attachment, or a cloud identity.
+Here is a GPU workload written once against the sketched `kube-vm.io/v1alpha1` type.
+
+```yaml
+# Portable, user-authored resource — the machine's full intent in one place.
+# Portable in shape, not in every value: the marked lines below resolve against
+# whatever each platform's administrator has published under those names.
+apiVersion: kube-vm.io/v1alpha1
+kind: VirtualMachine
+metadata:
+  name: inference-01
+  namespace: team-a
+spec:
+  powerState: PoweredOn
+  # A GPU-bearing sizing profile: a VirtualMachineClass on vSphere, a GPU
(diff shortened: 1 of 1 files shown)


external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -2,6 +2,398 @@
...
-[#1893](https://github.com/vmware-tanzu/vm-operator/pull/1893) to read it
-and leave comments/feedback.
+This is a design proposal, not an accepted API.
+The types it describes are
+implemented as a strawman in this repository so the shape can be reviewed
+concretely; see the repository README for what is and is not built.
+
+## Business Problem
+
+Kubernetes has become the default control plane for modern infrastructure, yet the ecosystem still lacks a cross-platform, VM-centric API: a single declarative surface through which any hypervisor or cloud can expose both the full lifecycle of a virtual machine and the hardware capabilities that demanding workloads depend on.
```
dims, line 12: How do we prevent this from being a least common denominator API? 

Also what are the thoughts on leaky abstractions because folks will ALWAYS want to configure/tweak things below especially calling out security or performance

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -2,6 +2,398 @@
...
+This gap is becoming urgent because a new class of workload is arriving faster than the tooling to run it.
+Agentic workloads — long-running processes that execute model-generated code and orchestrate tools — are increasingly deployed inside virtual machines, both for the strong isolation a VM provides around untrusted code and for direct access to the hardware accelerators, such as GPUs, SR-IOV network functions, and passthrough devices, that hypervisors already virtualize well.
```
dims, line 14: Is the effort here then ... consistent / exact / replicatable environment across providers?

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -2,6 +2,398 @@
...
+The community needs a credible way to run these workloads, and today it does not have one.
+
+The CNCF landscape today addresses adjacent needs but not this one.
+Kata Containers provides VM-strength isolation for individual workloads by wrapping a Pod in a lightweight micro-VM — well suited to isolating untrusted code at the granularity of a container.
+KubeVirt takes a different approach, converging the virtual machine into the container model by running a QEMU/KVM process inside a Pod, which is an excellent fit when Kubernetes is the sole infrastructure layer and rich, device-level VM modeling on Kubernetes nodes is the goal.
+Both are strong at their design point.
+What neither sets out to be is a portable, Kubernetes-native front door to a full-blown, hypervisor-native estate — an existing vSphere deployment or a public-cloud VM service — that exposes that platform's own lifecycle and hardware capabilities (GPUs, SR-IOV, passthrough) through one vendor-neutral API.
```
dims, line 21: Here are we saying we don't care where we are running? (like virtual kubelet creating things outside of the nodes in the k8s cluster itself?)

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -2,6 +2,398 @@
...
+That is the gap KubeVM fills, and it is complementary to both.
```
dims, line 22: What's the stance on making kubernetes resources being made available to things in the VM? (secrets may be handy? env variables? how about CSI volumes?)

Are we limited to what is available in the underlying provider API?

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -2,6 +2,398 @@
...
+
+This document proposes **KubeVM**, a generic and vendor-neutral `VirtualMachine` API served under the `kube-vm.io` group, together with a provider model that lets hypervisors and cloud VM services expose their machines — and, critically, their accelerators — through one portable, Kubernetes-native interface.
+KubeVM is intended to complement KubeVirt, not to replace it: it addresses the hypervisor-native design point that the VM-as-Pod model leaves unaddressed.
```
dims, line 25: Do the requested VM(s) run inside existing kubernetes nodes? do we have a stance on that?

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: Precisely.  The goal is to get to a model of what I am calling "thick core with thin provider".  Which basically means most of the fields land up in the generic KubeVM API layer with very minor platform specific things landing in the provider resources.
The value proposition is a portable resource.  A much stronger value proposition is abstractions built on top of these resource templates.  E.g., a VirtualMachineReplicaSet resource that knows how to orchestrate VM replicas from templates.

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: The proposal is agnostic to that.  If KubeVM is managing KubeVirt resources, then it will leave the Kubernetes layer orchestrate the scheduling of the VM.  If it is another provider such as vSphere, or AWS, the underlying hypervisor layer may want to orchestrate the provisioning and management of VMs.

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: Good point.  Thank you for calling this out.  KubeVM will provide the abstraction that lets providers accomplish these tasks.  For example: a `spec.volume` reference with a SecretRef can be consumed by the provider to mount the secret as a volume (however they choose to do so).

> Are we limited to what is available in the underlying provider API?
Yes, to some extent we will be limited by what a provider (and the underlying platform / hypervisor) can support.

There are opportunities to standardize the API for common lifecycle operations across all providers dealing with VMs.  Bootstrapping (cloud-init / Sysprep) comes to mind.  We will work with providers on this.

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: > How do we prevent this from being a least common denominator API?
As with any new standard, I expect some challenges with this early on.  But, as we start to show value by bringing in abstractions on top of the virtual machine layer, I expect more and more providers to converge into integrating into this API.

Taking the same example of VirtualMachineDeployments: If the KubeVM orchestration layer handles creation, rollout and deletion of the VMs of the Deployment, every provider gains this feature without doing any work at all.

> Also what are the thoughts on leaky abstractions because folks will ALWAYS want to configure/tweak things below especially calling out security or performance

This is a fair point and this will always be a discussion.  For a provider that does a certain feature in a completely bespoke manner, they can continue to handle that via the provider APIs.  However, in doing so, they would lose the benefits offered by the KubeVM API.

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: Exactly.  It is up to the provider to figure out where and how the virtual machines are instantiated.  For a provider like Kubevirt, it will be on one of the nodes.  For a provider like vSphere / EC2 / GCP, it can be a "true" VM directly on the hypervisor.

external/kubevm/docs/one-pager-kubevm-generic-api.md
dims: I hear you and this is fine for now. 

BUT i do want you to think from the point of view of the people who will use this, formulating this as a `Business Problem` is thinking from the vendor perspective. 

How is this going to help someone? Can they switch between multiple kubernetes clusters that run on different infra providers? how easy would it be for them to switch between them? How do they specify say the underlying image? Will it help manage capacity by bursting into another provider when their primary cluster is full?

Do you think that is worth a section as well above?

external/kubevm/docs/one-pager-kubevm-generic-api.md
dims: I really think that you all have to have a heart-to-heart with the **kubevirt** folks. give them a demo, walk them through your solution/api et al. figure out overlap areas, where you are different, talk about future plans etc.

external/kubevm/docs/one-pager-kubevm-generic-api.md
dims: see above. please talk to kubevirt folks

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: Hi @dims !
I went ahead and modified the doc since the value proposition was not immediately clear.  Please take a look and let me know if the doc expresses that more clearly now.

To summarize:
There's a user story section now (following the KEP model) that talks about how this will allow customers to port their virtual machines from one infra provider to the other with minimal changes.  The cross provider scheduling at-burst is also a very real use-case that you pointed out.

Regarding the "how easy would it be for them to switch":
This depends on the virtual machine.  If a virtual machine depends on a lot of platform specific features **not** supported by the KubeVM generic layer yet, then the porting to another provider will not be smooth.  But as we work with other providers to have common aspects ported to the generic layer, the cross-provider switch will get a lot easier.

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: I reached out to Kubevirt folks to share the proposal in the #virtualization channel in the Kubernetes Slack.  I will keep you posted.

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -1,7 +1,551 @@
...
+The types it describes are implemented as a strawman in this repository so the shape can be reviewed concretely; see the repository README for what is and is not built.
+
+## Summary
+
+This document proposes **KubeVM**, a generic and vendor-neutral `VirtualMachine` API served under the `kube-vm.io` group, together with a provider model that lets hypervisors and cloud VM services expose their virtual machines (along with any specialized hardware such as accelerators) through one portable, Kubernetes-native interface.
+KubeVM is intended to complement KubeVirt, not to replace it: it addresses the hypervisor-native design point that the VM-as-Pod model leaves unaddressed.
+
+The `VirtualMachine` resource is the starting point rather than the whole proposal.
+Standardizing the virtual machine is what lets the ecosystem build a set, a service, a rolling deployment, quota, policy, health checking, capacity-aware placement across providers, and eventually a workload model spanning both VMs and Pods.
+Each of those is then written once for every provider rather than once per platform.
```
brito-rafa, line 15: A casual reader will intuitively assume there is an 1:1 relationship between provider and platform. This sentence begs to differ. I would explain why.

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -1,7 +1,551 @@
...
+That trajectory is described in [Beyond a single machine](#beyond-a-single-machine), and it is where most of the long-term value of a generic virtual machine API lies.
+
+## Motivation
+
+Kubernetes has become the default control plane for modern infrastructure, yet the ecosystem still lacks a cross-platform, VM-centric API: a single declarative surface through which any hypervisor or cloud can expose both the full lifecycle of a virtual machine and the hardware capabilities that demanding workloads depend on.
+This gap is becoming urgent because a new class of workload is arriving faster than the tooling to run it.
+Agentic workloads (specifically, long-running processes that execute model-generated code and orchestrate tools) are increasingly deployed inside virtual machines, both for the strong isolation a VM provides around untrusted code and for direct access to the hardware accelerators, such as GPUs, SR-IOV network functions, and passthrough devices, that hypervisors already virtualize well.
+The community needs a credible way to run these workloads, and today it does not have one.
+
+The CNCF landscape today addresses adjacent needs but not this one.
```
brito-rafa, line 25: It may also be worth noting that non-CNCF solutions further fragment the landscape; for example, ACK manages AWS EC2 instances, while Google Config Controller handles GCP Compute Engine.

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: Thanks for the review, @brito-rafa !

And good catch.  Originally when I wrote the proposal I had multiple providers for each platform in mind.  For example, there are 3 ways to creates VMs using Kubernets on vSphere. 

However, to keep things simple, I would drop that for now. The intent is: higher-level logic (sets, services, quota, policy, placement, etc.) is written once, against the standard KubeVM API, and that single implementation then works across every provider/platform, instead of being rewritten separately for each platform's native API.

I'll reword to make that explicit.

external/kubevm/docs/one-pager-kubevm-generic-api.md
aruneshpa: Added a line calling this out.

external/kubevm/docs/one-pager-kubevm-generic-api.md
fabiand: FWIW  - I think this suggestion here is similar to https://deltacloud.apache.org/ - in the end the industry drifted towards vendor APIs in order to leverage platform specific features. OpenStack has tried to standardize on flavors, but at the same time it allows vendors to complement flavors with hypervisor specific features in order to achieve the same need: access to bleeding edge or unique hypervisor features.

In other projects such as ansible or terraform we have seen that abstraktions appear, but also there you have to tailor the pipeline towards a provider.

To me, in the end, this proposal needs to make clear what problem we want to address, what is in- and what out-of-scope.

Simple life-cycle management between hypervisors? Yes.
Basic advanced (live migration/vmotion)? Difficult. Because there we already have differences.
Provisioning? Difficult. because different hypervisors have different concepts of what resources you need, and how they are combined.
Aka th more complex it becomes, the more difficult it becomes to find the common denominator.

And then we are back to basic operations, and then I wonder: Does this merit a controller, an API?

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -1,7 +1,599 @@
...
+
+The `VirtualMachine` resource is the starting point rather than the whole proposal.
+Standardizing the virtual machine is what lets the ecosystem build a set, a service, a rolling deployment, quota, policy, health checking, capacity-aware placement across providers, and eventually a workload model spanning both VMs and Pods.
+Each of those is then written once against the standard API, rather than once per platform's native API — a provider is a thin conformance layer under that standard, not a new surface consumers have to target.
+That trajectory is described in [Beyond a single machine](#beyond-a-single-machine), and it is where most of the long-term value of a generic virtual machine API lies.
+
+## Motivation
+
+Kubernetes has become the default control plane for modern infrastructure, yet the ecosystem still lacks a cross-platform, VM-centric API: a single declarative surface through which any hypervisor or cloud can expose both the full lifecycle of a virtual machine and the hardware capabilities that demanding workloads depend on.
+This gap is becoming urgent because a new class of workload is arriving faster than the tooling to run it.
```
fabiand, line 21: Which specific gap?

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -1,7 +1,599 @@
...
+Agentic workloads (specifically, long-running processes that execute model-generated code and orchestrate tools) are increasingly deployed inside virtual machines, both for the strong isolation a VM provides around untrusted code and for direct access to the hardware accelerators, such as GPUs, SR-IOV network functions, and passthrough devices, that hypervisors already virtualize well.
+The community needs a credible way to run these workloads, and today it does not have one.
```
fabiand, line 23: I do not see how/hy this proposal is addressing it.

The runtime of VMs does not change, we still depend on hypervisors.

This proposal is just an abstractio, thus at best it would be "how we orchestrate and managet those workloads with a concistent API".

external/kubevm/docs/one-pager-kubevm-generic-api.md
fabiand: We are happy to hear oyu out!

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -1,7 +1,599 @@
...
+
+The fragmentation is not limited to the CNCF landscape: outside it, each cloud exposes its own vendor-specific Kubernetes controller for its VM service — AWS ACK manages EC2 instances, Google Config Connector manages Compute Engine instances — so the same "write it once per platform" problem shows up there too. The CNCF landscape today addresses adjacent needs but not this one.
+Kata Containers provides VM-strength isolation for individual workloads by wrapping a Pod in a lightweight micro-VM: well suited to isolating untrusted code at the granularity of a container.
```
fabiand, line 26: Kata is an isolation for container workloads, with container semantics.

KubeVirt is a VM runtime (and API).
ESXi is a VM runtime.
KVM is a VM runtime.

IOW - to me Kata is a differnt use-case (container API), and thus does not belong here.

Just FWIW.

external/kubevm/docs/one-pager-kubevm-generic-api.md
```diff
@@ -1,7 +1,599 @@
...
+KubeVirt takes a different approach, converging the virtual machine into the container model by running a QEMU/KVM process inside a Pod, which is an excellent fit when Kubernetes is the sole infrastructure layer and rich, device-level VM modeling on Kubernetes nodes is the goal.
+Both are strong at their design point.
+What neither sets out to be is a portable, Kubernetes-native front door to a full-blown, hypervisor-native estate, such as an existing vSphere deployment or a public-cloud VM service, that exposes that platform's own lifecycle and hardware capabilities (GPUs, SR-IOV, passthrough) through one vendor-neutral API.
```
fabiand, line 29: Well, to me the mismatch is: For this abstraction proposed here.
The resources provided by the external hypervisors, would just be reflected in Kube, but not be consumable through Kube workloads.

And to me this is the issue. Kube is about matching resources to a workload. But this proposal would reflect external resources inside kube, not really depending on the kube resource model and scheduling at all.

To me ie crossplane is a better fit for such a model where the paltform (kube here) is not really involed in computation.

fabiand (commented):
Just some 2 cts, with xrfes to other projects from the past ...

PR #1898, opened 2026-09-03
🐛 E2E fix: Wait for disk promotion to complete if IP not available after timeout. 

aniket-deole:
Wait for disk promotion to complete if IP not available after timeout. Does not wait fail on disk promotion, if disk promotion is not completed, it runs another round to check if IP would be available.

Fixes # 4155

```release-note
NA
```


test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -188,20 +190,67 @@ func WaitForVirtualMachineCreation(ctx context.Context, config *config.E2EConfig
...
+		return
+	}
+
+	vm, err := utils.GetVirtualMachine(ctx, svClusterClient, ns, vmName)
+	Expect(err).ToNot(HaveOccurred())
+	if !meta.IsStatusConditionTrue(vm.GetConditions(), vmopv1.VirtualMachineDiskPromotionStarted) {
+		Fail(fmt.Sprintf("Timed out waiting for VirtualMachine %s to get an IPv4 address", vmName))
+	}
+
+	By(fmt.Sprintf("VirtualMachine '%s/%s' IP not ready yet; wait for disk promotion in case it's the cause", ns, vmName))
```
aakashchan, line 239: Sometimes DiskPromotion tasks from other VMs would keep VMware tools down. Suggestion is, let's run it irrespective and wait for disk promotion to complete before proceeding to the next VM. 

This will ensure that the infra is not overloaded in uncontrolled manner.

aniket-deole: Are you suggesting to increase the timeout from 5m to 10m for all cases?
If that's so, is checking for disk promo needed?

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -188,20 +190,67 @@ func WaitForVirtualMachineCreation(ctx context.Context, config *config.E2EConfig
...
+	ipTimeout, ipPoll := intervalDurations(config.GetIntervals("default", "wait-virtual-machine-vmip"))
+
+	By(fmt.Sprintf("Verify that an IP (ipv4) is allocated to the VirtualMachine '%s/%s'", ns, vmName))
+	if hasVirtualMachineIP(ctx, svClusterClient, ns, vmName, ipTimeout, ipPoll) {
+		return
+	}
+
+	vm, err := utils.GetVirtualMachine(ctx, svClusterClient, ns, vmName)
+	Expect(err).ToNot(HaveOccurred())
+	if !meta.IsStatusConditionTrue(vm.GetConditions(), vmopv1.VirtualMachineDiskPromotionStarted) {
```
bryanv: I think this might be too misleading in some cases - if we're going down this path, we need to be checking `vm.spec.promoteDisksMode` (and maybe the power state itself), and encryption doesn't need promotion because that doesn't support linked clones

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -188,20 +190,67 @@ func WaitForVirtualMachineCreation(ctx context.Context, config *config.E2EConfig
...
+		Fail(fmt.Sprintf("Timed out waiting for VirtualMachine %s to get an IPv4 address", vmName))
+	}
+
+	By(fmt.Sprintf("VirtualMachine '%s/%s' IP not ready yet; wait for disk promotion in case it's the cause", ns, vmName))
+	promoTimeout, promoPoll := intervalDurations(config.GetIntervals("default", "wait-virtual-machine-disk-promotion"))
+	_ = wait.PollUntilContextTimeout(ctx, promoPoll, promoTimeout, true, func(ctx context.Context) (bool, error) {
```
bryanv, line 241: Should this be an eventually so it just fails it the promote doesn't complete? A 10+ min cumulative wait if the VM isn't going to get an IP for some other reason does seem like a long time to wait for.

test/e2e/vmservice/lib/vmoperator/vmoperator.go
```diff
@@ -188,20 +190,67 @@ func WaitForVirtualMachineCreation(ctx context.Context, config *config.E2EConfig
...
+			e2eframework.Logf("retry due to: %v", err)
+			return false, nil
+		}
+
+		return meta.IsStatusConditionTrue(vm.GetConditions(), vmopv1.VirtualMachineDiskPromotionSynced), nil
+	})
+
+	By(fmt.Sprintf("Verify that an IP (ipv4) is allocated to the VirtualMachine '%s/%s'", ns, vmName))
+	Expect(hasVirtualMachineIP(ctx, svClusterClient, ns, vmName, ipTimeout, ipPoll)).To(BeTrue(),
+		"Timed out waiting for VirtualMachine %s to get an IPv4 address", vmName)
```
bryanv: Maybe indicate that this one is after the promote wait

bryanv (approved):

test/e2e/vmservice/lib/vmoperator/vmoperator.go
aniket-deole: Worst case, it waits for disk promo to complete only if it is in progress for 5 minutes + another 5 minutes for IP. Ideally in most cases, it should be < 2-3 minutes.

I did not use eventually as that would fatally kill the test, i wanted to see if the IP is present regardless at the end.

PR #1907, opened 2026-09-07
e2e: Extend VM delete timeout for snapshot cleanup test

anwithaun-1:

**RCA**: This test intentionally deletes the VM while it still has two attached snapshots (one a memory snapshot), to verify they get garbage-collected. vCenter's Destroy_Task on a VM with outstanding snapshots consolidates the delta disks and memory-snapshot file as part of the same task, which takes noticeably longer than deleting a snapshot-free VM. The test was polling with the shared default/wait-virtual-machine-deletion interval (5m/3s), which isn't budgeted for that consolidation overhead — the VM deletion succeeded, just a couple of minutes past the 5-minute window, causing a spurious failure.

**Fix:**
Added a spec-scoped interval override, vm-snapshot/wait-virtual-machine-deletion: ["10m", "5s"], in wcp.yaml and kind.yaml for giving one spec more headroom without loosening the timeout for every other caller.
Extended VerifyVMDeleted to accept an *IntervalOptions param (same pattern as WaitForBootDiskPVC) so callers can opt into a spec-specific interval.
Only the "Delete vm" step in the VM-SNAPSHOT test opts into the longer budget; all other callers pass nil and are unaffected

Fixes # vmop-4164

```release-note

```

### DIFF
--- a/test/e2e/vmservice/config/wcp.yaml
+++ b/test/e2e/vmservice/config/wcp.yaml
@@ -50,6 +50,12 @@ intervals:
   # over 4 minutes under datastore/NFC contention; give it more headroom than
   # "default" without loosening the timeout for every other caller.
   register-vm-restore-disk-only/wait-virtual-machine-condition-update: ["12m", "5s"]
+  # This spec deletes the VM while snapshots are still attached (to verify
+  # they're garbage-collected), which makes vCenter's Destroy_Task
+  # consolidate delta/memory-snapshot files before it completes; give it
+  # more headroom than "default" without loosening the timeout for every
+  # other caller.
+  vm-snapshot/wait-virtual-machine-deletion: ["10m", "5s"]
   default/wait-virtual-machine-powerstate: ["5m", "3s"]
   default/wait-virtual-machine-resize: ["5m", "3s"]
   default/wait-virtual-machine-moid: ["5m", "3s"]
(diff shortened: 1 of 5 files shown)


test/e2e/vmservice/config/wcp.yaml
```diff
@@ -50,6 +50,12 @@ intervals:
   # over 4 minutes under datastore/NFC contention; give it more headroom than
   # "default" without loosening the timeout for every other caller.
   register-vm-restore-disk-only/wait-virtual-machine-condition-update: ["12m", "5s"]
+  # This spec deletes the VM while snapshots are still attached (to verify
```
bryanv: Let's just change the default. It is already doing a lot of heavy lifting, like typically waiting for a graceful guest power off

anwithaun-1: Done.

anwithaun-1: @bryanv pls take a look again with this fix

test/e2e/vmservice/config/wcp.yaml
```diff
@@ -43,7 +43,11 @@ intervals:
   # Please keep the wait-virtual-machine values in sync between the different
   # configuration files.
   default/wait-virtual-machine-creation: ["5m", "3s"]
-  default/wait-virtual-machine-deletion: ["5m", "3s"]
+  # VM deletion typically waits for a graceful guest power-off, and when
```
bryanv: I really don't think we need this comment. Also keep the poll period to be 3 seconds.

anwithaun-1: Fixed

faisalabujabal (approved):

PR #1917, opened 2026-09-09
Restore omitempty in quicktype-generated schemas

brito-rafa:

`verify-codegen` in CI is failing on unrelated PRs (e.g. #1909) because
`pkg/util/netplan/schema/netplan.go` drifts from the output of the
currently pinned quicktype version (`hack/quicktype/package-lock.json`
locks `quicktype@26.0.0`).

Root cause: quicktype's Go renderer only tags an optional field with
`,omitempty` when the field's JSON Schema type excludes `null`. Fields
declared nullable (as `schemars` emits for Rust `Option<T>`, used
throughout netplan's `schema.json`) lose `omitempty` even though they
are absent from the schema's `required` list. quicktype exposes
`--omit-empty` to restore the broader (all-non-required-fields)
behavior, but neither invocation in `hack/quicktype/Makefile` passed
it.

`netplan.go` itself was never regenerated after the quicktype version
bump, because the generation rule keys off file mtimes
(`schema.json` newer than `netplan.go`), and most local checkouts
happen to leave both files with the same mtime — so `make generate`
silently no-ops and never notices the drift. CI's `verify-codegen`
only surfaced it once a fresh `actions/checkout` happened to give
`schema.json` a newer mtime than `netplan.go`, forcing Make to
actually rerun quicktype.

`pkg/util/cloudinit/schema/cloudconfig.go` was already regenerated
with the same missing flag in 9b356ef0 (part of the
controller-runtime dependency bump, merged today), which silently
dropped `omitempty` from real fields (`ssh_pwauth`, `users`, `ntp`,
`wireguard`, `merge_how`, etc.) without any CI failure, since the
committed file was already "self-consistent" with the buggy output.
That's a live behavior change on `main`: those fields now always
serialize (as explicit `null` when unset) instead of being omitted.
This PR fixes that regression too.

Verified:
- `make -C pkg/util/netplan/schema generate-go` and
  `make -C pkg/util/cloudinit/schema generate-go` now reproduce the
  committed files exactly (byte-for-byte stable across repeated runs).
- Forcing a full `make generate` + `git diff --exit-code` (the same
  check `verify-codegen` runs) is now clean.
- `go test ./pkg/util/netplan/... ./pkg/util/cloudinit/...` pass.
- `go build ./...` succeeds.

Fixes the `verify-codegen` failure seen on
https://github.com/vmware-tanzu/vm-operator/actions/runs/34416144439/job/102681639604?pr=1909

This is a codegen/tooling-only fix (regenerated files + a Makefile
flag) with no hand-written logic changes, so it's treated as a
trivial fix per the SDD constitution's exemption list rather than
requiring a `.sdd/specs/` entry.

```release-note
Fix a regression where generated netplan and cloud-init config types
stopped omitting empty optional fields, causing extra `null` values
to be serialized into guest customization data.
```

### DIFF
--- a/hack/quicktype/Makefile
+++ b/hack/quicktype/Makefile
@@ -115,7 +115,7 @@ $(OUTPUT_GO): | $(QUICKTYPE)
 	$(QUICKTYPE) \
 	  --src $(SCHEMA_IN) --src-lang schema \
 	  --out $@ --lang go --package schema \
-	  --top-level $(START_TYP)
+	  --top-level $(START_TYP) --omit-empty
 	$(GOIMPORTS) -w $@
 else
 $(OUTPUT_GO): | $(QUICKTYPE_IMAGE_RECEIPT)
@@ -129,7 +129,7 @@ $(OUTPUT_GO): | $(QUICKTYPE_IMAGE_RECEIPT)
 	  --out /output/schema.go \
 	  --lang go \
 	  --package schema \
-	  --top-level $(START_TYP)
+	  --top-level $(START_TYP) --omit-empty
 	mv -f schema.go $@
 	$(GOIMPORTS) -w $@
 endif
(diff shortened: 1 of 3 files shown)


brito-rafa:
Closing this PR, it will be addressed on a different one.

PR #1922, opened 2026-09-10
Mixed Case assessment and support

anwithaun-1:

Fixes several cross-system case-sensitivity bugs where UUID and username values sourced from one system (CNS CnsNodeVmAttachment, vCenter cluster module annotations, privileged users list) were compared or indexed against values from another system (vCenter BiosUUID, live cluster module UUID, session username) using case-sensitive string equality or raw map keys. Since UUIDs and usernames are case-insensitive by convention, a casing mismatch between the two sources could cause attachment lookups to miss, stale cluster module reconciliation, or false privileged-user rejections.

Fixes # vmop-4101

- These are targeted fixes from a case-sensitivity compliance pass across the codebase; no behavioral change is intended other than making these UUID/username comparisons case-insensitive.
- scanned report attached in the jira

```release-note

```

### DIFF
--- a/controllers/virtualmachine/volume/volume_controller.go
+++ b/controllers/virtualmachine/volume/volume_controller.go
@@ -57,7 +57,7 @@ func AddToManager(ctx *pkgctx.ControllerManagerContext, mgr manager.Manager) err
 		"spec.nodeuuid",
 		func(rawObj client.Object) []string {
 			attachment := rawObj.(*cnsv1alpha1.CnsNodeVmAttachment)
-			return []string{attachment.Spec.NodeUUID}
+			return []string{strings.ToLower(attachment.Spec.NodeUUID)}
 		}); err != nil {
 		return err
 	}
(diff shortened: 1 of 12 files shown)


faisalabujabal (approved):

faisalabujabal (approved):
please add tests that would handle the different casings

hpannem:
we may need to run full set of pipelines. Also, check our @bryanv comment on user name comparison in the chat.

anwithaun-1:
> we may need to run full set of pipelines. Also, check our @bryanv comment on user name comparison in the chat.

@bryanv

Because userInfo.Username may originate from external IdP identity  and it's already compared case-insensitively  in this file (IsKubeAdmin, IsVMOperatorServiceAccount), so this keeps PRIVILEGED_USERS matching consistent with that precedent.

Also username finding is detected by the case_sensitivity_audit  skill added in the initiative. 

Do you see any concern here?

@hpannem I don't see pipeline issue wrt this except few already existing failures present even without my changes.

anwithaun-1:
Added the unit tests to cover the different casings

PR #1942, opened 2026-09-16
🐛 Randomize e2e SSO usernames for test runs

aniket-deole:

Packer and devops-namespaces e2e tests used a fixed SSO username, which caused collisions when tests ran in parallel. Add a random suffix to each username instead.

Fixes # 4197

NA

NA

```release-note

```


test/e2e/vmservice/vmservice_suite_test.go
```diff
@@ -293,7 +296,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 var _ = SynchronizedAfterSuite(func() {
 }, func() {
 	// Clean up cluster role bindings (best effort)
-	if svClusterProxy != nil && !skipCleanup {
+	if svClusterProxy != nil && !skipClusterRoleBindingCleanup {
```
bryanv: I think the scope & purpose of a flag like this should be expanded, since there are other things later one that can cause similar issues

aniket-deole: I agree, do you have anything in mind? I can take it up.
Alternatively, I can limit the scope of this PR to SSO usernames only.

aniket-deole: I have removed persisting of cluster role bindings after test complete.

bryanv (approved):

PR #1945, opened 2026-09-16
Mixed Case assessment and support

anwithaun-1:
Fixes several cross-system case-sensitivity bugs where UUID and username values sourced from one system (CNS CnsNodeVmAttachment, vCenter cluster module annotations, privileged users list) were compared or indexed against values from another system (vCenter BiosUUID, live cluster module UUID, session username) using case-sensitive string equality or raw map keys. Since UUIDs and usernames are case-insensitive by convention, a casing mismatch between the two sources could cause attachment lookups to miss, stale cluster module reconciliation, or false privileged-user rejections.

Fixes #

```release-note

```


anwithaun-1:
UUID is out of scope for crossporting so,dropping this PR and I'll create new PR excluding the UUID changes

PR #1948, opened 2026-09-17
🌱 Policy test fixes and additional tests

hpannem:

Fixes the `vm-eviction-policy` and `controlled-rebalancing-policy` E2E suites, which were failing because `automatic_vm_eviction` compute policies cannot be created through dcli's `compute policies create` verb server-side. The suite previously built these through `createVSphereInfraPolicy`/`CreateComputePolicy` like every other capability, which does not work for eviction.

Key changes:
- Added `CreateComputePolicyWithSpec`/`DeleteComputePolicy`/`DeleteInfraPolicy` to the dcli client, going through `compute policies createtagsandpolicies` for capabilities (like `automatic_vm_eviction`) whose spec fields must be nested in the JSON payload rather than passed as flags.
- Made `ComputePolicySpec.HostTagID` optional for capabilities that don't act on a specific host.
- Refactored `createVSphereInfraPolicy`/`createVSphereTag` in `computepolicies/utils.go` to share a single per-spec tag category (instead of one category per tag) and to accept a caller-supplied suffix, so related objects created by the same test case are easy to correlate in vCenter/dcli output.
- Rewrote the host-maintenance-mode scenarios in `vmevictionpolicy.go` from a single three-VM test relying on the soft/preferential `vm_host_affinity` capability (which a routine DRS pass could override) into four focused scenarios, using a real mandatory DRS VM/Host affinity rule to reliably pin a VM to a host.
- Added cleanup for compute/infra policies and tag categories so E2E runs no longer leak vCenter/WCP objects.

Fixes #

This is test-only; no `pkg`/`api`/`controllers` changes are included. The diff is large because the eviction/maintenance-mode `It` blocks were restructured rather than patched incrementally — worth reviewing scenario-by-scenario rather than as a line diff.

```release-note

### CODE REVIEW

File: test/e2e/vmservice/vmservice/computepolicies/utils.go [hunk 1]
```diff
@@ -21,62 +22,90 @@ import (
...
+	}
+
+	By("Creating a real vSphere tag to associate with the policy")
+	tagName := fmt.Sprintf("%s-tag-%s", prefix, suffix)
+	tagID, err := wcpClient.CreateTag(tagName, "e2e VM eviction policy test", tagCategoryID)
+	Expect(err).ToNot(HaveOccurred(), "failed to create tag")
+	Expect(tagID).NotTo(BeEmpty(), "tag ID should be returned")
+
+	DeferCleanup(func(cleanupCtx context.Context) {
+		_ = tagManager.DeleteTag(cleanupCtx, &tags.Tag{ID: tagID})
```
aniket-deole, line 128: Althought its not fatal to the test run, does the defer order matter? Would deleting the tag fail cause DeferCleanup for infra/compute policy has not run yet?

test/e2e/vmservice/vmservice/computepolicies/vmevictionpolicy.go
```diff
@@ -843,114 +1145,133 @@ func getVMHostMoRef(
...
-// transitioning InfraInMaintenance state. It is a no-op (returning nil) if
-// the host is already in maintenance mode. Callers must cancel the returned
-// task (if non-nil) before attempting to exit maintenance mode, since the
-// task may still be queued/running when it does so.
+// enterHostMaintenanceModeTimeout bounds how long enterHostMaintenanceMode
+// waits for a host to actually reach maintenance mode. On real vCenter, a
+// powered-on VM must be evacuated before the host fully enters maintenance
+// mode; this bounds that wait so a host that can never be evacuated (e.g. no
+// DRS/vMotion capacity) fails the spec instead of hanging indefinitely.
+const enterHostMaintenanceModeTimeout = 5 * time.Minute
```
aniket-deole: Would this be enough? What if it has other chatty VMs like Supervisor CPVM?

test/e2e/vmservice/vmservice/computepolicies/vmevictionpolicy.go
```diff
@@ -466,121 +491,330 @@ func Spec(ctx context.Context, inputGetter func() SpecInput) {
...
+				fmt.Sprintf("vm-eviction-mm-avep-policy-%s", suffix),
+				vspherepolv1.PolicyEnforcementModeMandatory, matchLabel, evictionTagID, nil)
+
+			By("Creating a VM matching the policy's label")
+			avepVMName := fmt.Sprintf("%s-avep-%s", specName, suffix)
+			createAndWaitForPoweredOnVM(ctx, input, svClusterClient, avepVMName, matchLabel)
+
+			hostMoRef := getVMHostMoRef(ctx, vCenterClient, svClusterClient, input.WCPNamespaceName, avepVMName)
+			if len(listClusterHostMoRefs(ctx, vCenterClient, hostMoRef)) < 2 {
+				Skip("this spec requires more than one host in the cluster to distinguish VM relocation " +
```
aniket-deole, line 511: Skip before vm creation?

test/e2e/vmservice/vmservice/computepolicies/vmevictionpolicy.go
```diff
@@ -466,121 +491,330 @@ func Spec(ctx context.Context, inputGetter func() SpecInput) {
...
+			By("Verifying the VM is relocated off the host and stays powered on")
+			Eventually(func(g Gomega) {
+				g.Expect(getVMHostMoRef(ctx, vCenterClient, svClusterClient, input.WCPNamespaceName, avepVMName)).
+					ToNot(Equal(hostMoRef), "VM should be relocated off the host entering maintenance mode")
+			}, input.Config.GetIntervals("default", "wait-policy-evaluation-compliant")...).Should(Succeed())
+			vmoperator.WaitOnVirtualMachineCondition(ctx, input.Config, svClusterClient, input.WCPNamespaceName, avepVMName,
+				metav1.Condition{Type: vmopv1.VirtualMachinePowerStateSynced, Status: metav1.ConditionTrue})
+
+			By(fmt.Sprintf("Taking host %s out of maintenance mode", hostMoRef.Value))
+			exitHostMaintenanceMode(ctx, vCenterClient, hostMoRef, enterTask)
```
aniket-deole: do we need along with defercleanup?

test/e2e/vmservice/vmservice/computepolicies/utils.go
hpannem: defercleanup runs in lifo order. So, the order is correct, So, the policies are deleted and then tags right?
tags are created first (delete registered for them)
policies are created (delete registered)

and reverse it happens.

test/e2e/vmservice/vmservice/computepolicies/vmevictionpolicy.go
hpannem: I exposed timeout via config and also skipped test if host does not enter MM.

test/e2e/vmservice/vmservice/computepolicies/vmevictionpolicy.go
hpannem: We are creating a vm and geting the host on which it was placed and checking the host's cluster.

test/e2e/vmservice/vmservice/computepolicies/vmevictionpolicy.go
hpannem: Done.

test/e2e/vmservice/vmservice/computepolicies/vmevictionpolicy.go
aniket-deole: Why not just place it in a zone with relevant number of hosts using zone label and skip if not enoguh hosts in any cluster/zone?

hpannem: am relying on DRS to pick up zone/rp and place it. It would be better as it knows the correct resources. If I pickup a zone, it may not have sufficient resources right?

aniket-deole (approved):

test/e2e/vmservice/vmservice/computepolicies/utils.go
```diff
@@ -21,62 +24,111 @@ import (
...
+	Expect(wcpClient.CreateInfraPolicy(infraPolicySpec)).
+		To(Succeed(), "failed to create infrastructure policy %q", infraPolicySpec.Name)
+
+	intervals := input.Config.GetIntervals("default", "wait-compute-policy-deletion")
+	timeout, err := time.ParseDuration(intervals[0].(string))
+	Expect(err).ToNot(HaveOccurred(), "failed to parse compute-policy-deletion timeout interval %q", intervals[0])
+	poll, err := time.ParseDuration(intervals[1].(string))
+	Expect(err).ToNot(HaveOccurred(), "failed to parse compute-policy-deletion polling interval %q", intervals[1])
+
+	DeferCleanup(func(_ context.Context) {
```
aruneshpa: The SDD is saying that this is supposed to be wrapped in an Eventually block.

test/e2e/vmservice/vmservice/computepolicies/utils.go
```diff
@@ -21,62 +24,111 @@ import (
...
+		// Delete the infra policy before the compute policy it references.
+		_ = wcpClient.DeleteInfraPolicy(infraPolicySpec.Name)
+
+		// Deleting the compute policy right after the infra policy that
+		// referenced it can transiently fail while WCP propagates the infra
+		// policy removal. Retry in the background rather than blocking this
+		// cleanup on it. This is best-effort cleanup, so a fire-and-forget
+		// retry that never blocks. Even it fails, this is fine as
+		// the policy is never reused.
+		go func() {
```
aruneshpa: Hmm.  If the parent process exits, this won't log anything in case there was a failure here.  Consider adding a log line here and emitting it to GinkgoWriter maybe. This would be the only clue in figuring out what went wrong if a stale compute policy is left around on a testbed.

While here, do you want to track the goroutines that you are firing here so you can wait for them to finish (with a timeout) in the after suite - in the spirit of idempotent tests?

aruneshpa (approved):
LGTM, one comment.

PR #1954, opened 2026-09-17
✨ Replica set unit tests

hpannem:

Adds the missing unit-test tier for `VirtualMachineReplicaSet` (previously only a thin 4-`It` envtest integration file existed, no unit coverage at all). Tests are written against `.sdd/specs/008-virtualmachinereplicaset/tds.md`'s 38 Given/When/Then scenarios rather than derived from the existing implementation, per that spec's task list (`tasks.md` T001–T007).

Scope covered: core reconciliation (create-from-scratch, replicas defaulting/explicit-zero, idempotency, scale-to-zero-and-back, rapid successive scale edits), status/condition correctness (`fullyLabeledReplicas` drift, `VirtualMachinesCreated` on create failure, `observedGeneration` freshness, `Resized` convergence), template/selector/label semantics (selector/template edits don't mutate existing replicas, label propagation rules, ownership-vs-discoverability), and non-goal absence checks (no rollout/revision fields, no built-in auto-scaling).

While writing tests against the TDS's intended contract, two real product gaps surfaced and are fixed here (not just documented):
- `ReplicaFailure` condition was defined on the API type but never set by the controller, even on sustained create/delete failures.
- `VirtualMachinesReady` condition was never set by `updateStatus` at all.

Both are now implemented in `virtualmachinereplicaset_controller.go`, with two new reason constants (`VirtualMachineDeletionFailedReason`, `VirtualMachinesNotReadyReason`) added to the API type. `tds.md` is updated to reflect both as "found and fixed" rather than open gaps, and to resolve two previously-`[NEEDS CLARIFICATION]` markers (§16 template-metadata-label propagation, §27 `replicas == 0` readiness semantics) with the behavior now codified as tests.

Also: registered `VirtualMachineReplicaSet` in `test/builder/fake.go`'s `KnownObjectTypes` so the fake client enforces the status subresource split per `testing-standards.md`; renamed the spec directory from `003-virtualmachinereplicaset` to `008-virtualmachinereplicaset` to resolve a numbering collision with `003-compute-config-reconcile`, and added it to `.sdd/INDEX.md` (it was missing from the index entirely).

Fixes #

Out of scope for this PR (deliberately deferred, tracked in `tasks.md` T008–T017): validation webhook tests (SC12/SC13), envtest integration tests for scaling/ownership/adoption/readiness (SC6–8, 11, 17–21, 24, 32, 33, 37, 38), mutation webhook baseline tests, and the real-vCenter E2E suite. Three scenarios remain intentionally unimplemented pending other tickets: SC12 (no negative-replicas validation exists), SC19 (adoption happens but has no persistent Condition — vmop-4017), and SC24 (`readyReplicas` counts every VM as ready regardless of actual status — has a `// TODO` in the controller). These live in the integration/webhook tiers, not touched here.

```release-note
NONE

### CODE REVIEW

File: controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go [hunk 1]
```diff
@@ -505,7 +508,14 @@ func (r *Reconciler) syncReplicas(
 		}
 
 		if len(errs) > 0 {
-			return apierrorsutil.NewAggregate(errs)
+			aggErr := apierrorsutil.NewAggregate(errs)
```
aruneshpa: A minor nit that this condition is set when the VM creation fails, but cleared when the poll (after creation) succeeds.  So, there's a period of time where this can be stale (if the creation succeeds in next reconcile, but poll times out).
I can't think of a super elegant way to handle that without peppering condition setting all over the place, so this is fine I guess.  Add a comment please.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -687,5 +704,20 @@ func (r *Reconciler) updateStatus(
 		// This means that we have sufficient number of VirtualMachine objects.
 		conditions.MarkTrue(rs, vmopv1.VirtualMachinesCreatedCondition)
 	}
-	// TODO: Set aggregate condition based on the condition of the individual Virtual Machines
+
+	switch {
+	// An empty desired set is trivially/vacuously ready.
```
aruneshpa: > vacuously

Nice.

aruneshpa (approved):
I think the PR description is stale: it calls out "008-virtualmachinereplicaset" instead of "009-virtualmachinereplicaset".

Otherwise looks good.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
```diff
@@ -687,5 +704,20 @@ func (r *Reconciler) updateStatus(
...
+	case desiredReplicas == 0:
+		conditions.MarkTrue(rs, vmopv1.VirtualMachinesReadyCondition)
+	case newStatus.ReadyReplicas >= desiredReplicas:
```
aruneshpa: Hmm.  If `desiredReplicas = 0` doesnt this hold true as well?  IOW, the first case can be removed?

hpannem: Done.

controllers/virtualmachinereplicaset/virtualmachinereplicaset_controller.go
hpannem: I delete the condition before wait. PTAL.

PR #1569, opened 2026-04-24
:seedling: Update controller runtime to 0.23.1 and otel 1.41

brito-rafa:
**What does this PR do, and why is it needed?**

This PR is based and incremental on top of https://github.com/vmware-tanzu/vm-operator/pull/1532

Update controller-runtime from v0.22.3 to v0.23.1 along with k8s.io/* dependencies from v0.34.1 to v0.35.4.

Ginkgo had to be upgraded to v2.27.2.

This patch includes updating opentelemetry dependency to v1.41 to address CVE-2026-29181 and CVE-2026-39883.

This includes the following breaking/deprecation changes (from @aruneshpa):
- `ctrl.NewWebhookManagedBy(mgr).For(&obj{}).Complete()` changed to `ctrl.NewWebhookManagedBy(mgr, &obj{}).Complete()` (object is now passed as a second argument instead of via `.For()` chaining).
- `mgr.GetEventRecorderFor()` is deprecated in favor of `mgr.GetEventRecorder()`, which returns the new k8s.io/client-go/tools/events.EventRecorder instead of the old k8s.io/client-go/tools/record.EventRecorder. Migrated pkg/record.Recorder to accept the new EventRecorder type and updated all call sites and test fakes accordingly. As per @aruneshpa "I am still going to keep our pkg/record because that is providing nice helpers on top of the upstream event recorder interface which only seems to support `Eventf.`"

As per investigation why some test-services failed after the upgrade,
it has been pointed that k8s v0.35.x client being ~100ms slower in cache sync.
On a couple of service tests, the `Eventually` default timeout "1s" had to be increased to "2s".

Lastly, as suggested by @bryanv during previous PR https://github.com/vmware-tanzu/vm-operator/pull/1532, this PR instruments `new-schema-version.py` with the new `ctrl.NewWebhookManagedBy(mgr, &obj{}).Complete()`.

**Which issue(s) is/are addressed by this PR?** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes # CVE-2026-29181 and CVE-2026-39883

**Are there any special notes for your reviewer**:

Code changes based on https://github.com/vmware-tanzu/vm-operator/pull/1532

**Please add a release note if necessary**:

```release-note
Update controller runtime to 0.23.1, opentelemetry to v1.41 and Kubernetes libraries to v0.35.4
Addresses CVE-2026-29181 and CVE-2026-39883
```

PR #1956, opened 2026-09-18
[x-port] [9.1.2] 🐛 Atomic VM publish status patching

brito-rafa:

* Atomic status patching and controller cleanups for VM publish operations (#1888)

Refactors the status patching logic across VirtualMachinePublishRequest and group publish controllers to prevent race conditions and incomplete status updates caused by split patch operations. Adds an option to the patch helper to issue a single status patch rather than splitting conditions and other status fields, ensuring fields like image names are set simultaneously with completion conditions.

Also switches those controllers to contextual loggers, corrects an invalid condition reason on invalid group publish specs, simplifies the publish request requeue check, and removes inaccurate CompleteTime documentation.

* Fix VirtualMachinePublishRequest Ready without ImageName (#1508)

When the conditions patch succeeded but the status patch failed, the object could have ImageAvailable condition true while ImageName was never persisted. A later reconcile would then set Ready without ImageName because checkIsImageAvailable returned early when ImageAvailable was already true.

Only skip the VMI lookup when both ImageAvailable is true and ImageName is already set; otherwise run the lookup and backfill Status.ImageName when a matching VMI is found. This repairs the inconsistent state on the next reconcile.

* Remove deprecated controller-runtime GetEventRecorderFor (#1569)

mgr.GetEventRecorderFor() is deprecated in favor of mgr.GetEventRecorder(), which returns the new k8s.io/client-go/tools/events.EventRecorder instead of the old k8s.io/client-go/tools/record.EventRecorder. Migrated pkg/record.Recorder to accept the new EventRecorder type and updated all call sites and test fakes accordingly.

Investigating why some test-services failed after the upgrade pointed out that the k8s v0.35.x and later client is ~100ms slower to sync its cache. The `Eventually` default timeout of "1s" had to be increased to "2s" in a couple of tests.

* Fix event recorder names to comply with events.k8s.io API validation (#1580)

The events.k8s.io API enforces that reportingController names are valid Kubernetes qualified names (regex `([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]`, under 64 characters). Previous event recorder names used the pattern "namespace/pod-name/component-name", which violated this constraint by containing multiple "/" characters and exceeding the length limit. Event recorders now use simple component names like "virtualmachine-controller" and "vmProvider" instead.

Adds the RBAC needed for the events.k8s.io API group.

Fixes #

#1569 and #1580 are partial cherry-picks adapted for this branch; see the individual commit messages for details.

```release-note
Adds RBAC for the events.k8s.io API group, required by the migration to mgr.GetEventRecorder().
```


aruneshpa (approved):
Approving.  Had @brito-rafa verify that this is a clean x-port of the 4 mentioned commits (with conflict resolution).

PR #1957, opened 2026-09-20
🌱 Add exponential backoff for Content Library download polls

brito-rafa:

This patch replaces the fixed-interval poll `(wait.PollUntilContextCancel)` used while waiting for a Content Library item's download-session file to become PREPARED with exponential backoff `(wait.ExponentialBackoffWithContext)`, from the already-vendored `k8s.io/apimachinery/pkg/util/wait`.

With current fixed interval, a slow or failing download-session floods vCenter with repeated fetch calls.

This patch leverages the existing ContentAPIWait parameter (1s by default) as the seed for the backoff, using industry-standard values: each retry doubles the interval (Factor: 2.0) with 10% jitter, and  plateaus at 120x that base — so a slow prepare backs off instead of hammering vCenter indefinitely, while a raised ContentAPIWait still scales the ceiling proportionally rather than leaving it pinned to an unrelated fixed value. The caller's context remains the real stop condition, so cancellation/timeout behavior is unchanged.

`k8s.io/apimachinery/pkg/util/wait`'s `ExponentialBackoffWithContext` has a quirk: `Backoff.Cap` zeroes `Backoff.Steps` the moment the growing interval would exceed the cap, and since Steps also gates that function's own retry loop, hitting the cap ends the whole loop with ErrWaitTimeout instead of leveling off the interval. To retry with a real grow-then-plateau shape, helper function `pollWithGrowthThenPlateau` calls ExponentialBackoffWithContext in two phases instead: a growth phase whose Steps is sized by growthStepsToReach to stop exactly at the plateau boundary (no Cap set), then a fixed-interval plateau phase (Factor: 0, so the Cap-crossing branch is never exercised) that keeps polling until the condition succeeds, a real error occurs, or ctx is done.

The new tests wrap the real vcsim server's REST transport with a scripted http.RoundTripper to simulate UNPREPARED responses, a terminal session error, and a stalled prepare — no production code changes were required to make that possible.

Fixes # vmop_6672

This is an internal retry/backoff change with no product-facing API/CRD/webhook behavior change, so it falls under the trivial-fix exemption in the constitution.

```release-note
NONE
```


pkg/providers/vsphere/contentlibrary/content_library_provider.go
```diff
@@ -46,14 +48,123 @@ type Provider interface {
...
+	})
+	if conditionErr != nil {
+		return conditionErr
+	}
+
+	//nolint:staticcheck // wait.Interrupted() also matches ctx cancellation/deadline,
+	// which must propagate immediately rather than fall through to the plateau phase;
+	// ErrWaitTimeout is the only way to detect steps-exhaustion specifically.
+	if !errors.Is(err, wait.ErrWaitTimeout) {
+		// ctx cancellation/deadline -- condition itself never errored.
```
bryanv: Isn't this also the happy path return too?

pkg/providers/vsphere/contentlibrary/content_library_test.go
```diff
@@ -5,19 +5,255 @@
...
+			multipliers := []int{1, 2, 4, 8, 16, 32, 64, 120, 120, 120}
+			for i, multiplier := range multipliers {
+				gap := pollTimes[i+1].Sub(pollTimes[i])
+				base := time.Duration(waitSeconds*multiplier) * time.Second
+				// Virtual time has no scheduling overhead. Retain production
+				// jitter and verify its bounds, including the phase transition.
+				Expect(gap).To(BeNumerically(">=", base), "poll interval %d", i+1)
+				Expect(gap).To(BeNumerically("<", base+base/10), "poll interval %d", i+1)
+			}
+		},
```
bryanv: Formatting looks a little off here

